PR #29, opened 2015-01-16
fix web3.toAscii returning empty strings

caktux:


debris:
can you also write tests for it? including tests like toAscii(fromAscii('xxx'))

PR #33, opened 2015-01-17
return strings instead of BigNumber, return on no output from transact...

caktux:
...to prevent BigNumber errors


debris:
i discussed that with Gavin, and we will probably stay with returning BigNumbers. what errors are you talking about?

debris:
is it parsing '0x' value by output parser? i will accept fix for that as a separate pr with test :)

PR #35, opened 2015-01-18
Export same module name when building bundle

### LINKED ISSUE #34: why are we renaming the module when we build?
This is unnecessary and causes confusion. Should be the same as the name in package.json
see #14 #17

kumavis:
Fixes #34

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,7 +25,7 @@ var build = function(src, dst, ugly) {
       detectGlobals: false,
       bundleExternal: false
     })
-    .require('./' + src + '.js', {expose: 'web3'})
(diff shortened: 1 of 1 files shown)


kumavis:
Would also need to:
* update the readme
* bump major version in `package.json` as this would be a breaking change. However, thats only true for people consuming the built version who likely not be following semver, but whatever.

kumavis:
related https://github.com/ethereum/ethereum.js/pull/56

caktux:
Why is this still pending? Mergeable or not, this is a trivial fix and needs to be addressed.

debris:
moved to #117

PR #47, opened 2015-01-28
added ethereumjs-utils

wanderer:
this seprates the utility functions to a sperate module that can be resused


jorisbontje:
ethereum.js requires bignumber.js, ethereumjs-utils uses bignum; can those be unified?

PR #48, opened 2015-01-28
added ethereumjs-utils

wanderer:
this seprates the utility functions to a sperate module that can be resused


jorisbontje:
ethereum.js requires bignumber.js, ethereumjs-utils uses bignum; can those be unified?

wanderer:
node-ethereum, ethereumjs-lib all use [bignum](https://github.com/justmoon/node-bignum) in node.js. When it get browserferied we use [bn.js](https://github.com/indutny/bn.js).  Also our elliptic curve operations relay on bn.js. So switching to a different bignumber.js is nontrival. How hard would it be to switch ethereum.js to bn.js ?

wanderer:
so maybe I missing something but I can't actully see where bignumber.js is being used.

debris:
I close, because it's outdated and it's not a priority right now.

PR #56, opened 2015-02-09
fixed require statement for node.js import

psalami:
In node.js, you have to require('ethereum.js'); the web3 module is not recognized.


ethers:
same here https://github.com/ethereum/ethereum.js/pull/14

debris:
moved to #117

PR #57, opened 2015-02-11
Externs file for optimizing compilers

sundbry:
Hi Marek,

Here is an externs file for including ethereum.js from projects built with the Google Closure compiler. You may want to include it for people building their Javascript projects with advanced compilation etc. I am personally using it here to provide ethereum.js to ClojureScript applications: https://github.com/sundbry/ethereum.cljs

Thanks for all the work you've done...

--Ryan


debris:
Hi!

Thanks for showing Google Closure Compiler. I didn't know about it until now. But I'm afraid, that I cannot accept this pr.

1. I like the buildsystem to be automated as much as possible. Changes in this pull request requires manual updates of .ext.js file.
2. After short research i found out, that we can use npm's [closurecompiler](https://www.npmjs.com/package/closurecompiler), but it requires JRE and downloads it, if it is not available. I don't want to introduce this dependency to our build system.

sundbry:
No problem, I also wish it were easier to auto-generate.

PR #67, opened 2015-02-19
changed most of the function names

frozeman:
Missing is still:

- eth.watch
- eth.contract (has different behaviour now)
- shh methods

correct return values need to be checked


frozeman:
The new api names according to https://github.com/ethereum/wiki/wiki/JavaScript-API-improvement-proposal are implemented besides `watch` and `contract`.

The new behavior of watch should be slightly different.
`watch` will be `filter` and when called does nothing, unitl one calls `filter.watch(cb)` or `filter.get()`.

`contract` should no instantiate an instance of a contract, but rather give the contract prototype so that one can get more contracts than one out of the object e.g.:

```js
var mycontract = web3.eth.contract(abi);

var myWalletX = new mycontract(myAddress);
var walletY = new mycontract(otherAddress);
```

frozeman:
BTW all old functions are still working stil, but show a deprcated warning. Sadly we can't hide them from the object as chrome doesnt support `__noSuchMethod__ ` :( -> http://stackoverflow.com/questions/994143/javascript-getter-for-all-properties

kumavis:
@frozeman can you set enumerable to false for the deprecated properties? or do they still show up in the debugger?

debris:
at first look, it seems ok. before merging, please rerun gulp once again and update `./test/web3.methods.js`, `./test/eth.methods.js` etc. accordingly to the changes. Because all methods are created dynamically I would like to have at least tests which check if they exist.

debris:
changes in filters/watches && contract should be separated to another pull request.

frozeman:
@kumavis sadly this didn't hide it from the debugger

lib/web3.js
```diff
@@ -70,13 +99,23 @@ var setupProperties = function (obj, properties) {
...
                 });
             };
         }
-        Object.defineProperty(obj, property.name, proto);
+
+        Object.defineProperty(obj, property.name, {
+            enumerable: (property.newProperty) ? false : true,
+            value: proto
+        });
+
```
debris: that's incorrect.
istead of 
```javascript
Object.defineProperty(obj, property.name, {
    enumerable: (property.newProperty) ? false : true,
    value: proto
});
```

should be:
```javascript
proto.enumerable = !property.newProperty;
Object.defineProperty(obj, property.name, proto);
```

PR #71, opened 2015-02-23
add formatter for gasPrice and balance

frozeman:


dist/ethereum.js
```diff
@@ -310,13 +310,13 @@ var exportNatspecGlobals = function (vars) {
...
     contract.call = function (options) {
-        contract._isTransact = false;
+        contract._isTransaction = false;
         contract._options = options;
         return contract;
     };
 
-    contract.transact = function (options) {
-        contract._isTransact = true;
+    contract.sendTransaction = function (options) {
```
debris: can `contract.transact` be also marked as deprecated?

dist/ethereum.js
```diff
@@ -452,24 +452,40 @@ var addEventsToContract = function (contract, desc, address) {
...
+ * @param abi - abi json description of the contract, which is being created
  * @returns contract object
  */
+var contract = function (abi) {
+
+    // return prototype
+    if(Object.prototype.toString.call(abi) === '[object Array]' && arguments.length === 1) {
+        return Contract.bind(this, abi);
+
+    // depreacted: auto initiate contract
```
debris: typo

lib/contract.js
```diff
@@ -180,24 +180,40 @@ var addEventsToContract = function (contract, desc, address) {
...
+ * @param abi - abi json description of the contract, which is being created
  * @returns contract object
  */
+var contract = function (abi) {
+
+    // return prototype
+    if(Object.prototype.toString.call(abi) === '[object Array]' && arguments.length === 1) {
+        return Contract.bind(this, abi);
+
+    // depreacted: auto initiate contract
```
debris: typo

dist/ethereum.js
frozeman: alright

lib/contract.js
```diff
@@ -180,24 +188,40 @@ var addEventsToContract = function (contract, desc, address) {
...
- * @param address - address of the contract, which should be called
- * @param desc - abi json description of the contract, which is being created
+ * @param abi - abi json description of the contract, which is being created
  * @returns contract object
  */
+var contract = function (abi) {
 
-var contract = function (address, desc) {
+    // return prototype
+    if(Object.prototype.toString.call(abi) === '[object Array]' && arguments.length === 1) {
```
debris: can you make utils methods for checking types? eg.

```javascript
var isFunction = function (..) {}
var isArray = functin (...) {}
// and so on
```

also i would prefer to use `x instanceof Array`. Or you should check how underscore or lodash is checking if something is array.

lib/formatters.js
```diff
@@ -99,6 +101,8 @@ var formatOutputInt = function (value) {
     return new BigNumber(value, 16);
 };
 
+
+
 /// Formats big right-aligned input bytes to uint
```
debris: why are you adding blank lines here?

lib/formatters.js
```diff
@@ -90,7 +90,9 @@ var signedIsNegative = function (value) {
 /// Formats input right-aligned input bytes to int
 /// @returns right-aligned input bytes formatted to int
 var formatOutputInt = function (value) {
+
     value = value || "0";
+
```
debris: why are you adding blank lines here?

test/eth.methods.js
```diff
@@ -8,7 +8,7 @@ describe('web3', function() {
         u.methodExists(web3.eth, 'getState');
         u.methodExists(web3.eth, 'getStorage');
         u.methodExists(web3.eth, 'getTransactionCount');
-        u.methodExists(web3.eth, 'getCode');
+        u.methodExists(web3.eth, 'getData');
```
debris: `getData` is not verbose enough for me. `getCode` implicitly told me, that I want to get it from some address. `getData` doesn't do that.

lib/contract.js
```diff
@@ -180,24 +188,40 @@ var addEventsToContract = function (contract, desc, address) {
...
+        return Contract.bind(this, abi);
```
debris: it's not necessary to bind this. it can be also null

debris:
generally it looks ok. please fix these few things that I mentioned. I also made jshint on master branch more strict. Please merge or rebase and try to apply style changes to your code (if it is necessary).

lib/contract.js
frozeman: i changed to `instanceof Array`. I don't see the point for creating our own check functions here, if we really need that than lets use underscore etc. for its ok like that i say

lib/formatters.js
frozeman: visibility :)

PR #79, opened 2015-02-26
Api overhaul - value transformations

frozeman:
Don't merge yet, i opened it already so that i can see the increased test coverage.


kumavis:
Looking good! 
Standard callback signature is `cb(err, result)`. Do you think this is not appropriate for ethereum.js?
https://github.com/ethereum/ethereum.js/pull/79/files#diff-f9ee6b5f84e0ac1cde088e0cc86f045fR67

frozeman:
i agree with the error first approach

chevdor:
:clap:

kumavis:
:fireworks: :+1:

PR #81, opened 2015-02-27
constants should be before array contents, fixes #30

### LINKED ISSUE #30: array support
need to be able to call contracts where some input parameters are arrays.

please add an example too, or extend 
https://github.com/ethereum/ethereum.js/blob/develop/example/contract.html

ethers:
PR as previously requested by @gavofyork 

3 tests added.  formatInput function has the fix.  (my editor auto cleaned up extraneous whitespaces)


ethers:
forgot some things in the first commit, thus sorry for the extra commits

debris:
great job!

PR #87, opened 2015-02-27
Move browserify transforms to package.json

### LINKED ISSUE #36: Browserify and CORS
https://github.com/ethereum/ethereum.js/blob/master/lib/httprpc.js#L26
This line inadvertently causes some problems when using ethereum.js with browserify.
Though we're in the browser, `XMLHttpRequest` gets overridden with `xmlhttprequest` module's version.
This would be fine, but it has a bug where it behaves as if `xhr.withCredentials` is set to `true`.
This disables wildcard ('*') CORS headers. Which is annoying.

### LINKED ISSUE #86: Cannot be externally browserified
######  EDIT:

There are transforms necessary as part of the build process. This is not a problem, but external browserify processes are not aware of them. Here is a [simple fix](https://github.com/ethereum/ethereum.js/pull/87).

###### original description below:

The way the tree trimming is being done with envify is preventing the module from being included and browserified externally. This means even modular projects using browserify must expose `ethereum.js` as a global.

[here is xhr](https://github.com/ethereum/ethereum.js/blob/master/lib/httpsync.js#L24-L26) as an example of one of the environment specific code that is causing problems. [And here is Bignum] (https://github.com/ethereum/ethereum.js/blob/c184728ee8391ef6d381d685f36af8f6684703ac/lib/const.js#L24-L26), both are problems for subtle reasons. E.g. in the case of `BigNum` even if you don't enter that if statement, the var BigNum now blocks the globally exposed BigNum. 

There are a couple solutions:

Add the necessary transforms to the package.json, so that browserify knows to use those when processing your module.
```
"browserify": {
  "transform": [ 'envify', 'unreachable-branch-transform' ]
}
```
~~Although specifying options is a little tricky~~
https://github.com/substack/module-deps#transforms

Another way is to rethink why those things are there in the first place. Its a weird mismatch of expectations, where you expect globals to be present in one environment and not another. Why not always expect `BigNum` to be present? or never expect it to be present and provide your own.

This problem can be resolved **without any breaking changes**, it just requires a little communication so we can figure out a solution we're both comfortable with.

So to proceed, can you explain the rational for expecting globals like BigNum?

kumavis:
Browserify will automatically read the `package.json` and run the transforms. **This produces the exact same build** as specifying the transforms in the `gulpfile.js`. Additionally, now when someone tries to externally browserify `ethereum.js`, it will use the same transforms as when you use `gulp`. Yay! Consistency!

Fixes #86
Fixes #36

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -47,8 +47,6 @@ gulp.task('build', ['clean'], function () {
     return browserify(browserifyOptions)
         .require('./' + src + '.js', {expose: 'web3'})
         .add('./' + src + '.js')
-        .transform('envify', { NODE_ENV: 'build' })
-        .transform('unreachable-branch-transform')
         .bundle()
         .pipe(exorcist(path.join( DEST, dst + '.js.map')))
         .pipe(source(dst + '.js'))
--- a/package.json
+++ b/package.json
@@ -45,6 +45,19 @@
   "bugs": {
     "url": "https://github.com/ethereum/ethereum.js/issues"
   },
+  "browserify": {
+    "transform": [
+      [
+        "envify",
+        {
+          "NODE_ENV": "build"
+        }
+      ],
+      [
+        "unreachable-branch-transform"
+      ]
+    ]
+  },
   "keywords": [
     "ethereum",
     "javascript",


wanderer:
:+1:

kumavis:
I'm happy to provide further explanation, just ask

kumavis:
@debris if there is anything I can do to help get this merged, let me know :smile_cat:

debris:
looks good! thanks!

PR #98, opened 2015-03-04
Browserify transforms must be in normal dependencies

kumavis:
Move `envify` and `unreachable-branch-transform` to normal dependencies for external browserify support.

This does not affect builds, but fixes external browserify support. Sorry for the trouble.


kumavis:
Fixes https://github.com/ethereum/ethereum.js/issues/86

PR #100, opened 2015-03-05
Add Karma support for In-Browser testing

chevdor:
To test, ensure you do have Chrome, Firefox and Safari installed. If this is not the case, you may either remove the missing ones (or test some more with IE for instance) either from the karma.conf.js file or by running the tests from the command line.

The default is set to watch the test files, changing any will re-rerun the tests.

To make a single run with selected browsers, you may run:
```karma start --singleRun=true --browsers="Chrome,Firefox" ```

At the moment, 128/128 tests pass for Chrome, Safari, Firefox.
In action: https://asciinema.org/a/17349

screen shot 2015-03-05 at 23 31 54


chevdor:
e46f6f156b53858212f9ab0abf1df96cc171b375 brings the missing tests for toEth()

debris:
great job! thanks!

chevdor:
Thanks @debris 
At least now we can test locally with any browser but the next step would be to check if we can run the karma test during the CI. Not sure though what browsers are available in travis. I will try and see if that passes or fails.

debris:
this would be cool, cause right now some of the karma tests are failing in browser for develop branch (while mocha is ok)

chevdor:
I am happy to see this is already useful!
I saw that issues were reported for *some* browsers and not others and that´s exactly why I setup those tests. I will look into adding that to Travis.

PR #103, opened 2015-03-06
Fix #97 (web3.fromDecimal)

### LINKED ISSUE #97: negative hexadecimal numbers
`web3.fromDecimal(-1)` returns `0x-1`, this should be `-0x1` instead. At least that is the format supported by JavaScript, Python, Ruby, etc.

chevdor:
Mainly a fix for #97 , some formatting and a set of new tests for fromDecimal.
singleRun=false by default as it is much more convenient. Travis is specifically using --singleRun=true so it won´t hurt.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,7 +1,7 @@
 // Karma configuration
 // Generated on Thu Feb 19 2015 19:57:47 GMT+0100 (W. Europe Standard Time)
 
-module.exports = function (config) {
+module.exports = function(config) {
     config.set({
 
         // base path that will be used to resolve all patterns (eg. files, exclude)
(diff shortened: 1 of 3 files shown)


debris:
I will not accept this pull request right now, cause i didn't like the way formatting fromHex/toHex, fromDecimal/toDecimal was done before. I refactored this in ethereum/apiOverhaul branch. Tests are available [here](https://github.com/ethereum/ethereum.js/blob/apiOverhaul/test/utils.fromDecimal.js)

chevdor:
It is your pick but why not merging the pull request into master as a temporary solution (it is rather an issue for users) and refactoring when time is right in the apiOverhaul branch?

Ah and I see you are using chai now? I want to ask why not using it... I have my answer :smile:

PR #107, opened 2015-03-12
Api overhaul the second!

frozeman:
add missing formatters, cleaned up and add version properties.


frozeman:
oh i see conflicts, will solve

PR #108, opened 2015-03-12
Api overhaul missing features

frozeman:
Add versions and fixed in and outputs (e.g. dont hexify solidity)


lib/web3.js
```diff
@@ -19,10 +19,12 @@
  *   Jeffrey Wilcke <jeff@ethdev.com>
  *   Marek Kotewicz <marek@ethdev.com>
  *   Marian Oancea <marian@ethdev.com>
+ *   Fabian Vogelsteller <fabian@ethdev.com>
  *   Gav Wood <g@ethdev.com>
  * @date 2014
  */
 
+var packagejson = require('../package.json');
```
debris: please don't require package json

lib/web3.js
```diff
@@ -129,7 +135,18 @@ var setupProperties = function (obj, properties) {
...
 
         proto.enumerable = !property.newProperty;
-        Object.defineProperty(obj, property.name, proto);
+
+        if(objectProperties.length > 1) {
+            if(!obj[objectProperties[0]])
+                obj[objectProperties[0]] = {};
+
+            Object.defineProperty(obj[objectProperties[0]], objectProperties[1], proto);
+            // obj[objectProperties[0]][objectProperties[1]] = callFunction;
```
debris: don't add commented out code

lib/web3.js
```diff
@@ -99,7 +104,8 @@ var setupMethods = function (obj, methods) {
 /// setups api calls for these properties
 var setupProperties = function (obj, properties) {
     properties.forEach(function (property) {
-        var proto = {};
+        var objectProperties = property.name.split('.'),
+            proto = {};
         proto.get = function () {
```
debris: would be better to do:

```javascript
var name = property.name;
var objectProperties = name.split('.');
if (objectProperties.length > 1) {
  obj = objectProperties[0];
  name = objectProperties[1];
}
```

also please be consistent with defining variables in the rest of the code.
```javascript
var objectProperties = property.name.split('.');
var proto = {};
```

would be better

debris: or even:
```javascript
var objectProperties = name.split('.');
var name = objectProperties[1] || property.name;
var object = objectProperties.length > 1 ? objectProperties[0] : obj;
```

frozeman:
your improvement leads to:
`/Users/frozeman/Sites/meteor-dapp-wallet/app/packages/ethereum.js/lib/web3.js: line 106, col 33, This function's cyclomatic complexity is too high. (5)`

i think actually my version was better. i agree yours have 1 or so line less, but more cyclomatic complexity and doesn't work if web3.version is not already defined...

PR #109, opened 2015-03-13
Renaming hash/stringXX to bytes

LefterisJP:
We have to rename the etherum.js types to conform with the changes in Solidity.

Is needed due to https://github.com/ethereum/cpp-ethereum/pull/1284

For now I only changed the tests.

### DIFF
--- a/test/abi.inputParser.js
+++ b/test/abi.inputParser.js
@@ -227,56 +227,6 @@ describe('abi', function() {
 
         });
 
-        it('should parse input hash', function() {
-
-            // given
-            var d = clone(description);
-
-            d[0].inputs = [
-                { type: "hash" }
-            ];
-
-            // when
-            var parser = abi.inputParser(d);
-
-            // then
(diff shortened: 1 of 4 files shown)


test/abi.inputParser.js
```diff
@@ -294,13 +244,13 @@ describe('abi', function() {
...
         });
 
-        it('should parse input string', function () {
+        it('should parse input fixed bytes type', function () {
 
             // given
             var d = clone(description);
 
             d[0].inputs = [
-                { type: "string" }
```
chriseth: This seems not correct. `string` is a dynamic type while `bytes32` is a fixed type. I think either leaving the string as it is or changing it to `bytes` would be the right way.

LefterisJP: But in solidity there is no strings type, so bytes would probagbly be more correct

debris:
looks good ;) I will not merge directly, but I will base my merge on this branch, cause there are few minor things that still needs to be done ;) But anyway, thanks! :+1:You saved me a lot of time.

PR #117, opened 2015-03-16
changed bundle name web3 -> ethereum.js

debris:
long awaited fix.


kumavis:
@debris long time coming! seems it needs a rebase tho.

PR #125, opened 2015-03-16
changed back to getCode

frozeman:
add the change, which gav made in RPC and renamed `getData` back to `getCode`


jorisbontje:
Think this is already merged as part of https://github.com/ethereum/ethereum.js/pull/121

@frozeman your change is missing the `newMethod` in https://github.com/frozeman/ethereum.js/blob/apiOverhaul/lib/web3/eth.js#L116

frozeman:
Youre right. thanks!

PR #126, opened 2015-03-17
Made batch processing the default and removed getStorage()

frozeman:
This PR makes batch processing the default and mainly uses it for the filter polls. This means no matter how much filters are created only one request is made per interval cycle..


dist/ethereum.js
```diff
@@ -2063,14 +2062,19 @@ var filter = function(options, implementation, formatter) {
...
         callbacks.push(callback);
     };
 
     var stopWatching = function() {
         implementation.stopPolling(filterId);
+        callbacks = [];
+    };
+
+    var uninstall = function() {
+        implementation.stopPolling(filterId);
```
debris: what is the difference between `stopWatching` and `uninstall`?

dist/ethereum.js
```diff
@@ -2580,23 +2583,31 @@ var requestManager = function() {
...
+
+            // FORMAT BASED ON the input FORMATTER ARRAY
+            } else if(item.inputFormatter instanceof Array) {
+                item.params = Array.prototype.map.call(item.inputFormatter, function(formatter, index){
+                    // format everything besides the defaultblock, which is already formated
+                    return (!item.addDefaultblock || index+1 < item.addDefaultblock) ? formatter(item.params[index]) : item.params[index];
+                });
+            }
+
+            payload.push(jsonrpc.toPayload(item.method, item.params));
```
debris: 1. You should have used `map`, not `forEach`.
2. I don't like calling `typeof` and `instanceof` directly. We have utils methods `isArray`, `isFunction` to do that.
3. I don't like the fact, that we are checking if formatters are array or functions. You are overusing lack of types in javascript.
4. This function is called !send! Do you think that bringing formatters here is proper?

dist/ethereum.js
```diff
@@ -2604,39 +2615,65 @@ var requestManager = function() {
         }
 
         // HTTP ASYNC (only when callback is given, and it a HttpProvidor)
-        if(typeof callback === 'function' && provider.name === 'HTTP'){
+        if(typeof callback === 'function' && provider.name === 'RPC over HTTP'){
```
debris: Checking provider name and creating provider specific logic here is against all good practices. Let's assume, that we have 10 async providers. Would you create `if` for each of them?

debris: I don't know, why I accepted this before

dist/ethereum.js
```diff
@@ -2604,39 +2615,65 @@ var requestManager = function() {
...
-                        }));
-                    }
-                    return null;
+                if(!utils.isArray(data)) {
+                    callback(new Error({
+                        status: status,
+                        error: result,
+                        message: provider.name +' didn\'t repsond with an array of method responses'
+                    }));
+                    return;
```
debris: why only if it is not Array? also return is redundant

dist/ethereum.js
```diff
@@ -2604,39 +2615,65 @@ var requestManager = function() {
...
+                            console.error(item.error.message);
+                            callback(item.error);
+                        } else {
+                            callback(new Error({
+                                status: status,
+                                error: item,
+                                message: provider.name +' Bad Request'
+                            }));
+                        }
+                        return null;
```
debris: This is completely unacceptable!

1. isValidResponse already checks expected properties of the object
2. On error we should throw OR call a callback, not both depending on params!
3. 
```
if(typeof item === 'object' && item.error && item.error.message) {
```

dist/ethereum.js
```diff
@@ -2604,39 +2615,65 @@ var requestManager = function() {
...
+                    if(typeof item === 'object' && item.error && item.error.message)
+                        console.error(item.error.message);
+                    return null;
+                }
+
+                // format the output
+                return (typeof data[index].outputFormatter === 'function') ? data[index].outputFormatter(item.result) : item.result;
+            });
+
+            return (result.length === 1) ? result[0] : result;
```
debris: result.length > 1 will fail here. also stop overusing lack of types!

debris:
I wanted to merge this pull request, but I have a feeling that this is not a right thing to do. I generally like the improvements that this pr introduces, but not the way it is done. Since we started refactoring the API we are constantly decreasing the quality of the code and bringing unnecessary complexity. This commit brings another ~40 lines to `requestmanager.js` file. And the file itself just looks terrible. Just looking at it, I cannot !guess! what are the input/output param types of particular functions. Please compare [requestmanager.js v0.0.16](https://github.com/ethereum/ethereum.js/blob/v0.0.16/lib/requestmanager.js)  to [new requestmanager.js](https://github.com/frozeman/ethereum.js/blob/apiOverhaul/lib/web3/requestmanager.js). The send function itself has 97 lines!!!

Also requestmanager.js on master && develop branch needs big improvements in quality. We cannot continue creating unmaintainable code!

frozeman:
Reqgestmanager is more complex as it does ways more things than before. It basically now formats in and output parameters based on given formatters. additionally it now sends and parses batch processes by default. You might be right that we can find a cleaner way, but as it stands we are refactoring the API and need top make it work in the first place. 

If you go a better idea how to make the code shorter, you're welcome to look at it i will also take a look.

Current develop doesnt work with the new RPC API fully and has minor bugs. This PR fixes a few.
IMHO we should first make the RPC work and ethereum as well and look later for improvements. All files in ethereum.js a very modular and simplistic still, so i wouldn't worry about complexity.

dist/ethereum.js
frozeman: Stop watching just stops sending FilterChanges requests, and you can start watching agiain. uninstall uninstalls the filter completely.

frozeman: I write you all this in skype already ;)

dist/ethereum.js
frozeman: 1. agree i will take a look
2. true, but better we use undescore anyway ad we introduce more and more of thos funcitons.
3. i thinks thats actually clever, as you allow the specify one inputformatter for all parameters, or one for each single parameter. This way the developer of the method can choose what he wants. And its actually a better way of using the types here, as different types (array or funciton) mean different behaviour.
4. You would like to move the formating into 30 different places spread all over ethreuem.js?
Its called send, but actually its called Requestmanager.send(), means its a module and what the requestmanager need to do before he sends it is not important for other modules to know, so i think it belongs here. Though i agree we could split up the functionality more, inside requestmanager.

dist/ethereum.js
frozeman: The problem here is that only HTTP has currently callback style responses, all others are return values. Other providers could use callbacks later too, and we just add those to the list, but we won't need to introduce more complexity anyway.

dist/ethereum.js
frozeman: Because we are always sending batch request, so we expect arrays back at all times. If its not an array its already a wrong response. 

The return is necessary, as he would otherwise try to iterate over the non-array type. (the error before is not thrown, but passed to the callback)

dist/ethereum.js
frozeman: Yes but if its now valid, we want to present either the error we got back, or show at least ageneral erro, if the request failed for other reasons. Though i agree we could put it in an extra funciton. Its only in two places, and won't be anywhere else ever i guess.

dist/ethereum.js
frozeman: Why would it fail? If its not an array we return already before.

Issue #128, opened 2015-03-17
call() with integer gas value


ethers:
develop branch commit de5de9e built ethereum.js and got this error from line 2634 "could not decode, json: cannot unmarshal number into Go value of type string"
```
            if (!jsonrpc.isValidResponse(result)) {
                if(typeof result === 'object' && result.error && result.error.message)
line 2634                    console.error(result.error.message);
```

This was repro by call() with gas number (instead of string)
`` contract.call({from:gCoinbase, gas:500000}).foo('ab', 3); ``
(Changing gas to string '500000' leads to correct behavior)

https://github.com/ethereum/wiki/wiki/JavaScript-API#web3ethsendtransaction says "gas (integer), the amount of gas to purchase for the transaction" so call() should also be consistent and allow integers.

PR #130, opened 2015-03-17
fix call() with integer gas value #128

ethers:
or do we want the fix to be
var inputCallFormatter = inputTransactionFormatter ?


frozeman:
If call accepts the same as sendTransaction we should just use the `inputTransactionFormatter`.

Though it will add empty gas and gasprice if no gas or gasprice was specified, which should the node use the default gas/price internally.

debris:
fixed in e0d83d065ef7c88361129ee1b4ca59edd1c99a84

PR #158, opened 2015-04-13
Watch improvement 2

frozeman:
Re did the PR, with a cleaner commit history


dist/web3-light.js
```diff
@@ -2416,13 +2452,18 @@ HttpProvider.prototype.sendAsync = function (payload, callback) {
     };
 
     request.open('POST', this.host, true);
-    request.send(JSON.stringify(payload));
+
+    try {
+        request.send(JSON.stringify(payload));
+    } catch(error) {
+        throw errors.NoConnection(this.host);
```
debris: instead of throwing error, it should send an error in callback:

```javascript
callback(errors.NoConnection(this.host));
```

frozeman: you refactored that ;)

PR #164, opened 2015-04-16
Fix #162: allow empty topic list in messages

### LINKED ISSUE #162: web3.shh.filter.watch broken?
Similar to the previous report, post now worked, but when I created an empty filter and started watching it, the API crashed once again:

```javascript
var filter = web3.shh.filter({});
filter.watch(function(res) {
    console.log(res);
});
```

```
/usr/lib/node_modules/web3/lib/web3/formatters.js:195
    post.topics = post.topics.map(function(topic){
                              ^
TypeError: Cannot call method 'map' of undefined
    at outputPostFormatter [as formatter] (/usr/lib/node_modules/web3/lib/web3/formatters.js:195:31)
    at /usr/lib/node_modules/web3/lib/web3/filter.js:123:50
    at Array.map (native)
    at Object.callback (/usr/lib/node_modules/web3/lib/web3/filter.js:122:36)
    at /usr/lib/node_modules/web3/lib/web3/method.js:152:21
    at /usr/lib/node_modules/web3/lib/web3/requestmanager.js:104:9
    at request.onreadystatechange (/usr/lib/node_modules/web3/lib/web3/httpprovider.js:59:13)
    at dispatchEvent (/usr/lib/node_modules/web3/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:572:25)
    at setState (/usr/lib/node_modules/web3/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:591:14)
    at IncomingMessage.<anonymous> (/usr/lib/node_modules/web3/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:431:13)
```

karalabe:


frozeman:
The fix should actually not be required.

In the `inputPostFormatter` it will already create an topics array and even accept strings, which will then get add as the first topic.

the `outputPostFormatter` assumes that there is always an topic array coming, which the node should send.

lib/web3/formatters.js
```diff
@@ -159,6 +159,9 @@ var inputPostFormatter = function(post) {
     post.priority = utils.fromDecimal(post.priority);
 
     // fallback
+    if (post.topics == undefined) {
+        post.topics = []
+    }
     if (!utils.isArray(post.topics)) {
```
frozeman: Can you remove this? As the line below already takes care of creating and array if non is given.

lib/web3/formatters.js
```diff
@@ -192,6 +195,9 @@ var outputPostFormatter = function(post){
     }
 
     // format the following options
+    if (post.topics == undefined) {
+        post.topics = []
+    }
     post.topics = post.topics.map(function(topic){
```
frozeman: Can reformat thiss please to:

```js
    if (!post.topics) {
        post.topics = [];
    }
```

(Missing `;` and using `!` is better as the node could return us also `topics = null`)

lib/web3/formatters.js
karalabe: If I remove this check, then the sample code crashes with

```
TypeError: Cannot read property 'length' of undefined
    at toHexNative (/usr/lib/node_modules/web3/lib/utils/utils.js:121:27)
    at Object.fromAscii (/usr/lib/node_modules/web3/lib/utils/utils.js:139:15)
    at /usr/lib/node_modules/web3/lib/web3/formatters.js:168:22
    at Array.map (native)
    at inputPostFormatter (/usr/lib/node_modules/web3/lib/web3/formatters.js:167:31)
    at /usr/lib/node_modules/web3/lib/web3/method.js:86:28
    at Array.map (native)
    at Method.formatInput (/usr/lib/node_modules/web3/lib/web3/method.js:85:32)
    at Method.toPayload (/usr/lib/node_modules/web3/lib/web3/method.js:130:23)
    at Method.send (/usr/lib/node_modules/web3/lib/web3/method.js:148:24)
```

karalabe: this is the contents of post.topics: `[ undefined ]` if I don't specify any in my code.

frozeman: True, can you refactor the part like this:

```js
    // fallback
    if (!utils.isArray(post.topics)) {
        post.topics = post.topics ? [post.topics] : [];
    }

    // format the following options
    post.topics = post.topics.map(function(topic){
        return utils.fromAscii(topic);
    });
```

karalabe: Sure :)

PR #170, opened 2015-04-20
abi refactor

debris:
*changes*
* removed nested, unreadable loops
* encoding && decoding solidity params is now structured

*todo*
* refactor interface of `abi.js`


frozeman:
I looked over everything it looks good as far as i can tell.

BTW we should also change all the appearances of `ethereum.js` in the files comments to `web3.js`

debris:
haha, yes we should :+1:

frozeman:
an please fetch develop, so i can automatically merge

PR #174, opened 2015-04-20
events refactor

debris:
**changes**
* `event.js` && `contract.js` refactored
* new file `function.js`
* indexed params are now null if not specified


lib/utils/utils.js
```diff
@@ -143,6 +143,22 @@ var fromAscii = function(str, pad) {
...
+ */
+var transformToFullName = function (json) {
+    if (json.name.indexOf('(') !== -1) {
+        return json.name;
+    }
+
+    var typeName = json.inputs.map(function(i){return i.type; }).join();
+    return json.name + '(' + typeName + ')';
+};
+
```
frozeman: I would probably call this `abi` instead of `json` for clarity.

lib/web3/contract.js
```diff
@@ -166,40 +74,37 @@ var contract = function (abi) {
...
-    addEventsToContract(result, abi, address);
+    addFunctionsToContract(this, abi);
+    addEventsToContract(this, abi, this.address);
+};
+
+Contract.prototype.call = function (options) {
+    this._isTransaction = false;
+    this._options = options;
+    return this;
+};
```
frozeman: Is there a way that we can structure this so that we call:
`myCotract.myMethod().sendTransaction()` it looks more natural to me and looks more logical structured.

debris: not really. Unless we want to force people to always use `call()` and `sendTransaction()` explicitly

We can also do
```javascript
myContract.myMethod.sendTransaction()
```

but it also looks ugly.

lib/web3/event.js
```diff
@@ -14,125 +14,160 @@
...
+    options.topics.push('0x' + this.signature());
+
+    var indexedTopics = this._params.filter(function (i) {
+        return i.indexed === true;
+    }).map(function (i) {
+        var value = indexed[i.name];
+        if (value !== undefined) {
+            return '0x' + coder.encodeParam(i.type, value);
+        } 
+        return null;
```
frozeman: Not sure if i fully understand that, but you first filter, means you remove all non given index parameters. And the  you return either the value, or null.

But when you filter them first, you would never return null right?
and `i` in the filter and map callback is the value here, not the index, so i probably would rename it, because i is commonly used for the index, which would be the second parameter.

lib/web3/function.js
```diff
@@ -0,0 +1,134 @@
...
+
+var web3 = require('../web3');
+var coder = require('../solidity/coder');
+var utils = require('../utils/utils');
+
+/**
+ * This prototype should be used to call/sendTransaction to solidity functions
+ */
+var SolidityFunction = function (json) {
+    this._inputTypes = json.inputs.map(function (i) {
```
frozeman: Why solidity specific functions in the web3 folder?

lib/web3/event.js
debris: > Not sure if i fully understand that, but you first filter, means you remove all non given index parameters. And the you return either the value, or null. But when you filter them first, you would never return null right?

Exactly. Cause I don't want to create topic for non indexed param.

lib/web3/function.js
debris: it was always here, but not as separate object prototype. I would be happy to move it to solidity folder, but I won't do it when it depends on web3 (sha3 encoding) :p

debris: Maybe in future we will add `crypto-js` to our project... and fix this :p

lib/web3/contract.js
frozeman: I actually prefer that, looks clean to me, as the current still looks odd :)

lib/web3/event.js
frozeman: but you need then to fill that space in the send topics los with `null`, as the indexed params are ordered.

frozeman:
Please see my answers on you comments. 

Additionally we need to fix the events output to match the docs in event.js line 120 from:

```js
    var result = {
        event: this.displayName(),
        number: data.number,
        hash: data.hash,
        args: {}
    };
```

to:

```js
    var result = {
        event: this.displayName(),
        args: {},
        logIndex: data.logIndex,
        transactionIndex: data.transactionIndex,
        transactionHash: data.transactionHash,
        address: data.address,
        blockHash: data.blockHash,
        blockNumber: data.blockNumber
    };
```

And probably run `data` first through the log output formatter.

lib/web3/event.js
debris: I'm filling the space with nulls, but only for indexed params. Not indexed params are just ignored.

eg, the event params are:
`indexed int a`, `indexed int b`, `int c`, `indexed int d`
and we want to filter `b` matching `5`, we will create a topics array like this:

```javascript 
[signature, null, 0x0...05, null]
// signarure, a = null, b = 5, d = null
```

debris:
fields are fixed in latest pr

debris:
now I will do
```javascript
myContract.myFunction.sendTransaction({value: 20});
// and
myContract.myFunction.call();
```

debris:
once this pr is merged we will upgrade version to `0.3.0`

frozeman:
The contract tests are still the same showing e.g.:

`contract.call({from: address, gas: 50000}).balance(address);`

This shouldn't work anymore, or do we have backwards compatibility?

frozeman:
events.js:120 is still the same, and has not the necessary properties.

debris:
> Can we add contract tests which also use the ..myMethod.sendTransaction() and ..myMethod.call() syntax?

done

PR #178, opened 2015-04-22
String topics

frozeman:
Refactored input topics filter, to format everything as to HEX strings except `null`


karalabe:
Seems to be working ok for me :)

debris:
:+1:

PR #190, opened 2015-04-30
Use async contract calls when a callback is passed

niran:


niran:
I put this together when I thought that synchronous contract calls were making my app slow, which was incorrect. This might not be worth pursuing in the near term, but at the very least, the documentation for `Contract.call` should mention that it should be so fast on localhost that you shouldn't really need an async version.

niran:
Never mind, the synchronous calls are actually making the app slow. What do I need to do to get something like this merged?

debris:
This looks good and I like the idea of having asynchronous calls, but unfortunately it interferes with `toPayload` method in `SolidityFunction.prototype` (see [transactionObject](https://github.com/ethereum/wiki/wiki/JavaScript-API#contract-methods) in docs). We need a better way to solve passing arguments,callback and 'transactionObject' to solidity functions.

Suggestions are welcomed. :+1:

frozeman:
Please direct your PR into develop no master. @debris can you take a look?

debris:
I'm closing, cause it's gonna be a part of #199. Thanks @niran !

PR #191, opened 2015-04-30
Handle asynchronous errors properly (fixes a TODO)

tcoulter:
Not handling errors properly could cause applications to stop functioning due to a runtime exception. For instance, if for some reason the application cannot connect to the RPC client, JSON won't be able to parse the result (an empty string), which would result in a RuntimeException.


lib/web3/httpprovider.js
```diff
@@ -55,8 +55,16 @@ HttpProvider.prototype.sendAsync = function (payload, callback) {
     var request = new XMLHttpRequest();
     request.onreadystatechange = function() {
         if (request.readyState === 4) {
-            // TODO: handle the error properly here!!!
-            callback(null, JSON.parse(request.responseText));
+            var result = request.responseText;
+            var error = null;
+
+            try {
+                result = JSON.parse(result);
```
frozeman: This should also be add to the synchronous `send` function. (modified to work for sync and the errors should be thrown)

tcoulter:
@frozeman How's this?

frozeman:
can you pull develop first, there seem some conflicts. Also can you please run `npm test` to see if all test pass still?

frozeman:
Ah and you PR should go into develop not master

debris:
merged into develop. Thanks!

PR #193, opened 2015-05-07
add newblockfilter Proposal

frozeman:
Add proposal changes from https://github.com/ethereum/wiki/wiki/newBlockFilter-Improvement-Proposal#new-proposal

to `eth_newBlockFilter` and add `eth_newPendingTransactionFilter`


frozeman:
Its fully implemented in GO and defined in the docs. though if cpp doesn't have it its a breaking change for CPP. We need to merge that soonish (nowish) ;)

PR #195, opened 2015-05-08
fixed #194

### LINKED ISSUE #194: web3.js ABI is not following wiki spec when encoding sizes of dynamic parameters
Found via https://github.com/ethereum/go-ethereum/issues/871

https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI sam example was updated (all dynamic sizes are at the start, and then all contents are in the order they appear)

debris:
fixed web3.js encoding
https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI


debris:
@ethers please verify tests :)

ethers:
@debris tests look good :)  merge it :+1:   (I will also try further when Serpent is also updated)

PR #199, opened 2015-05-11
web3.eth.contract overhaul

debris:
this pr contains new way of creating contract objects:

- `param1`, `param2` ... `paramX` - contract constructor params (optional)
- `{code: 0x00...df}` - transaction object (required for method `new`)
- `function (err, contract) {}` - callback object (optional)

```javascript
var MyContract = web3.eth.contract(abi);

// create new contract on a blockchain
var contract = MyContract.new({code: 0x00...df});
var contract = MyContract.new(param1, param2, {code: 0x00...df});
MyContract.new({code: 0x00...df}, function (err, contract) {});
MyContract.new(param1, param2, {code: 0x00...df}, function (err, contract) {});

// reuse existing contract
var contract = MyContract.at(address);
MyContract.at(address, function (err, contract) {})
```

there is also possibility to call && sendTransaction to contract asynchronously:

```javascript
// myCall is a contract const function
var result = contract.myCall([3]);
var result = contract.myCall([3], {from: 0x...ff});
contract.myCall([3], function (err, result) {});
contract.myCall([3], {from: 0x...ff}, function (err, result) {});

// myTransact is a contract function
contract.myTransact([3]);
contract.myTransact([3], {from: 0x...ff});
contract.myTransact([3], function (err) {});
contract.myTransact([3], {from: 0x...ff}, function (err) {});
```


niran:
This is slightly tangential, but I think the API for calling contracts needs to be focused on the batch call scenario. Most non-trivial apps will be doing several function calls at a time, which can lead to a combinatorial explosion of calls when dealing with lists of objects that each have related lists of objects.

I ended up adding async contract calls to Augur, but we're forced to minimize the concurrency to avoid choking geth at the network level with too many open sockets. We'll be able to ask for data sooner once we have a batch API available, though we've started using direct RPC calls instead of web3. I'm hoping that the lessons we learn will get rolled into web3 so all dapp builders can benefit, and so we can move back to web3 someday.

debris:
I understand. And what do you think about following syntax for batch calls?

```javascript
contract.batch([
{name: myCall, params: [], options: {}, callback: cb},
{name: myCall1, params: [], options: {}, callback: cb},
{name: myCall2, params: [], options: {}, callback: cb},
{name: myTransact, params: [], options: {}, callback: cb}
]);
```


or even something more general:

```javascript
var batch = web3.newBatch();
batch.add(contract.myCall.request(param1, {from: 0x...ff}, function (err, result) {}));
batch.add(contract.myTransact.request(param1, param2, {from: 0x...ff}, function (err, result) {}));
batch.add(web3.eth.getBalance.request(0x000..1f, function (err, result) {})
batch.add({jsonrpc: '2.0', method: 'eth_superMethod', params: 0x1231, callback: function (err, result) {} });
batch.execute()

```

frozeman:
Please re-open with merge to develop not master

niran:
Something like the second example is what I think would work best. I'm assuming that people are only going to want to batch calls, not transactions, so `contract.request` to generate the equivalent of a call to batch sounds good to me.

PR #200, opened 2015-05-11
Contract overhaul

debris:


dist/web3-light.js
```diff
@@ -1588,65 +1528,106 @@ var addEventsToContract = function (contract, desc) {
...
+    }
+
+    // throw an error if there are no options
+
+    var bytes = encodeConstructorParams(this.abi, args);
+    options.data += bytes;
+
+    if (!callback) {
+        var address = web3.eth.sendTransaction(options);
+        return this.at(address, callback);
```
frozeman: Wouldn't be `return this.at(address);` efficient already?

dist/web3-light.js
```diff
@@ -2641,19 +2627,41 @@ SolidityFunction.prototype.signature = function () {
...
-    var output = web3.eth.call(payload);
-    output = output.length >= 2 ? output.slice(2) : output;
-    var result = coder.decodeParams(this._outputTypes, output);
-    return result.length === 1 ? result[0] : result;
+    var args = Array.prototype.slice.call(arguments);
+    var callback = this.extractCallback(args);
+    var payload = this.toPayload(args);
+
+    if (!callback) {
+        var output = web3.eth.call(payload, callback);
```
frozeman: Same here what not `web3.eth.call(payload);` if callback is undefined anyway?

debris: sure :+1:

dist/web3-light.js
debris: :+1:

PR #202, opened 2015-05-12
async properties

debris:
added: 
- async way of getting properties

removed:
- property setter (it was unused)


dist/web3.js
```diff
@@ -3211,15 +3218,20 @@ Property.prototype.get = function () {
...
- * @method set
- * @param {Object} new value of the property
+ * @method asyncGet
+ * @param {Function}
  */
-Property.prototype.set = function (value) {
-    return RequestManager.getInstance().send({
-        method: this.setter,
-        params: [this.formatInput(value)]
+Property.prototype.asyncGet = function (callback) {
```
frozeman: Though i would habe called it `getAsync`

debris: fixed

frozeman:
I really like the changes, but i makes myself question the whole use of "properties". Why we don't plainly use:

```js
// sync
var res = web3.eth.blockNumber();

// async
web3.eth.blockNumber(function(e,res) { ... })
```

You can still use the properties synchronously, but you need to add a `()`. It at least makes everything more unified. Also in terms of the docs.

debris:
but we will break another thing then :) I would prefer to do it later, after merging this pr and releasing new version :)

frozeman:
ok. lets do it later then. it will anyway be very much recommended to use only async, except you know what you're doing.
so we can later get "rid" of the pure property ones..

PR #205, opened 2015-05-16
sha3 in browser

debris:
- this will fix sending call && sendTransaction in batch


kumavis:
@debris is this correct? seems like you'll need to pluck off any '0x' prefix on the input and add it back on the result. Also, what kind of input is it expecting? hex string, big num, buffer?

debris:
@kumavis, Yes, it's correct. It expects plain string and the output is a string without `0x` prefix. For backwards compatibility we also can provide support for hex form of the string as an input. Although I don't think it's really necessary (We still haven't released).

**was**

```javascript
var x = web3.sha3(web3.fromAscii('helloWorld'));
console.log(x); // 0xfff...
```

**is**

```javascript
web3.sha3('helloWorld');
console.log(x); // fff...
```

kumavis:
@debris i mean that `web3.fromAscii('helloWorld') // 0xfff` has the hex prefix, which will be part of the content that gets hashed

kumavis:
ah sorry reading again I got it

kumavis:
@debris even though you "haven't released" many developers including myself do rely on web3. if you can keep at least a primitive migration notes, that would be much appreciated.

debris:
@kumavis, before merging to master I will prepare migration notes.

kumavis:
yaay :joy_cat:

PR #207, opened 2015-05-18
namereg and icap

debris:


frozeman:
@debris why `web3.eth.icapTransfer(iban, from, value);`

I would prefer "from", "to", "value", to make it more intuitive. And maybe we can call it `sendIBANTransaction`, or `sendICAPTransaction ` to make it more similar to sendTransaction?

frozeman:
I though a bit more about it and i think having a separate `sendIBANTransaction` is not a good idea, but instead it should be part of the default `sendTransaction`.

This means you can pass not only and address but also a IBAN to the `{to: ...}`. This allows also to use the IBAN to send through a contract and send with data.

The inputFormatter of the sendTransaction could simply replace the IBAN for the address, but we need to make sure that when people use `sendTransaction` async, that we call also async for the IBAN.

debris:
>I though a bit more about it and i think having a separate sendIBANTransaction is not a good idea, but instead it should be part of the default sendTransaction.

I don't agree. We should not merge two completely different functions into one. This is misleading and error prone.

frozeman:
And how would i then use IBAN in a contract method? Or send data with it?

Allowing IBAN in a `to` is IMHO just making it optional to use either addresses OR IBAN.

debris:
You **can't** use IBAN in contract. That's not a purpose of it. Also nor cpp neither go implementation will allow to use `IBAN` in `to`. It's part of higher level interface, whereas `sendTransaction` is more `low-level`.

`IBAN` was created to unify **money** transfers between different institutions. You send nothing besides money.

Also in decent languages you don't create method which accepts all different types of input params. Javascript allows it, but it doesn't mean, that it is a good way of solving problems.

frozeman:
I think you misunderstand me here. With allowing to pass IBAN into the `to` field i mean that we in web3.js transform it into an address. The nodes never deal with the IBAN.

But this way we can easier convince people to use the IBAN. 

Having an simple `sendIBAN..` method won't do that, as most money won't be coming from simple accounts, but instead contracts (See gavin's wallet2.sol, which i am building the wallet with)

If we don't add this auto convert feature into web3.js, i don't see the reason why adding it at all (e.g. let developers themselves retrieve the address for an IBAN using the deployed contract and simple contract methods). There are enough IBAN.js libraries out there for validation.

I hope you see my point now.

So in short:

I only suggested to add the "get address from given iban" to the inputTransactionFormatter function and send addresses to the nodes. Be it in contract methods, or simple `eth.sendTransaction`

debris:
I don't agree with that. One function should always do one thing and only this thing. That is:

*sendTransaction*
- sendsTransaction from a to b appropriate gas, value, data... (1 API call)

*sendIBANTransaction*
- get IBAN institution name for indirect transaction (1st API call)
- send direct or indirect IBAN transaction (2nd API call)

frozeman:
Ok could you please give me an example of sendIBANTransaction with data?

debris:
you can't sendIBANTransaction with data.

lib/web3/transfer.js
```diff
@@ -0,0 +1,94 @@
...
+ * Should be used to deposit funds in ClientReceipt contract
+ *
+ * @method deposit
+ * @param {String} address
+ * @param {String} from (address)
+ * @param {Value} value to be tranfered
+ * @param {String} client unique identifier
+ * @param {Function} callback, callback
+ */
+var deposit = function (from, address, value, client, callback) {
```
frozeman: Could you explain me what this function does?

debris: I updated function description in latest commit. It's used to deposit funds to generic exchange contract (*which implements deposit(bytes32) function*).

PR #212, opened 2015-05-20
add estimateGas to contract methods and fixed sendTransaction return value

frozeman:
This PR adds the estimateGas function to contract methods like `myContract.myMethod.estimateGas()`.

Additional it fixes the sendTransaction return value, which currently didn't return the tx hash when used like `myContract.myMethod.sendTransaction()`


debris:
looks good :+1: Please fix travis build before merging :) There is small typo (`;`)

PR #224, opened 2015-05-28
Prefix name reorg

alexvandesande:
Every other week someone comes to us and ask why we don’t stick to SI units. Sure, no one remembers which is supposed to be babbage, but what’s the deal with Kwei with a uppercase K, kilo doesn’t start with uppercase! Why are unit names written in full but prefixes just the letter? It’s not Mbyte or mliter, it’s either the Megabyte or the ml. And grand, really? It’s not a SI unit it’s a number.

I always say that unit names are just suggestions and we should only use ether and wei, but in practice that doesn’t work.

Most things are in ether. Gas prices are usually in finney or szabo. Basic functions are in wei.

Then it dawns on me that the only that actually matter are ether, finney and szabo, and they have a direct 1000 relationship. Here’s a nice mnemonic: they’re the micro and the mili. A Finney is a mili-ether, szabo is the micro-ether. Ether, mili, micro. Now I usually don’t memorize the micro but since I’m having to remember the szabo anyway, why not the micro-?

So I did some cosmetic changes, made the comments clearer on this, and added support for SI prefixes femto-, pico-, nano-, micro- and mili-

Just giving developers more choice, since so many of them asked for this


kumavis:
I need 200 mi-croethers, stat!

alexvandesande:
a szabo is a micro-ether. A thousand szabo make a finney, which is a milliether. Which one you prefer to use is a matter of personal preference which could be left to the final user.

Issue #219, opened 2015-05-25
fromAscii on unicode strings


chriseth:
fromAscii in lib/utils/utils.js should either reject non-ascii unicode strings or hex-encode them in e.g. utf8 (and then be renamed).
The string `"xǺ"` (= `"x\u01fa"`) is encoded to `"0x781fa"`, i.e. the `Ǻ` is not correctly padded to an even number of hex characters.

Issue #218, opened 2015-05-25
toAscii stops at zero byte


chriseth:
The `toAscii` function in https://github.com/ethereum/web3.js/blob/25d9f5121735b76481b2518fe49ae285b75bac4d/lib/utils/utils.js#L86 stops once it encounters a zero byte.

Since javascript strings do support zero bytes and the current decoding throws away data, is this really justified?

PR #228, opened 2015-06-01
abi string type, fixes #216, #218, #219

### LINKED ISSUE #216: Don't auto transform bytes(X) to string and introduce string as a new type for solidity contracts
Currently bytes sometimes get decoded wrong, which make it impossible to use bytes as non-strings, because you can't reliable encode them back to bytes.

```js
web3.toAscii('731a3afc00d1b1e3461b955e53fc866dcf303b3eb9f4c16f89e388930f48134b')
"s:ü"

web3.toAscii('731a3afc')
"s:ü"
```

Instead we need to re-introduce the `string` type in solidity and only decode this one to strings and keep bytes as is: https://www.pivotaltracker.com/n/projects/1189488/stories/95173586

debris:
This pr brings back `string` type for solidity.

@ethers @chriseth @frozeman can you review tests?


chriseth:
Would be great to have tests for actual unicode / utf8.

chriseth:
Would be great to have tests for en- and decoding the following:
 - correctly encoded utf-8 string containing non-ascii characters
 - incorrectly encoded utf-8 string terminating prematurely (e.g. "\xc3")
 - utf-8 strings containing internal zeros (e.g. "\xc3\xa4\x00\x00\xc3\xa4")

LefterisJP:
May I suggest [this](http://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt) UTF8 encoding/decoding stress testing collection? I use it in [one](https://github.com/LefterisJP/refuclib/blob/master/test/test_utils_unicode.c) of my projects.

frozeman:
@debris please merge this PR asp, as solidity has now string and array support and the wallet dapp depends on that.

Additionally we should give a note to the community that auto conversion of bytes was removed.

frozeman:
@debris URF8 encoding seems to work fine, but decoding seems broken:
```js
"\xc3\xa4\x00\x00\xc3\xa4"
"Ã¤Ã¤"

web3.fromAscii("\xc3\xa4\x00\x00\xc3\xa4")
"0xc3a40000c3a4"

web3.toAscii('0xc3a40000c3a4')
"Ã¤"
```

frozeman:
@debris ok changed, could please also take a look and as well at the encoding issue i mentioned above above:
```js
> "\xc3\xa4\x00\x00\xc3\xa4"
"Ã¤Ã¤"

> web3.fromAscii("\xc3\xa4\x00\x00\xc3\xa4")
"0xc3a40000c3a4"

> web3.toAscii('0xc3a40000c3a4')
"Ã¤"
```

debris:
>@debris ok changed, could please also take a look and as well at the encoding issue i mentioned above above:
```js
> "\xc3\xa4\x00\x00\xc3\xa4"
"Ã¤Ã¤"

> web3.fromAscii("\xc3\xa4\x00\x00\xc3\xa4")
"0xc3a40000c3a4"

> web3.toAscii('0xc3a40000c3a4')
"Ã¤"
```


decoding solidity strings does not use `web3.toAscii` internally. I added few tests in 93c8006537d0529ff3603aca6df93ded74c4c049 .

**Let's merge!**

frozeman:
Ok but then lets actually rename `toAscii` to `toString`, as its no converting to ascii but latin something..

PR #229, opened 2015-06-04
add extend method and tests

frozeman:
adding an extend method, which allows to extend web3.js with additional RPC endpoints:

```js
web3.extend({
  property: 'admin',
  methods: [new web3.extend.Method({
       name: 'getBalance',
       call: 'eth_getBalance',
       params: 2,
       inputFormatter: [web3.extend.utils.toAddress, web3.extend.formatters.inputDefaultBlockNumberFormatter],
       outputFormatter: web3.extend.formatters.outputBigNumberFormatter
  })],
  properties: [new web3.extend.Property({
        name: 'gasPrice',
        getter: 'eth_gasPrice',
        outputFormatter: web3.extend.formatters.outputBigNumberFormatter
  })]
})

// Will give you
web3.admin.getBalance(address, cb);
web3.admin.gasPrice;
web3.admin.getGasPrice(cb);
```

You can even leave the `property: 'admin'` property out to add functionality to the web3 object itself.


debris:
As I told you, I would prefer it to be more modular, but since it's not a lot of code let's leave it as you propose :+1: Also I would prefer to call it `_extend`, cause method, property, utils and formatters are a part of private api and I would not like people to use them, cause they may change in future.

stvenyin:
OK，See..

frozeman:
Just as reference in the #1.0 version extend is exposed as follows: https://web3js.readthedocs.io/en/1.0/web3.html#extend

Issue #213, opened 2015-05-20
removing filter, seems to break polling


frozeman:
After calling `MyFilter.stopWatching()` i get:

```js
Uncaught TypeError: Cannot read property 'callback' of undefined    web3.js:3689
```

pointing to:
```js
RequestManager.prototype.poll = function () {
    this.timeout = setTimeout(this.poll.bind(this), c.ETH_POLLING_TIMEOUT);

    if (!this.polls.length) {
        return;
    }

    if (!this.provider) {
        console.error(errors.InvalidProvider());
        return;
    }

    var payload = Jsonrpc.getInstance().toBatchPayload(this.polls.map(function (data) {
        return data.data;
    }));

    var self = this;
    this.provider.sendAsync(payload, function (error, results) {
        // TODO: console log?
        if (error) {
            return;
        }
            
        if (!utils.isArray(results)) {
            throw errors.InvalidResponse(results);
        }

        results.map(function (result, index) {
            result.callback = self.polls[index].callback; // <- ERROR is here
            return result;
        }).filter(function (result) {
            var valid = Jsonrpc.getInstance().isValidResponse(result);
            if (!valid) {
                result.callback(errors.InvalidResponse(result));
            }
            return valid;
        }).filter(function (result) {
            return utils.isArray(result.result) && result.result.length > 0;
        }).forEach(function (result) {
            result.callback(null, result.result);
        });
    });
};
```

It seems like the results returned still have the same index, which does fit with the callback index anymore.
A quick fix would be to check `if(self.polls[index])`, but we probably should better check if the poll is not firing the non-existing filter anymore, as i got 101 errors of the above type.

debris:
:+1: I will fix it.

Issue #159, opened 2015-04-13
Impossible to pass default block when calling contract function


Georgi87:
I can pass the default block to the call function like described in the documentation:

web3.eth.call(callObject [, defaultBlock] [, callback]);

However if I call a function on my contract I dont see any possibility to pass the defaultblock:

contract.call().function_name();

frozeman:
Shouldn't you be able to call then:

`contract.call({} [, defaultBlock] [, callback]).function_name();`
?

Btw we are restructuring the way to call contract method.

frozeman:
We restructured it, but you still cant pass the defaultBlock or callback, we will fix it see https://github.com/ethereum/web3.js/issues/192

Georgi87:
Thanks! I saw you changes and noticed it didn't have any effect, when I passed the "pending" block.

PR #230, opened 2015-06-05
asyncified filters fixes #220 #213 #159

### LINKED ISSUE #220: Make all communication async.
I just adapted my application to work asynchronously using callbacks. The DAPP is way more responsive now. But it still lags sometimes. It could be due to the sha3 calls that are made for every request to get function signatures.

frozeman:
Made all internal requests async to stop interface lagging.


lib/web3/filter.js
```diff
@@ -83,51 +84,69 @@ var Filter = function (options, methods, formatter) {
...
+    var intervalId = setInterval(function(){
 
-        messages.forEach(function (message) {
-            message = self.formatter ? self.formatter(message) : message;
-            self.callbacks.forEach(function (callback) {
-                callback(null, message);
-            });
-        });
-    };
+        if(self.filterId || self.filterError)
```
debris: won't work if the filterId === 0

lib/web3/filter.js
```diff
@@ -83,51 +84,69 @@ var Filter = function (options, methods, formatter) {
...
+            clearInterval(intervalId);
 
-    // call getFilterLogs on start
-    if (!utils.isString(this.options)) {
-        this.get(function (err, messages) {
-            // don't send all the responses to all the watches again... just to this one
-            if (err) {
-                callback(err);
+        if(!self.filterId)
```
debris: won't work if the filterId === 0

lib/web3/filter.js
```diff
@@ -83,51 +84,69 @@ var Filter = function (options, methods, formatter) {
...
+            return;
+
+        self.callbacks.push(callback);
```
debris: why callback is pushed after 3 seconds?

lib/web3/function.js
```diff
@@ -46,6 +47,12 @@ SolidityFunction.prototype.extractCallback = function (args) {
     }
 };
 
+SolidityFunction.prototype.extractDefaultBlock = function (args) {
```
debris: this is different functionality, please separate it to another pr

lib/web3/requestmanager.js
```diff
@@ -226,9 +226,16 @@ RequestManager.prototype.poll = function () {
         }
 
         results.map(function (result, index) {
-            result.callback = self.polls[index].callback;
-            return result;
+            // make sure the filter is still installed after arrival of the request
+            if(self.polls[index]) {
```
debris: this is not a solution. `self.polls` is an Array and you may still call callback for different filter.

test/contract.js
```diff
@@ -218,6 +241,31 @@ describe('web3.eth.contract', function () {
...
+                    data: '0x' + sha3(signature).slice(0, 8) + '0000000000000000000000001234567890123456789012345678901234567890',
+                    to: address,
+                    from: address,
+                    gas: '0xc350'
+                }, '0xb']);
+            });
+
+            var contract = web3.eth.contract(desc).at(address);
+
+            var r = contract.balance.call(address, {from: address, gas: 50000}, 11);
```
debris: Looking at the code above, async `calls` won't work. Please fix this and make a test. Also remember to move it to separate pr :)

lib/web3/requestmanager.js
```diff
@@ -226,9 +226,16 @@ RequestManager.prototype.poll = function () {
...
+                result.callback = self.polls[index].callback;
+                return result;
+            } else
+                return false;
         }).filter(function (result) {
+            if(!result)
+                return false;
```
debris: also, it's usually better and cleaner to make a separate filter for this kind of statements

test/contract.js
debris: update*

ok, i see that it will work, but 

```javascript
contract.balance.call(address, {from: address, gas: 50000}, 11, function (result) { } )
contract.balance(address, {from: address, gas: 50000}, 11, function (result) { } )
```

juz doesn't look nice :/

what do you think about?

```javascript
contract.atBlock(11).balance.call(address, {from: address, gas: 50000}, function (result) { } )
contract.atBlock(11).balance(address, {from: address, gas: 50000}, function (result) { } )
```

although it's still not perfect

lib/web3/filter.js
```diff
@@ -83,51 +84,69 @@ var Filter = function (options, methods, formatter) {
...
     var self = this;
 
-    var onMessage = function (error, messages) {
-        if (error) {
-            return self.callbacks.forEach(function (callback) {
-                callback(error);
-            });
-        }
+    // check inf an interval of 10ms if the filter id has arrived
+    var intervalId = setInterval(function(){
```
debris: please make sure, that `setInterval` is available in go environment. I expect it to be missing.

lib/web3/filter.js
```diff
@@ -83,51 +84,69 @@ var Filter = function (options, methods, formatter) {
...
 
-        messages.forEach(function (message) {
-            message = self.formatter ? self.formatter(message) : message;
-            self.callbacks.forEach(function (callback) {
-                callback(null, message);
-            });
-        });
-    };
+        if(self.filterId || self.filterError)
+            clearInterval(intervalId);
```
debris: also `clearInterval` may be missing in go jsenv.

test/contract.js
frozeman: I would rather keep the above as its in-line with normal `call(options [, defaultBlock] [, callback])` calls. It does look perfect, but its guessable.

lib/web3/filter.js
frozeman: setInterval as setTimeout was implement in otto, otherwise out filters wouldn't work anyway.

lib/web3/requestmanager.js
```diff
@@ -220,14 +229,21 @@ RequestManager.prototype.poll = function () {
...
-            return result;
+            var key = pollsKeys[index];
+            // make sure the filter is still installed after arrival of the request
+            if(self.polls[key]) {
+                result.callback = self.polls[key].callback;
+                return result;
+            } else
+                return false;
+        }).filter(function (result) {
+            return (!result) ? false : true;
```
debris: minor, could be also:
```javascript
return !!result;
```

debris:
looks much better now! :+1: :+1:

PR #233, opened 2015-06-08
Set requests content type to application/json

asinyagin:
Hey,

I think web3.js should set requests' content type to application/json. I need it for my Ethereum sandbox. The framework I use doesn't parse body of requests with content type text/plain (https://github.com/cloud9ide/frontdoor).

I haven't added tests for content type checking, because I didn't see any easy way to do it. Do you have any suggestions?


debris:
:+1:

PR #239, opened 2015-06-17
add currentProvidor and mist fixes

frozeman:
I add the `web3.currentProvidor` property, so that dapps can check if a  provider is already set (e.g. by mist).

Electron, which has node integration in the render process, has `require`, but no `process` global variable available. Therefore packages like `xmlhttprequest` think they are in an node environment and will fail to load. The fix using:

```js
var XMLHttpRequest = (typeof window !== 'undefined' && window.XMLHttpRequest) ? window.XMLHttpRequest : require('xmlhttprequest').XMLHttpRequest; // jshint ignore:line
```

Is the most viable option, as it checks whether the `window` object and `window.XMLHttpRequest` is available and prevents requiring an unnecessary module (which would fail).


debris:
Can you also remove [browser-xhr.js](https://github.com/ethereum/web3.js/blob/master/lib/utils/browser-xhr.js) and this line from [package.json](https://github.com/ethereum/web3.js/blob/master/package.json#L15-L17). They are no longer necessary.

Also, can you give me an example when you need to check if provider is already set? I'm just curious.

frozeman:
Can do.

Ifs necessary e.g. in mist.

Mist creates the web3 objects and sets the provider (and in the future we even might provide a different provider over IPC),

If you as a dapp developer want to make your dapp work in mist and somewhere else, you should check before if you need to set a provider, or if its already set.

The same applies for the meteor package, it won't be used when web3 is already defined, this way dapp developer can keep it in their codebase, but know that when in mist, the mist version is used.

```js
// set providor
if(!web3.currentProvidor)
    web3.setProvider(new web3.providers.HttpProvider("http://localhost:8545"));
```

debris:
great! :+1:

PR #240, opened 2015-06-18
unit names per White Paper

ethers:
A very early version of the White Paper (Dec 10 2014) already clarified / simplified unit names.
https://github.com/ethereum/wiki/wiki/White-Paper/502e0e01355a42c8e6f5463c49cce496ae423179


frozeman:
These unit names are just possibilities you can use when converting wei. They are not returned a names itself anywhere, so having more doesn't hurt.

@alexvandesande you want to give your opinion on that, as you add even more :)

alexvandesande:
Having support for unit names is very different than suggesting them in the user interface. While I strongly advise all Ðapp programmers to only use ether and at most one other, we should support all of them on the web3 object.

Actually, the reason I added milliethers and microethers was a push to slowly move away from personality names and use standard SI units. Effigies are to be considered nicknames: finney can be treated as a nickname of the milliether, like an american might call the 5 dollar bill a "Lincoln".

I am in favor of moving away from all these units, but I'm against this pull request.

ethers:
SI units seem fine as they are standard and self-explanatory.  The reason for removing some of the effigies even though we don't show them in the UI, is others may build UIs with effigies that are only mentioned in this code, and not part of any other Ethereum literature.

@alexvandesande What do you mean by "I am in favor of moving away from all these units" ?  Remove all effigies?  Or remove the SI units too?

alexvandesande:
@ethers I am in favor of discouraging Ðapp developers from using other
units that are not Ether. Finney (milliether) may or may not be a useful
unit when discussing micropayments or gas prices, let's see if the
community adopts it. Wei is useful for debating protocol issues.

Having said that, I'm against this Pull Request. We should support it on
the web3.js object, but discourage it's use.

On Fri, Jun 19, 2015 at 5:00 AM, ethers <notifications@github.com> wrote:

> SI units seem fine as they are standard and self-explanatory. The reason
> for removing some of the effigies even though we don't show them in the UI,
> is others may build UIs with effigies that are only mentioned in this code,
> and not part of any other Ethereum literature.
>
> @alexvandesande <https://github.com/alexvandesande> What do you mean by
> "I am in favor of moving away from all these units" ? Remove all effigies?
> Or remove the SI units too?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ethereum/web3.js/pull/240#issuecomment-113420587>.
>



-- 

Alex Van de Sande
UX Designer

ethers:
I think I agree about discouraging units that are not Ether.  One relatively visible place we could start with is Alethzero, replace szabo with microether and finney with milliether.

To recap, this PR does keep/support finney and szabo in the web3.js object.  The PR simply removes ada, babbage, shannon, einstein (the first 3 were removed promptly in the 2nd version of Vitalik's whitepaper; I don't recall where einstein came from).  I don't see much good reason to "support" nicknames that are not referenced anywhere, no one is probably going to use, makes web3.js longer and slightly more difficult to understand.

frozeman:
I actually agree

frozeman:
Thanks for your PR, but i think we keep the names for compatibility reasons. Many use them already.

ethers:
https://github.com/ethereum/EIPs/issues/33

ethers:
Revisiting in ethereum/EIPs#33

frozeman:
@alexvandesande looks fine to me, @ethers can you resolve the merge conflicts? rebase on develop

alexvandesande:
I've solved the conflicts and added that as this PR: https://github.com/ethereum/web3.js/pull/399

frozeman:
@alexvandesande this is now in right? so we can close this.

alexvandesande:
Github automatically closes PRs when it detects commits have gone in elsewhere

PR #241, opened 2015-06-19
Added sendRawTransaction RPC

SilentCicero:
sendRawTransaction RPC support for web3.js. This is a relatively new RPC cmd that has only been merged in develop go-ethereum. This should provide basic coverage for 'eth_sendRawTransaction'. Please let me know your thoughts on formatters.

### DIFF
--- a/lib/web3/eth.js
+++ b/lib/web3/eth.js
@@ -160,6 +160,13 @@ var getTransactionCount = new Method({
     outputFormatter: utils.toDecimal
 });
 
+var sendRawTransaction = new Method({
+    name: 'sendRawTransaction',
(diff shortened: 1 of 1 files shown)


lib/web3/eth.js
```diff
@@ -160,6 +160,13 @@ var getTransactionCount = new Method({
     outputFormatter: utils.toDecimal
 });
 
+var sendRawTransaction = new Method({
+    name: 'sendRawTransaction',
+    call: 'eth_sendRawTransaction',
+    params: 1,
+    inputFormatter: [null]
```
frozeman: Please just use an empty array `[]`

PR #244, opened 2015-06-24
all events filter

debris:
filter on all contract's events

usage:
```javascript
var Contract;
var filter = Contrat.allEvents();   // similar to Contract.myEvent();
filter.watch(function (err, res) {
});
```


frozeman:
:+1:

PR #245, opened 2015-06-24
Add possible callback to filters

frozeman:
This PR allows to pass in a callback directly to the filter to start watching.

This is mostly a convenient feature and a common mistake by developers, but should be allowed:

```js

web3.eth.filter({toBlock: 'pending'}, function(e, log){

})

// works also for events

mycontract.MyEvent({}, {toBlock: 'pending'}, function(e, log){

})

```


dist/web3-light.js
```diff
@@ -1429,22 +1429,26 @@ web3.version.api = version.version;
 web3.eth = {};
 
 /*jshint maxparams:4 */
-web3.eth.filter = function (fil, eventParams, options, formatter) {
+web3.eth.filter = function (fil, eventParams, options, callback) {
+
+    if (utils.isFunction(arguments[arguments.length - 1])) {
+        callback = arguments[arguments.length - 1];
```
debris: this line is not necessary.

frozeman: It is, because you can create filters and leave the options part out:

```js
web3.eth.filter({toBlock: 1}, function(e, res){ ...})
```

dist/web3-light.js
```diff
@@ -1429,22 +1429,26 @@ web3.version.api = version.version;
...
     // if its event, treat it differently
     // TODO: simplify and remove
     if (fil._isEvent) {
-        return fil(eventParams, options);
+        return fil(eventParams, options, callback);
     }
 
     // output logs works for blockFilter and pendingTransaction filters?
-    return new Filter(fil, watches.eth(), formatter || formatters.outputLogFormatter);
+    return new Filter(fil, watches.eth(), callback, formatters.outputLogFormatter);
```
debris: not sure if this is necessary / valid:
- `new XXX` returns new object directly, not in the callback
- `new Filter` doesn't do anything synchronously anyway

frozeman: Sure, but you probably want to get the filter and then use it later in the callback to stop it:

```js
var filter = web3.eth.filter({}, funciton(e, log){

   ...
   filter.stopWatching();
})
```

dist/web3-light.js
```diff
@@ -1429,22 +1429,26 @@ web3.version.api = version.version;
...
+
+    if (utils.isFunction(arguments[arguments.length - 1])) {
+        callback = arguments[arguments.length - 1];
+    }
 
     // if its event, treat it differently
     // TODO: simplify and remove
     if (fil._isEvent) {
-        return fil(eventParams, options);
+        return fil(eventParams, options, callback);
```
debris: I think we can remove this line. I've never seen anyone doing:

```javascript
web3.eth.filter(myContract.myEvent);
```

dist/web3-light.js
debris: ok :)

If we remove `if (fil._isEvent...` (see comment below) we will be able to remove these 2 redundant params (`eventParams` and `options`) and this line :)

dist/web3-light.js
```diff
@@ -2323,10 +2329,15 @@ SolidityEvent.prototype.decode = function (data) {
  * @param {Object} options
  * @return {Object} filter object
  */
-SolidityEvent.prototype.execute = function (indexed, options) {
+SolidityEvent.prototype.execute = function (indexed, options, callback) {
+
+    if (utils.isFunction(arguments[arguments.length - 1])) {
+        callback = arguments[arguments.length - 1];
```
debris: you also need to set `options` to `null` here :)

dist/web3-light.js
```diff
@@ -2323,10 +2329,15 @@ SolidityEvent.prototype.decode = function (data) {
...
+    }
+    
     var o = this.encode(indexed, options);
     var formatter = this.decode.bind(this);
-    return web3.eth.filter(o, undefined, undefined, formatter);
+    return new Filter(o, watches.eth(), callback, formatter);
```
debris: :+1:

dist/web3-light.js
```diff
@@ -2178,6 +2182,8 @@ var coder = require('../solidity/coder');
 var web3 = require('../web3');
```
debris: I think, that this line may no longer be required in `events.js`.

dist/web3.js
```diff
@@ -2477,7 +2488,7 @@ var pollFilter = function(self) {
 
 };
 
-var Filter = function (options, methods, formatter) {
+var Filter = function (options, methods, callback, formatter) {
```
debris: It might be sensible to put `callback` at the end of params list. Also using `if (argumets...` might be good here

dist/web3-light.js
debris: I saw auto start watch later. It's cool :+1:

dist/web3-light.js
frozeman: I thought about that too., Ok lets remove it. Its not even documented

debris:
It's good PR. It simplifies few dirties functions that we had. All the issues are minor and we should merge pr as soon as they are fixed :+1:

frozeman:
Ok all done!

PR #246, opened 2015-06-24
Add ipc provider

frozeman:
This adds the IPC provider, which receives a pic file path and a and optional `net` module (used for remote require in mist)

The main difference to HTTP provider is that it has no sync method and there is no way to make a sync request work, without hacking and blocking the UI thread.


frozeman:
Example for mist:
```js
const remote = require('remote');
web3.setProvider(new web3.providers.IpcProvider('/tmp/geth.ipc', remote.require('net')));

```

frozeman:
Additionally this adds the `isConnected()` method to the http provider and web3, so that developers can use this to check if a connection is established or broken.

frozeman:
@debris It seems he takes bignumber.js into consideration for the test and thats why the coverage dropped so much. Any idea how to fix that?

dist/web3-light.js
```diff
@@ -3327,6 +3348,145 @@ module.exports = ICAP;
...
+"use strict";
+
+var utils = require('../utils/utils');
+var errors = require('./errors');
+
+
+var IpcProvider = function (path, net) {
+    var _this = this;
+    this.responseCallbacks = {};
+    this.path = path;
```
debris: is there any default path in geth? If yes, please make it default here

dist/web3-light.js
```diff
@@ -3327,6 +3348,145 @@ module.exports = ICAP;
...
+    
+    net = net || require('net');
```
debris: why `net` is passed as a param? Can it be anything else than standard `net` package?

dist/web3-light.js
```diff
@@ -3327,6 +3348,145 @@ module.exports = ICAP;
...
+
+
+    try {
+        this.connection = net.connect({path: this.path});
+    } catch(error) {
+        throw errors.InvalidConnection(path);
```
debris: does `net.connect` throw an error? Just asking cause [documentation](https://nodejs.org/api/net.html#net_net_connect_options_connectionlistener) doesn't say anything about it.

dist/web3-light.js
```diff
@@ -3327,6 +3348,145 @@ module.exports = ICAP;
...
+    }
+
+    // this.connection.on('error', function(e){
+    //     throw errors.InvalidConnection(path);
+    // });
```
debris: please remove unused code

dist/web3-light.js
```diff
@@ -3327,6 +3348,145 @@ module.exports = ICAP;
...
+    var id = payload.id || payload[0].id;
+
+    this.responseCallbacks[id] = callback;
+};
+
+
+IpcProvider.prototype.isConnected = function() {
+    // try reconnect, when connection is gone
+    if(!this.connection._handle)
+        this.connection.connect({path: this.path});
```
debris: method called `isConnected` shouldn't do anything besides checking if there is existing connection.

dist/web3-light.js
```diff
@@ -3327,6 +3348,145 @@ module.exports = ICAP;
...
+
+    return !!this.connection._handle;
```
debris: `_handle` is private, undocumented property. It should not be used. `socket.connect` is asynchronous and may create `_handle` just after connection, later or never. We do not know.

debris: Instead it might be better to store connection status as a property of IpcProvider prototype.

eg.

```javascript
var connected = false;

var net = require('net');
var client = net.connect({port: 8124}, function() {
  connected = true;
});

client.on('end', function() {
  connected = false;
});
```

dist/web3-light.js
```diff
@@ -3327,6 +3348,145 @@ module.exports = ICAP;
...
+};
+
+IpcProvider.prototype.send = function (payload) {
+
+    if(this.connection.writeSync) {
+
+        // try reconnect, when connection is gone
+        if(!this.connection._handle)
+            this.connection.connect({path: this.path});
```
debris: same as earlier. `_handle` is private and `connect` is asynchronous. also it should not be a part of `send` method.

dist/web3-light.js
```diff
@@ -3327,6 +3348,145 @@ module.exports = ICAP;
...
+
+    } else {
+        throw new Error('You tried to send "'+ payload.method +'" synchronously. Synchronous requests are not supported by the IPC provider.');
+    }
+};
+
+IpcProvider.prototype.sendAsync = function (payload, callback) {
+    // try reconnect, when connection is gone
+    if(!this.connection._handle)
+        this.connection.connect({path: this.path});
```
debris: same as earlier. `_handle` is private and `connect` is asynchronous. also it should not be a part of `send` method.

dist/web3-light.js
```diff
@@ -3327,6 +3348,145 @@ module.exports = ICAP;
...
+
+    return !!this.connection._handle;
+};
+
+IpcProvider.prototype.send = function (payload) {
+
+    if(this.connection.writeSync) {
```
debris: I cannot find `writeSync` in [net documentation](https://nodejs.org/api/net.html). Are you using any other library?

dist/web3-light.js
```diff
@@ -3327,6 +3348,145 @@ module.exports = ICAP;
...
+            throw errors.InvalidResponse(result);                
+        }
+
+        var id;
+
+        // get the id which matches the returned id
+        if(utils.isArray(result)) {
+            result.forEach(function(load){
+                if(_this.responseCallbacks[load.id])
+                    id = load.id;
```
debris: there is no need to iterate over an array. According to logic from `_getResponse` id will always be an id of `[0]` array element.

dist/web3-light.js
```diff
@@ -13192,8 +13352,11 @@ module.exports = BigNumber; // jshint ignore:line
 
 },{}],"web3":[function(require,module,exports){
 var web3 = require('./lib/web3');
+
 web3.providers.HttpProvider = require('./lib/web3/httpprovider');
 web3.providers.QtSyncProvider = require('./lib/web3/qtsync');
+web3.providers.IpcProvider = require('./lib/web3/ipcprovider');
```
debris: Can wrap it into `if` statement? We do not need `IpcProvider` in browser.

dist/web3-light.js
debris: I read your comment under PR.  Is [this](https://www.npmjs.com/package/remote) a `remote` package that you used in example?

dist/web3-light.js
frozeman: There will be a default path, but it ts probably depending on the operating system and the user folder, so it has to be calculated. 

If i have a solid way to determine this i will add it if not, it needs to passed in (still figuring this part out)

dist/web3-light.js
frozeman: Yes, but in mist its even using a net wrapper package, which tunnels all `net` calls, through the electron backend to filter it over and internal IPC.

So the default net package makes only sense of you're in a node environment. In a browser you always would need to use HTTP or a custom `net` wrapper of some sort.

dist/web3-light.js
frozeman: Its unclear, because some errors are watched by electron before i can catch them and i have to figure out a way to make that reliable and presentable to the user..

dist/web3-light.js
frozeman: The PR is not yet ready and i might use this part again.

dist/web3-light.js
frozeman: It works a little different here. As the IPC connections can close and can be reopened immediately, thats why this does auto connect before it returns.

If the connection would then still fail for some reason (e.g. `geth` is closed) it wouldn't be able to re-open the connection and return false.

But i'm also still in the process to figure out a reliable way of doing this.

dist/web3.js
```diff
@@ -1444,6 +1444,9 @@ web3.setProvider = function (provider) {
     this.currentProvider = provider;
     RequestManager.getInstance().setProvider(provider);
 };
+web3.isConnected = function(){
```
debris: I'm against this method. 

- First line of [JSONRPC spec](http://www.jsonrpc.org/specification#overview) says:

  > JSON-RPC is a stateless...

  and providers should also be stateless. If a provider like IpcProvider requires persistent connection it should be handled internally inside such provider.
- also you can't tell that `HttpProvider` is `connected`. You can only say that it responded to some request.

dist/web3-light.js
frozeman: Its not supported by the native `net` package, but part of the mist net wrapper.

The mist wrapper is able to do synchronous IPC connections to the electron backend (which then sends of the real write) and therefore we can still provide synchronous send in mist. When you use it in node directly we would throw an error stating that sync is not supported as IPC connections in node a inherently async.

dist/web3-light.js
frozeman: The result can be a batch array and could technically come in  a different order than the send batch. Therefore iterating over the batch results until one items matches the noted ID, is more secure.

dist/web3-light.js
frozeman: We do. in mist.

dist/web3.js
frozeman: The main and important point of this methods is that dapp developers have a reliable way to check if a node i connected, to e.g. show the user some popup exlpaining how to start a node. It has nothing to do with the provider per se, just the way to check if its running can be different per provider.

kumavis: I think that the dapp should not care or know if its connected, that is the job of the browser.

frozeman: I want that dapps running in mist are still compatible with normal browsers, and there is no reason they shouldn't. So when you use the in a normal browser, the they have to inform the user to start a node. 

And if they aren't able to check the connection to a node, they will have be able to deal with unexpected response behaviour etc. So what will happen is that people will do things like checking for `eth.accounts` and then show a window, but there is no reason why we should not offer a standard to check for connections.

kumavis: if dapps will just fall back to heuristics to determine connection status, I suppose its better to provide an explicit way to look it up

frozeman:
The example in mist is now:
```js
const ipcProviderWrapper = require('../ipc/ipcProviderWrapper.js');
web3.setProvider(new web3.providers.IpcProvider('', ipcProviderWrapper));
```

We pass in an empty connection path, as the ipc backend of mist, will set the path, so that the dapps can't manipulate this connection.

They are basically not even able to require any package, as the code above is running in a preloaded file, with node integration, before the actual web view content is loaded.

This way the web3 object appears as native as possible, with an already set and working connection.

dist/web3-light.js
debris: But mist is using electron. Electron is a subset of node. You can't do [`ipc`](https://en.wikipedia.org/wiki/Inter-process_communication) in normal browser (like chromium). It should be wrapped in the `if` statement.

dist/web3.js
debris: I still do not like this method here. It's not reliable and it gives you nothing. You still have to take care of random unpredicted errors (eg. geth node going down) in your code.

.versions
```diff
@@ -1,3 +1,3 @@
-ethereum:web3@0.5.0
+ethereum:web3@0.7.0
```
debris: it's first time I noticed this file. What it does and why we need it? If it's necessary it might be sensible to add version upgrade to `gulpfile`.

dist/web3-light.js
```diff
@@ -3495,6 +3516,201 @@ module.exports = ICAP;
...
+        /*jshint maxcomplexity: 6 */
+        data = data.toString();
+
+        // DE-CHUNKER
+        var dechunkedData = data
+            .replace(/\}\{/g,'}|--|{') // }{
+            .replace(/\}\]\[\{/g,'}]|--|[{') // }][{
+            .replace(/\}\[\{/g,'}|--|[{') // }[{
+            .replace(/\}\]\{/g,'}]|--|{') // }]{
+            .split('|--|');
```
debris: Why is it like that? Can't you fix backend to send proper JSON? If not, it would be better to do:

```js
var dechunkedData = JSON.parse('[' + data + ']');
```

dist/web3-light.js
```diff
@@ -3495,6 +3516,201 @@ module.exports = ICAP;
...
+
+
+        dechunkedData.forEach(function(data){
+
+            // prepend the last chunk
+            if(_this.lastChunk)
+                data = _this.lastChunk + data;
```
debris: why are you prepending this?

dist/web3-light.js
frozeman: eletctron has can preload files in a web view (type of iframe), which will have node integration, but are in a browser environment (e.g. having the window object).
In this stage i set the provider wrapper as IPC provider. The web page later loaded doesn't have the node integration and couldn't use the IPC, but even they could provide some kind of wrapper for an IPC/websocket behaviour, se even here the IPC provider makes sense, as its current behaviour very much matches the ones of web sockets too.

But for mist we need it available in the browser for sure and i don't see a problem having it visible to users.

dist/web3.js
frozeman: Sure you do, but e.g. the wallet dapp will work in mist and normal browser (where it asks the user to start a node) and therefore a connection checker method is required and useful.
We can always later improve the way, how we check for connections:
```js
// set providor
if(!web3.currentProvider)
    web3.setProvider(new web3.providers.HttpProvider("http://localhost:8545"));
```

.versions
frozeman: this file is generated by meteor, when you publish as it contains all package dependencies.
We should remove it and i wouldn't rely on it as its updated by meteor, when you publish.

dist/web3-light.js
frozeman: The chunks which come back, can be cut, objects or array if batch requests and they can be all of this in one string e.g.:

```js
"{....}[{..},{..}]{...}{...}{.."
```

So i agree the separator of `|--|` is debateable, but the solution is the one which works very well and covers all cases.

dist/web3-light.js
```diff
@@ -3495,6 +3516,201 @@ module.exports = ICAP;
...
+
+            var result = data,
+                id = null;
+
+            try {
+                result = JSON.parse(result);
```
debris: if I understand the code corretly, result might be one of:

- `{}`  - which is correct
- `[]`  - which is correct
- `"null"` - which is result of `null + []` - line 3594 and 3570
- `"null[object Object]"` -  which is result of `null + {}` - line 3594 and 3570
- `""` - which is result of `[] + []`
- `"[object Object]"` - which is result of `[] + {}`
- `0` - which is result of `{} + []`

+ multiple other combinations

If it behaves differently, it is totally unreadable

dist/web3-light.js
frozeman: because a request can be half complete if the last buffer was full, so i se the `_this.lastChunk` and prepend it to the next buffer piece i get on the next call of `on('data'`.

This is all tested and works.

dist/web3-light.js
```diff
@@ -3495,6 +3516,201 @@ module.exports = ICAP;
...
+};
+
+/**
+Timeout all requests when the end/error event is fired
+
+@method _timeout
+*/
+IpcProvider.prototype._timeout = function() {
+    for(var key in this.responseCallbacks) {
+        if(this.responseCallbacks.hasOwnProperty(key)){
```
debris: This line is redundant. Please try:

```js
x = {a: 1};
for (key in x) { console.log(key) }
```

dist/web3-light.js
```diff
@@ -3495,6 +3516,201 @@ module.exports = ICAP;
...
+            this.responseCallbacks[key](errorTimeout.replace('__id__', key).replace('__method__', this.responseCallbacks[key].method));
```
debris: please use function instead

dist/web3-light.js
frozeman: See the comment above.

And if you comment on lines, it would be good if we use the real files, not the generated dist files :)

dist/web3-light.js
frozeman: https://blog.javascripting.com/2014/09/19/when-should-javascript-developers-use-hasownproperty/

dist/web3-light.js
```diff
@@ -3495,6 +3516,201 @@ module.exports = ICAP;
...
+IpcProvider.prototype.isConnected = function() {
+    var _this = this;
+
+    // try reconnect, when connection is gone
+    setTimeout(function(){
+        if(!_this.connection.writable)
+            _this.connection.connect({path: _this.path});
+    }, 0);
+
+    return !!this.connection.writable;
```
debris: I commented on this before. Method called `isConnected` should not handle reconnection.

dist/web3-light.js
frozeman: Its also part of the mozilla docs example as well:
```js
for (var prop in obj) {
  if( obj.hasOwnProperty( prop ) ) {
    console.log("o." + prop + " = " + obj[prop]);
  } 
}
```

dist/web3-light.js
debris: I'm talking about [this](https://github.com/ethereum/web3.js/blob/ipcProvider/lib/web3/ipcprovider.js#L51-L111).

Data should be:

1. received
2. parsed
3. processed 

step by step. This function is a mix of everything together.

dist/web3-light.js
frozeman: It needs to, as the connection is timeouted after 15s by the geth node.
So before every request we need to re-establish the connection (see the send function too).

If we wouldn't do that, and the connection timed out, you appear to be offline, even though you actual have a connection.

dist/web3-light.js
frozeman: no sure if that makes sense, as its the only time its done here.
Reducing cyclomatic complexity this way is fake reducing it, as it still has to execute the same amount of complexity, just jslint doesn't see it.

dist/web3-light.js
frozeman: why splitting this up, whats the benefit?

This function does receive and parse and then calls the correct response callbacks. If you can tell me whats the benefit of moving these 50 lines into separate functions would be nice.

PR #255, opened 2015-07-06
add new contract.new behaviour

frozeman:
This adds the new contract deployment behaviour, as now the sendTransaction only returns the tx hash and not the contract address anymore.

Usage:
```js
var contract = web3.eth.contract([{"constant":true,"inputs":[{"name":"greeting","type":"string"}],"name":"something","outputs":[{"name":"","type":"string"}],"type":"function"}]);

var returnedContract = contract.new({
   data: '6060604052610118806100136000396000f30060606040526000357c010000000000000000000000000000000000000000000000000000000090048063d1f9cf0e1461003957610037565b005b6100896004803590602001906004018035906020019191908080601f01602080910402602001604051908101604052809392919081815260200183838082843782019150505050505090506100f7565b60405180806020018281038252838181518152602001915080519060200190808383829060006004602084601f0104600302600f01f150905090810190601f1680156100e95780820380516001836020036101000a031916815260200191505b509250505060405180910390f35b6020604051908101604052806000815260200150819050610113565b91905056',
   from: '0x343c98e2b6e49bc0fed722c2a269f3814ddd1533',
   gas: 3000000
}, function(e, contract){
   console.log(e, contract);
   // contract instance
   // contract === returnedContract
})

// returnedContract will later get the address attached to it automatically. (undefined at the beginning)
returnedContract.address;

// Additionally the contract will have the `transactionHash` property, which is the current transactions hash. When one want to use to check itself.
returnedContract.transactionHash;
```

There are two types of errors, one is fired, if after 50 blocks the contract tx still has no receipt.
And another if the receipt is present, but the `web3.eth.getCode()` returns `0x`, means the tx was executed, but the code was never stored.


dist/web3-light.js
```diff
@@ -1821,6 +1821,56 @@ var contract = function (abi) {
...
+                    callback(new Error('Contract couldn\'t be deployed'));
+
+                filter.stopWatching();
+
+            } else {
+
+                web3.eth.getTransactionReceipt(contract.transactionHash, function(e, receipt){
+                    if(receipt) {
+
+                        web3.eth.getCode(receipt.contractAddress, function(e, code){
```
debris: is it necessary? can't we just use `receipt.contractAddress`?

package.json
```diff
@@ -57,29 +57,29 @@
...
-    {
       "name": "Marek Kotewicz",
       "email": "marek@ethdev.com",
       "url": "https://github.com/debris"
     },
     {
+      "name": "Fabian Vogelsteller",
+      "email": "fabian@ethdev.com",
+      "homepage": "http://frozeman.de"
+    },
```
debris: sneaky alteration :+1:

dist/web3-light.js
frozeman: Yes this is necessary, as the contract address can be created and the code executed, but there was not enough gas to store the code.

I had these issues see here: https://github.com/ethereum/yellowpaper/issues/132

package.json
frozeman: I just ordered it in the order, of contributors, no reason that jeff is on the top here :)

PR #259, opened 2015-07-07
fixed contract deploy

frozeman:
Fixes an issue, where the contract address is not properly set to methods, and events after using `contract.new(...)`.

Instead of the address, we are passing a reference of the contract instance and we are therefore always use the right address, even if one would manually update `contract.address`.


debris:
I understand why we need this pr, but I would prefer to do it differently. Not all the objects that you called `contracts` are `contracts`. This might be misleading, especially when the prototype contianing field `_contract` has a method `attachToContract`.

We should refactor this in near future with something else eg. lazy evaluation of `address`.

PR #283, opened 2015-08-05
namereg && iban stuff prepared for end of thawing

debris:
*i should have opened this pr against develop, but merging into master && develop is also fine.*

**changes**: 

- new contract `web3.eth.ibanNamereg`
- removed `web3.isIBAN`
- new field `web3.eth.iban`
  - `new web3.eth.iban(string)` -> Iban object
  - `web3.eth.iban.isValid` -> bool
  - `web3.eth.iban.fromAddress(string)` -> Iban object
  - `web3.eth.iban.fromBban(string)` -> Iban object
  - `web3.eth.iban.createIndirect(object)` -> Iban object
  - `iban.toString()`
  - and few more...

**TODO**:
- after thawing phase ends and nameregs will be created, we will need to update their addresses


frozeman:
There are some errors:
```
lib/solidity/coder.js: line 166, col 23, This function's cyclomatic complexity is too high. (5)
lib/utils/utils.js: line 112, col 31, 'escape' is not defined.
lib/utils/utils.js: line 123, col 11, 'unescape' is not defined.
```

Could you elaborate on the two different name regs a bit?

Additionally does using these nameregs do anything sync (it does look like), if so i would recommend we add a async version, as sync is really a no go, especially in mist where on request could block another and cause a lock down.

frozeman:
additionally its not rebuild.

debris:
> There are some errors:
```
lib/solidity/coder.js: line 166, col 23, This function's cyclomatic complexity is too high. (5)
lib/utils/utils.js: line 112, col 31, 'escape' is not defined.
lib/utils/utils.js: line 123, col 11, 'unescape' is not defined.
```

Errors are not related to this pr.

> Could you elaborate on the two different name regs a bit?

One of them will be used as global (main) registar. Everyone will be able to register in it for free (or almost for free), whereas second will be used by institutions such as exchanges. It has fixed registration fee (probably 69ETH). It will be also used by users (underneath) to look up exchange address before sending funds to exchange.

> Additionally does using these nameregs do anything sync (it does look like), if so i would recommend we add a async version, as sync is really a no go, especially in mist where on request could block another and cause a lock down.

There is sync and async version.

lib/web3/transfer.js
```diff
@@ -49,7 +50,7 @@ var transfer = function (from, iban, value, callback) {
         return deposit(from, address, value, icap.client());
     }
 
-    namereg.addr(icap.insitution(), function (err, address) {
+    namereg.addr(icap.institution(), function (err, address) {
         return deposit(from, address, value, icap.client(), callback);
```
frozeman: We should probably add a check for `if(err)` here, otherwise we send to `undefined`

debris:
pr is updated and ready to merge

Issue #270, opened 2015-07-20
Cannot resolve module 'net'


caktux:
```
ERROR in ./~/web3/lib/web3/ipcprovider.js
Module not found: Error: Cannot resolve module 'net' in [...]/web3/lib/web3
 @ ./~/web3/lib/web3/ipcprovider.js 36:17-31
```

simondlr:
Getting the same. Shouldn't it be:

require('./net');

?

frozeman:
hm net is a standard module in node. In which environments you're getting this error?

caktux:
Using webpack, and if webpack complains then there's something fishy ;)

frozeman:
The IPC provider is only useable in node environments and mist frontend, therefor probably you web pack complain.

frozeman:
if you have an idea how we can hide that module from web pack, would be great.

caktux:
added extra configs to webpack to solve it:
```
  node: {
    fs: 'empty',
    net: 'empty',
    tls: 'empty'
  }
```
having a `net` module in web3 made this issue rather confusing..

debris:
@frozeman I think that it is good moment to clean up everything. Can we move `ipcprovider` from `web3.js` to `mist`? Or to it's own repo? It just doesn't belong here. We should make web3 a backbone, not double-edged sword.

frozeman:
the IPC provider is necessary for node based applications, which want to talk t a local node.
So every hybrid dapp, would need it. Though it doesn't work in the browser, thats true. We can hide it from there and i add it in mist myself.

Would it be enough to add a check in the `index.js`?
```js
if(type window !== 'undefined')
 // add IPC provider
```

PR #284, opened 2015-08-06
cleanup

debris:
plain cleanup and fixes:
- `httpprovider`
- #270 
- ipcprovider's `errorTimeout`


lib/web3/requestmanager.js
```diff
@@ -161,9 +159,8 @@ RequestManager.prototype.setProvider = function (p) {
  * @todo cleanup number of params
  */
 RequestManager.prototype.startPolling = function (data, pollId, callback, uninstall) {
-    this.polls['poll_'+ pollId] = {data: data, id: pollId, callback: callback, uninstall: uninstall};
+    this.polls[pollId] = {data: data, id: pollId, callback: callback, uninstall: uninstall};
```
frozeman: the point of this is to prevent keys starting with numbers, as some browser have problems with that.

test/ipcprovider.js
```diff
@@ -33,13 +32,13 @@ describe('lib/web3/ipcprovider', function () {
 
     describe('isConnected', function () {
         it('should return a boolean', function () {
-            var provider = new IpcProvider();
+            var provider = new IpcProvider('', net);
```
frozeman: It should also with if you ant providing a net package, as it should load the nom default one

frozeman: But it might make sense here, so it uses a stub

frozeman:
Needs rebase, but otherwise looks good.
I couldn't see the `A-F` included in `isAddress`. I checked with jeff and go is fine with uppercase addresses.

PR #288, opened 2015-08-08
+ add personal API support

almindor:
Adds personal API support to use. Might need some security warnings to use only via IPC provider.

### DIFF
--- a/lib/web3.js
+++ b/lib/web3.js
@@ -27,6 +27,7 @@
 var version = require('./version.json');
 var net = require('./web3/net');
 var eth = require('./web3/eth');
+var personal = require('./web3/personal');
 var db = require('./web3/db');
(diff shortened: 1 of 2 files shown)


frozeman:
Thanks for the PR, but we have on purpose not included any admin API in web3 by default, we have first to discuss on how we want to do this. (e.g. separate module, or shipped with web3)

almindor:
OK, how about exposing the setupParameters and setupMethods so people can add their own into your library without modifying the original code? That'd solve it for my use for not.

frozeman:
We have a semi official method called web3._extend(). See here for an example: https://github.com/ethereum/mist/blob/master/modules/preloader/web3Admin.js

almindor:
Ah I see, thanks for the info. I'll use that for the time being.

PR #307, opened 2015-08-26
fromUtf8 and toUtf8 should ignore null characters

caktux:


frozeman:
@debris could you please take a look

johnmcdowall:
@caktux @frozeman I just got tripped up with this on the toAscii method to. Let me know if you're going to expand your PR @caktux or if I should open a separate one. 

Example of a hex (0x444f574e4c4f4144000000000000000000000000000000000000000000000000) converted using toAscii:

```
ps[0]
"download"
ps[0].length
32
```

PR #316, opened 2015-09-15
Is syncing

frozeman:
adds the new `web3.eth.syncing` property as well as a the `web3.eth.isSyncing(cb)` function, which gets called as follows:

```js
web3.eth.isSyncing(function(error, sync){
 if(!error) {
    if(sync === true) {
       // stop all activity
       web3.reset();
    } else if(sync) {
       // show sync info, stop app calls etc
       console.log(sync.highestBlock);
    } else {
       // regain app operation
    }
  }
});
```
1. `sync === true`,  called once when the node starts syncing
2. `sync === web3.eth.syncing` object, gets called everytime there is a change in the sync status (new imported block, etc)
3. `sync === false`, gets called once the syncing stopped


lib/utils/utils.js
```diff
@@ -391,7 +391,7 @@ var isStrictAddress = function (address) {
  * @return {Boolean}
 */
 var isAddress = function (address) {
-    return /^(0x)?[0-9a-f]{40}$/.test(address);
+    return /^(0x)?[0-9a-f]{40}$/i.test(address);
```
frozeman: This doesn't fully belong here, but its a necessary change so allow also addresses like `0xFFAA123123FFF...`, @debris can you please cake if cpp is flexible with the case for addresses?

frozeman:
@karalabe just tested this in the go console:
```js
> web3.eth.isSyncing(function(error, sync){
......  if(!error) {
..........     if(sync === true) {
..............        // stop all activity
..............        console.log('started syncing')
..............     } else if(sync) {
..............        // show sync info, stop app calls etc
..............        console.log(sync.highestBlock);
..............     } else {
..............        // regain app operation
..............       console.log('stopped syncing')
..............     }
..........   }
...... });
{
  callbacks: [function(error, sync)],
  implementation: {
    call: "eth_syncing",
    inputFormatter: undefined,
    name: "isSyncing",
    params: 0,
    attachToObject: function(obj),
    extractCallback: function(args),
    formatInput: function(args),
    formatOutput: function(result),
    getCall: function(args),
    outputFormatter: function(result),
    request: function(),
    send: function(),
    toPayload: function(args),
    validateArgs: function(args)
  },
  pollId: "syncPoll_940",
  addCallback: function(callback),
  stopWatching: function()
}
> started syncing
237129
[...]
237129
stopped syncing
```

debris:
Looks fine. :+1:  In future, it might be good to add some generic tool for polling. 

Please merge with latest develop

frozeman:
In the future we should switch from JSONRPC to DPP and introduce proper pub/sub for all of this :) fuck polling :P

debris:
DPP?

Issue #297, opened 2015-08-13
Multiple instances of web3 object


asinyagin:
Hey,

I want to work with 2 web3 providers at the same time. As I see, web3 is not supposed to be used with multiple instances. Can you help me with some workaround for it?

d11e9:
:+1: yeah I don't like to current global approach either. global == bad

aakilfernandes:
+1. Working on a testchain with an oracle on the mainchain, this would be super useful

debris:
:+1:

d11e9:
:+1:

emprisupriatna:
I looking for the right etherwallet address

Adyfamba:
Support web3

PR #317, opened 2015-09-16
Add support of multiple providers

asinyagin:
Hey,

I'd like to add support of multiple providers to web3. I was asking about workaround for it in #297. I've created this PR to discuss the idea, it changes only `eth.getBalance()`, so now the method can work through multiple providers.

I'll add comments to the diff to mark the main points.

### DIFF
--- a/lib/web3.js
+++ b/lib/web3.js
@@ -38,6 +38,7 @@ var c = require('./utils/config');
 var Property = require('./web3/property');
 var Batch = require('./web3/batch');
 var sha3 = require('./utils/sha3');
+var Eth = require('./web3/methods/eth1');
 
 var web3Properties = [
     new Property({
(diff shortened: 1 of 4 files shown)


lib/web3.js
```diff
@@ -78,56 +79,60 @@ var setupProperties = function (obj, properties) {
 };
 
 /// setups web3 object, and it's in-browser executed methods
-var web3 = {};
-web3.providers = {};
-web3.currentProvider = null;
-web3.version = {};
-web3.version.api = version.version;
-web3.eth = {};
+function Web3(provider) {
```
asinyagin: That's a constructor for `web3` object. The object stores links to `requestManager` and namespaces like `eth` (`eth1` in this case).

lib/web3/method.js
```diff
@@ -168,5 +168,30 @@ Method.prototype.send = function () {
     return this.formatOutput(RequestManager.getInstance().send(payload));
 };
 
-module.exports = Method;
+Method.prototype.attachToObject1 = function (obj) {
```
asinyagin: It uses closure to attach method to namespace.

lib/web3/method.js
```diff
@@ -168,5 +168,30 @@ Method.prototype.send = function () {
...
+        obj[name[0]] = func; 
+    }
+};
+
+Method.prototype.buildCall = function() {
+    var method = this;
+    return function send() {
+        var payload = method.toPayload(Array.prototype.slice.call(arguments));
+        if (payload.callback) {
+            return this.web3.requestManager.sendAsync(payload, function (err, result) {
```
asinyagin: It sends a request through namespace's `requestManager`.

lib/web3/methods/eth1.js
```diff
@@ -0,0 +1,42 @@
+var formatters = require('../formatters');
+var utils = require('../../utils/utils');
+var Method = require('../method');
+var Property = require('../property');
+var c = require('../../utils/config');
+
+function Eth(web3) {
```
asinyagin: Here's the way to define a namespace with methods and properties.

lib/web3.js
frozeman: So you need to instantiate `Web3`?
So you would like to have multiple instances, with different providers running in the same app?

web3 is designed to be an object like the `chrome` or later `mist` object. It gives "native" functions and the dapp shouldn't even care how it connects.

lib/web3/methods/eth1.js
```diff
@@ -0,0 +1,42 @@
...
+var methods = [
+    new Method({
+        name: 'getBalance',
+        call: 'eth_getBalance',
+        params: 2,
+        inputFormatter: [formatters.inputAddressFormatter, formatters.inputDefaultBlockNumberFormatter],
+        outputFormatter: formatters.outputBigNumberFormatter
+    })
+];
+methods.forEach(function(method) { method.attachToObject1(Eth.prototype) });
```
frozeman: I like attaching it here, rather than in `web3.js`. @debris any opinion?

lib/web3.js
asinyagin: Yeah, our ide works with 2 and more JSON RPC in parallel.

Okay, I see the problem. Yeah, that could be unnecessary complication. But if you think it's useful to work with multiple providers, you could get rid of `RequestManager` singleton and current global `web3` and define global object like `window.web3 = new Web3()`.

debris:
I like this pr a lot. I was thinking about this kind of refactor for some time :+1: 

And after we finish that, I would like to refactor even further. I would like to have strongly typed input params. I even did a small [demo](https://github.com/debris/web3.ts) in typescript. We could have similar types to [these](https://github.com/debris/web3.ts/blob/develop/lib/types.ts). It would also solve issues similar to #322. What do you think about this? @frozeman @asinyagin

debris:
Of course we would to this in 2 (or more) separate steps

asinyagin:
@debris nice! Looking forward to test the version with multiple providers. And I agree it's a good idea to validate args. I have such validators to check params before giving them to web3.

frozeman:
How would an example look like for the strict params? Would you simply throw an error if they don't match?

asinyagin:
I'm not sure what's a good way in general, but in my case validator (for uintN from Solidity) returns an array with errors:
```javascript
define(function() {
    return function(type) {
        var size = parseInt(type.substr(3));
        return {
            validate: function(value) {
                var errors = [];
                if (!value.match(/^-?\d+$/))
                    errors.push('int must contain only sign and digits.');
                else {
                    var max = new BigNumber(2).pow(size).minus(1),
                        val = new BigNumber(value, 10);
                    if (!val.lessThan(max))
                        errors.push(type + ' must be less than 2^' + size + '.');
                }
                return errors;
            }
        };
    };
});
```

Issue #225, opened 2015-05-29
Expose a function that will enable encoding of a solidity function call


alexvandesande:
Many apps, like the multisig wallet and the democracy DAO, allows the users to ask the contract to execute a custom solidity call to another contract. This is a very powerful feature that allows, for example the user to have a second authentication for any transaction in the wallet contract; or allows the DAO to own and maintain names in the registrar, or the DAO to own custom coins. 

This is usually done with the solidity .call feature that will have this parameters:

``` js
someAddress.call.value(_amount)(_data);
```

Where `someAddress` is an address to another contract, _amount is amount of ether in wei and _data is a custom abi encode of function that the user wants to call. The issue is that there is no user exposed function that will help build that _data field. 

This is already half baked into web3.js in the Solidity Functions Prototype:

``` js
options.data = '0x' + this.signature() + coder.encodeParams(this._inputTypes, args);
```

I propose that we expose a function in web3 that would have the following parameters:

``` js
web3.encodeSolidityFunctionCall('myFunctionName',
     [{
       name: '_parameter1',
       type: 'address',
       value: '0x2345678..'
    },{
       name: '_parameter2',
       type: 'uint',
       value: '901'
    }])
    ...
```

This would return an ABI encoded string that could be passed as data on to any contracts methods, which use this data in calls to other contracts.

frozeman:
+1

frozeman:
Should be fairly easy to implement as every piece is already there.

debris:
I think that this problem should be solved differently. You can 'include' interface of the contract that you want to call in your solidity contract. Example:

```javascript

// included solidity interface 
contract IncludedContract {
  function someMethod(bytes32 name) {}
}

// contract
contract MyContract {
  function MyContract(address a, bytes32 name) {
    IncludedContract(a).someMethod(name)
  }
}
```

alexvandesande:
Yes but this only works for specific contracts. Some contracts like the DAO and multisig wallet are built to be able to handle any transaction to any contract including those that have not been written yet.

frozeman:
Thats actually a very valid use case. The same is valid for gavins wallet contract, which could be used to sign other contracts or execute contract functions, so the above proposal makes fully sense e.g.:

```js

var callData = web3.encodeSolidityFunctionCall('myFunctionName',
     [{
       name: '_parameter1',
       type: 'address',
       value: '0x2345678..'
    },{
       name: '_parameter2',
       type: 'uint',
       value: '901'
[...]

PR #328, opened 2015-10-08
Refactor

debris:
still in progress, although most of the tests are already passing. #297 #317


debris:
@asinyagin @frozeman can you review? I know that it's quite a big pr, but these are only required changes to make multiple instances of web3 work.

In addition, @frozeman can you look at `test/web3.eth.isSyncing.js` I really couldn't make it work. It was randomly failing. I don't know what logic is inside or what should it do, so I haven't touched it

debris:
in the meantime I will try to make jshint happy :)

lib/web3.js
```diff
@@ -24,168 +25,106 @@
...
+var Batch = require('./web3/batch');
+var Property = require('./web3/property');
+var HttpProvider = require('./web3/httpprovider');
+var IpcProvider = require('./web3/ipcprovider');
+
+
+function Web3 (provider) {
+    this._requestManager = new RequestManager(provider);
+    this.eth = new Eth(this);
+    this.db = new DB(this);
```
frozeman: I would keep it `Db` to adhere the camel case standard

lib/web3/batch.js
```diff
@@ -45,7 +45,7 @@ Batch.prototype.add = function (request) {
  */
 Batch.prototype.execute = function () {
     var requests = this.requests;
-    RequestManager.getInstance().sendBatch(requests, function (err, results) {
+    this.requestManager.sendBatch(requests, function (err, results) {
```
frozeman: Far better, than this weird getInstance thingy..

lib/web3/contract.js
```diff
@@ -57,7 +56,7 @@ var addFunctionsToContract = function (contract, abi) {
     abi.filter(function (json) {
         return json.type === 'function';
     }).map(function (json) {
-        return new SolidityFunction(json, contract.address);
+        return new SolidityFunction(contract._web3, json, contract.address);
```
frozeman: All the things we call solidly, are actually not solidly specific and work well for other languages like LLLL, as they are byte encoded EVM code. I suggest we rename all those to `evmEncoding` or something.

lib/web3/contract.js
```diff
@@ -129,10 +116,10 @@ var checkForContractAddress = function(contract, abi, callback){
 
             } else {
 
-                web3.eth.getTransactionReceipt(contract.transactionHash, function(e, receipt){
+                contract._web3.eth.getTransactionReceipt(contract.transactionHash, function(e, receipt){
                     if(receipt && !callbackFired) {
 
-                        web3.eth.getCode(receipt.contractAddress, function(e, code){
+                        contract._web3.eth.getCode(receipt.contractAddress, function(e, code){
```
frozeman: you call it here `_web3` but in the `ContractFactory` `this.web3`, are you sure thats correct? (I might miss something..)

lib/web3/contract.js
```diff
@@ -207,33 +201,30 @@ ContractFactory.prototype.new = function () {
...
 
         // wait for the contract address adn check if the code was deployed
-        web3.eth.sendTransaction(options, function (err, hash) {
+        this.web3.eth.sendTransaction(options, function (err, hash) {
             if (err) {
                 callback(err);
             } else {
                 // add the transaction hash
-                contract.transactionHash = hash;
+                contract._transactionHash = hash;
```
frozeman: This is a property people are supposed to access from the callback, we can't prefix that with `_`.

This is an unnecessary breaking change..

lib/web3/event.js
```diff
@@ -30,7 +30,8 @@ var watches = require('./methods/watches');
 /**
  * This prototype should be used to create event filters
  */
-var SolidityEvent = function (json, address) {
+var SolidityEvent = function (web3, json, address) {
```
frozeman: These events maybe solidly specific, but we have to check that with christian. Otherwise they should also be renamed to something more general.

lib/web3/methods/db.js
```diff
@@ -22,35 +22,45 @@
...
-    params: 2
-});
-
-var methods = [
-    putString, getString, putHex, getHex
-];
-
-module.exports = {
-    methods: methods
+var DB = function (web3) {
```
frozeman: We might get rid of this all together not a Dapp independent storage. So people should rather use the local storage or indexedDB of the browser.

lib/web3/methods/eth.js
```diff
@@ -74,223 +56,273 @@ var uncleCountCall = function (args) {
...
-    name: 'sendTransaction',
-    call: 'eth_sendTransaction',
-    params: 1,
-    inputFormatter: [formatters.inputTransactionFormatter]
+function Eth(web3) {
+    this.web3 = web3;
+
+    var self = this;
+
+    methods().forEach(function(method) {
```
frozeman: Why you define `methods` as a function, if it could be simple a static array?

lib/web3/property.js
```diff
@@ -76,7 +81,7 @@ Property.prototype.extractCallback = function (args) {
  */
 Property.prototype.attachToObject = function (obj) {
     var proto = {
-        get: this.get.bind(this),
+        get: this.buildGet()
```
frozeman: I don't see the advantage of moving this into its own function, if you simply write:
```js
var proto = {
        get: function get() {
              return property.formatOutput(property.requestManager.send({
                  method: property.getter
             })); 
        }
}
```

frozeman:
I added some comments.

I would like to see the solidity function exposed as discussed in #225, maybe we go with Christians example, but don't call it `web3.funcition` but `web3.composeFunctionCall` or something and pass the args as a third parameter, not an extra `encode` function.

index.js
```diff
@@ -1,19 +1,8 @@
...
-
-web3.eth.contract = require('./lib/web3/contract');
-web3.eth.namereg = namereg.namereg;
-web3.eth.ibanNamereg = namereg.ibanNamereg;
-web3.eth.sendIBANTransaction = require('./lib/web3/transfer');
-web3.eth.iban = require('./lib/web3/iban');
+var Web3 = require('./lib/web3');
 
 // dont override global variable
 if (typeof window !== 'undefined' && typeof window.web3 === 'undefined') {
```
frozeman: `window.web3` is still lowercase.

And shouldn't we expose already a `web3` to avoid backwards incompatibility?

frozeman:
Also, could you give an example of how to integrate it?

like this?
```js
const web3 = require('web3');
window.web3 = new Web3(new web3.providers.IpcProvider('', ipcProviderWrapper));
```

frozeman:
You also seem to break the web3_extend support..

frozeman:
You also removed `web3.currentProvider`, which is absolutely necessary for dapps to check if they are in mist, where the provider is already set.

debris:
web3._extend is not broken

debris:
why do you need web3.currentProvider exposed? You can have multiple web3 objects now, so you will know which web3 object uses which provider :p

frozeman:
With the current implementation you make a full breaking change, as currently web3 was by default exposed. But now its not at all anymore, just `Web`, which needs to initiated.

To don't break backwards compatibility, we should still instantiate one by default and let users set a provider later. But they need to be able to check if one is not already set.

Having multiple providers is cool, but an edge case and not the normal usage.

The `web3` object acts like the `chrome` object, its there and just works.

We could also go the route to tell people to explicitly instantiate web3, but check first if a web3 object exists (which would then be mist)

frozeman:
The test now have a lot of:
```
․[Error: Provider not set or invalid]
․[Error: Provider not set or invalid]
․[Error: Provider not set or invalid]
․․[Error: Provider not set or invalid]

  ․․․․․․․․․․․․․․․․․․․․․․․․․[Error: Provider not set or invalid]
․[Error: Provider not set or invalid]
```

So i assume the web3 object is not properly instantiated with a fake provider.

frozeman:
I randomly get the following error:
```js
  1) web3.eth getBalance async test: 0:
     Uncaught AssertionError: expected [ { data: '0xb' }, { data: '0x11' } ] to equal [ Array(2) ]

```

There seems some kind of race condition

frozeman:
The isSyncing test by itself passes, but not when run as a whole, i assume other test don't clean up properly.

debris:
> The test now have a lot of:
```
․[Error: Provider not set or invalid]
․[Error: Provider not set or invalid]
․[Error: Provider not set or invalid]
․․[Error: Provider not set or invalid]
  ․․․․․․․․․․․․․․․․․․․․․․․․․[Error: Provider not set or invalid]
․[Error: Provider not set or invalid]
```
So i assume the web3 object is not properly instantiated with a fake provider.


I check, and it's becuase the test check if web3.eth properties exist before setting the provider. Actually it's easy to fix. I will do it.

> We could also go the route to tell people to explicitly instantiate web3, but check first if a web3 object exists (which would then be mist)

Yes, that's what we should do in my opinion. Initialising everything globally was not a good practice imo.

> I randomly get the following error:
```js
  1) web3.eth getBalance async test: 0:
     Uncaught AssertionError: expected [ { data: '0xb' }, { data: '0x11' } ] to equal [ Array(2) ]
```
There seems some kind of race condition

I will investigate it

> The isSyncing test by itself passes, but not when run as a whole, i assume other test don't clean up properly.

It may be related with the race condition that you described. I will look at it.

frozeman:
Ok i will push my changes then, please test on those.

frozeman:
Sometimes i also get
```
 2) web3.eth getBalance async test: 1:
     Uncaught AssertionError: expected [ Array(2) ] to equal [ Array(2) ]
```

frozeman:
There is something odd, even when i comment out the whole test content and web3 and the FakeHTTPProvider, it gives me this error:
```
  1) web3.eth isSyncing property test: 0:
     Uncaught AssertionError: "undefined" == "2.0"
```

How is this possible,?

frozeman:
Ok the problem occurs, because some filter polls, are not yet arrived, so even calling `web3.reset()` might reset the poll cycle, but the ones that are already send aren't "returned" yet.

A few weeks ago i add a `setTimeout 1` to the `sendAsync` functions, to make sure they re really async. This may cause the issues here.

What i don't understand is, why the request managers, are not isolated from each other, they shouldn't interfere actually...

frozeman:
One solution would be to create a new requestManager when we set the provider, but this lets a lot of test fail:
```js
Web3.prototype.setProvider = function (provider) {
    this._requestManager = new RequestManager(provider);
    this.currentProvider = provider;
};
```

Not sure if our tests not good because, they are not as isolated as they should be, or this approach is flawed.

lib/web3/methods/eth.js
asinyagin: Also, i think you can attach methods to `Eth.prototype`. Not sure if it's useful here (we don't need inheritance), but it could be a little faster, cause you don't need to do it on `Web3` creation.

debris: > Why you define `methods` as a function, if it could be simple a static array?

Not exactly. We cannot use global objects ('static array', `prototype`) for anything that has a state. In our case, the state is the provider which is currently set. Web3 keeps a reference to it.

The initial refactor from @asinyagin attached methods to `Eth.prototype`. Then `Method` class called `this.web3` assuming that function will be executed in context where `this.web3` is available. This was a workaround which made it possible to attach methods to a prototype. But we cannot make an assumption that `web3` will always be available. It's error prone and if something fails difficult to debug. Also it cannot be used with `extensions` without spoiling extended object with `web3` property.

lib/web3/methods/db.js
debris: I agree. But it's not only up to me and I didn't want to make it a part of this pull request.

lib/web3/event.js
debris: Events are just higher layer of abstraction of the logs (and that's why they are using filters internally). I find the name fine. Also if we decide to change it, it should not be a part of this pr.

lib/web3/contract.js
debris: I might be wrong, but I think that all the things that we call `solidity` in source code, **were** initially part only of solidity ABI specification. They are not a part of evm, nor of yellowpaper. Certain implementations of other languages (eg. LLL) adopted this ABI. Am I right @chriseth ?

debris: Also renaming is not a part of this pr. If we decide to rename all of this, I suggest to move all abi-related stuff to it's own library. I initially wanted to this with [solidity.js](https://github.com/ethereum/solidity.js), but was stopped by more urgent things.

lib/web3/property.js
debris: it was not moved, cause it always was it has always been it's own function. We just do not do binding here.

Also, I find it more readable (and possibly testable), when it's separated.

debris:
I fixed most of the issues with tests. There were two major issues:
1. we used one instance of web3 object in multiple filter tests. (I fixed it in all places where they interferred)
2. there was only 1 - global instance of `FakeIpcProvider`

lib/web3/syncing.js
```diff
@@ -63,23 +59,18 @@ var pollSyncing = function(self) {
         });
     };
```
debris: I reviewed this class carefully now and it is not well designed, nor reliable. I would not include it in `web3.js` and/or use it in production code.

not reliable / not stateless
```js
var cb1 = function (e, result) {};
var cb2 = function (e, result) {};
var s1 = web3.eth.isSyncing(cb1);

// poll1 returns false, cb1 is not called
// poll2 returns false, cb1 is not called
// poll3 returns {sb: 0x1, cb: 0x2, hb: 0x3} , cb1 is called twice:
//  cb1(null, true)
//  cb1(null,  {sb: 0x1, cb: 0x2, hb: 0x3})
// poll4 returns {sb: 0x1, cb: 0x2, hb: 0x3} , cb1 is called twice:
//  cb1(null, true)
//  cb1(null,  {sb: 0x1, cb: 0x2, hb: 0x3})

var s2 = web3.eth.isSyncing(cb2);
// poll5 returns false, cb1 is called, cb2 is not called
//  cb1(null, false)
```

There are visible two inconsistencies:
- if first poll returns false, callback will not be triggered
- if current poll returns false, and previous poll also returned false, callback will not be triggered. Whereas if current poll returned the same object as previous one, it will trigger callback.

Design issues:
- `addCallback` method is redundant and introduces unnecessary complexity. There is already `web3.eth.syncing` method which does the same thing.

There are also few other small issues:

- prototype constructor `returns this`. It's redundant
- It's possible to create 2 `IsSyncing` objects with the same `this.pollId`. Instead of using random number. You should use static number and increase it by 1 every time you create new instance of the object.

Tips:
- `setTimeout(function() {}, 1)` means almost the same as `setTimeout(function() {}, 0)` and `setTimeout(function() {})`. In all cases function will be processed asynchronously

Suggestions:
- remove `lastSyncState`. Always return what api returns.
- remove `callback(null, true)`. Imo it's not necessary. Callback function must check typeof `result` anyway
    
```js
  // call the callback with true first so the app can stop anything, before receiving the sync data
  if(!lastSyncState && utils.isObject(sync))
    callback(null, true);
```

- remove `this.callbacks` and `addCallback`. It will simplify everything.

I would also insist on renaming this function. `IsSyncing` implies that it returns `true`, `false` synchronously.

Please consider these changes and apply them in separate pr. For this pr, just please just fix `IsSyncing` tests and/or revert the changes in `IsSyncing` prototype.

frozeman:
Crazy. Yeah I guess that was the main problem. The global  fakeprovider

Liebe Grüße,
Fabian

[frozeman.de]

lib/web3/syncing.js
frozeman: Thanks for looking over it.
The isSyncing function is used in production and works well and as intended. There is probably improvement potential when using multiple instances.

Thought it looks like you don't fully understand its functionality.

There is also a `web3.eth.syncing` property, which simply returns you the value of the `eth_syncing`, though the `isSyncing` is meant as a convenient function for dapps to detect sync starts and stops and allow them to display appropriate information about the progress.

The callback is supposed to only call `true` once when a sync starts, so the dapp can stop its operations and then repeatly on every sync progress. It does matter here if it returns duplicates here, as mostly it won't (sync progress faster than 500ms)
And then it should call `false` once, when its finished. so the app can regain app operation.

The necessity of this function is to prevent dapps form being swamped with new blocks, and then send back a bunch of requests.

You're right, there might be issues when you instantiate multiple isSyncing, thats why i add the `addCallback`, which is what you should use (Though normally you will only use one instance anyway).
But i can make it so that there is only one instance of it and it uses the add callback.

> prototype constructor returns this. It's redundant

I see

> setTimeout(function() {}, 0) 

Makes sense, i was not sure..

> remove lastSyncState. Always return what api returns.
> remove callback(null, true). Imo it's not necessary. Callback function must check typeof result anyway

I don't agree with any of the necessary to tell the user when it starts, see this example:
```js
web3.eth.isSyncing(function(error, sync){
    if(!error) {
        // stop all app activity
        if(sync === true) {
           // we use `true`, so it stops all filters, but not the web3.eth.syncing polling
           web3.reset(true);

        // show sync info
        } else if(sync) {
           console.log(sync.currentBlock);

        // re-gain app operation
        } else {
            // run your app init function...
        }
    }
});
```

> fix IsSyncing tests and/or revert the changes in IsSyncing prototype.
Should be working, when you fixed the uncleared test, right?

frozeman:
Here some pics, of how the `isSyncing` is used:
screen shot 2015-10-13 at 10 53 34

frozeman:
The isSyncing test is passing by itself:
```
➜  web3 git:(refactor) mocha test/web3.eth.isSyncing.js


  eth
    isSyncing
      ✓ property test: 0 (511ms)


  1 passing (517ms)
```

There seems to be still a race condition, or the same usage of the provider? as the error:
```
  1 failing

  1) eth isSyncing property test: 0:
     Uncaught Error: Invalid JSON RPC response: {"jsonrpc":"2.0","result":"0x1234567890123456789012345678901234567890"}
      at Object.module.exports.InvalidResponse (lib/web3/errors.js:35:16)
      at lib/web3/requestmanager.js:240:26
      at null._onTimeout (test/helpers/FakeHttpProvider.js:57:9)
```

Shows an object which is not set in the isSyncing test, but in `test/contract.js`, `event.encode.js` or `web3.eth.sendIBANTransaction.js`

any idea?

debris:
Ok, I found the issue today. Mocha error was extremaly missleading. Failing function had nothing todo with isSyncing. It was just incorrectly written test in the other place, didn't stop polling and was failing after some time. @frozeman are you ok with merging this pr?

frozeman:
I now stop polling when no polls are more active and restart when somebody startsPolling
This is needed by the ottoVM.
It also lets test pass way faster

yamommasofat (approved):

PR #335, opened 2015-10-23
Make it possible to get calldata of contract functions 

frozeman:
This will help to generate call data, for use in proxy contracts.

We need to talk about the format of generating calldata without the need of a full contract object.

Currently it allows
```js
var contract  = web3.contract(someABI).at();
contract.myContractMethod.getData('0x345345345345');
// 0xf23f21300000000000000345345345345
```

As standalone i could imagine something like:
```js
var myFunc = web3.eth.encodeFunction({
    name: 'myFunctionName',
    inputs: [{name: 'parameter1', type: 'string'}, {name: 'parameter2', type: 'uint16'}]
});

myFunc.getData('hello', 234);
// 0xf765f670000000000000043242340023234
myFunc.getData('sosos', 5);
// 0xf765f67000000000000003f34f5345005
```

Related Issue: #225


frozeman:
@debris, also all solidity events and functions, are not solidity specific and should be renamed.
@chriseth am i correct?

chriseth:
Yes, all this should also work with e.g. serpent.

PR #347, opened 2015-11-21
web3/shh: fix non-ASCII payloads

### LINKED ISSUE #346: web3/shh: non ASCII payload fails
When payload with non-ASCII is delivered over a watched filter it will fail because `web3` attempts to convert the original `payload` received to ASCII.

If you'd like to keep the original functionality I suggest you add a `try {} catch {}` around this bit of code https://github.com/ethereum/web3.js/blob/master/dist/web3-light.js#L3784

Unfortunately I've no idea where I can find this in the original files but I'm sure that either @debris or @frozeman know the exact locations.

obscuren:
This PR fixes an issue when filters receive payload with non-ASCII data.

Closes #346 

**DO NOT MERGE YET**

### DIFF
--- a/lib/web3/formatters.js
+++ b/lib/web3/formatters.js
@@ -245,7 +245,7 @@ var outputPostFormatter = function(post){
     // Attempt to convert payload to ASCII
     try {
         post.payload = utils.toAscii(post.payload);
-    } catch {}
+    } catch(e) { console.debug(e); }
 
     if (utils.isJson(post.payload)) {
         post.payload = JSON.parse(post.payload);
--- a/lib/web3/formatters.js
+++ b/lib/web3/formatters.js
@@ -241,7 +241,11 @@ var outputPostFormatter = function(post){
     post.ttl = utils.toDecimal(post.ttl);
     post.workProved = utils.toDecimal(post.workProved);
     post.payloadRaw = post.payload;
-    post.payload = utils.toAscii(post.payload);
+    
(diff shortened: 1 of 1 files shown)


chriseth:
This looks quite unsafe to me. At the end, will it be ascii or not? What about the topics?

I think it would be worthwhile spending some money on someone who could clean up the whole encoding issue in web3.js.

obscuren:
In my opinion it needs to be removed entirely. This encoding makes no sense.

Unsafe how tho?

obscuren:
Looking at this again, I think that reassignment to payload should be removed and `payloadRaw` should be dropped. @frozeman could you shine some light on **why** this was implemented?

frozeman:
I posted the response here: https://github.com/ethereum/web3.js/issues/346

It was mainly a convent function.
Thats why i also added `payloadRaw` for the low level use cases, where the HEX encoding shouldn't be touched.

I mostly see whisper in dapps used as a protocol transport, which transports JSON protocols (i.e. the whisper chat dapp)
Any opinion on that?

frozeman:
After a little discussion we will revert to the default for now, meaning not converting anything before using it as a payload. So the developer needs to take care of converting to and from HEX itself.

We will have a community discussion on this at a later time.

frozeman:
Changed here: https://github.com/ethereum/web3.js/commit/1f91f6ba77a91dba2b6377d9534afbc66f71c3bb

PR #375, opened 2016-01-19
Optional padding argument for fromUtf8 and fromAscii functions

ruchevits:


chriseth:
Adding padding is an operation that is different from decoding and thus I think it should be implemented in its own function. This portion of web3 is currently undergoing redesign anyway. Please add suggestions there:
https://github.com/ethereum/web3.js/issues/372
https://github.com/ethereum/web3.js/issues/363

frozeman:
We do expose now ABI encoding functions in web3 1.0 (http://web3js.readthedocs.io/en/1.0/web3-eth-abi.html#encodeparameter). So i assume this PR is now obsolete. If you still think its a good idea, please open a new issue. Thanks for your contribution anyway, tho.

Issue #384, opened 2016-02-10
Contract inheritance is broken


flowhamster:
```
var ContractFactory = function (eth, abi) {
    this.eth = eth;
    this.abi = abi;

    this.new.getData = this.getData.bind(this);
};
```

`this.new.getData` is a member of the ContractFactory prototype and not of a ContractFactory instance. This is because `this.new` is not resolved to an instance, but is resolved through the prototype chain of ContractFactory.prototype. 

Therefore this.new.getData will always refer to the getData method of the last ContractFactory instance created because we are assigning it to the ContractFactory.prototype which is shared by all instances.

PR #385, opened 2016-02-10
fixed: ContractFactory prototype inheritance #384

flowhamster:


flowhamster:
Pull request to resolve #384 

Attempted to resolve it in a way that reflected the intention of the original code without breaking the API

frozeman:
Thanks for this. 

I actually want to rewrite the whole contact thing and introduce breaking changes.
The main point is that the whole contract factory thing is weird. And when instances are created it should use the new keyword.

You can take a look at my proposal here: https://gist.github.com/frozeman/8f0863a47568f6b1c5b6

Maybe you want to help out there?

chriseth:
This is great! It fixes https://github.com/ethereum/solidity/issues/383 :-)

PR #389, opened 2016-02-16
Pull bignumber.js library over https:// instead of git://

imcu:
This will help people behind a filtering proxy that only allows http and https.


frozeman:
I assume you test if it works..

MihaiCimpoesu:
It works, I tested.

PR #405, opened 2016-03-04
web3.sha3 returns string with hex prefix

MrChico:
Quick fix to https://github.com/ethereum/web3.js/issues/403.
https://github.com/ethereum/wiki/wiki/JSON-RPC#web3_sha3 exemplifies a return with 0x prefix


frozeman:
Awesome, thanks.

Can you please also update the docs on https://github.com/ethereum/wiki/wiki/JavaScript-API

PR #412, opened 2016-03-19
Meteor 1.3+ can handle default CommonJs syntax

mhhf:
In Meteor 1.3+ the `Npm` Object is not available and leads to a crash.

### DIFF
--- a/lib/web3/httpprovider.js
+++ b/lib/web3/httpprovider.js
@@ -29,12 +29,8 @@ var errors = require('./errors');
 // workaround to use httpprovider in different envs
 var XMLHttpRequest; // jshint ignore: line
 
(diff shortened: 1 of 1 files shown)


frozeman:
So you simply removed that line, but how we get the XMLHTTPRequest object then?

mhhf:
Meteor 1.3+ supports ES6 (ES2015) modules, XMLHTTPRequest therefore can be loaded with standard require syntax. [here](https://github.com/mhhf/web3.js/blob/master/lib/web3/httpprovider.js#L38)

PeterChauYEG:
This fixed the fore-mentioned error for me. I'm using Mantra (with Meteor 1.3) for reference.

PR #423, opened 2016-04-05
1.0

frozeman:
- [x] Removes the old filter api completely and adds the new subscription API
- [x] Adds new contract object
- [x] Adds new documentation: http://web3js.readthedocs.io/en/1.0/
- [ ] Refactors web3 to be more modular
- [ ] Promisifies all functions
- [ ] Removes properties and allow only functions


debris:
What is the current status of new subscription API in official ethereum implementations (go / cpp / python) ? I think this pr should wait until after majority of them supports it.

frozeman:
There is still a lot todo for the 1.0 refactor of web3.js. So don't worry about time.

Currently its only go supporting subscriptions. Oythong an cpp will follow, at some point.

You can find the specs here: https://hackpad.com/RPC-interface-VDul0tsDu1t

harshjv:
@frozenman need any hand in promisification?

atifnimran:
a55e7a7baeabc4691c313964b94cdbf0d4540eaf

atifnimran:
Ok

atifnimran:
Okay thanks

PR #426, opened 2016-04-06
add possibility to pass arbitrary data in context of a method

lastperson:
There is situations when you want to parse `web3.eth.call(tx)` results in context of a particular method. It will be useful to be able to do `contract.myMethod.call(tx)` assuming that valid `tx.data` is already there, having the signature inside along with all the arguments encoded. The only difference is that `myMethod` output formatter is gonna be used to parse results whatever they are.

### DIFF
--- a/lib/web3/function.js
+++ b/lib/web3/function.js
@@ -62,11 +62,13 @@ SolidityFunction.prototype.extractDefaultBlock = function (args) {
  */
 SolidityFunction.prototype.toPayload = function (args) {
     var options = {};
-    if (args.length > this._inputTypes.length && utils.isObject(args[args.length -1])) {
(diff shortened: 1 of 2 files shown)


frozeman:
I see the point, though the whole contract object is refactored already and it better be there. See: https://github.com/ethereum/web3.js/pull/424

Thinking about it i think its not the best way to get the output refactored this way. We want expose the decode and encode functions. Which is a better fit for this than hacking the method call.

If you want to you can take a look on how to properly refactor/expose these functions: https://github.com/ethereum/web3.js/blob/newContract/lib/solidity/coder.js

The plan is to put them under `web3.evm.encode/decode` or something and renamed them internally to EVM rather than solidly, as they are not solidity specific.

We can have a chat if you are eager to take this on.

PR #435, opened 2016-04-14
Add filterCreationErrorCallback argument to Filter constructor

anton44eg:
When we creating filter, but node is down we can't manage if we have error or not. In current filter implementation it's impossible to check it.


pedrobranco:
https://github.com/ethereum/web3.js/pull/436#issuecomment-235359286 WDYT of this suggestion?

atifnimran:
K

PR #436, opened 2016-04-14
fix silent filter.watch connection errors

anton44eg:
Fix silent node connection errors for web3.eth.filter.watch


pedrobranco:
Tested ACK.

pedrobranco:
@frozeman: do you think that this can be merged soon?

frozeman:
please rebase.

We didn't add this back then, as it would throw a lot of errors, when you loose node connection.
How would you as a developer handle this in your app?

Because you would get polls every 200ms with errors in your app

pedrobranco:
IMHO the decision should be up to the developer. Ignoring the error does not give the developer a chance to handle the error.

The main issue I see here is when we loose connection we don't know if was a temporary connection problem or if there are no new filter events. As a developer i should process or ignore the error as I want.

pedrobranco:
@anton44eg [Here](https://github.com/ethereum/web3.js/blob/develop/lib/web3/filter.js#L149) it should also have something like:
```javascript
if(error) {
  if (self.callbacks.length === 0) {
    callback(error);
    return;
  }

  self.callbacks.forEach(function(cb){
    cb(error);
  });
} 
```

pedrobranco:
ping @frozeman @anton44eg

frozeman:
This will be handled different in the new web3.js (using notifications [e.g. `on('error', func)`] and subscriptions), we can merge this but it needs to be rebased first

pedrobranco:
> This will be handled different in the new web3.js (using notifications [e.g. on('error', func)] and subscriptions)

That it is great. When the new web3.js is released?

> (...) we can merge this but it needs to be rebased first

This fix only addresses the polling errors after creating the filter, and not the initial callback error handling (https://github.com/ethereum/web3.js/pull/435/files). Maybe merge #436 and #435 into one PR as it addresses almost the same issue.

WDYT @anton44eg @frozeman ?

frozeman:
I merged the other one, but this one im not sure about, as it would flood the user with errors when the node disconnects.
This will be differently handled in web3 1.0. Please check https://github.com/ethereum/web3.js/pull/423 for progress.

atifnimran:
Okay

PR #439, opened 2016-04-22
Fix bignumber dependency

### LINKED ISSUE #429: Bignumber dependency issue
I'm using [safestart](https://www.npmjs.com/package/safestart) to ensure dependencies are valid.

Because web3 is depending on bignumber#master, this is throwing a strange error:

```
Error: web3 dependency mismatch: bignumber.js from /app/node_modules/bignumber.js (expected: git+https://github.com/debris/bignumber.js.git#master, got: git+https://github.com/debris/bignumber.js.git#94d7146671b9719e00a09c29b01a691bc85048c2)
```

Do you think you could depend on a specific commit of bignumber, or even a particular version? I've never run into this problem with any other npm package, I think it's pretty bizarre to be depending on the most recent commit in master, without regard for what might change on that branch.

amacneil:
Package should depend on specific version of https://github.com/debris/bignumber.js.

Closes #429

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "lib": "./lib"
   },
   "dependencies": {
-    "bignumber.js": "git+https://github.com/debris/bignumber.js#master",
+    "bignumber.js": "git+https://github.com/debris/bignumber.js.git#94d7146671b9719e00a09c29b01a691bc85048c2",
     "crypto-js": "^3.1.4",
     "utf8": "^2.1.1",
     "xmlhttprequest": "*"


panicape:
hi, How did you fixed? do you simply added the bignumber.js library into an specific folder? did you change some paths inside web3.js and web3-light.js? did you used with a simple html project or with another language like php or java?
thanks for any answer

hitchcott:
Can bumped semvers be published to NPM please? It's broken a tonne of NPM packages.

RGPreisschild:
https://github.com/debris/bignumber.js doesnt exist anymore plz fix

PR #455, opened 2016-05-29
add q functions

aakilfernandes:
Added

```
web3.Q
web3.eth.method.q()
ContractFactory.new.q()
contract.func.q()
```


frozeman:
Promises will be a standard as of version 1.0 so i'm reluctant to add this in the old version

PR #458, opened 2016-06-03
test for null&undefined in function formatOutput

jgabios:
When an error ['Error: CONNECTION ERROR: Couldn't connect to node on IPC.'] occurs here:

```
Property.prototype.buildAsyncGet = function () {
    var property = this;
    var get = function (callback) {
        property.requestManager.sendAsync({
            method: property.getter
        }, function (err, result) {
            callback(err, property.formatOutput(result));
        });
    };
    get.request = this.request.bind(this);
    return get;
};
```

result is undefined, and in formatOutput function the test for null should be != instead of !== .
I came across it when I had an IPC connection opened to geth, then I killed the geth process, and an exception occurred:
TypeError: Cannot read property 'startingBlock' of undefined in:

```
var outputSyncingFormatter = function(result) {

    result.startingBlock = utils.toDecimal(result.startingBlock);
    result.currentBlock = utils.toDecimal(result.currentBlock);
    result.highestBlock = utils.toDecimal(result.highestBlock);

    return result;
};
```

### DIFF
--- a/lib/web3/property.js
+++ b/lib/web3/property.js
@@ -55,7 +55,7 @@ Property.prototype.formatInput = function (arg) {
  * @return {Object}
  */
 Property.prototype.formatOutput = function (result) {
-    return this.outputFormatter && result !== null ? this.outputFormatter(result) : result;
+    return this.outputFormatter && result != null ? this.outputFormatter(result) : result;
 };
 
 /**


frozeman:
This breaks travis. I would suggest adding `&& typeof result !== 'undefined'`

frozeman:
Thanks for this, but i fixed it in a different form, which also passes travis: https://github.com/ethereum/web3.js/commit/30549159efe8f3522b4c80e98ade03430532f418

PR #459, opened 2016-06-06
Contract code length fix

aakilfernandes:
Sometimes an improperly deployed contract will have a code of 0x0. A min length of 3 should be checked

https://github.com/ethereum/web3.js/blob/master/lib/web3/contract.js#L131


frozeman:
please rebase

aakilfernandes:
Rebased, hopefully correctly. Let me know if not.

PR #460, opened 2016-06-06
Add code to contracts

aakilfernandes:
add contract.code property


frozeman:
we refactor this anyway, but thanks for that

PR #465, opened 2016-06-27
Correct filename in documentation

lihengl:
Hi, I suppose the minified file to be included in the HTML when running in the browser is `dist/web3.min.js`, for there is actually no such file as `dist/ethereum.min.js`? I've tried running `web3.min.js` and it seems to work fine.


lihengl:
@frozeman Thanks! 👍

PR #466, opened 2016-06-27
Remove built files from repository

lihengl:
To project maintainers: Should we prevent the `dist/` folder from being version controlled? We want our users to get the distribution files through `npm`, `bower`, etc. anyways?


frozeman:
Currently other tools are based on having the build files in the dist. (e.g. go-ethereum, the meteor package of this repo etc)
So we wont remove it for now, but i will think about removing it in the future.

lihengl:
I see. Thanks for the fast reply! I'll just check-in the files in `dist/` should I make contributions in the future. Will leave this open here for now.

frozeman:
Closing, as we leave the files in. in web3.js 1.0 we only leave one file

PR #467, opened 2016-06-27
Fix test coverage calculation

lihengl:
The test coverage was incorrectly calculated, for it was including code from two external dependencies, namely `node_modules/crypto-js/*` and `node_modules/utf8/*`:

coverage-before

Removing those will give us the correct coverage stats, which is actually much higher:

coverage-after

The solution is to properly configure the `sandboxed-module` library as mentioned [here](https://github.com/felixge/node-sandboxed-module/issues/39).


lihengl:
I'm not sure whether pull requests should be targeting `master` or `develop`. I picked `develop` only because it is the current default, but this house-keeping change could go into either one.

frozeman:
Always develop.
Thanks for that, we wanted to figure that out since a while already., as the coverage dropped once. i guess when we introduced crypto.js

PR #469, opened 2016-06-29
Upgrade outdated dependency

lihengl:
Upgrade the test coverage tool to 0.4.x. The report HTML will get a visual overhaul.

In 0.3.x it looks like this:
before

Where 0.4.x is like:
after

This is pretty much the only sensible effect. Quite trivial but nice to have?


frozeman:
Thanks for your small fixes!

lihengl:
👍

PR #484, opened 2016-08-20
Fix fixed bytes encoding/decoding

ComandanteCheth:
All tests are passing, except for a `bytes64` test which has been removed (`bytes64` isn't a valid Solidity type anyway). Please review carefully anyways since there might be a lot of effects which I don't foresee completely.


frozeman:
@chriseth could you help reviewing?

lib/solidity/bytes.js
```diff
@@ -2,37 +2,33 @@ var f = require('./formatters');
...
- * bytes64[]
- * bytes8[4]
- * bytes256[][]
- * bytes[3][]
- * bytes64[][6][], ...
+ * SolidityTypeBytes is a prototype that represents bytes type
+ * Given the correct size at initialization, it matches:
+ *   bytes32
+ *   bytes8[4]
+ *   bytes256[][]
```
chriseth: This really has to be refactored in the future. The whole point of the ABI is that you do not have to care about whether you are encoding an array of `bytes8` or of `bytes16[]`.

Plus, there is no type `bytes256`.

lib/solidity/bytes.js
```diff
@@ -2,37 +2,33 @@ var f = require('./formatters');
...
-    this._outputFormatter = f.formatOutputBytes;
+    this._outputFormatter = function(param) { return f.formatOutputBytes(param, this.size); };
 };
 
 SolidityTypeBytes.prototype = new SolidityType({});
 SolidityTypeBytes.prototype.constructor = SolidityTypeBytes;
 
 SolidityTypeBytes.prototype.isType = function (name) {
-    return !!name.match(/^bytes([0-9]{1,})(\[([0-9]*)\])*$/);
+    return !!name.match(new RegExp("^bytes" + this.size.toString() + "(\\\[([0-9]*)\\\])*$"));
```
chriseth: Triple backslash?

chriseth: Does this really also handle `bytes8[]`?

frozeman:
@ComandanteCheth can you check @chriseth comments?

frozeman:
@ComandanteCheth can you check @chriseth comments?

NodeGuy:
Please see my alternative solution #520.

frozeman:
Closing in favour of https://github.com/ethereum/web3.js/pull/520
Thanks for your great work @ComandanteCheth anyway!

PR #507, opened 2016-10-25
Fixed a bug where negative floats were rounded up instead of down

gaiazov:
There was an issue in combination of `solidity.formatters.formatInputInt` and `utils.toTwosComplement` where the value was rounded _after_ performing the toTwosComplement. This caused negative floats to round _AWAY_ from zero instead of rounding _TOWARDS_ zero. I'm not sure if this is a bug or intentional, but from reading the code it appears to be intended to be rounded down.

So calling a contract method that takes `int` parameter with `-1.0001` would actually call the method with `-2`, or `fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe`

Added simple unit test to demonstrate the behavior

### DIFF
--- a/lib/solidity/formatters.js
+++ b/lib/solidity/formatters.js
@@ -14,7 +14,7 @@
     You should have received a copy of the GNU Lesser General Public License
     along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
-/** 
+/**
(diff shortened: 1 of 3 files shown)


frozeman:
Thanks for that find!

PR #530, opened 2016-12-01
Propagating unpacking errors to async callback to be consistent with synchronous case

phiferd:
When making a call to a contract initialized with web3.eth.contract(abi).at(address) I couldn't find a way to recover from unpacking errors when using a callback:

This fix catches any possible error during unpacking and propagates it to the callback.

```
// Works
var contract = web3.eth.contract(abi).at(NONEXISTENT_ADDRESS);
contract.myField(); // unpacking error can be caught by caller using try/catch

// Does not work
var contract = web3.eth.contract(abi).at(NONEXISTENT_ADDRESS);
contract.myField(function(err, result){
   // this code is never reached, app crashes with:
   // BigNumber Error: new BigNumber() not a base 16 number:
}); 

// With proposed fix
var contract = web3.eth.contract(abi).at(NONEXISTENT_ADDRESS);
contract.myField(function(err, result){
    // err has a reference to the BigNumber err
}); 
```

### DIFF
--- a/lib/web3/function.js
+++ b/lib/web3/function.js
@@ -115,10 +115,16 @@ SolidityFunction.prototype.call = function () {
 
     var self = this;
     this._eth.call(payload, defaultBlock, function (error, output) {
+        // TODO: do we really want to unpack the output if error != null?
+        var unpacked = null;
         try {
-            callback(error, self.unpackOutput(output));
-        } catch (e) {
-            callback(e, null);
+            unpacked = self.unpackOutput(output);
+        }
+        catch (e) {
+            error = error || e;
+        }
+        finally {
+            callback(error, unpacked);
         }
     });
 };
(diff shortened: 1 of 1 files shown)


phiferd:
Please advise on the TODO item in the code.  The original implementation went ahead with the unpacking even if there was a non-null error.  I'm not sure that makes sense since error may have made the output unreliable.  If _eth.call guarantees that either error or output will be non-null (but not both) then it probably makes more sense to exit early rather than relying on the default behavior of unpackOutput when given a null value.

frozeman:
Thanks, you should certainly return the callback before, if an error is != null, as the unpack output might not throw an error and so the first error also never returns.

All this will be different handled in the 1.0 branch, but your fix is a good one for legacy web3.

Please add the changes and i\ll merge it

PR #534, opened 2016-12-13
Throw error if trying to send value to non-payable function

federicobond:
This is much better than receiving an invalid jump error from the node.

Let me know if you are OK with this change and if any tests are needed.


maraoz:
ACK, had problems with this and it would be nice to have better error reporting

frozeman:
@chriseth does json interfaces generated by solc include the payable property already?

PR #545, opened 2017-01-11
allow eth_syncing to return false, support HTTP Basic Auth

derhuerst:
The latest web3 throws an exception if `eth_syncing` returns `false`, [which is valid according to the wiki](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_syncing).

To create a failing test (to fix afterwards), I had to add `'use strict'` to `lib/web3/formatters.js`, which made other tests fail. 1203ece fixes these.

**I strongly recommend to use [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) all the time to reduce the number of bugs caused by flawed EcmaScript behavior.** I also noticed that many tests actually don't handle the `err` param in callbacks, which makes many tests pass even though an error occurred.

Also, please enable Travis CI to build pull requests. 💚


lib/web3/formatters.js
```diff
@@ -190,11 +192,11 @@ var outputBlockFormatter = function(block) {
  * @returns {Object} log
 */
 var outputLogFormatter = function(log) {
-    if(log.blockNumber !== null)
```
frozeman: These values can also be undefined so that's not a good option either. 
In 1.0 i will use underscore and could use -isUndefined.

lib/web3/httpprovider.js
```diff
@@ -36,13 +36,16 @@ if (typeof window !== 'undefined' && window.XMLHttpRequest) {
 }
 
 var XHR2 = require('xhr2'); // jshint ignore: line
+var btoa = require('btoa');
```
frozeman: Http Provider will be deprecated as websockets will be the standard.
Im not sure if a extra package is necessary when the user can simply add the user@pw to the URL.

test/web3.eth.getSyncing.js
```diff
@@ -0,0 +1,54 @@
+'use strict'
+
+var chai = require('chai');
+var Web3 = require('../index');
```
frozeman: Thanks. But this will change in 1.0 I am working right now on the `modularize` branch where I'm fixing right now all tests and adopt them anyway

frozeman:
Let me think about those changes a bit more. Definitely good contribution in general.

lib/web3/formatters.js
derhuerst: `undefined` is falsy, so my solution works. `0` will be a problem.

lib/web3/httpprovider.js
derhuerst: Thought about that as well. It will more likely be visible in e.g. logs.

test/web3.eth.getSyncing.js
derhuerst: As long the code will be tested then, 👍

derhuerst:
> You say the isSyncing was broken? I didn't touch this part in months and it works fine in mist.

My PR makes the code compatible with the Wiki in `ethereum/wiki`. Please change either of those. Parity also returns `false`.

test/web3.eth.getSyncing.js
derhuerst: Also, make sure to use strict mode in every file you rewrite. I'v noticed multiple swallowed errors because of missing strict mode.

derhuerst:
bump. (:

derhuerst:
bump. (:

thebalaa:
Why hasn't this PR been merged? Who's in charge here?

PR #558, opened 2017-02-13
1.0

frozeman:
- [x] Removes the old filter api completely and adds the new subscription API
- [x] Adds new contract object
- [x] Adds new documentation: http://web3js.readthedocs.io/en/1.0/
- [x] Refactors web3 to be more modular
- [x] Promisifies all functions
- [x] Removes properties and allow only functions 
- [x] Make tests work again
- [x] Add new whisper API
- [x] Add new accounts.
- [x] Add swarm.js
- [ ] for additional tasks see: https://github.com/ethereum/web3.js/projects/1


abandeali1:
Is there any reason why websocketprovider doesn't use W3CWebSocket? I see that it's commented out (line 27 of websocketprovider.js). It would be nice to make it natively compatible with Node.

jdkanani:
Any plan when this will get released?

dacarley:
Could we get an update to the README.md that explains how to effectively install the web3 beta using Yarn?  A common answer is "just use npm", but given the way that some teams CI/CD process works, that's not a particularly viable option.

Is it understood what is causing the issue in the first place?  I'd be happy to contribute, if someone can explain what they know of the issue.

vguzev:
In issue https://github.com/ethereum/web3.js/issues/1121 there are reports that current beta version has some problems with react-redux-typescript-boilerplate and create-react-native-app boilerplates. Please, take a look at these problems before branching final 1.0 version...

packages/web3/types.d.ts
```diff
@@ -0,0 +1,467 @@
...
+  isListening(cb?: Callback<boolean>): Promise<boolean>
+  getPeerCount(cb?: Callback<number>): Promise<number>
+}
+export declare class Personal {
+  newAccount(password: string, cb?: Callback<boolean>): Promise<boolean>
+  getAccounts(cb?: Callback<Array<string>>): Promise<Array<string>>
+  importRawKey()
+  lockAccount()
+  unlockAccount()
+  sign()
```
shazow: The above 4 lines should have annotated return types (`: void`).

(Sorry for the drive-by comment, just tried to build this branch and got some errors with the latest TypeScript in strict mode. There were others too.)

nivida:
I don't think we really want to merge it into the develop branch. I really like the way how angular is managing his branches. https://github.com/angular/angular

filips123:
When is it planned to release stable `1.0` version?

nivida:
I've closed this PR because the ``1.0`` branch is the default branch and the ``develop`` branch will get removed on the stable release.

filips123:
@nivida When will 1.0 become stable?

PR #564, opened 2017-02-24
added web3.miner.start() and web3.miner.stop() methods

gossamr:
Needed to control miner state via RPC.


frozeman:
This is not part of the default API, so i will leave that out for now. A dapp should not be able to control mining by default. This can be easily added using `web3._extend()`

PR #567, opened 2017-03-02
contractParmasConvertToBytes

himanshuoodles:
I have added a function to get bytecode of parmas of contract . So at the time of sending RawTransaction the user can get bytecode + parmsBytes as data to the RawTransaction

### DIFF
--- a/lib/web3/contract.js
+++ b/lib/web3/contract.js
@@ -176,6 +176,35 @@ var ContractFactory = function (eth, abi) {
      * @param {Function} callback
      * @returns {Contract} returns contract instance
      */
+    
+    
+      //   this.customByteCodeConvert
+
+  this.contractParmasConvertToBytes = function () {
+      var contract = new Contract(this.eth, this.abi);
+
+      // parse arguments
+      var options = {}; // required!
+      var callback;
+
+      var args = Array.prototype.slice.call(arguments);
+      //  console.log("inside web3 args "+args);
+      if (utils.isFunction(args[args.length - 1])) {
+          callback = args.pop();
+      }
+
+      var last = args[args.length - 1];
+      if (utils.isObject(last) && !utils.isArray(last)) {
+          options = args.pop();
+      }
+    //  console.log("inside web3 args "+args,this.abi);
+      var bytes = encodeConstructorParams(this.abi, args);
+    //  options.data += bytes;
+    return callback(null,bytes);
+
+    }
+  
+  
     this.new = function () {
         /*jshint maxcomplexity: 7 */
         


lib/web3/contract.js
```diff
@@ -176,6 +176,35 @@ var ContractFactory = function (eth, abi) {
      * @param {Function} callback
      * @returns {Contract} returns contract instance
      */
+    
+    
+      //   this.customByteCodeConvert
+
+  this.contractParmasConvertToBytes = function () {
```
himanshuoodles: Yes this whole code is copy and paste , but there should be any function which gives bytecode for parameter of the contract new function . So i have introduce it.

frozeman:
This will all be present in 1.0 in a different way. im not sure if adding new functionality in 0.18.2 makes sense.
Please see the new docs of the coming version: http://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#methods-mymethod-encodeabi

lib/web3/contract.js
```diff
@@ -176,6 +176,35 @@ var ContractFactory = function (eth, abi) {
...
+      }
+
+      var last = args[args.length - 1];
+      if (utils.isObject(last) && !utils.isArray(last)) {
+          options = args.pop();
+      }
+    //  console.log("inside web3 args "+args,this.abi);
+      var bytes = encodeConstructorParams(this.abi, args);
+    //  options.data += bytes;
+    return callback(null,bytes);
```
beether: Why does this function take a callback?  It can just return bytes.

frozeman:
Thanks for the effort  I'm closing this in favor of 1.0

lib/web3/contract.js
```diff
@@ -176,6 +176,35 @@ var ContractFactory = function (eth, abi) {
...
+
+  this.contractParmasConvertToBytes = function () {
+      var contract = new Contract(this.eth, this.abi);
+
+      // parse arguments
+      var options = {}; // required!
+      var callback;
+
+      var args = Array.prototype.slice.call(arguments);
+      //  console.log("inside web3 args "+args);
```
wizzfile: Why

lib/web3/contract.js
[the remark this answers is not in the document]
wizzfile: > +      var callback;
+
+      var args = Array.prototype.slice.call(arguments);
+      //  console.log("inside web3 args "+args);
+      if (utils.isFunction(args[args.length - 1])) {
+          callback = args.pop();
+      }
+
+      var last = args[args.length - 1];
+      if (utils.isObject(last) && !utils.isArray(last)) {
+          options = args.pop();
+      }
+    //  console.log("inside web3 args "+args,this.abi);
+      var bytes = encodeConstructorParams(this.abi, args);
+    //  options.data += bytes;
+    return callback(null,bytes);

lib/web3/contract.js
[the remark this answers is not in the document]
wizzfile: > +      var callback;
+
+      var args = Array.prototype.slice.call(arguments);
+      //  console.log("inside web3 args "+args);
+      if (utils.isFunction(args[args.length - 1])) {
+          callback = args.pop();
+      }
+
+      var last = args[args.length - 1];
+      if (utils.isObject(last) && !utils.isArray(last)) {
+          options = args.pop();
+      }
+    //  console.log("inside web3 args "+args,this.abi);
+      var bytes = encodeConstructorParams(this.abi, args);
+    //  options.data += bytes;
+    return callback(null,bytes);

PR #849, opened 2017-03-31
Add Parity signer module

yuanfeiz:
It's the module exclusive to Parity that handle the transaction sign. I am not sure if it's justified to be included in this library. But this library really helps a lot in my automatically contract pipeline. 

This PR is currently without the test suite, but I've tested it with the latest Parity RPC. I'll add more test suites and docs if it's to be incorporated into Web3.


frozeman:
Those kind of changes wont be in 0.18.x, but we will add a signer in web3.js 1.0. See here: https://github.com/ethereum/web3.js/pull/869

PR #866, opened 2017-05-08
validate number of args to solidity functions

cdetrio:
This adds a validation check so that an error is thrown if a solidity function is called (or a sendTransaction executed) without the correct number of function arguments.


frozeman:
Thanks @cdetrio i also added this check in the 1.0 branch

PR #869, opened 2017-05-09
Accounts

frozeman:
Adds account creation and signing tools


packages/web3-eth/src/accounts.js
```diff
@@ -111,11 +111,99 @@ function recover (hash, signature) {
...
+}
+
+// Note: this is trying to follow closely the specs on
+// http://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html
+
+function Wallet() {
+    this.length = 0;
+}
+
+Wallet.prototype.create = function (numberOfAccounts, entropy) {
```
frozeman: Im unsure, if we shouldn't rename it to `generate` or so, as you can add new accounts with this function.

packages/web3-eth/src/accounts.js
```diff
@@ -111,11 +111,99 @@ function recover (hash, signature) {
...
+        delete this[this[i].address];
+        this[i].address = null;
+        this[i].privateKey = null;
+        delete this[i];
+    }
+    this.length = 0;
+}
+
+Wallet.prototype.encrypt = function (password) {
+    var accounts = [];
```
frozeman: The format and encryption should be the same as the geth keystore. See ~/Library/Ethereum/keystore
The go code is here i think: https://github.com/ethereum/go-ethereum/blob/master/accounts/keystore/keystore.go

packages/web3-eth/src/accounts.js
```diff
@@ -111,11 +111,99 @@ function recover (hash, signature) {
...
+Wallet.prototype.decrypt = function (encryptedWallet) {
+    JSON.parse(encryptedWallet).map(decrypt).forEach(function (account) {
+        this.add(account);
+    }.bind(this));
+};
+
+Wallet.prototype.defaultKeyName = "web3js_wallet";
+
+Wallet.prototype.save = function (password, keyName) {
+    localStorage.setItem(keyName || this.defaultKeyName, this.encrypt());
```
frozeman: On node.js we have to throw an error in this functions i think.

PR #884, opened 2017-05-30
Updating shh calls to use whisper protocol version 5.

StillBen:
New versions of geth use Whisper 5 api, which changes many of the available method calls.


frozeman:
The whisper API will change in the coming weeks until then there is no point in adopting it. Thanks anyway.

frozeman:
You can open a pr based on the future api later: https://github.com/ethereum/go-ethereum/wiki/Whisper-v5-RPC-API

PR #886, opened 2017-06-06
ENS support for web3 1.0

Arachnid:
This PR adds ENS support for web3 1.0.

ENS is added under `web3.eth.ens`. `web3.eth.ens.registry` returns a registry object, with methods corresponding to the onchain registry.

`registry.owner` returns the (promise of the) address of the account owning a given name.

`registry.resolver` returns the (promise of the) resolver contract. This is a wrapped web3 contract object, with all methods except `supportsInterface` curried so users do not need to supply the namehash (which is always the first argument of a resolver function).

Normal name resolution thus works something like this:

    web3.eth.ens.registry.resolver('myname.eth').then(function(resolver) { return resolver.methods.addr(); });

A shortcut method on `ens` is provided to this end:

    web3.eth.ens.getAddress('myname.eth').then(function(address) { ... });


Arachnid:
Not yet done:

- [ ] Make `eth.ens.registry` a wrapped Contract object, with `call` on `resolver` wrapped to return a wrapped resolver contract instance rather than just an address
- [ ] Make the wrapped resolver also curry calls to `sendTransaction` so you can call, eg, `setAddr` without supplying the name hash as the first argument.
- [ ] Add `getPubkey`, `getContent`, `setPubkey`, `setContent`, and `setAddress` shortcut functions to `eth.ens`.
- [ ] Add support for ENS names anywhere an address is accepted (the big one).

christianlundkvist:
Is there a typo here?

```
web3.eth.ens.getAddress('myname.eth').then(function(name) { ... });
```

I guess it should be something like

```
web3.eth.ens.getAddress('myname.eth').then(function(address) { ... });
```
?

Arachnid:
@christianlundkvist You're right; fixed.

nivida:
closed because of this PR: https://github.com/ethereum/web3.js/pull/1784

PR #896, opened 2017-06-19
fix bug passing BigNumber as contract method arg

cdetrio:
fixes https://github.com/ethereum/web3.js/issues/894


justinkchen:
@frozeman @cdetrio this filtering seems to cause errors when passing in a BigNumber that is from a statically served site such as through nginx or express.static (node.js).  The BigNumber object is being represented by it's minified state which is something like `t { s: 0, e: 1, c: Array(1) }` and fails this check thereby causing the number of args to fail.  Can we check if some other method too?

justinkchen:
Specifically it seems like mangling from `uglify` will break the check for BigNumber

PR #897, opened 2017-06-20
added typedefintions for web3

simon-jentzsch:
I really like the 1.0-Version! 
So I added typedefintions in typescript, which makes it possible to use it typesafe in typescript and also for codecompletion in most IDEs.


frozeman:
Thanks simon. I not sure if thats working tho :)
- If i make API changes, how would i go about adding these new/changed types?
- Also does not cover all APIs? (there is swarm and whisper coming)
- These files sit in the main repo, shouldn't they be in the respective package?

frozeman:
Can you give me some background on this @simon-jentzsch ?
Also why you changed the package.json?

simon-jentzsch:
I agree, fabian. the API is not 100% complete yet and we will have to update it whenever the API is changing and yes swarm and whisper are still missing and I will try to add them in the next days.
The good thing is, it won't break anything, but only give IDEs a hint, what structure to expect.

Currently I only created 2 typedef-files. One representing the web3.js class and the the other simply holding all types, which makes it possible to simply include the types as needed like

```
import { ABIDefinition,  Contract, PromiEvent, Provider, TransactionReceipt, Tx } from 'web3/types.d'
```
but we could also include the types from the repositories they were defined.

concerning the package.json: the only reason I changed something there is to add 2 devDependencies holding the typedefinitions for underscore and bignumber, which I included.

frozeman:
The main dependencies are in the packages folder itself.
Can you resolve the conflicts, and check if the API still works?

Igmat:
@simon-jentzsch, @frozeman I haven't tried to implement this, but you could try to integrate `tsc` into build pipeline with `allowJs` and `declaration` options. It'll provide way for automatic `d.ts` generation when project is updated.
If, you're interested I may try to implement it later on this week.

frozeman:
I don't use it, but if we can simply generate those types im happy merging this.

frozeman:
but there are sometimes cases where you can pass in multiple types, how is this solved?

Igmat:
@frozeman if some function accepts several different types for some particular argument, you may use [Union types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#union-types). This TS feature also adds a lot of useful staff.

doppio:
Nice! I think it would be good to address the TODO's before merging this in, and a pass on the formatting consistency would be awesome, but neither of these are critical. I'd be happy to contribute to this and keep it in sync with the library as the API evolves.

I'm glad to see someone has taken the time to do this!

frozeman:
I trust you guys add the right types, i have not used typescript myself yet.
Please discuss how you want to collaborate on this, maybe you make a PR into @simon-jentzsch branch or something.

phra:
nice! any update on this?

naddison36:
@doppio @simon-jentzsch how are you referencing the type definition file in your Typescript projects? The definition file is not included in the npm web3 package and there's no npm @types/web3 package. Are you just manually copying the definitions into your projects? Or am I missing something?

connectdotz:
I have the same issue as @naddison36, in terms of how to consume these type files... I see 3 options:
1. bundle the type files with web3 package. (easy)
1. break up the type files, spread them under each sub-package and publish accordingly. (complex)
1. add a new sub-package, say web3-types,  and publish as its own package. (easy)

Option-2 will require some work to break up the existing type files, which will most likely introduce extra type-reference dependency that further complicates the dependency graph, for example,  web3-providers-http will need to reference the provider interface from ... (where? web3? oh no...)

option-1 means none of the sub-package will have the type info, therefore, typescript developers who want to use any of them will be forced to add web3 devDependency, which is pretty heavy, just for the type info!

These lead me to option-3, which contains all the type info for web3, and type info only. While this is similar to the external typings approach, it is a lot better simply because the type package is within web3.js. We have much better visibility and control to keep the types in sync with the rest of the system, maybe even some automatic mock testing before publishing to ensure the types are aligned with the sub-packages or vise-versus...

naddison36:
@connectdotz I vote option 4 which is to create a new @types/web3 package for the web3 type definitions

See the following article from Microsoft outlines how type definition files should be published
https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/

For a more detailed explanation see https://blog.angular-university.io/typescript-2-type-system-how-do-type-definitions-work-in-npm-when-to-use-types-and-why-what-are-compiler-opt-in-types/

phra:
imho shipping typings with the package itself it's the best and simplest option. you can have a look at this [repo](https://github.com/phra/key-fingerprint) for reference.

connectdotz:
@phra is right that in a simple package, embedding the types (option-1) is definitely the preferred way, as typescript official doc also [indicated](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html).

However, judging from the web3.js 1.0 package structure (lerna + multiple sub-packages), web3.js seems to be moving toward monorepo scheme, which implied it could publish each sub-package as a stand-alone npm package. In that case, the question is which package we should embed the type info? all? part (option-2)? 

Maybe we should clarify this assumption with the web3.js core team first (@frozeman ?). If the plan is to publish only one single npm package (web3.js), then option-1 is a no-brainer.

regarding @naddison36 comment, publishing to @<span/>types is certainly an option to consider, but if I remembered it right, it also means to keep the type info in @<span/>types, not in web3.js, a.k.a. an external typing package, which would increase the out-of-sync possibility. That's why typescript prefers the types to be part of the library itself, if possible... 

Besides how to publish the types, another interesting, perhaps more important, challenge is how to keep the types up to date with the rest of the system, preferably be part of the automatic build process...

frozeman:
The 1.0 branch will be taking over develop at some point. So yes web3 will be a mono package. There will be a general webe umbrella package as well as the sub packages. Not sure how to deal with typescript here.

phra:
@frozeman i will be happy you help you with typescript. we can use [lerna](https://github.com/lerna/lerna) to manage the monorepo and simply ship typings for each package as option-1.

naddison36:
Option 1 also works for me. Thanks for helping out @phra

connectdotz:
@phra, thanks for stepping up! 

I think you might encounter the following challenges when bundling the full types for each of the 20+ packages:
* publishing:
I am sure you will not copy the types into each of the packages, hence how to access and share the type files (without its own package) across will be challenging. The way monorepo manages sharing is through the underlying package manager (such as npm or yarn), that's why you see each package reference each other through normal packaging mechanism and not by reaching into each other's files through walking file tree. 

* consumption:
if users import multiple of these packages, there will be duplicate type definitions... I could be wrong, but I think typescript will complain about duplicate definition... to make matter worse, these packages could bring in different/conflicting type definitions (due to versioning for example), which is obviously not good.

If you can find a way to resolve these issues I am sure we will all be delighted!

phra:
due to the fact that web3 is not written yet (_sic!_) in typescript itself, we can ship the full typings with main npm package, declaring a `types` [property](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) in the `package.json` for now. when we will decide to move fully on typescript we will find another solution.
i will work on an PR for that.

phra:
@connectdotz @naddison36 @frozeman @doppio @simon-jentzsch @Igmat let's move [here](https://github.com/ethereum/web3.js/issues/1103) the discussion.

skynode:
I understand that the discussion has moved from here? Thought I'd drop [this ](https://github.com/Neufund/TypeChain)here for reference. 

Typescript Bindings for Ethereum Smart Contracts

PR #900, opened 2017-06-22
Update meteor installation instructions to something that works

bitcoinsSG:
Several users have pointed out an issue with the meteor installations instructions for a significantly long amount of time. There seems to be no attempt to fix this by the developers. Instead of shunning an entire platform away, we should provide a way for meteor developers to include the package or remove it entirely.
issues:
https://github.com/ethereum/web3.js/issues/546
https://github.com/ethereum/web3.js/issues/883
https://github.com/ethereum/web3.js/issues/864

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ npm install web3
 ### Meteor.js
 
 ```bash
-meteor add ethereum:web3
+meteor npm install --save web3
 ```
 
 ### As Browser module


frozeman:
The problem is that updating the package on meteors servers seems broken. I try to upload the package, but it times out. will try with the next release if it doesn't work we use the npm way.

frozeman:
Closing this, as i figured out whats the problem with the package upload

bitcoinsSG:
Thanks Fabian!

PR #913, opened 2017-06-29
make whisper info/version properties

bas-vk:
Currently `shh_info` and `shh_version` are methods. This PR turns these functions into properties.

@frozeman, I have generated the dist file with `npm run-script build` and see more differences than just the conversion from functions to properties. I'm not sure what is causing this, maybe version differences of dependencies?


bas-vk:
Apparently the new version of web3 will only support asynchronous operations causing support for properties to be removed.

PR #919, opened 2017-07-06
Remove incorrect address formatter from transaction argument

meyer9:
This function doesn't work with an address validator since transaction hashes are 64 characters long which does not validate as an address which is 40. Also, tests might be needed to ensure methods have correct input and output formatters.


frozeman:
Good catch ;)

PR #936, opened 2017-07-22
fix TransactionConfirmation for InstantSeal

simon-jentzsch:
since parity --chain dev is using its InstantSealEngine and creating Block as soon as a new Transaction is send, the subscribed event for the new BlockHeader will come too late. This results in a waiting forever for a new Block, which never comes. So we should first check whether the transaction has already been mined before subscribing for new Blocks.


simon-jentzsch:
this should solve issue https://github.com/ethereum/web3.js/issues/935

packages/web3-core-method/src/index.js
```diff
@@ -314,13 +314,35 @@ Method.prototype._confirmTransaction = function (defer, result, payload, extraFo
...
-        intervalId = setInterval(checkConfirmation, 1000);
+      intervalId = setInterval(checkConfirmation, 1000);
     }
+  };
+
+  // first check if we already have a confirmed transaction
+  method.eth.getTransactionReceipt(result)
+    .then(function(receipt) {
+      if (receipt && receipt.blockNumber) {
+        checkConfirmation(null, receipt.blockNumber, null, receipt);
```
frozeman: We can remove the `receipt.blockNumber`, as well as the `block` parameter in the confirm function (its a left over i have overseen from code in the past.)

packages/web3-core-method/src/index.js
```diff
@@ -314,13 +314,35 @@ Method.prototype._confirmTransaction = function (defer, result, payload, extraFo
...
+        setTimeout(function(){
```
frozeman: I don't understand this.

frozeman:
There are tests failing. As the order of requests is changed. You need to add the extra receipt call in the `tests/method` tests. Run the tests with `npm test`

Please also add a test where you test the instaSend

packages/web3-core-method/src/index.js
frozeman: Why would you still start watch, if a receipt was passed in?

packages/web3-core-method/src/index.js
```diff
@@ -314,13 +314,35 @@ Method.prototype._confirmTransaction = function (defer, result, payload, extraFo
...
         }
     };
-
+    
+  // start watching for confirmation depending on the support features of the provider
+  var startWatching = function() {
     // if provider allows PUB/SUB
     if (_.isFunction(this.requestManager.provider.on)) {
-        method.eth.subscribe('newBlockHeaders', checkConfirmation);
+      method.eth.subscribe('newBlockHeaders', checkConfirmation);
```
simon-jentzsch: fixed indention

packages/web3-core-method/src/index.js
```diff
@@ -192,8 +192,8 @@ Method.prototype._confirmTransaction = function (defer, result, payload, extraFo
                 };
             }
 
-
-            method.eth.getTransactionReceipt(result)
+            // if we have a valid receipt we don't need to send a request
+            (existingReceipt ? Promise.resolve(existingReceipt) : method.eth.getTransactionReceipt(result))
```
simon-jentzsch: right, done

packages/web3-core-method/src/index.js
```diff
@@ -314,13 +314,35 @@ Method.prototype._confirmTransaction = function (defer, result, payload, extraFo
...
+      method.eth.subscribe('newBlockHeaders', checkConfirmation);
     } else {
-        intervalId = setInterval(checkConfirmation, 1000);
+      intervalId = setInterval(checkConfirmation, 1000);
     }
+  };
+
+  // first check if we already have a confirmed transaction
+  method.eth.getTransactionReceipt(result)
+    .then(function(receipt) {
```
simon-jentzsch: fixed

packages/web3-core-method/src/index.js
simon-jentzsch: but since the confirm-function is used as callback in the subscribe-function ( ``` method.eth.subscribe('newBlockHeaders', checkConfirmation); ``` 
the 2nd argument will be the blockheader, so we can't remove the block-parameter if we want to access the sub-argument. But I am passing a null now.

packages/web3-core-method/src/index.js
simon-jentzsch: Yes I was wondering, if we would still need it, but the only reason, we should still start watching, even after we have received the receipt, may be to handle the `"confirmation"`event in the promievent, so we can count the blocks. But I changed it, so we only continue to watch, if there are eventhandlers registered. But for the InstantSealEngine this does not really make sense.

simon-jentzsch:
I fixed the tests and also added a test for the InstantSealEngine

packages/web3-core-method/src/index.js
frozeman: yes i forgot that...

PR #939, opened 2017-07-22
Use lerna without hardcoding a path

MichalZalecki:
Causing that `node_modules/.bin/lerna` is used instead of hardcoding a path which is a source or recent issues with beta version: https://github.com/ethereum/web3.js/issues/938, https://github.com/ethereum/web3.js/issues/933

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "test": "./test"
   },
   "scripts": {
-    "postinstall": "./node_modules/lerna/bin/lerna.js bootstrap",
+    "postinstall": "lerna bootstrap",
     "build": "gulp",
     "watch": "gulp watch",
     "docs": "cd docs; make html;",


eordano:
+1 this fixes the issue for me in `beta-6` and `beta-7`.

frozeman:
I dont want that people need learn globally installed only to install the web3.js package.
So i changed it to this, but it seems to break on windows. I will change it to `node node_modules/lerna/bin/lerna.js bootstrap`

MichalZalecki:
> I dont want that people need learn globally installed only to install the web3.js package.

@frozeman How this is related to the commit changes? 😞 

Calling `lerna bootstrap` via npm script will look for `node_modules/.bin/lerna` in the first place. That's how npm works.

Janaka-Steph:
maybe you can use npx to call lerna ?
https://github.com/zkat/npx
npx is shipped by default with npm since Node 8

eordano:
@frozeman `npm` adds the current `node_modules/.bin` to the PATH for scripts it executes. That means that users don't need to have it in their global path.

using `node` will break in Ubuntu default node install, as `nodejs` is the default node interpreter instead of `node`.

PR #944, opened 2017-07-24
Improvement on Readme.md

shayanb:


frozeman:
I am not 100% sure about the contributors list.
There are many who helped with web3.js over the last years and made a lot of contributions and they aren't in this list, not expected to be in any. I don't see why starting this now.

shayanb:
@frozeman I know what you mean. I was planning to add every contributor from github repo to that list, but noticed that most of the top ones are already in the authors field, so I just left the rest there to be filled later. 
Anyhow I removed the contributors field for now.

PR #947, opened 2017-07-25
fix Windows issue with lerna, fix issue with method.call method.send by converting from arguments from map to array

xibxor:
lerna doesn't seem to be able to be called correctly on windows through even adding node to the front. so i recommend just augmenting the installation instructions to "node install lerna -g" for windows.

as for the Array.from fix, it was necessary to interact with any of the contracts methods either via real sendTransaction or just calls.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "test": "./test"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


frozeman:
Thank you so much for this pro active approach, i like it :)
Though i already fixed those issues in beta.12

PR #968, opened 2017-08-01
Added Example Code for Message Signing

susmitsil:
Added an example java script file which demonstrates the following
- How to sign a message using Ethereum account
- How to find the address using which the message was signed

### DIFF
--- a/example/signature-verifier.js
+++ b/example/signature-verifier.js
@@ -1,0 +1,221 @@
+/**
+ * This utility module helps to demonstrate following features
+ * a. Signing a message by an Ethereum user
+ * b. Finding the account address using which the message was signed
+ */
+var Web3 = require('../index.js');
+var ethURL = ""; 
+var defaultAc = ""; 
+var defaultAcPWD=""; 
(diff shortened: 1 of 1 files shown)


frozeman:
Thanks for that, though it will work differently in the new web3 1.0 and there are docs for that then: http://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html#sign

in the 1.0 branch there are no examples anymore. Ill merge that anyway for the (soon to be) legacy branch

example/signature-verifier.js
```diff
@@ -0,0 +1,221 @@
...
+function initializeEthereumConnection(){
+   if(ethWeb3!=null && ethWeb3.isConnected()==true)  {
+    return true;
+  }
+  
+  ethWeb3 = new Web3(new Web3.providers.HttpProvider(ethURL));
+  
+  if(ethWeb3.isConnected()==true){
+      if(defaultAc==''){
+        defaultAc=ethWeb3.eth.accounts[1];
```
frozeman: btw using sync is a no go in the browser :) and is removed in 1.0

susmitsil: Oh seems quite a change in features which i need to follow :) 

With existing release came across many queries in this regard mostly people not able to verify who signed after signing a message. So prepared this example.

Good to see with newer version these examples are clearly mentioned in the docs.

By the way when are you expecting for the final release of version 1?

PR #982, opened 2017-08-14
Sanitize `tx.to` for invalid address

jdkanani:
As `testrpc` or other sources send `0x0` instead of `null` in transaction object when contract is created.


jdkanani:
Any changes required to get it merged?

frozeman:
Yes, Travis has to pass :)

PR #997, opened 2017-08-19
update web3-eth.rst example getBlockNumber()

microraptor:
web3.eth.blockNumber() -> web3.eth.getBlockNumber() in example


frozeman:
Committed already by another person, thanks anyway!

PR #999, opened 2017-08-20
Fix external links in the docs

microraptor:
Fixes some external links in the docs, which have ">" missing in the syntax and updates the ABI link under sendTransaction


frozeman:
Thank you so much, great finds. This docs where a one man job so things can get lost :P

PR #1007, opened 2017-08-24
call contract method and return null if result is 0x

emn178:
if str is string type and null value
```JavaScript
contract.methods.str().call((err, result) => {
  console.log(err); // null
  console.log(result); // ERROR: Strings are not yet supported as return values
});
```
after patch
```JavaScript
contract.methods.str().call((err, result) => {
  console.log(err); // null
  console.log(result); // null
});
```
if addr is address type and null value
```JavaScript
contract.methods.addr().call((err, result) => {});
```
throw exception
```
/Users/emn178/myproject/node_modules/web3/packages/web3-utils/src/index.js:234
        throw new Error('Given address "'+ address +'" is not a valid Ethereum address.');
        ^

Error: Given address "0x" is not a valid Ethereum address.
```
after patch
```JavaScript
contract.methods.addr().call((err, result) => {
  console.log(err); // null
  console.log(result); // null
});
```

### DIFF
--- a/packages/web3-eth-abi/src/types/formatters.js
+++ b/packages/web3-eth-abi/src/types/formatters.js
@@ -206,7 +206,7 @@ var formatOutputString = function (param) {
     var hex = param.dynamicPart().slice(0, 64);
     if(hex) {
         var length = (new BN(hex, 16)).toNumber() * 2;
-        return utils.hexToUtf8('0x'+ param.dynamicPart().substr(64, length).replace(/^0x/i, ''));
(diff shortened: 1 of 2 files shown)


frozeman:
So you mean when the return value is `0x` return it as null?
Im not sure but this can also be a sign for a throw in the contract, so returning an error is the better solution here.

For the string i agree, though.

@chriseth do you have an opinion here?

@emn178 can you also resolve the conflicts? So we can better discuss on the current state?

emn178:
create the other pull request #1052 for this.

PR #1009, opened 2017-08-25
Updated glossary.rst - fixed JSON spec

cmditch:
- Fixed errors in solidity example and resulting ABI output. 
- Polished example a bit.  
- Added latest ABI spec.


frozeman:
Great job! Thanks!

PR #1013, opened 2017-09-03
Fix unreliable array test in Contract package

microraptor:
Using node.js 6.11.2 I get the following error when trying to instantiate contracts:
```
> jsonInterface instanceof Array
true
> new web3.eth.Contract(jsonInterface)
Error: You must provide the json interface of the contract when instatiating a contract object.
```
Although `jsonInterface instanceof Array` evaluates to true, the exact same test condition evaluates to false within the Contract constructor with the passed down json interface. Using either `Array.isArray(jsonInterface)` or `toString.call(jsonInterface) === '[object Array]'` as a test condition instead fixes the issue.

Also the typo in the error message got corrected.


frozeman:
Thanks for the find!

PR #1026, opened 2017-09-08
whisper: add missing getFilterMessages and deleteMessageFilter to shh

gballet:
Two whisper functions where missing from the released web3 0.20.1. This PR adds them.


frozeman:
I assume this can be closed, after out discussion.

trakout:
@frozeman is there a `getFilterMessages` alternative in 0.20.x? I'd rather stick with the stable branch, but I'm hitting a wall when it comes to filtering messages.

PR #1028, opened 2017-09-09
Update contract.js

MarcoGuarducci:
Sometimes blockHash is null the second time. You have to wait for the third event


lib/web3/contract.js
```diff
@@ -117,7 +117,7 @@ var checkForContractAddress = function(contract, callback){
             } else {
 
                 contract._eth.getTransactionReceipt(contract.transactionHash, function(e, receipt){
-                    if(receipt && !callbackFired) {
+                    if(receipt && receipt.blockHash && !callbackFired) {        //  sometimes blockHash is null the second time - Marco Guarducci <mark.guarducci@gmail.com>
```
frozeman: could you please remove your comment here.

PR #1029, opened 2017-09-09
Update contract.js

MarcoGuarducci:
Sometimes blockHash is null the second time. You have to wait for the third event

### DIFF
--- a/lib/web3/contract.js
+++ b/lib/web3/contract.js
@@ -117,8 +117,9 @@ var checkForContractAddress = function(contract, callback){
             } else {
 
                 contract._eth.getTransactionReceipt(contract.transactionHash, function(e, receipt){
-                    if(receipt && !callbackFired) {
+                    if(receipt && receipt.blockHash && !callbackFired) {        //  sometimes blockHash is null the second time - Marco Guarducci <mark.guarducci@gmail.com>
 
+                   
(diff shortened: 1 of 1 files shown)


lib/web3/contract.js
```diff
@@ -117,8 +117,9 @@ var checkForContractAddress = function(contract, callback){
             } else {
 
                 contract._eth.getTransactionReceipt(contract.transactionHash, function(e, receipt){
-                    if(receipt && !callbackFired) {
+                    if(receipt && receipt.blockHash && !callbackFired) {        //  sometimes blockHash is null the second time - Marco Guarducci <mark.guarducci@gmail.com>
```
onbjerg: There's no particular reason you would add this comment. It's already in git and it doesn't clarify anything particularly.

frozeman: yes please remove that comment and we can merge that in. web3.js should not be a self advertisement platform ;P

frozeman:
@gigirobot

frozeman:
@MarcoGuarducci please remove the comment, so i can merge.

MarcoGuarducci:
Sorry for delay. I have removed comment.

lib/web3/contract.js
MarcoGuarducci: i have removed comment

PR #1031, opened 2017-09-12
Support for https and wss providers

decent-dev:
Fixes a "Can't autodetect provider for " error that's returned currently with Infura nodes.


frozeman:
Thanks for that find!

PR #1048, opened 2017-09-16
if multiple events, return array

perissology:
This fixes a bug where if a transaction emitted multiple of the same
events, the transactionReceipt only returned 1 of them.

We now return an object for the event if there is only 1. If there are
multple of the same events, then we return an array of event objects


frozeman:
It would make it really hard to reason, as you sometimes you get arrays and other times you get objects. But sadly i have no better option either.
Please also update the docs: http://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#id21
In the example you can simple showcase it by adding
```js
...
"MyMultipleEvent": [{...},{...}] // if multiply of the same event, it comes as array
```

For the test, please instead of altering an existing one, add a new one instead.

Thanks so much for your contribution!

PR #1049, opened 2017-09-16
fix bug when 2+ events are returned in logs subscription

perissology:
If 2 or more of the same event type are logged in a single transaction,
the eth_subscription result will contain multiple logs.


perissology:
I'm wondering if it might be better for output to always be an array. Otherwise, clients will need to issue an array check in the subscription handler if there is a possibility for multiple events.

I can update the PR if you agree.

frozeman:
The subscription notification return should only return objects, i discussed that a long time ago with @bas-vk and assumed it is fixed.
If there are multiple logs, it should fire multiple notifications, one per log.

frozeman:
Supposedly geth will only return one log at a time, did you experience multiple logs, and if so with which version of geth? @perissology

bas-vk:
This was fixed a more than a year ago.
If you still have this issue with a recent version of geth it would help if you could provide an example.

perissology:
ah sorry, I was actually using parity v1.7.0-beta-5f2cabd-20170727.

not sure about geth

perissology:
so is this a parity bug then?

perissology:
@frozeman ^

frozeman:
They will fix that.

PR #1052, opened 2017-09-18
fixed bugs when calling contract string method

emn178:
case 1: if str is string type and contract has no this property

origin
```JavaScript
contract.methods.str().call((err, result) => {
  console.log(err); // null
  console.log(result); // "ERROR: Strings are not yet supported as return values"
});
```
after patch
```JavaScript
contract.methods.str().call((err, result) => {
  console.log(err); // an Error object
                    // Error: ERROR: Strings are not yet supported as return values....
  console.log(result); // undefined
});
```
case 2: if str is string type and value is empty
```JavaScript
contract.methods.str().call((err, result) => {
  console.log(err); // null
  console.log(result); // "0x00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000"
});
// and something error message in console 
Unhandled rejection Error: The parameter "0x" must be a valid HEX string....
```
after patch
```JavaScript
contract.methods.str().call((err, result) => {
  console.log(err); // null
  console.log(result); // ""
});
```


frozeman:
When the type is wrong it should rather return an error. @chriseth what do you think about the above?

chriseth:
@frozeman you have to give me more context.

frozeman:
Could you please put the ABI in your example above, and add test cases to the `tests/contract.js` ?

frozeman:
I fixed this myself with added tests. Thank you so much for the find, well done!

PR #1055, opened 2017-09-18
Change default and add comma

rstormsf:
I blindly copy/pasted initialization of new contract and it has 20,000 gwei for GasPrice.
In order to save some money to other people, I think we should set 20 gwei as default.
I also added missing comma there.
Also, we don't need numbers in strings, it could be just numbers.

```
web3.toWei(20, 'gwei')
"20000000000"
```


frozeman:
Thank you for the hint. The numbers in strings is required, as wei values can be large numbers and to make sure they are converted properly to HEX they need to be as strings. i will close this as i will do the corrections myself. Thank you again!

PR #1068, opened 2017-09-22
ES6 + rollup (Experimental)

jdkanani:
Idea is to benefit from tree shaking and reduce build size for dapps using `web3` or any other `web3` packages. 

**What this pull request contains:**
- Converted `web3-utils` to es6 - fetch only methods you want: 
```es
 import { sha3, isAddress, BN } from "web3-utils"
 ```
- Cleaner build system with babel, rollup and gulp. Generates cjs files in `lib` folder.
```bash
gulp build
gulp web3-utils # Generates web3-utils.js and web3-utils.min.js dist files
```
- eslint for source code
```es
gulp lint
```
- Example of loading only needed functions from peer package (which reduce dist file sizes)

This whole experiment yields `web3-utils.min.js` with size of `63k` (almost half than before `121K`)
Plus, Source code looks much cleaner, easy to read, maintainable and extendable.

Any feedback/suggestions would be appreciated.


wlzla000:
According to #570, it seems _Web3.js_ is not currently ES6.
I have no idea if they have any plan to use ES6.

jdkanani:
@wlzla000 This pull request convert `web3.js` to ES6.

packages/web3-bzz/src/index.js
```diff
@@ -20,62 +20,60 @@
...
+    //   provider = 'http://swarm-gateways.net'; // default to gateway
     // }
 
-
-    if(_.isString(provider)) {
-        this.currentProvider = provider;
-    } else {
-        this.currentProvider = null;
-        return false;
+    if (!_.isString(provider)) {
```
frozeman: Here you are altering the functionality.
Further down it should be `this.currentProvider` not `this.provider`

packages/web3-bzz/src/index.js
```diff
@@ -20,62 +20,60 @@
...
+      this.pick = swarm.pick;
     }
 
     this.setProvider(provider);
-};
+  }
 
-// set default ethereum provider
-/* jshint ignore:start */
-Bzz.givenProvider = null;
```
frozeman: The point of this is to detect an existing `window/global.ethereumProvider`
You seem to have removed that altogether...

jdkanani: This line here: https://github.com/jdkanani/web3.js/blob/504541d414d8b5f01d4d2d942920a5ea5c6a6256/packages/web3-bzz/src/index.js#L28 will take care of this.

packages/web3-bzz/src/index.js
jdkanani: Sorry about that. Fixed it.

jdkanani:
@frozeman Converted to 4 spaces

jdkanani:
Changed base branch to `1.0ES6` now.

frozeman:
Thanks. Are there any code alterations besides let and const chages?

jdkanani:
No as currently just converting through tool.

jdkanani:
Web3-core-method has changed a lot though and core-method's tests are failing.

pkieltyka:
do you guys think this PR will land before 1.0?

jdkanani:
@pkieltyka Plan is to do so :) Any help would be appreciated.

frozeman:
@dacarley started to change to es6 incremental in #1068 and other prs. I am not sure how we should go about implementing your work. Could you check how he did it and contribute there? Especially on the build phase.

syzer:
Short comment: 
@jdkanani updated web3 to year 2015, etherum team cannot merge.
Short opinion:
:D
Does etherum team live in 2009 ? (https://en.wikipedia.org/wiki/ECMAScript)
Much innovation, so progressive... very modern..

PR #1073, opened 2017-09-23
Add personal.signTypedData

ukstv:
The method signs the data along with the schema, so that a user is well informed on what she signs.
As proposed in https://github.com/ethereum/EIPs/pull/712


frozeman:
As this is not accepted standard yet, and not in geth or parity i can't merge that, also it should be on the `eth_` namespace, as it doesn't receive a password as last argument.

frozeman:
Im happy adding that, if there is a broad consensus.

elenadimitrova:
Isn't there also a requirement for `eth_signTypedData` in addition to `personal_signTypedData` according to https://github.com/ethereum/EIPs/pull/712

ukstv:
@elenadimitrova It is indeed. It doesn’t matter though as eip712 is not widely accepted. Neither geth nor parity support it.

Amxx:
+1

nivida:

PR #1078, opened 2017-09-26
Ensure websocket connection is open before sending

moeadham:
to prevent unhandled error.


moeadham:
@frozeman any thoughts on this one? We are running off a fork in prod now because of this. We aren't alone if you check the issues list.

packages/web3-providers-ws/src/index.js
```diff
@@ -209,6 +209,16 @@ WebsocketProvider.prototype.send = function (payload, callback) {
     // try reconnect, when connection is gone
     // if(!this.connection.writable)
     //     this.connection.connect({url: this.url});
+    if (this.connection.readyState !== this.connection.OPEN) {
+        console.trace()
```
frozeman: Please remove the console trace here

moeadham: @frozeman done

moeadham:
@frozeman This test failed due to something in shh. Should I change anything or can we move this forward?

```
1) web3.shh.subscribe
       should create a subscription for "syncing":
      Uncaught AssertionError: expected { Object (startingBlock, currentBlock, ...) } to deeply equal { Object (startingBlock, currentBlock, ...) }
      + expected - actual
       {
         "currentBlock": 786193
      -  "highestBlock": 786193
      +  "highestBlock": 712483
         "knownStates": 698915
         "pulledStates": 35
         "startingBlock": 786211
       }
```

frozeman:
If tests don't pass, then we can't move forward :)
Can you see if you can reproduce the failing test using `npm test` locally? You can trigger the rebuild, by pushing an empty commit. There are a few failing tests, can you please see why this is the case, it may have to do with your changes.

moeadham:
@frozeman Sorry about that. All good now, dummy commit worked.

frozeman:
Thanks!

Issue #1077, opened 2017-09-26
web3.eth.accounts.wallet.create can overwrite existing account(s) 


cmditch:
Bug: Removing account from a wallet where `account's wallet index < wallet.length -2`, then running wallet.create will overwrite the tail end of the wallet.

Reproduce error:
```javascript
wallet = web3.eth.accounts.wallet;
wallet.create(3);  //  wallet.length === 3, three index instances
wallet.remove("0"); // wallet.length === 2, two index instances
lastWallet = wallet["2"].address;
console.log("Wallet address before create: ", lastWallet);

wallet.create(1);  // wallet.length === 3, TWO index instances, wallet["2"] was overwritten
console.log("Wallet address after create: ", wallet["2"].address);

wallet["2"].address === lastWallet;
> false
```

Browser : Chrome 60.0.3112.113 
Web3.js Version : 1.0.0-beta.18

Apologize for all the issue title renames. This bug kept rearing an uglier head each time I looked closer.

EDIT - Turns out this same bug affected `wallet.encrypt`, `wallet.clear` as well. See PR below.

PR #1079, opened 2017-09-26
Decouple wallet.length from other wallet functions (add, encrypt, clear)

cmditch:
See issue here #1077 

This adds a new function `web3.eth.accounts.wallet._findSafeIndex` which recursively counts up from 0 and finds the first unused index in the wallet, where `web3.eth.accounts.wallet.add()` will inject a new account.

An alternative fix for this error can be seen [here](https://github.com/cmditch/web3.js/commit/fd18778fdcc44e0c16f0ac0f8b2c410cd8c02f18), where instead of a new function, a new variable is added `wallet.pointerIndex` to keep track of a safe lower limit for future indexes.

Despite this recursive fix being O(n), and the alternative fix (linked above) and original `web3.eth.accounts.wallet.add()` being O(1), they all benchmark around the same @ ~1000ms for `wallet.create(500)`.

EDIT - 

Turns out `wallet.length`'s coupling to other primary wallet functions causes issues if `wallet.remove` has been used. 

This PR adds two new functions, `wallet._findSafeIndex` & `wallet._currentIndexes`, to help address these issues. `wallet.add`, `wallet.enrypt`, & `wallet.clear` refactored accordingly.

Tests have been added as well.

Chai version number was bumped to utilize chai's new array equality assertions.

### DIFF
--- a/packages/web3-eth-accounts/src/index.js
+++ b/packages/web3-eth-accounts/src/index.js
@@ -320,6 +320,15 @@ function Wallet(accounts) {
     this.defaultKeyName = "web3js_wallet";
 }
 
+Wallet.prototype._findSafeIndex = function (pointer) {
+    pointer = pointer || 0;
+    if (_.has(this, pointer)) {
+        return this._findSafeIndex(pointer + 1);
+    } else {
+        return pointer;
+    }
+};
+
 Wallet.prototype.create = function (numberOfAccounts, entropy) {
     for (var i = 0; i < numberOfAccounts; ++i) {
         this.add(this._accounts.create(entropy).privateKey);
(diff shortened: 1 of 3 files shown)


packages/web3-eth-accounts/src/index.js
```diff
@@ -320,6 +320,24 @@ function Wallet(accounts) {
...
+    } else {
+        return pointer;
+    }
+};
+
+Wallet.prototype._currentIndexes = function () {
+    var keys = Object.keys(this);
+    var indexes = keys
+        .map(function(key) { return parseInt(key); })
+        .filter(function(n) { return (n < 9e20); });
```
cmditch: Filters out NaN's and hex addresses (which will be parsed to int's). Makes the assumption that a user will not have many quintillions of wallets.

frozeman:
Thanks for that fix!

PR #1081, opened 2017-09-27
isHex: '0x' prefix changed to be optional

chatch:
added unit tests: utils.isHex.js
added all cases documented in web3-utils.rst - most were breaking but now pass

### DIFF
--- a/packages/web3-utils/src/utils.js
+++ b/packages/web3-utils/src/utils.js
@@ -358,7 +358,7 @@ var toHex = function (value, returnType) {
  * @returns {Boolean}
  */
 var isHex = function (hex) {
-    return ((_.isString(hex) || _.isNumber(hex)) && /^(-)?0x[0-9a-f]*$/i.test(hex));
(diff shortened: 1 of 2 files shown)


chatch:
ok so this breaks a bunch of code that is expecting the isHex call to fail if there is no '0x' prefix.  one example: utils.soliditySha3.

so maybe it's the web3-utils.rst doc that should change rather then this function.

frozeman:
Could you please point me to examples in the doc where it states hex should be without prefix?

chatch:
https://github.com/ethereum/web3.js/blob/1.0/docs/web3-utils.rst#ishex

web3.utils.isHex('c1912');

frozeman:
I added the `web3.utils.isHexStrict` function which expects `0x` and it being a string, and `isHex` is now more flexible and adhers to the docs. This is potentially a breaking change, for apps who relied on 'isHex'

frozeman:
Thanks for your finding!

PR #1104, opened 2017-10-10
chore(types): move typings to main npm package

### LINKED ISSUE #1103: Move typescript typings to main npm package
continuation of https://github.com/ethereum/web3.js/pull/897#issuecomment-335610741

phra:
closes #1103


connectdotz:
I obviously have too much time, or care too much about the integrity of this project... hopefully this will be my last post regarding this topic... I do have a real job that desperately needs my attention ;-)

Let's summarize our discussion in the past few days:

publishing

|criteria|embed in web3 only|embed in all|@<span/>types|web3-types|
|---|---|---|---|---|
|code management|✔️|✔️|❌|✔️|
|isolation|✔️|❌|✔️|✔️|
|guarantee in-sync|❌|❌|❌|❌|

* code management: can types and code be managed consistently and be part of the same development lifecycle. 
* isolation: each package is an independent/encapsulated package, sharing only through package managers.
* guarantee in-sync: ensure types are in sync with code (none does automatically, actually the types are already out of sync while we discussing where to put them...)

consumption

|typescript project use cases|embed in web3 only|embed in all|@<span/>types|web3-types|
|---|---|---|---|---|
|web3 + sub-packages|✔️|❌|✔️|✔️|
|sub-packages only|❌|❌|✔️|✔️|
|types only|❌|❌|✔️|✔️|

use cases:
1. typescript projects depend on web3 + sub-packages (web3 will always bring in other sub-packages)
1. typescript projects depend on only the sub-packages (such as projects only need the web3-utils package)
1. typescript projects only need the web3 types (such as a custom provider library only needs 
the provider interface).

**types are for all of us**
web3.js will probably not migrate to typescript anytime soon, but it is irrelevant because the types are for the typescript projects out there (like yours and ours), not just web3.js itself. We should do our best to help web3.js to serve the Ethereum development community. What we do here does matter, let's try not to put in the code that we can't defend.

frozeman:
Guys, i assume this type stuff works. And thanks so much for doing this, even though i don't care about typescript right now :P

PR #1116, opened 2017-10-15
Doco updated to removed privateKey from web3.eth.signTransaction and web3.eth.accounts.signTransaction

naddison36:
The code does not take a private key in the web3.eth.signTransaction or web3.eth.accounts.signTransaction functions so removing from the doco.


cmditch:
Are you sure?

https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth-accounts/src/index.js#L115

naddison36:
I know if I pass different private keys in I get the exact same transaction back. So this is either a bug in the code or doco

cmditch:
You able to provide some console logs or screenshots?

Thanks

naddison36:
I've added a unit test to my forked web3.js repo on the 1.0 branch to show the private key is not used
https://github.com/naddison36/web3.js/blob/1.0/test/eth.accounts.signTransaction.js#L66

Here's the declaration of the accounts.signTransaction function which does not take a privateKey
https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth-accounts/src/index.js#L92

I think this makes sense. If you want to sign a transaction using a different private key then just create a new account with that key and call signTransaction

cmditch:
I think the confusion here is the difference between:
```
web3.eth.account.signTransaction(txObj, privateKey)
```   
and
```
accountPrototype = web3.eth.account.create(1)    

accountPrototybe.signTransaction(txObj)
```
They are two different functions. The latter is built using the former. (As can be seen [here](https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth-accounts/src/index.js#L93))

naddison36:
@cmditch you are right. I have confused Accounts with Account.

I think web3.eth.signTransaction is still a problem, yes? It's been configured to only have one parameter, not two
https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth/src/index.js#L290

cmditch:
@naddison36 
`web3.eth.signTransaction` becomes a JSON-RPC call that asks the eth node itself to sign the tx with whatever accounts it's managing. It does not rely on locally stored/created accounts like you'd find in `web3.eth.accounts` and `web3.eth.accounts.wallet`.

Hope that helps 🤓

frozeman:
Hi,
i don't fully understand what it still confusing, or incorrect?

`web3.personal.sendTransaction(tx, pw)` receives a tx and a password, `web3.eth.sendTransaction(tx)` just an object, while the from in that object can also be a local account created with `web3.eth.accounts.wallet.create()`

naddison36:
Closing as doco was ok

PR #1130, opened 2017-10-21
Feature/exact version dependencies

dacarley:


gulpfile.js
```diff
@@ -107,88 +107,96 @@ var browserifyOptions = {
...
+
+    var version = lernaJSON.version;
+    var jsonPattern = /\"version\"\: \"[\.0-9\-a-z]*\"/;
+    var jsPattern = /version\: \'[\.0-9\-a-z]*\'/;
+    var meteorPattern = /ethereum:web3@[\.0-9\-a-z]*/;
+    var glob = [
+        './package.json',
+        './bower.json',
+        './package.js',
+        './.versions'
```
frozeman: This will be generated by meteor and can’t be modified directly.

package.json
```diff
@@ -12,7 +12,7 @@
     "postinstall": "lerna bootstrap",
     "build": "gulp",
     "build-all": "gulp all",
-    "release": "lerna bootstrap; gulp; lerna publish --skip-git; gulp version;",
+    "release": "lerna bootstrap; lerna publish --yes --cd-version prerelease --exact --skip-npm --skip-git; gulp version; gulp; git checkout lerna.json; lerna publish --cd-version prerelease  --exact --skip-git",
```
frozeman: This one I don’t understand.if you call publish twice it will increase by two versions. Seems the only thing you had todo is add —exact
We can’t hardcode prelrelease, as in the future I will also release other versions.
For what is the git checkout lerna.json?
This needs to be simplified, currently it will probably create more issues than it solves.

gulpfile.js
```diff
@@ -107,88 +107,96 @@ var browserifyOptions = {
...
-    .pipe(gulp.dest('./'));
+gulp.task('version', function () {
+    if (!lernaJSON.version) {
+        throw new Error("version property is missing from lerna.json");
+    }
+
+    var version = lernaJSON.version;
+    var jsonPattern = /\"version\"\: \"[\.0-9\-a-z]*\"/;
+    var jsPattern = /version\: \'[\.0-9\-a-z]*\'/;
+    var meteorPattern = /ethereum:web3@[\.0-9\-a-z]*/;
```
frozeman: This we also can remove, as we can’t modify the versions Json, this will be done by meteor publish.

Currently this is a manual thing,as meteor doesn’t allow to exclude certain folders, I have to manually copy them out to publish a release.

dacarley: OK, removed.

gulpfile.js
dacarley: Ditto.

package.json
dacarley: OK, now I'm confused as well.  I added the complexity as a way to work around an issue I was seeing where the built web3.min.js contained out-of-date version numbers for some packages.

I've reverted to the simpler command-line with just the addition of `--exact`, and I can't reproduce the issue I had earlier, so I'll commit the simpler line.

package.json
```diff
@@ -12,7 +12,7 @@
     "postinstall": "lerna bootstrap",
     "build": "gulp",
     "build-all": "gulp all",
-    "release": "lerna bootstrap; gulp; lerna publish --skip-git; gulp version;",
+    "release": "lerna bootstrap; lerna publish; gulp version; gulp",
```
frozeman: Hm, but now we basically publish before we change the version globally? and the `--exact` is missing, which is quite essential.

dacarley: See the changes in lerna.json.

PR #1136, opened 2017-10-23
Update web3-shh.rst

Boehrner:
Correction of the shh subscription example.


frozeman:
Whisper v5 only has one topic, this might change in v6. @gballet any comment on this?

gballet:
@frozeman @Boehrner Messages have only one topic, but filters can cover several topics. If you create a filter, you must be able to specify several topics and therefore the array is correct.

PR #1139, opened 2017-10-24
Fixes in types definition

icaroharry:
Complements #1131 and fixes some issues that broke my code after implementation of type definitons

- Create default export 
- Change `bignumber.js` to `bn.js`
- Change `Provider` interface to `IProvider` and create `Provider` type, which is a union of all the 3 providers (HttpProvider & IpcProvider & WebsocketProvider)
- Implement WebsocketProvider interface according to the current version (branch 1.0) of `web3-providers-ws`

### DIFF
--- a/packages/web3/index.d.ts
+++ b/packages/web3/index.d.ts
@@ -25,4 +25,4 @@ declare class Web3 {
 
 }
 
-export = Web3
+export default Web3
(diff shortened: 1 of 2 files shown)


icaroharry:
Following [this answer](https://github.com/Microsoft/TypeScript/issues/12815#issuecomment-266250230), I decided to make the `Provider` type a union between `WebsocketProvider, HttpProvider and IpcProvider`

frozeman:
Hoping this makes sense, as i didn't test the typescript compiler

PR #1157, opened 2017-11-02
correct personal.newAccount Promise resolve value

eyezick:
Docs currently show boolean, but it is returning a string


JeroenMerks:
I can confirm

frozeman:
Thanks!

PR #1167, opened 2017-11-12
include 0x prefix to signedtransaction

eyezick:


frozeman:
actually this whole example should be changed to using `web3.eth.accounts.signTransaction`. Want to do that?

PR #1174, opened 2017-11-14
Fixing compiler errors - pragma solidity

stefek99:
Tried to run the examples, encountered some errors.


frozeman:
Thanks, but these are outdated anyway, as the 1.0 branch is the one which gets developed

stefek99:
Yeah, totally get that... Didn't realise 1.0 is on the way!

Issue #1121, opened 2017-10-17
Problems with TypeScript bindings of web3.js


vguzev:
# **Steps to reproduce the problem:**
1. Clone any boilerplate with TypeScript/Webpack. For example, this one:
https://github.com/rokoroku/react-redux-typescript-boilerplate
```
git clone https://github.com/rokoroku/react-redux-typescript-boilerplate.git
cd react-redux-typescript-boilerplate
```
2. Install all dependencies:
```
npm install
```
3. Make sure that it is working correctly:
```
npm start
```
4. Install web3:
```
npm install web3
```
5. Add the following lines to file <code>src/index.tsx</code> somewhere before the <code>ReactDOM.render</code>:
```Typescript
import * as Web3 from 'web3';
const wa = window as any;
if (window.hasOwnProperty('web3')) {
  wa.web3 = new Web3(wa.web3.currentProvider);
} else {
  wa.web3 = new Web3(new Web3.providers.HttpProvider("http://localhost:8545"));
}
```
6. Start the application:
```
npm start
```

# **Expected results:**
It shouldn't throw any errors.

# **Actual results:**
It gives the following errors in console:
```
ERROR in ./node_modules/scrypt/index.js
Module not found: Error: Can't resolve './build/Release/scrypt' in '/Users/vguzev/Projects/test2/react-redux-typescript-boilerplate/node_modules/scrypt'
 @ ./node_modules/scrypt/index.js 3:19-52
 @ ./node_modules/scrypt.js/node.js
 @ ./node_modules/web3-eth-accounts/src/index.js
 @ ./node_modules/web3-eth/src/index.js
 @ ./node_modules/web3/src/index.js
 @ ./src/index.tsx
 @ multi (webpack)-dev-server/client?http://localhost:3000 webpack/hot/dev-server ./index.tsx

ERROR in ./node_modules/got/index.js
Module not found: Error: Can't resolve './package' in '/Users/vguzev/Projects/test2/react-redux-typescript-boilerplate/node_modules/got'
 @ ./node_modules/got/index.js 22:12-32
 @ ./node_modules/swarm-js/lib/files.js
 @ ./node_modules/swarm-js/lib/api-node.js
 @ ./node_modules/web3-bzz/src/index.js
 @ ./node_modules/web3/src/index.js
 @ ./src/index.tsx
 @ multi (webpack)-dev-server/client?http://localhost:3000 webpack/hot/dev-server ./index.tsx

[...]

PR #1184, opened 2017-11-19
Using 'export =' instead of 'export default' in main typescript definition for fixing #1121

matrushka:
At the moment if you are using TS and importing web3 with:
* `import Web3 from 'web3';`  => TS definitions work but you get `TypeError: web3_1.default is not a constructor` in runtime.
* `import Web3 = require('web3');` or `import * as Web3 from 'web3';` => the code works but TS definitions doesn't work and you end up with `error TS2351: Cannot use 'new' with an expression whose type lacks a call or construct signature.` in TS compilation.

The cause for this is TS expects a ES6 module definitions inside the `index.js` file. Although it's returning the class itself through CommonJS (this is the standard node behaviour, normally preprocessors like Babel can work around this but in TS it's optional and can cause some problems with environments which are not using any loaders). With this fix the typescript definitions point out the right output type (which is the class not the module) therefore it can be imported properly by using `import Web3 = require('web3');`. I've tried to make it work as `import * as Web3 from 'web3';` which is more appropriate but typescript doesn't allow classes as direct and throws out `error TS2309: An export assignment cannot be used in a module with other exported elements.`

**TL;DR** TS doesn't allow exporting a class as the module hence creating various problems with a module compiled in CommonJS but imported as ES6. With this PR it's possible to use all the type definitions with the `import Web3 = require('web3');` import instruction just like mentioned in #1121.

### DIFF
--- a/packages/web3/index.d.ts
+++ b/packages/web3/index.d.ts
@@ -22,7 +22,6 @@ declare class Web3 {
   providers: t.Providers
   setProvider(provider: t.Provider): void
   utils: t.Utils
-
 }
 
-export default Web3
+export = Web3;


frozeman:
Does any of the typescripters have an opinion on that?

connectdotz:
@frozeman this particular fix is right if we consider supporting the syntax
```
import Web3 = require("web3")
```
The way web3.js is written, in typescript term, is a [module-class](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html), which indeed should export the class with `export =` (there are other ways to import it in a typescript module but we don't need to get into that here)

There are many other problems with the current type definition, but it is probably beyond the scope of this PR... I think it is ok to merge this as it is, we could address other issues in a separate PR...

frozeman:
Thanks!

eepstein:
I don't think this is sufficient.
There's currently no clean way to import this module using es2015 settings in TypeScript.

PR #1185, opened 2017-11-19
Handle overload functions

jbaylina:
This PR tries all the functions until it finds a one that matches the number of parameters.

Solves https://github.com/ethereum/web3.js/issues/924

### DIFF
--- a/packages/web3-eth-contract/src/index.js
+++ b/packages/web3-eth-contract/src/index.js
@@ -123,8 +123,16 @@ var Contract = function Contract(jsonInterface, address, options) {
 
 
                     // add method only if not one already exists
-                    if(!_this.methods[method.name])
+                    if(!_this.methods[method.name]) {
                         _this.methods[method.name] = func;
+                    } else {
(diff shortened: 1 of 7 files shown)


frozeman:
This is definitely a useful fix, thought because of this reason there is also the function signature as well as the function name. As even with this fix, we have the problem there can be overloaded functions with the same parameter amount, just different types.

packages/web3-eth-contract/src/index.js
```diff
@@ -123,8 +123,16 @@ var Contract = function Contract(jsonInterface, address, options) {
...
 
                     // add method only if not one already exists
-                    if(!_this.methods[method.name])
+                    if(!_this.methods[method.name]) {
                         _this.methods[method.name] = func;
+                    } else {
+                        var cascadeFunc = _this._createTxObject.bind({
+                            method: method,
+                            parent: _this,
+                            next: _this.methods[method.name]
```
frozeman: Could we please rename this to: `next` -> `nextMethod`

btw we have the same issue for events :0

PR #1189, opened 2017-11-21
return contract instance when deploying with web3 wallet

perissology:
Fix a bug when deploying a contract using the wallet sign tx, the tx
receipt was returned instead of the contract instance.


frozeman:
Great catch will merge!

frozeman:
Thanks for this catch!

Issue #1188, opened 2017-11-21
Web3 1.0 instances seem to share state (provider overwritten)


benjamincburns:
Observed this issue on 1.0-beta.26 - haven't tested other versions.

When I create an instance of web3, use it for a bit, then create a second instance with a different provider, the first instance's provider is overwritten.

I've modified `test/setProvider.js` to demonstrate the issue. These changes are given below as a diff taken against c4c92c35.

```javascript
diff --git a/test/setProvider.js b/test/setProvider.js
index d0f4caf..52a3a15 100644
--- a/test/setProvider.js
+++ b/test/setProvider.js
@@ -54,7 +54,11 @@ describe('lib/web3/setProvider', function () {
         provider1.bzz = 'http://localhost:8500';
         provider2.bzz = 'http://swarm-gateways.net';
 
+        var provider3 = new FakeHttpProvider();
+        provider3.bzz = 'http://localhost2:8500';
+
         var lib = new Web3(provider1);
+        var lib2 = new Web3(provider3);
 
         assert.equal(lib.eth.currentProvider.constructor.name, provider1.constructor.name);
         assert.equal(lib.eth.net.currentProvider.constructor.name, provider1.constructor.name);
@@ -71,6 +75,21 @@ describe('lib/web3/setProvider', function () {
         assert.equal(lib.eth.accounts._requestManager.provider.constructor.name, provider1.constructor.name);
         assert.equal(lib.shh._requestManager.provider.constructor.name, provider1.constructor.name);
 
+        assert.equal(lib2.eth.currentProvider.constructor.name, provider3.constructor.name);
+        assert.equal(lib2.eth.net.currentProvider.constructor.name, provider3.constructor.name);
+        assert.equal(lib2.eth.personal.currentProvider.constructor.name, provider3.constructor.name);
+        assert.equal(lib2.eth.Contract.currentProvider.constructor.name, provider3.constructor.name);
+        assert.equal(lib2.eth.accounts.currentProvider.constructor.name, provider3.constructor.name);
+        assert.equal(lib2.shh.currentProvider.constructor.name, provider3.constructor.name);
+        assert.equal(lib2.bzz.currentProvider, provider3.bzz);
+
+        assert.equal(lib2.eth._requestManager.provider.constructor.name, provider3.constructor.name);
+        assert.equal(lib2.eth.net._requestManager.provider.constructor.name, provider3.constructor.name);
[...]

PR #1191, opened 2017-11-22
Fixes ethereum/web3.js#1188

benjamincburns:
Expands on existing `setProvider` test, and then uses some inheritance hacks to track `web3.eth.Contract.currentProvider` and similar per `Web3` instance.


packages/web3-eth/src/index.js
```diff
@@ -138,8 +140,18 @@ var Eth = function Eth() {
...
 
+    var ContractWrapper = function ContractWrapper() {
+      Contract.apply(this, arguments);
+    };
+
+    ContractWrapper.setProvider = function() {
+      Contract.setProvider.apply(this, arguments);
+    };
+
+    util.inherits(ContractWrapper, Contract);
```
benjamincburns: To me, I think this is the most contentious part. 

From the [current node docs](https://nodejs.org/docs/latest/api/util.html#util_util_inherits_constructor_superconstructor):

> Note: Usage of util.inherits() is discouraged. Please use the ES6 `class` and `extends` keywords to get language level inheritance support. Also note that the two styles are [semantically incompatible](https://github.com/nodejs/node/issues/4179).

So, @frozeman my question here is whether we'd rather do things the ES6 way, per those docs. Is Web3 1.0 babelified? If so, perhaps that'd be more idiomatic.

test/setProvider.js
```diff
@@ -89,6 +108,21 @@ describe('lib/web3/setProvider', function () {
         assert.equal(lib.eth.accounts._requestManager.provider.constructor.name, provider2.constructor.name);
         assert.equal(lib.shh._requestManager.provider.constructor.name, provider2.constructor.name);
 
+        assert.equal(lib2.eth.currentProvider.constructor.name, provider3.constructor.name);
+        assert.equal(lib2.eth.net.currentProvider.constructor.name, provider3.constructor.name);
+        assert.equal(lib2.eth.personal.currentProvider.constructor.name, provider3.constructor.name);
+        assert.equal(lib2.eth.Contract.currentProvider.constructor.name, provider3.constructor.name);
```
benjamincburns: Test fails here without above changes.

packages/web3-eth/src/index.js
frozeman: We are currently working on the 1.0ES6 branch. When we have that we can use it. Currently we do some Babel trans compiling, but I didn’t try if it works, as all the code is es5

frozeman: Also did you try the standalone lnstantiation? According to the docs?

benjamincburns: Would you rather I submitted a PR against that branch? Happy to do so - I think that answers the `util.inherit` vs `class` and `extend` question as well, then?

I didn't try standalone instantiation, as I figured existing tests covered that. This PR definitely won't fix shared state between standalone instances of `Contract`, however.

If you want to remove that shared state, then I suggest we ditch the dynamic inheritance hacks altogether and go with the usual `setProvider` thing post construction and revert back to the old `contract` factory method on `web3.eth` like in the later pre-1.0 releases.

benjamincburns: @frozeman Just tested and this commit cherry-picks cleanly against 1.0ES6 and all tests pass - I'm happy to do that if you like and resubmit?

frozeman: Important is that standalone still works  as expected. We currently don’t have tests for that. Feel free to add some. 

Concerning the es6 branch, I will anyway merge this branch  back into 1.0ES6 so it should be fine.

benjamincburns: @frozeman Provided "as expected" means "as documented," then I'll be happy to add some tests for that.

Otherwise a test like the following will not pass under this branch. Please let me know if fixing this issue is a condition for acceptance.

```javascript

Contract.setProvider(provider1);
let contract1 = new Contract(...);
Contract.setProvider(provider2);
let contract2 = new Contract(...);

assert.deepEqual(provider1, contract1.currentProvider);
assert.deepEqual(provider2,contract2.currentProvider);
```

benjamincburns: Actually, forget I asked - I'll add that test, make it pass, and then I think we'll be in good shape. :-)

benjamincburns:
@frozeman after merging latest 1.0 branch into my PR, I see the following test failure.  Not sure why this is happening, however. Would appreciate a pointer.

```
1) typical usage
       with data
         should deploy a contract, sign transaction, and return contract instance:

      AssertionError: expected 'eth_sendTransaction' to equal 'eth_sendRawTransaction'
      + expected - actual

      -eth_sendTransaction
      +eth_sendRawTransaction

      at test/contract.js:2829:24
      at IpcProvider.FakeIpcProvider.send (test/helpers/FakeIpcProvider.js:53:9)
      at RequestManager.send (packages/web3-core-requestmanager/src/index.js:129:66)
      at sendRequest (packages/web3-core-method/src/index.js:542:42)
      at send (packages/web3-core-method/src/index.js:563:13)
      at Object._executeMethod (packages/web3-eth-contract/src/index.js:888:24)
      at Context.<anonymous> (test/contract.js:2879:16)
```

benjamincburns:
Broken test is now fixed.

packages/web3-eth/src/index.js
```diff
@@ -138,8 +140,18 @@ var Eth = function Eth() {
     this.personal = new Personal(this.currentProvider);
     this.personal.defaultAccount = this.defaultAccount;
 
+    var ContractWrapper = function ContractWrapper() {
```
frozeman: The problem is that we want the "class" to be "Contract" not "ContractWrapper" when somebody logs the `web3.eth.contract` object.

I also dopant fully understand this wrapper.

packages/web3-eth/src/index.js
```diff
@@ -36,6 +36,8 @@ var Iban = require('web3-eth-iban');
 var Accounts = require('web3-eth-accounts');
 var abi = require('web3-eth-abi');
 
+var util = require('util');
```
frozeman: Is this module available in the browser? As we compile to browser, we need to make sure its available there as well.

frozeman:
Thanks. This is a very valuable contribution. But there are still a few things unclear to me. maybe we want to setup a call. Can you reach out to me on gitter?

packages/web3-eth/src/index.js
benjamincburns: The problem the wrapper solves:

Given two instances of `eth`, called `eth1` and `eth2`, and two contract instances, `contract1`, created by calling `new eth1.Contract(...)` and `contract2` created by calling `new eth2.Contract(...)`, we want to ensure that calling `eth1.setProvider` does not mutate `contract2.currentProvider`.

benjamincburns: Also, it's trivial to rename the `ContractWrapper` type to `Contract`. I can do that if need be. I just wanted to avoid the namespace collision w/ the `Contract` type defined in the `web3-eth-contract` module.

benjamincburns:
@frozeman Sorry, didn't see your last comment before I started responding above. I reached out on gitter. I'll keep an eye out for your response, but I often miss notifications from there. If I don't respond quickly, feel free to DM me on the ConsenSys Slack as well (I assume you're @frozeman over there as well), as I rarely miss those notifications.

packages/web3-eth/src/index.js
benjamincburns: Good question. I'm not familiar w/ how you're building for the browser yet, but since `util` is part of the node runtime, I'd expect most tools like `browserify` and `webpack` include shims for it. I'll double check and get back to you.

benjamincburns: It appears that you're using browserify, and it appears (per browserify/browserify#1503) that browserify will pull in the entire `util` package in this case.

If you'd rather not pull in the entire util module, we can do here what was suggested on that issue, use the `inherits` package instead of `util.inherits`. In the ES6 branch you can ditch that as well, and just use the `extends` keyword.

benjamincburns:
@frozeman Since we aren't likely to connect today, I finished responding to your comments above. I'm still happy to have a call on this, but given the timezones and that I'll be traveling all next week, we might not connect for a while.

Please be aware that this is a very important issue for Ganache. I won't be even the slightest bit offended (in fact, I'd be delighted) if you wanted to reject this PR and redo it yourself in some way that's more aligned to your vision of how these things should fit together, especially if that's the quicker path to getting #1188 closed.

packages/web3-eth/src/index.js
frozeman: It would be great if we can avoid using util here and use something natively existing in ES5 already

packages/web3-eth/src/index.js
frozeman: Yes please change to `Contract`

benjamincburns:
@frozeman I too am sorry for the delay! :blush:

Unless I missed something, your requests were that I

1. Rename the proxied `ContractWrapper` type to, simply, `Contract` (I assume to avoid potentially breaking type checks which relied on that name)
2. Ditch `util.inherits` in favour of proper ES5 inheritance.

Both are done in the latest push. Also thanks for pushing me to ditch `util.inherits` - I'd apparently leaned on it as a crutch for too long and never bothered to learn the proper way to do prototype chaining!

PR #1196, opened 2017-11-23
empty string handling for asciiToHex

vibhumbhardwaj:
Will return 0x00 for empty inputs.

0x was giving me inconsistent results while sending the contract deploy (if multiple 0x arguments in array). So, I used 0x00.


frozeman:
Good point. Thanks

PR #1201, opened 2017-11-27
web3-core-requestmanager: Use sendAsync when available in sendBatch

iamdoron:


frozeman:
In new web3.js version the `send` IS async. so only send should be used. When the provider is added it should take care of making sure `sendAsync` is translated to `send`. So this logic should be rather there.

iamdoron:
@frozeman Thanks for the reply. The problem is that some providers in the wild are not compatible, and you already doing it for regular send: https://github.com/iamdoron/web3.js/blob/48b73c70fe00745176a836725442f45ab5378e83/packages/web3-core-requestmanager/src/index.js#L129

my actual problem is when using the Ganache test provider, as it doesn't support send

frozeman:
Youre right, it added that as a fallback, when as i only change that for the `web3.currentProvider` which is already existing. Will merge,

PR #1205, opened 2017-11-29
fix: don't silently decode missing return values

lionello:
When converting the output of a failed `eth_call` to JS types, some types would get converted to a default value whereas others would show an error during conversion.

This change makes the behavior consistent for all types: converting missing output will always result in a conversion failure. In particular, converting the result of failed `require`, `assert`,  `throw` or `revert` will cause an error being thrown in web3. (The exception is calling a method that returns `void` for which a successful call cannot be distinguished from a failed one.)


packages/web3-eth-abi/src/formatters.js
```diff
@@ -142,7 +142,11 @@ var signedIsNegative = function (value) {
  * @returns {BN} right-aligned output bytes formatted to big number
  */
 var formatOutputInt = function (param) {
-    var value = param.staticPart() || "0";
+    var value = param.staticPart();
```
frozeman: We have to make sure the this function in return nothing when there is an empty value returned. could you run some tests?

frozeman:
i see you have some tests here. I am not hundred percent sure, that it will return not `0x` on the empty values. Technically it shouldnt

packages/web3-eth-abi/src/formatters.js
lionello: I'm confused why there's this special case for `int`? `uint` for example does not have this default value. My tests cover all of them.

lionello:
@frozeman Please have a look at the test cases. Try running them against the unpatched web3 and you'll notice the many inconsistencies between how the types are handled. After the patch empty is handled correctly.

RPC will always return `0x` for empty param. Either way, the 0x is cut off immediately by web3.

frozeman:
for whatever reason travis is not finishing, or reporting back

frozeman:
are the tests passing?

lionello:
They are passing locally. Not sure what's up with travis not reporting anything though. I only extended already existing tests.

frozeman:
Hm i have one failure locally:
```js
 17 passing (25ms)
  1 failing

  1) decodeParameters should not convert 0x to uint256:
     AssertionError: expected [Function] to throw an error
      at Context.<anonymous> (test/eth.abi.decodeParameters.js:89:20)
```

frozeman:
In Uint formatter, you forgot: `!param.rawValue`

The code was
```js
```

lionello:
It appears commit 15ffd55b46ae6500a1ca44f65327cdc5415c2bff changed the behavior again.

lionello:
So the "fix" to https://github.com/ethereum/web3.js/issues/1089 introduced the `rawValue` check, but I don't think that's the right fix. Certainly it's not right to do this for just `uint256`. Please reconsider that fix.

frozeman:
I agree, and answered there. i also tested the return value behaviour locally.

frozeman:
though the rawValue check is useful, as otherwise 0 values would also error out

Issue #989, opened 2017-08-16
web3.eth.Contract.allEvents does not work in Web3 1.0.0-beta16 using WebSockets


onbjerg:
Using ``allEvents`` in Web3 1.0.0 does not return any events. My code is as follows:

```js
this.contract.events.allEvents({
  fromBlock: 0,
  toBlock: 'latest'
}, (err, event) => {
  console.log(err, event)
})
```

Using ``getPastEvents`` works. It seems like the subscription is never given an ID(?):

```js
Subscription {
  id: null,
  callback: [Function],
  arguments:
   [ { toBlock: 'latest',
       address: '0x59c3b17749502874f237467818dfe46170ba92d8',
       topics: [] } ],
  _reconnectIntervalId: null,
  options:
   { subscription:
      { params: 1,
        inputFormatter: [Array],
        outputFormatter: [Function: bound ],
        subscriptionHandler: [Function: subscriptionHandler] },
     type: 'eth',
     requestManager: RequestManager { provider: [Object], providers: [Object], subscriptions: {} },
     params:
      { toBlock: 'latest',
        address: '0x59c3b17749502874f237467818dfe46170ba92d8',
        topics: [] } },
  subscriptionMethod: 'logs' }
```

I am using geth 1.6.7-stable.

Forbu:
I am using web3 1.0.0 and I have somewhat the same problem.
I did some test with this simple contract (using testrpc as the rpc provider) :
`pragma solidity ^0.4.11;
contract eventc {

event Top(string yolo);

function event_testing(string test){
  Top(test);
}
}`

`contract_event.methods.event_testing("tessssst").send({from: address_pers,gas: 1500000,
      gasPrice: 15000000}).then(console.log);

contract_event.events.Top({
    fromBlock: 0,
    toBlock: 'latest'
}, function(error, event){
console.log("event : ",event);
});`

The console only return : `event :  null` despite multiple events in the blockchain.

Forbu:
I am using web3 1.0.0 and I have somewhat the same problem.
I did some test with this simple contract (using testrpc as the rpc provider) :
```solidity
contract eventc {

event Top(string yolo);

function event_testing(string test){
  Top(test);
}
}
```

```javascript
contract_event.methods.event_testing("tessssst").send({from: address_pers,gas: 1500000,
      gasPrice: 15000000}).then(console.log);

contract_event.events.Top({
    fromBlock: 0,
    toBlock: 'latest'
}, function(error, event){
[...]

PR #1209, opened 2017-12-01
Add call to callback in WebsocketProvider.prototype.send on error

gabmontes:
When calling `send(payload, callback)` on a `WebsocketProvider` object, if the connection is not open the error is logged to the console but the `callback` is never called with such an error. As a consequence, the error is not bubbled up in the callbacks chain and the original caller never gets any indication of such an error condition.

This can be reproduced by trying to connect to an incorrect/misspelled URL or if the node is not up or running the WebSocket interface. 

Currently, only "connection not open on send()" is logged to console. With this change, `callback` is called with the error 'connection is not open'.

Related issues: #989


frozeman:
Thanks!

PR #1214, opened 2017-12-06
fix: make signTransaction return a promise, fixes #1213

### LINKED ISSUE #1213: signTransaction should always return promises
_related to web3 1.0_

i strongly suggest to adopt this pattern for every use case because it will make the dev ux smoother.

phra:


frozeman:
This is breaking 7 tests, can you please run `npm test` yourself?

This was a decision i made to allow `accounts.signTransaction()` to return synchronous when chain, id gasPrice etc is given. I am also not happy with the inconsistency it creates, but deemed this more convienient. As described here: http://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html#id5

I am still torn, but it might be better to return a promise at all times. Can you fix the tests as well?

phra:
@frozeman i will try later.

frozeman:
Please fix the test :)

phra:
sorry, i forgot. i will do it now.

phra:
@frozeman tests and conflicts are fixed. have a look at the `void async function(done)` trick, it can be used to gradually migrate the codebase to promises and async functions. :beers:

phra:
tests are failing on node < 7.
native async function are missing in those runtime versions.
node LTS 6 will reach end of life in april 2018.

frozeman:
Sorry for the delay.
There is no reason why we should use `void async` here, please just call `done()` so these tests works also on legacy systems

frozeman:
Currently there is a 1.0ES6 branch, but the current version is ES5 and we cant introduce now ES6 in one place here.

phra:
@frozeman done, but i strongly suggest to implement a transpiling step in the build to enable es6 usage on legacy versions. or even better port the code to typescript! :rocket:

phra:
CI is green. :white_check_mark:

phra:
the PR is stale again..

PR #1224, opened 2017-12-11
Adds checksum isAddress() test cases

uzyn:
Adding some test cases now that `web3.isAddress()` validates checksum addresses.


frozeman:
thanks

PR #1228, opened 2017-12-11
Extends HttpProvider with custom http headers

eugenone:
It is often necessary to include additional http headers with the request.


eugenone:
@frozeman can you check this pool request please?

frozeman:
Thanks! good addition. Though you could also make a PR against the 1.0 branch

frozeman:
Ok have done it myself.

Issue #1071, opened 2017-09-23
web3.eth.accounts.decrypt: createDecipheriv is not a function


narcis2007:
I get "TypeError: crypto.createDecipheriv is not a function" when I try to do web3.eth.accounts.decrypt. A possible fix: https://github.com/nwjs/nw.js/issues/880

ERROR TypeError: crypto.createDecipheriv is not a function
    at Accounts.webpackJsonp.../../../../web3-eth-accounts/src/index.js.Accounts.decrypt (index.js:252)
    at FileReader.myReader.onloadend (

frozeman:
Can you please test if the latest 1.0 branch fixes that?

frozeman:
will release a version beta.22 today, please test

narcis2007:
Ok, I will test it, thanks

2017-09-25 17:50 GMT+03:00 Fabian Vogelsteller <notifications@github.com>:

NarcisCreamCapital:
I still get core.es5.js:1020 ERROR TypeError: cryp.createDecipheriv is not a function
    at Accounts.webpackJsonp.../../../../web3-eth-accounts/src/index.js.Accounts.decrypt (index.js:252)

narcis2007:
I still get the same error

narcis2007:
Found the problem, it should be crypto = require('crypto-browserify'); instead of crypto = require('crypto');

AdrielGS:
Hi !
I get another error when using the method `web3.eth.accounts.encrypt(privateKey, password);`
The source of the error is the same that @narcis2007 commented about.

`core.es5.js:1020 ERROR Error: Uncaught (in promise): TypeError: cryp.randomBytes is not a function
TypeError: cryp.randomBytes is not a function
    at Accounts.webpackJsonp.../../../../web3-eth-accounts/src/index.js.Accounts.encrypt (index.js:263)
    at signUpService.signUp.then.namespace (sign-up.component.ts:60)
    at ZoneDelegate.webpackJsonp.../../../../zone.js/dist/zone.js.ZoneDelegate.invoke (zone.js:365)
    at Object.onInvoke (core.es5.js:3890)
    at ZoneDelegate.webpackJsonp.../../../../zone.js/dist/zone.js.ZoneDelegate.invoke (zone.js:364)
    at Zone.webpackJsonp.../../../../zone.js/dist/zone.js.Zone.run (zone.js:125)
    at zone.js:760
[...]

PR #1238, opened 2017-12-16
change crypto library reference

nicolaslawdune:
fixes crypto library reference #1071


frozeman:
As you dont add the package itself to the package.json, i wonder if that still works in node.js, and the browser? Can also please build the browserify files and test it in the browser?

frozeman:
I made the change myself, thanks for the PR anyway

PR #1245, opened 2017-12-22
Update the Documentation of methods.myMethod.send

nisdas:
Add a value option, so its is clear that this function can also be used to send ether in conjunction with contract method calls.


frozeman:
Thanks for adding this.

Issue #1248, opened 2017-12-23
Web3 typescript definition has no default export


nicolaslawdune:
Hello,

Web3 Typescript definition on branch 1.0 doesn't contain "default export".

### Context

- Typescript version :  2.6.2
- Converting ES2015(ES6) to ES5

After importing :
`import Web3 from 'web3';`

### Compilation Error 
Module '"xxxx/node_modules/web3/index"' has no default export.

## Update 1
After importing like this `import Web3 = require('web3');` I got a compilation error :

```
error TS1202: Import assignment cannot be used when targeting ECMAScript modules. Consider using 'import * as ns from "mod"', 'import {a} from "mod"', 'import d from
"mod"', or another module format instead.
```
This error is due to Module mode from tsconfig which is ES6.

## Update 2

I remove module mode from tsconfig and it works now.

matrushka:
Have you seen https://github.com/ethereum/web3.js/pull/1184 ?

connectdotz:
the type definition reflects the current module implementation [line 78](https://github.com/ethereum/web3.js/blob/6816ce526981257bc1c81942366b71be1de357fe/packages/web3/src/index.js#L78)
```
module.exports = Web3;
```
Based on typescript [document](https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html), it is recommended to adopt [module-class](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html), which is why the type is exposed with "=" and not default.

We have been going back and forth on this issue as you can see with #1184, which was exactly to undo the default import that undo the "=" earlier...  I am sure we don't want to keep repeating this cycle,  will the following import syntax work for you? 

```
import Web3 = require("web3")
```

nicolaslawdune:
Thank you.

nicolaslawdune:
Using "require" keyword, I got this error on compilation

```
error TS1202: Import assignment cannot be used when targeting ECMAScript modules. Consider using 'import * as ns from "mod"', 'import {a} from "mod"', 'import d from
"mod"', or another module format instead.
```
I use typescript : 2.6.2

connectdotz:
web3.js is still using es5 module system, so it's best to use the compiler option that supports it... try module: "commonjs"

kespeart:
Doing ```import Web3 from 'web3'``` works for me in version 1.0.0-beta.26 but when i upgrade to 1.0.0-beta.27 I receive the error mentioned in this post.

[...]

PR #1249, opened 2017-12-23
Fixes typescript definition 

nicolaslawdune:
Fixes web3 typescript definition #1248

### DIFF
--- a/packages/web3/index.d.ts
+++ b/packages/web3/index.d.ts
@@ -24,4 +24,4 @@ declare class Web3 {
   utils: t.Utils
 }
 
-export = Web3;
+export default Web3;


MaxBlaushild:
👍 🌮 🥇

hodlbank:
It breaks 
`import Web3 from 'web3';`
with 
`"module": "commonjs",`
because:
`TypeError: web3_1.default is not a constructor`

hodlbank:
$ node -v
v6.13.0
$ tsc -v
Version 2.6.2

```
{
  "compilerOptions": {
    "target": "esnext",
    "module": "commonjs"
  }
}
```
Minimal code to reproduce:
```
// .ts file
import Web3 from 'web3';

let web3 = new Web3(new Web3.providers.HttpProvider('http://localhost/'));
```
Transpiling result:
```
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
const web3_1 = require("web3");
let web3 = new web3_1.default(new web3_1.default.providers.HttpProvider('http://localhost/'));

```

PR #1258, opened 2017-12-27
treat RLP 'r' and 's' fields as quantity, not string

### LINKED ISSUE #1170: eth.accounts.signTransaction: expected r & s to be treated as quantity, not data
According to the docs, [r & s are returned from signTransaction with leading zeros](https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html#id6), implying that they are always 32 bytes of binary data. In many other implementations that return r & s (geth, parity, ethereum-keys), they are treated as integers, without leading zeros.

I would expect r & s in web3.eth.accounts.signTransaction to follow the same pattern.

jellegerbrandy:
This fixes #1170 by using the patch provided by @ldn0x7dc in that issue

### DIFF
--- a/packages/web3-eth-accounts/src/index.js
+++ b/packages/web3-eth-accounts/src/index.js
@@ -151,15 +151,15 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
         var rawTx = RLP.decode(rlpEncoded).slice(0, 6).concat(Account.decodeSignature(signature));
 
         var trimLeadingZero = function (hex) {
-            while (hex && hex.startsWith('0x0')) {
-                hex = '0x' + hex.slice(3);
+            while (hex && hex.startsWith('0x00')) {
+                hex = '0x' + hex.slice(4);
             }
             return hex;
-        };
+        }
         rawTx[7] = trimLeadingZero(rawTx[7]);
         rawTx[8] = trimLeadingZero(rawTx[8]);
 
-        var rawTransaction = RLP.encode(rawTx);
+        var rawTransaction = RLP.encode(rawTx)
 
         var values = RLP.decode(rawTransaction);
         var result = {
--- a/packages/web3-eth-accounts/src/index.js
+++ b/packages/web3-eth-accounts/src/index.js
@@ -146,10 +146,20 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
 
         var hash = Hash.keccak256(rlpEncoded);
 
-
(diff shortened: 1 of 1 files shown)


carver:
This change in output seems like the right thing to do, but it doesn't completely fix #1170. `r` and `s` fields should have single leading `0`s trimmed as well (when returned individually in https://github.com/ethereum/web3.js/pull/1258/files#diff-7bb2a20126193b9ecfe4723f83429c49R168 ).

jellegerbrandy:
You are right of course; should be better now

malissa:
Would love to get this merged in!

jellegerbrandy:
Applying the patch in our code, and running it against parity, returns the following error (from Parity):

```Invalid params: Invalid bytes format. Expected a 0x-prefixed hex string with even length.```

Apparently, the issue is more complex than I had hoped. This issue seems related as well: https://github.com/ethereumjs/ethereumjs-tx/issues/51

In our code, we have gone back to the "strip two leading zeros" code, and that seems to work fine against parity.

malissa:
@jellegerbrandy I also received an error (using geth) when stripping zeros resulted in an odd numbered string. I will try removing two leading zeros - should at least improve the number of times I receive this error which is A LOT.

jellegerbrandy:
I reverted the commit stripping _all_ leading zeros, which clearly was mistaken. Apparently, `geth` and `parity` expect strings with an even number of characters that do not start with `0x00`... the documentation does not help me much here, unfortunately..

carver:
RLP encoding is one byte at the smallest unit. Represented as a hex string, that's two hex characters at a time. So for the rlp encoding here, it should only trim 0 **bytes** (double 0 hex characters): https://github.com/ethereum/web3.js/pull/1258/files#diff-7bb2a20126193b9ecfe4723f83429c49R162

But for the `r` and `s` values, which are not RLP-encoded when returned individually, they should trim down all leading hex 0 characters. https://github.com/ethereum/web3.js/pull/1258/files#diff-7bb2a20126193b9ecfe4723f83429c49R168

frozeman:
Thanks for the fix! And sorry for the delay, i was not available over christmas.

malissa:
Woot! Thanks for this!

frozeman:
It would be good to add a few test cases, does somebody have a tx and key which results in the 0x00.. r, s values?

malissa:
Of course now I am unable to get an error. I will comment here if I come across one.

carver:
I just posted short r & s values in #1170

frozeman:
@malissa you could use a previous commit :)

PR #1283, opened 2018-01-10
Handle 0x conversion

aakilfernandes:
The bignumber library doesnt understand that  '0x' is 0. This fixes that.


aakilfernandes:
Actually going to move this to toBigNumber

PR #1284, opened 2018-01-10
To big number 0x

aakilfernandes:
Fixes issue where toBigNumber('0x') throws an error


frozeman:
This actually breaks the tests, and im not sure if an empty data value, like what `0x` represents should convert to 0.

nivida:

PR #1288, opened 2018-01-11
Return r, s, and v without leading 0's

carver:
These values are [quantities, not data](https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding). Other nodes return r, s, and v as hex-encoded quantites.

See #1170

### DIFF
--- a/packages/web3-eth-accounts/src/index.js
+++ b/packages/web3-eth-accounts/src/index.js
@@ -42,14 +42,14 @@ var isNot = function(value) {
 };
 
 var trimLeadingZero = function (hex) {
-    while (hex && hex.startsWith('0x00')) {
-        hex = '0x' + hex.slice(4);
+    while (hex && hex.startsWith('0x0')) {
+        hex = '0x' + hex.slice(3);
     }
     return hex;
 };
 
 var makeEven = function (hex) {
-    if(hex % 2 === 1) {
+    if(hex.length % 2 === 1) {
         hex = hex.replace('0x', '0x0');
     }
     return hex;
@@ -165,17 +165,17 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
 
         var rawTx = RLP.decode(rlpEncoded).slice(0, 6).concat(Account.decodeSignature(signature));
 
-        rawTx[7] = trimLeadingZero(rawTx[7]);
-        rawTx[8] = trimLeadingZero(rawTx[8]);
+        rawTx[7] = makeEven(trimLeadingZero(rawTx[7]));
+        rawTx[8] = makeEven(trimLeadingZero(rawTx[8]));
 
         var rawTransaction = RLP.encode(rawTx);
 
         var values = RLP.decode(rawTransaction);
         var result = {
             messageHash: hash,
-            v: makeEven(values[6]),
-            r: makeEven(values[7]),
-            s: makeEven(values[8]),
+            v: trimLeadingZero(values[6]),
+            r: trimLeadingZero(values[7]),
+            s: trimLeadingZero(values[8]),
             rawTransaction: rawTransaction
         };
         if (_.isFunction(callback)) {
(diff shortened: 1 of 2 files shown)


frozeman:
I would actually like to get a third opinion on that. As bytes are technically invalid if they are uneven.
@chriseth, @axic any opinion?

packages/web3-eth-accounts/src/index.js
```diff
@@ -42,14 +42,14 @@ var isNot = function(value) {
 };
 
 var trimLeadingZero = function (hex) {
-    while (hex && hex.startsWith('0x00')) {
-        hex = '0x' + hex.slice(4);
+    while (hex && hex.startsWith('0x0')) {
```
frozeman: The problem with this is what if the rs, value is `0x00001234` Then a even number is made uneven

packages/web3-eth-accounts/src/index.js
```diff
@@ -165,17 +165,17 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
 
         var rawTx = RLP.decode(rlpEncoded).slice(0, 6).concat(Account.decodeSignature(signature));
 
-        rawTx[7] = trimLeadingZero(rawTx[7]);
-        rawTx[8] = trimLeadingZero(rawTx[8]);
+        rawTx[7] = makeEven(trimLeadingZero(rawTx[7]));
+        rawTx[8] = makeEven(trimLeadingZero(rawTx[8]));
```
frozeman: And are you sure passing a modified value into `RLP.decode` is not going to break anything?

frozeman:
The tests pass here.
But this would actually be an arguement for padding the r,s values at all times with leading `0`s

frozeman:
Additionally allowing the odd numbers again will make this problem reappear: https://github.com/ethereum/web3.js/pull/1258

Though its very important that signatures are correctly validated!

packages/web3-eth-accounts/src/index.js
carver: Do you mean `RLP.encode`?

It looks like it dynamically calculates the length, so I don't see any problems: https://github.com/MaiaVictor/eth-lib/blob/master/lib/rlp.js#L27

In fact, I believe leaving out 0 bytes at the beginning of an int in the RLP-encoded data is a more standard approach. (not sure if it's *required* though)

carver:
> But this would actually be an argument for padding the r,s values at all times with leading 0s

> Additionally allowing the odd numbers again will make this problem reappear: #1258

Let's separate the two places where `r` and `s` values are used:

1. Just before rlp encoding: https://github.com/ethereum/web3.js/pull/1288/files/ac28015f159b8bece403bcac356008cc19cb037c#diff-7bb2a20126193b9ecfe4723f83429c49R168
2. When returning from `signTransaction`: https://github.com/ethereum/web3.js/pull/1288/files/ac28015f159b8bece403bcac356008cc19cb037c#diff-7bb2a20126193b9ecfe4723f83429c49R177

This PR intends to not change the functionality of # 1 - the change there was only a refactor. That means that the #1258 problem should not reappear, and I agree that `r` and `s` values should be provided with an even number of hex characters. (because that's what this rlp implementation expects)

This PR only changes the functionality of # 2, trimming all leading 0s. This brings it in line with what the nodes themselves return when returning (non-rlp-encoded) values of `r` and `s`.

frozeman:
Ill merge this.
@malissa and @jellegerbrandy can you test the 1.0 branch when its merged?

frozeman:
@carver this also should fix #1289, right?

carver:
@frozeman No, some more work would have to be done, maybe in web3 or maybe in eth-lib. #1290 is the test that should pass.

Some options I see, highest preference first:

- web3 could zero-pad r and s to 32 bytes before concatenating and handing off the signature to eth-lib
- eth-lib could add a new recovery option that accepts r, s, and v as individual arguments (splitting them is the only thing it has trouble with IIRC)
- (more of a strawman than a good idea:) eth-lib could detect short r and s, maybe by doing some modified RLP decoding

carver:
On further investigation, it looks like eth-lib itself is responsible for encoding the signature. So the best solution is to fix eth-lib's `Account.encodeSignature()`

carver:
Submitted a PR at https://github.com/MaiaVictor/eth-lib/pull/4

That repo has 3 open pull requests that go back to October, and has never merged or closed an incoming pull request. Maybe eth-lib needs to be forked into the ethereum org to be maintained?

frozeman:
I will talk to @MaiaVictor, but we could also leftpad this before sending it to eth-lib

frozeman:
I merged eth-lib and the tests pass. Ill create a new version soon

Issue #1276, opened 2018-01-08
contractInstance.methods.someMethod(123).send(...) produce an error Invalid JSON RPC response: ""


vldmkr:
### Description

Using binary search I found out that the problem appeared after the [6cd3885](https://github.com/ethereum/web3.js/commit/6cd3885103e69f9a0a251e73446d45e16847e9d0) commit.
The problem is that **ethAccounts** is *undefined* at the stage when it is used. For example, when the **eth_sendTransaction** Method is being created at [884](https://github.com/ethereum/web3.js/blob/6cd3885103e69f9a0a251e73446d45e16847e9d0/packages/web3-eth-contract/src/index.js#L884) line.

### Temporary solution
In *package.json* you need to set the version of **web3-eth-contract** and everything that depends on it up to version **1.0.0-beta.26**.
```json
"dependencies": {
  "web3": "1.0.0-beta.26",
  "web3-eth": "1.0.0-beta.26",
  "web3-eth-contract": "1.0.0-beta.26"
}
```

vldmkr:
https://github.com/vldmkr/web3-beta27-problem

I am adding a repository with a SSCCE so you can be sure. There are two files **package.json**. Make sure that you delete **node_modules** and **package-lock.json** before you change the version of **package.json**

VoR0220:
Also having issues with this. Not sure if what I'm doing is correct for npm installation. Trying a couple things. This project is cluttered.

Issue #1274, opened 2018-01-05
"Returned error: unknown account" after upgrading from 1.0.0-beta.26 to 1.0.0-beta.27


starspire:
Hello,
After web3js package update from 1.0.0-beta.26 to 1.0.0-beta.27 I faced following issue. Anytime I execute state changing contract method (like this):
```
exampleToken.methods
	.mint(toAddress, tokenNumber)
	.send({from: web3.eth.defaultAccount, gas: 100000});
```

I get an error: 
```
Unhandled rejection Error: Returned error: unknown account
    at Object.ErrorResponse (/Users/krzysztof/Documents/work/projects/Level01/level01_ethereum_backend_integration/node_modules/web3-core-helpers/src/errors.js:29:16)
    at /Users/krzysztof/Documents/work/projects/Level01/level01_ethereum_backend_integration/node_modules/web3-core-requestmanager/src/index.js:137:36
    at XMLHttpRequest.request.onreadystatechange (/Users/krzysztof/Documents/work/projects/Level01/level01_ethereum_backend_integration/node_modules/web3-providers-http/src/index.js:66:13)
    at XMLHttpRequestEventTarget.dispatchEvent (/Users/krzysztof/Documents/work/projects/Level01/level01_ethereum_backend_integration/node_modules/xhr2/lib/xhr2.js:64:18)
    at XMLHttpRequest._setReadyState (/Users/krzysztof/Documents/work/projects/Level01/level01_ethereum_backend_integration/node_modules/xhr2/lib/xhr2.js:354:12)
    at XMLHttpRequest._onHttpResponseEnd (/Users/krzysztof/Documents/work/projects/Level01/level01_ethereum_backend_integration/node_modules/xhr2/lib/xhr2.js:509:12)
    at IncomingMessage.<anonymous> (/Users/krzysztof/Documents/work/projects/Level01/level01_ethereum_backend_integration/node_modules/xhr2/lib/xhr2.js:469:24)
    at IncomingMessage.emit (events.js:164:20)
    at endReadableNT (_stream_readable.js:1062:12)
    at process._tickCallback (internal/process/next_tick.js:152:19)
```

I dived deeper and checked how it is transformed behind to json-rpc:

Request:
```
[...]

PR #1296, opened 2018-01-16
fixed the problem that ethAccounts is undefined. #1276 #1274

baryon:
fixed the problem that ethAccounts is undefined. #1276 #1274


frozeman:
Thanks for the fix!

PR #1308, opened 2018-01-19
Fixing examples and typos in 1.0 docs

corymsmith:
Some of the examples were console logging variables in the subscribe callback that only existed in the `data` callbacks.


frozeman:
thanks!

Issue #1310, opened 2018-01-20
Missing module in 1.0.0-beta2


Andromelus:
Hi.

In nodejs, importing` "web3": "^1.0.0-beta2"`, the module `web3-requestManager` is missing.

![web3jsbeta2missingmanager](https://user-images.githubusercontent.com/10050024/35181251-3903a22c-fdf9-11e7-8481-56af9743d683.png)

justinmchase:
I get the same thing.

lsaether:
This is a problem with how the package is registered (?) or being downloaded from NPM. To fix it locally you can specify the exact version. First uninstall web3 from your project, then install it again and specify the exact version you want, e.g. `npm i web3@1.0.0-beta.28`.

It would be nice to get a fix to this. I want to distribute a package over NPM that bundles web3 1.0.0 and if I try, NPM will always install the erroneous `web3@1.0.0-beta2` leading to this error.

justinmchase:
Basically instead of doing `beta.28` it should just be `1.0.28-beta`. Then npm would pick up the patch change version.

justinmchase:
How can I run a 1.0.0-beta node? I'm trying to use INFURA but its not working.

```js
var Web3 = require('web3')
var web3 = new Web3()
web3.setProvider(new web3.providers.HttpProvider('https://ropsten.infura.io/sH34tSfDKm34ecQwz43N'))
web3.shh.getVersion((err, version) => {
  if (err) return console.error(err.message) // err, 
  console.log(version)
})
```

suspended:
Same problem too

Andromelus:
As @lsaether said, you can "fix" this specifying the version you want. works fine. But having a fix would be great!

jellegerbrandy:
you can only "fix" it liek this using npm, though.

With `yarn`, I know of no fix (except using npm)

frozeman:
I guess i should release this soon as 1.0.0

PR #1316, opened 2018-01-23
The version should be semver compatible

justinmchase:
The version should be semver compatible and instead when publishing to npm you should specify the beta tag.

e.g.
```
npm publish --tag beta
```

This will allow users to opt-into this package by installing either this version specifically or this tag and still get patch level updates as they are available:

```
npm install web3@beta
```
or
```
npm install web3@1.0.0
```

Users who simply do the basic `npm install web3` will get the version currently published as `latest`.


justinmchase:
Addresses issue #1310

frozeman:
web3.js 1.0.0 is not yet released, so this would release it.

justinmchase:
No, I don't think so. Publishing it to NPM as latest would do that. When you publish it, specify the beta or next tag instead of latest. That way people can opt into beta and still get updates as patches are published without it ever being officially released. To release you just remove the beta tag and replace it with the latest tag.

https://docs.npmjs.com/getting-started/using-tags

KaelWD:
@justinmchase You're right that they should be using tags, but there's nothing wrong with the version number.

justinmchase:
@KaelWD Well I should say that I created this because I specifically encountered an issue with updating to newer versions with npm. New versions are published but you can't tell from npm commands like `npm outdated` because they were not being published with proper semver compatible versions.

Then after looking at it it became apparent that the reasoning behind the odd versioning scheme just seemed to not necessary if the tag feature was used instead. It would make working with this library a lot smoother.

justinmchase:
Here is an example of a popular npm project (gulp) on which uses this technique:
https://www.npmjs.com/package/gulp#versions

Notice they have `latest` and `next` tags. If you just do an `npm install gulp` it installs `^3.9.1` but one could install `npm install gulp@next` and it would give you `^4.0.0` instead but you'd know you're opt'ing in to the beta track.

KaelWD:
> you can't tell from npm commands like npm outdated because they were not being published with proper semver compatible versions

 - the versions are semver compatible: https://semver.org/#spec-item-9
 - `npm outdated` works for me, but shows the wrong version in "Wanted". I guess someone forgot a `.` when publishing. They can fix this just by unpublishing that version, although it's probably too late now thanks to the whole left-pad thing. 

image

KaelWD:
Tags would be good though, I use a script that automatically sets the release tag from the version. Anything  with a pre-release version goes to `next`, everything else goes to `latest`.

justinmchase:
Interesting, got it. I do think I was getting hit by that `beta2` issue. May be worth unpublishing it if you're already on 34. I was aware of the `-foo` being legit but I didn't realize you could tag on `.n` also and have it resolve correctly so I thought that was the issue.

Issue #1168, opened 2017-11-12
eth.sign allow {encoding: 'hex'} parameter


christoph2806:
For `web3.sha3()`, you can specify the that the data is hex-encoded by adding the `{encoding: 'hex'}` parameter. 
For `web3.eth.sign()`, you don't have that option. It would be good to have that. 
The current implementation makes it impossible to sign a hex encoded string, for example a message hash, because a hash usually contains bytes > 127 and therefore this hash cannot be encoded as string (javascript will interpret bytes > 127 as unicode chars and transform them in a 2-byte sequence).

`web3.eth.sign(account, msg)` is basically implemented as `sign(account, web3.sha3('\19Ethereum Signed Message\n'+msg.length+msg)`. It should be possible to call `web3.eth.sign(account, msg, {encoding: 'hex'})` which would be implemented as `sign(account, web3.sha3(Buffer.from('\x19Ethereum Signed Message\n'+(msg.length-2)/2).toString('hex')+msg.slice(2), {encoding: 'hex'}))`

Usecase: 
In the upcoming multisigs which work with presigned transactions, you need to pass a signature to `ecrecover` which is calculated on a hash of the transaction parameters (source, destination, value, data, nonce). 
This is impossible at the moment with `web3.eth.sign`; you need to use non-standard signing functions e.g. from lightwallet (which do not prefix the message), which should be avoided .

enderphan94:
This `web3.utils.sha3(lengthSlot, {encoding: 'hex'})` does not return the same thing as `web3.sha3(lengthSlot, {encoding: 'hex'})`. Any ideas?

PR #1325, opened 2018-01-26
Ability to sign hex string

jstoxrocky:
**What was broken:**
Referencing #1168. Trying to sign a raw hex string (results of a keccak256 hash for example) with `web3.eth.accounts.sign` will throw an `Invalid continuation byte` error. This is because of this line: https://github.com/ethereum/web3.js/blob/dd55daa6cd54a3907457169425bcd5e29bf32ec4/packages/web3-eth-accounts/src/index.js#L219 The above line assumes that any hex string passed to the `hashMessage` function is utf8 encodable. The results of a keccak256 hash are not utf8 encodable (Maybe the number is too large or something. I'm not an expert on bytes and encodings so I may not be saying this correctly!). 

However, the ability to sign a hex string (specifically the results of a keccak256 hash) is important functionality. Say I have a state-channel contract and I want to hash some pre-image values, and then sign the resulting hash. Attempting this with the current implementation throws an error since `msg` is not a utf8 encodable hex string.

```javascript
var msg = web3.utils.soliditySha3(
      {type: 'address', value: some.address}
      {type: 'uint256', value: some.value},
);
var sig = web3.eth.accounts.sign(msg, some.privateKey);
``` 

**How it was fixed:**
Two fixes were made in this PR. 
(1) `web3.eth.accounts.hashMessage` will now break data into Buffers and performs a keccak256 hash on on the Buffer data.
(2) The existing implementation of`web3.eth.accounts.recover` allows users to pass in either the pre-image message or the hashed-message. This is incompatible with allowing hex strings that are not utf8 encodable to be pre-image values. This method was changed to only allow hashed-messages to be passed as arguments. This can be seen as the mirror image of the `web3.eth.accounts.sign` method, which only allows pre-images to be passed as arguments.
To sum up, users should pass pre-images to `sign` and message-hashes to `recover`. I believe that This is in line with how `web3.py` works.

### DIFF
--- a/packages/web3-eth-accounts/src/index.js
+++ b/packages/web3-eth-accounts/src/index.js
@@ -216,13 +216,15 @@ Accounts.prototype.recoverTransaction = function recoverTransaction(rawTx) {
 /* jshint ignore:end */
 
 Accounts.prototype.hashMessage = function hashMessage(data) {
-    var message = utils.isHexStrict(data) ? utils.hexToUtf8(data) : data;
-    var ethMessage = "\x19Ethereum Signed Message:\n" + message.length + message;
+    var message = utils.isHexStrict(data) ? utils.hexToBytes(data) : data;
+    var messageBuffer = Buffer.from(message);
+    var preamble = "\x19Ethereum Signed Message:\n" + message.length;
+    var preambleBuffer = Buffer.from(preamble);
+    var ethMessage = Buffer.concat([preambleBuffer, messageBuffer]);
     return Hash.keccak256s(ethMessage);
 };
 
 Accounts.prototype.sign = function sign(data, privateKey) {
-
     var hash = this.hashMessage(data);
     var signature = Account.sign(hash, privateKey);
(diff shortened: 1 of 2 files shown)


frozeman:
Thanks for that.
In this case one can only pass in hex data, to be signed, no string directly, is that so in web3.py?

jstoxrocky:
@frozeman You can also pass a string. All the existing tests on regular old strings still pass. They are still in the `test/eth.accounts.sign.js` file. This PR just provides the _added_ functionality to sign hex strings of the form `0x1234...6789` that can't be utf-8 encoded.

Web3.py forces the user to submit a bytes array actually. Strings cannot be passed. The user mist convert them to bytes first. I guess the equivalent of forcing the user to convert their data into a Buffer for Web3.js.

frozeman:
Sorry for the late response, i was on holidays.
You changed the tests to pre-convert from UTF to HEX, so its not working as before actually.

PR #1330, opened 2018-01-27
fix(allevents): update decode logic to handle no topics

marcusmolchany:
the default for `data.topics` on line 53 was causing an exception on line 55:
```
TypeError: Cannot read property 'slice' of undefined
```

### DIFF
--- a/lib/web3/allevents.js
+++ b/lib/web3/allevents.js
@@ -50,7 +50,11 @@ AllSolidityEvents.prototype.encode = function (options) {
 
 AllSolidityEvents.prototype.decode = function (data) {
     data.data = data.data || '';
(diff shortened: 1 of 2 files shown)


marcusmolchany:
not sure the best way to add tests for this one

frozeman:
A test case would be great. There should be some decoder tests, just add one with empty data, or create a new unit test for that. Thanks for the find though!

frozeman:
Thanks for the fix, i'll fix the other one myself.

frozeman:
The `this._json` is the jsonInterface of the contract which will fire this event. its needed to decode the event.

marcusmolchany:
thanks!

PR #1336, opened 2018-01-30
build dist into packages/web3/dist,fix #1041

### LINKED ISSUE #1041: Where is `dist` folder and  `web3.min.js`?
After `npm install --save web3`, I only see `packages` and `src` foldesr inside `node_modules/web3`. Where is the `dist` folder containing `web3.min.js`?

gasolin:
The new web3 1.x package is published via `packages/web3`. 
The `dist/` is missing in new `web3` module because of the generated `dist/` is not located in `web3/packages/web3/dist`.

The patch 

* fixed links in packages/web3/README.md
* move `web3/dist` to `web3/packages/web3/dist` so npm publish will include `dist/` folder in web3 module.


frozeman:
Why would you want the mist folder in the npm module?

gasolin:
@frozeman for example if a web developer want to make a DApp to query on-chain block status, they just need to create a HTML page and include the dist/web3.min.js file into it.

guix77:
Actually I tend to agree with @gasolin, it's convenient not to have to build it and just include it from /dist. Why do you .npmignore it on the 1.0 branch and not on the stable branch ?

gasolin:
Rebased the patch. To avoid the conflict, require run `npm build` command to generate `packages/web3/dist/` after apply this patch.

@frozeman  how do you think?

slyfox42:
Hi guys, any updates on this? I need to inject web3 directly in the browser when running feature tests with Puppeteer since I cannot have MetaMask. Since the dist folder and the `web3.min.js` file is not included I have to use [RawGit](https://rawgit.com/) to point to the minified file on the repo, which works but is not ideal.

nivida:
Hey all,

This should be fixed with the PR #2000. I'll provide on the next release for each module a UMD, commonJS and ESM bundle.

Issue #1341, opened 2018-01-31
Invalid RLP-encoding produced (web3-eth-accounts)


holiman:
Allows `-1` as `nonce`:

```

var Web3EthAccounts = require('web3-eth-accounts');
var account = new Web3EthAccounts('ws://localhost:8546');

> account.signTransaction({'nonce':-1, "value":0,"gas":21000,"gasPrice":0, "chainId":1},"0x5384837ece911af85b8b2f2a0e6fe73f5f2add257fc25e1efbeebd4ee72226b7")
{ messageHash: '0x71488a00b0ac6c3f13f230787de8e2cde99e018a9037209ae89f81baf7e7f471',
  v: '0x26',
  r: '0xc0d893f83137b9c82d3b1a602edda9a2d9c92f6714f421f96be08fa200a0b7d8',
  s: '0x3fd99522cf6ea496c7bc4ae2de4807e1d9fb136919462bbfb99012979664ff34',
  rawTransaction: '0xf84c81x18082520880808026a0c0d893f83137b9c82d3b1a602edda9a2d9c92f6714f421f96be08fa200a0b7d8a03fd99522cf6ea496c7bc4ae2de4807e1d9fb136919462bbfb99012979664ff34' }
```

But in fact, it produces an _erroneous_ `rawTransaction`: `0xf84c81x1...` with an `x` in the middle of the data payload.

Same with 
- negative `value`
- negative `gas`
- negative `gasPrice`

Using negative `chainId` makes the RLP even more strange: 
```javascript

var Web3EthAccounts = require('web3-eth-accounts');
var account = new Web3EthAccounts('ws://localhost:8546');

> account.signTransaction({'nonce':1, "value":0,"gas":1,"gasPrice":0, "chainId":-1},"0x5384837ece911af85b8b2f2a0e6fe73f5f2add257fc25e1efbeebd4ee72226b7")
{ messageHash: '0xa090f9d4887116465e139c4ce08be7c4835b97cb5ea10866bbf8694ca0d9fd90',
  v: '0x.8',
  r: '0x12',
  s: '0x5a',
  rawTransaction: '0xf94b.801800180808081.8125a03ba111521e2400ce9a0ac20d27082aa7119a8f6a230495202a5cdf704f9a419ba05079ac7d2bfa62c7f1d2ce8a339d8ad76c2ec3faf2a29c04e0992419f6df2c70' }

```

holiman:
[...]

PR #1348, opened 2018-02-02
Check hex strings before encoding #1341

ahaensler:
Fix for #1341 and a bug where "v" is missing a leading zero.


frozeman:
Thanks for the fix, i fixed it on my side already

Issue #1372, opened 2018-02-12
Newly deployed contract has wrong provider


rumkin:
web3: v1.0.0-beta29
node: v9.3

## Description

New contract instance throws no provider error on method call:
```
Error: Provider not set or invalid
      at Object.InvalidProvider (node_modules/web3-core-helpers/src/errors.js:38:16)
      at RequestManager.send (node_modules/web3-core-requestmanager/src/index.js:125:32)
      at sendRequest (node_modules/web3-core-method/src/index.js:542:42)
      at send (node_modules/web3-core-method/src/index.js:563:13)
      at Object._executeMethod (node_modules/web3-eth-contract/src/index.js:819:24)
      at Context.<anonymous> (test/registry.spec.js:47:14)
```

## Expect

Method properly calls. No error thrown.

## Reproduce

1. Deploy new contract:
```
const contract = new web3.eth.Contract(abi, options);

const registry = await contract.deploy({
    data: bytecode,
})
.send();
```

2. Call any contract method:
```
await registry.methods.addEntry('name').send();
```

congxinUSC:
Met the same problem, my solution is to save the previous provider and reassign it after deploy.

frozeman:
Will be fixed in the next release.

PR #1373, opened 2018-02-12
Fix contract instance clone method #1372

rumkin:
This PR fix issue with new deployed contract #1372 caused by contract's clone method implementation.


frozeman:
Thanks for the find

PR #1381, opened 2018-02-17
private key should be prefixed with `0x`

kevinwucodes:
The first example is correct, but the second example does not have `0x` prefixed


frozeman:
Thanks

PR #1389, opened 2018-02-21
Fix typo in WS provider

gabmontes:
The websocket provider incorrectly creates errors saying `IPC`, which might be misleading.


frozeman:
Thanks!

PR #1390, opened 2018-02-21
Add option to WS provider for setting a custom timeout

gabmontes:
The underlaying websocket will eventually timeout at 120 seconds on network failure so web3 calls using this provider will not settle to error for 2 minutes.

This PR's proposal is to add an option to the provider so the lib users can set a custom timeout to better suit their own needs.

This code shows the issue:

```js
const web3 = new Web3(new Web3.providers.WebsocketProvider(url))

let time

web3.eth.getBalance(address)
  .then(function () {
    console.log('Disconnect the net in the following 5 seconds!')
    return new Promise(function (resolve) {
      setTimeout(function () {
        time = Date.now()
        resolve()
      }, 5000)
    })
  })
  .then(() => web3.eth.getBalance(address))
  .then(function (balance) {
    console.log('Forgot to disconnect the net?')
  })
  .catch(function (err) {
    console.log('Timeout!', Date.now() - time) // 2 minutes
  })
```

The proposal is to allow the user

```js
const web3 = new Web3(new Web3.providers.WebsocketProvider(url, { timeout: 5000 }))
```

and let the web3 calls timeout in i.e. 5 secs instead of the default 120 secs.


packages/web3-providers-ws/src/index.js
```diff
@@ -171,6 +173,18 @@ WebsocketProvider.prototype._addResponseCallback = function(payload, callback) {
 
     this.responseCallbacks[id] = callback;
     this.responseCallbacks[id].method = method;
```
frozeman: Doesnt look like youre clearing the custom timeout when the response came. also keep in mind that certain calls, like for past logs etc take very long to answer. Sometimes 10+ seconds

gabmontes: @frozeman thanks for the comment!

Actually, clearing the timeout is not needed and would add complexity IMHO. Once the actual response comes in, the function at `this.responseCallbacks[id]` will be executed and deleted from the dictionary in L79. As long as ids are unique, there will be no issues. And, in fact, if ids are not unique, many other things would break.

Re calls with long answer time, it would be up to the user to select a proper timeout. This PR does not enforce nor provides any default. Having said that, is a very good observation. Do you think changing the README so the example is i.e. 30s instead of 5s would clear any doubts it may generate?

frozeman: default of 30s would be better yes

frozeman: and i would simply make the sample as such: `var ws = new Web3WsProvider('ws://localhost:8546', {timeout: 30000});`

frozeman:
I fixed it and merged your changes, thanks!
I changed you can also now pass headers in the options object.

gabmontes:
@frozeman thanks a lot!

PR #1393, opened 2018-02-22
Re-enable repeated firing of confirmation events

cgewecke:
At the moment when a transaction receipt is immediately available after a method call, the confirmation event only fires once. It seems like the subscription logic for this case might have been accidentally disabled while making `web3-core-method` [compatible with Parity's instant-seal](https://github.com/ethereum/web3.js/pull/936/files/87af1b39cb01cf3e4b65193ab741ea0bfcac36d2#diff-b87e4847acaf5b45ad88f78ed46732db). This issue primarily affects testing clients like ganache but possibly also Parity and Geth in dev mode. 

Because `startWatching` is set on a delay [here](https://github.com/ethereum/web3.js/compare/1.0...cgewecke:fix/repeat-confirmations?expand=1#diff-b87e4847acaf5b45ad88f78ed46732dbL403) and only if the promise hasn't resolved, the subscription never initiates and repeated firings of the confirmation handler never happen. 

This PR tries to re-enable multiple confirmations by 
+ removing the promise condition and delay
+ passing the immediately retrieved receipt to `checkConfirmation` so it can be emitted
+ adding logic to handle the case where the direct call to `checkConfirmation` required by instant-seal is doubled by a call made by the `startWatching` method.
+ adding a test for multiple confirmations.


frozeman:
Thanks for the find!

cgewecke:
Thanks :)

Issue #1386, opened 2018-02-20
bluebird warnings about promises not being returned


mdcoon:
My D-App runs extremely slow because of the many interactions I am making with web3js from redux-saga. Here are versions:

web3js: 1.0.0-beta.30,
redux-saga: 1.0.0-beta.0,
redux: 3.7.2

For some reason, when I call any web3js function from inside a saga "yield call(...)", bluebird spews warnings "Warning: a promise was created in a handler but was not returned from it". There are literally thousands of these being spewed at runtime when I interact with web3js for anything including getting accounts, balances, blocks, transactions, anything. I tried disabling the warning per bluebird's docs and it doesn't work. It says configuration is reset for each instance of bluebird, so I'm guessing web3js has its own instantiation internally with its own config. So either we need a way to turn off the warnings or we need to get to the bottom of why redux-saga doesn't play well with web3js 1.0.

DiscRiskandBisque:
Have you tried implementing #1415 ? I had this same issue and swapping out `any-promise` for `bluebird` got rid of the warnings.

PR #1415, opened 2018-03-02
Use any-promise for everything

lionello:
`any-promise` was already in use in some submodules, except these two were using bluebird. This patch uses it uniformly.


onetom:
One reason for this PR is that we noticed that sometimes promise rejections were not triggered.

As we were debugging we saw some DOM elements were created within promise management code by the [bluebird](http://bluebirdjs.com) library, instead of using the native Promise object available in the browsers ([except IE 11 & Opera Mini](https://caniuse.com/#search=promise).

The `[any-promise](https://github.com/kevinbeaty/any-promise)` library also says:
> NOTE: You probably want native promises now
> ...
> If no preference is registered, defaults to the global Promise for newer Node.js versions. The browser version defaults to the window Promise, so polyfill or register as necessary.

We were concerned that such DOM trickery - which is not really necessary in 2018 - might interfere with web frontend frameworks (eg Vue.js), especially in their development mode with hot-code-reload.

So far we couldn't prove such interference because our issues seems stem from a combination of regressions and bugs in `geth` and `web3.js`, but it definitely helps the debugging experience if we don't have to wade through ancient DOM-magic-promise code in addition to the million other layers of the software.

DiscRiskandBisque:
+1 because this fixed #1386 for me.

PR #1425, opened 2018-03-06
Fix: invalid params error while batch execution for contract methods.

joujiahe:
[Root Cause]
While sending transaction to contract methods will always occurred
error: 'Invalid params: invalid length 2, expected fewer elements in
array.'

The problem is method 'eth_sendTransaction' only accepts one parameter,
however , it will get two parameters while making a batch request payload.


frozeman:
Good catch, thanks!

PR #1435, opened 2018-03-07
Add main file to package.json

ianaya89:
The `package.json` file doesn't have the main file, so is not possible to use this code directly from the repo (without npm). And that is useful especially when you are working with forks or testing functionality.


frozeman:
good one

PR #1436, opened 2018-03-07
Add support for multiple events in subscriptions

ianaya89:
The subscription handler was forcing the response to be a unique result, even when the result was an array of multiple objects (due multiple events).

This PR allows the handler to manage an array of responses. So any time multiple events were registered, the subscription will be able to handle all of them properly and emit the corresponding events.


frozeman:
Only parity seems to send multiple events in one push, but good addition.

PR #1437, opened 2018-03-08
Add Missing err param onmessage function (ws provider)

ianaya89:
During the `onmessage` function of ws provider, a callback is called with a success result but the err param (`null`) is missing. So the success result is treated as an error.


frozeman:
This is actually correct, as this is the handler for subscriptions, which is not error first. 
Where did you discover that issue? ryan found an issue, and i discovered that web3.js is handling this totally wrong currently internally.

I fixed it in the 1.0 branch. So this would be a bad change

ianaya89:
The issue I found is when I listen to subscription events, without that change I am getting the new events as errors.

```
this.newEvents.on('data', e => {
  // Is supposed to enter here
})
.on('error', err => {
  // But I always get new events as errors. the err object is technically an event
}) 
```

In this case, I am using `allEvents` contract function, but the same issue with `getPastEvents` and even with some internal contract methods.

PR #1453, opened 2018-03-14
ERC-147: process.env.ETHEREUM_SOCKET default provider

### LINKED ISSUE #492: Add ERC 147 to web3 init
When a web3.js instance is cerated and no provider is given, the system tries to find it using the `ETHEREUM_SOCKET` environment variable.

bitpshr:
This PR implements [ERC-147](https://github.com/ethereum/EIPs/issues/147) from [Milestone 1.0](https://github.com/ethereum/web3.js/milestone/3) and adds associated tests. Specifically, creating a new `Web3` instance uses `process.env.ETHEREUM_SOCKET` as a provider if no value is explicitly passed into the constructor.

Resolves #492

### DIFF
--- a/lib/web3.js
+++ b/lib/web3.js
@@ -44,9 +44,8 @@ var HttpProvider = require('./web3/httpprovider');
 var IpcProvider = require('./web3/ipcprovider');
 var BigNumber = require('bignumber.js');
 
-
-
 function Web3 (provider) {
+    provider = provider || process.env.ETHEREUM_SOCKET;
     this._requestManager = new RequestManager(provider);
     this.currentProvider = provider;
     this.eth = new Eth(this);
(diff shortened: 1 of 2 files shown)


bitpshr:
Is this pending until the actual ERC-147 is approved? This would be useful.

bitpshr:
Is this still a relevant change for [Milestone 1.0](https://github.com/ethereum/web3.js/milestone/3)? Thanks!

frozeman:
Thanks for that that, but 3 things:
1. The socket does return a provider, but rather a URL, which need to be used to instantiate a provider
2. we have the `web3.givenProvider` as a place where we put given providers, this could also look up this ENV var.
3. the process variable may not be available in the browser, so it needs to be correctly checked the it is available, ideally with `typeof process !== 'undefined'`

bitpshr:
Thanks @frozeman. I will update this PR accordingly.

PR #1473, opened 2018-03-17
Set XHR object timeout;

Fi1osof:


frozeman:
Why setting it to 0 by default?

Fi1osof:
This reason: https://github.com/Fi1osof/web3.js/blob/6816ce526981257bc1c81942366b71be1de357fe/packages/web3-providers-http/src/index.js#L33
and this https://github.com/Fi1osof/web3.js/blob/6816ce526981257bc1c81942366b71be1de357fe/packages/web3-providers-http/src/index.js#L66

PR #1488, opened 2018-03-23
Add status to getTransactionReceipt in docs

gurbraj:
As per https://ethereum.stackexchange.com/a/6003, this PR adds status as a field to the return of web3.eth.getTransactionReceipt in docs.


frozeman:
We should actually convert the status object into a boolean in the `formatters.js`, you could do that, and please remove the lock file changes from your PR.

frozeman:
I fixed it myself in 85f45341

PR #1494, opened 2018-03-25
MaxListenersExceededWarning fix

devMYC:


frozeman:
I'm declined to merge this, as it would silently fail and not tell the developers that the subscription was not created.
Also this assumes that all providers which support `.on()` are event emitters with a `maxListener..` function.
We should think about a way to return proper warnings in such cases so that the dapp dev can take care of cleaning up subscriptions.

devMYC:
@frozeman yeah totally agree, definitely need a better solution to solve this problem. I'm currently just using this hack to ignore the warning when running tests.

frozeman:
Then lets think about a real solution, Ill close this.

PR #1500, opened 2018-03-27
Add personal_signAndSendTransaction call for backwards compatibility with go-ethereum

dukle:


frozeman:
As this is not a commonly accepted method im declined to add it to web3.js, ideally we only add methods that a generally adopted by most nodes.
If you want to add it locally you can simply do that using:
```js
```

frozeman:
If you convince parity to add it, we can add it here too ;)

dukle:
Cool, then I'll extend web3. Thanks!

PR #1503, opened 2018-03-27
Rejects send() promise if receipt status == "0x0"

vanruch:
Since Byzantium hard fork the `status` field inside the transaction receipt was introduced ([EIP 689](https://github.com/ethereum/EIPs/pull/658)). However, web3 still doesn't check it after a transaction was completed. 

This PR rejects promise if `status=='0x0'` fails similarly how it was before when `gasProvided == receipt.gasUsed`.

### DIFF
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -348,7 +348,8 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
                 if (!isContractDeployment && !promiseResolved) {
 
                     if(!receipt.outOfGas &&
-                       (!gasProvided || gasProvided !== receipt.gasUsed)) {
+                       (!gasProvided || gasProvided !== receipt.gasUsed) &&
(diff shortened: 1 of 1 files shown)


packages/web3-core-method/src/index.js
```diff
@@ -361,7 +362,14 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
                         if(receipt) {
                             receipt = JSON.stringify(receipt, null, 2);
                         }
-                        utils._fireError(new Error("Transaction ran out of gas. Please provide more gas:\n"+ receipt), defer.eventEmitter, defer.reject);
+                        if (receipt.status === '0x0') {
```
frozeman: Thanks for that, but could you add the status parsing filed to the web3-core-helpers formatters, so that it returns `status: true` or `false`, instead of the 0x1, and then change this code here to check for that.

frozeman:
I fixed it myself in 85f45341

PR #1506, opened 2018-03-28
Add CDN link

LukasDrgon:
I added a [jsDelivr CDN](https://www.jsdelivr.com/package/npm/web3) link to your readme, so that people can use it directly without bower or component. jsDelivr is the fastest open-source CDN and updates automatically from npm. 

I also added dist folder to npm so that this continues to work after v1.0.0 is released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,13 @@ meteor add ethereum:web3
 ```
 
 ### As Browser module
+
+CDN
+
+```html
+<script src="https://cdn.jsdelivr.net/npm/web3@0.20.6/dist/web3.min.js"></script>
+```
+
 Bower
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ meteor add ethereum:web3
 CDN
 
 ```html
-<script src="https://cdn.jsdelivr.net/npm/web3@0.20.6/dist/web3.min.js"></script>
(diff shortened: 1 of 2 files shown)


frozeman:
This is a nice idea, but would increase the package size for every user by the `web3.min.js` file. So we would need to find a better way to link to the dist files, like GitHub raw files or such.
Also this link would need to be updated then.

So better link to: `https://raw.githubusercontent.com/ethereum/web3.js/develop/dist/web3.min.js`

frozeman:
But shipping it in NPM grows it by a lot, i would prefer getting it from GitHub. Same for 1.0.
For this link `https://cdn.jsdelivr.net/gh/ethereum/web3.js/dist/web3.min.js` it uses the default branch (develop)?
Better would be here to point to the branch like `master` here.

MartinKolarik:
It uses the latest release: https://github.com/ethereum/web3.js/releases

frozeman:
That would be 1.0 version

MartinKolarik:
Actually it's 0.19 right now because beta releases are not considered when selecting the latest version: https://cdn.jsdelivr.net/gh/ethereum/web3.js/

frozeman:
And why not 0.20.6? Btw you want to adjust your PR?

PR #1521, opened 2018-04-04
Fix #1520: fix soliditySha3 for v:0

### LINKED ISSUE #1520: soliditySha3 throws "Error: undefined is not a number" for value 0
`Web3.utils.soliditySha3({t:'uint',v:0})` throws because the logic in `soliditySha3.js` does `value = arg.v || arg.value`, evaluating to `undefined` when `v` is 0.

lionello:
Fixes #1520


frozeman:
Thanks, good catch

PR #1541, opened 2018-04-11
Add Basic Auth support in wss:// Websocket URL

peterbroadhurst:
Basic Auth is not currently supported for the WebsocketProvider in Node.js (it works fine in HttpProvider).

I can see the Node.js package used does not support providing the Basic Auth within the URL syntax, so this PR parses the URL and adds an `Authorization` header explicitly.


peterbroadhurst:
This PR would address https://github.com/ethereum/web3.js/issues/1524

PR #1545, opened 2018-04-12
update shh doc for `post` to promise usage

gballet:
The documentation for `post` in `web3-shh` is still mentioning callbacks and a `Boolean` return value. This is no longer the case and this PR fixes the documentation on this point.


frozeman:
Thanks

PR #1546, opened 2018-04-13
Auto re-connect when connection is closed

songlinm:
This attempts to fix https://github.com/ethereum/web3.js/issues/1354 by automatically reconnecting every second.

Manually tested with the following steps:
Start up a proxy script that forwards port 9999 to 8546 ([Originally by](https://notes.tweakblogs.net/blog/7955/using-netcat-to-build-a-simple-tcp-proxy-in-linux.html) )

> export LC_CTYPE=C
> export LANG=C
> 
> TMP=`mktemp -d`
> BACK=$TMP/pipe.back
> SENT=$TMP/pipe.sent
> RCVD=$TMP/pipe.rcvd
> trap 'rm -rf "$TMP"' EXIT
> mkfifo -m 0600 "$BACK" "$SENT" "$RCVD"
> sed 's/^/ => /' <"$SENT" &
> sed 's/^/<=  /' <"$RCVD" &
> nc -l 127.0.0.1 "$SRC_PORT" <"$BACK" | tee "$SENT" | nc "$DST_HOST" "$DST_PORT" | tee "$RCVD" >"$BACK"

Run the following
`node`
`> const Web3 = require('web3');`
`> const web3 = new Web3('ws://localhost:9999')`
`> web3.eth.getBlock('latest').then(console.log)`
This should print out the latest block. Now kill the above proxy script. Re-do:
`> web3.eth.getBlock('latest').then(console.log)`

This should print
`> (node:48280) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: connection not open`

Now start up the same proxy again and repeat
`> web3.eth.getBlock('latest').then(console.log)`

This will print out the latest block again.

Unfortunately, it doesn't seem easy to add a unit test case for this.

### DIFF
--- a/packages/web3-providers-ws/src/index.js
+++ b/packages/web3-providers-ws/src/index.js
@@ -50,54 +50,7 @@ var WebsocketProvider = function WebsocketProvider(url, headers)  {
     var _this = this;
     this.responseCallbacks = {};
     this.notificationCallbacks = [];
-
-    // The w3cwebsocket implementation does not support Basic Auth
-    // username/password in the URL. So generate the basic auth header, and
-    // pass through with any additional headers supplied in constructor
-    var parsedURL = parseURL(url);
-    headers = headers || {};
-    if (parsedURL.username && parsedURL.password) {
-        headers.authorization = 'Basic ' + _btoa(parsedURL.username + ':' + parsedURL.password);
-    }
-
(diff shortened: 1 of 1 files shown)


frozeman:
In order to create tests you would need to create stubs for the `ws.connect` function and fire `ws.onclose`.

PR #1550, opened 2018-04-14
checking that window is defined is not always enough, we also need to…

julien51:
… check that window.WebSocket is actually a thing

I am using jsdom in our test env and `window` is defined in that context, without a `WebSocket`. Adding this check means that we revert to the node module if that is the case.


frozeman:
Thanks!

PR #1557, opened 2018-04-17
Use Node.js 6 API where possible for WSS URL

peterbroadhurst:
Sorry @frozeman - the very last change in https://github.com/ethereum/web3.js/pull/1541 to allow the build to run with Node 5, broke the fix.
It seems that the legacy `parse()` API in Node 5 does not support username/password, so we need to use the `new URL()` API wherever it is available


frozeman:
Thanks for the find

PR #1565, opened 2018-04-19
Allow web3.js without 'unsafe-eval' Content Security Policy rule

jonashaag:


Powersource:
This works for me, but I had to install `lerna` too.

ErikBjare:
Wait, why isn't this merged? We've been going crazy in our own issue to use `npm ci` to install from git which totally depends on this issue: https://github.com/SuperuserLabs/thankful/pull/170 (we need npm ci to get reproducible builds, a requirement for Web Extensions published on Mozilla Addons)

Ping @frozeman @nivida

frozeman:
Sorry have overseen this. I merged it. It still need testing for build files to run in the browser with this change

PR #1571, opened 2018-04-20
Fix typo

takp:
I think `achived` should be `achieved`


nivida:
Please remove the dist files from this PR :)

takp:
@nivida Thanks! I've removed the dist files 👍

PR #1580, opened 2018-04-24
Run tests on NodeJS 5, 6, 7 and 8

serima:
Run TravisCI tests on NodeJS version 5, 6, 7, and 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: node_js
 node_js:
   - "4.1"
   - "4.0"
+  - "5"
+  - "6"
+  - "7"
+  - "8"
 before_script:
   - npm install
   - npm install jshint


volschin:
6 and 8 are the lts versions, that really should be supported.

serima:
Yes, at least I think we should test other versions than 4.x

zachlysobey:
It also might be good to specify the `engines` field in the `package.json`

https://docs.npmjs.com/files/package.json#engines

frozeman:
Sure. Please test and make a PR. I assume it works :)

zachlysobey:
@frozeman I'll take a stab it it. :)

volschin:
10 as lts is not decided iirc because of missing OpenSSL lts version of 1.1.

PR #1581, opened 2018-04-24
setter and getter method for TIMEOUTBLOCK

Harzu:
Hello everyone!
We've got a problem with your library.
In test network, transactions sometimes do not run in 50 blocks. 
And when we run functions on the contract we've received an error.

Suggestion: Please allow users to choose the timeout like 50 seconds of the blocks themselves 

I made some changes. Please have a look and if you don't mind send us your feedback how it can be improved

For now, we've just added the getter and setter to variable TIMEOUTBLOCK and made the method web3.setTimeoutBlock (count)


jimmy-collazos:
@Harzu check your PR and remove unnecesary files (lerna-debug.log, packages/web3-core-method/.gitignore, ....) and don't change de git url of project.

frozeman:
This should not be on the `web3` object itself, as blocks are only part of `web3.eth`, better would be as a getter setter using `web3.eth.blockTimeout`, and people can set it or change it.
Same as here: https://web3js.readthedocs.io/en/1.0/web3-eth.html#defaultblock
Please also add docs to it. 

This should also simplify your code a lot

nivida:
Hay I've already implemented it in the ethereumProvider branch (PR #2000). Because of this, I will close this PR now.

Issue #1570, opened 2018-04-20
Wrong encoding tests for two-dimensional variadic arrays


biboran:
Encoding of two-dimensional variadic arrays in these tests is wrong:

https://github.com/ethereum/web3.js/blob/f15caf13c91682cccd7b7762b80feadb9fb969dc/test/coder.encodeParam.js#L43-L64

I took these tests for my own implementation of ABI encoding and found out that contracts would not deploy when I wrote some integration tests. These tests are built on the premise that offset of dynamic elements in a dynamic collection should be counted from the beginning of the whole encoding. The correct encoding is that offset is counted from the beginning of the dynamic collection regardless of the position of the dynamic collection in the whole encoding.

Here is a deployed contract using a correct encoding in the ctor: https://kovan.etherscan.io/address/0x812b72fa1fb3b842cb193d743a44ec8439c9a060#readContract

Here is my corrected version of the tests https://github.com/BlockStoreApp/Web3Swift/commit/8ab1f12ec3eb32fabf7ad67913fd39e505876f5f
I would be very grateful if you star my project if that information was of any help : )

biboran:
You can also observe the correct encoding if you call the `data()` for a simple getter contract like:

```solidity
pragma solidity ^0.4.23;
pragma experimental ABIEncoderV2;

contract SimpleGetter {
    
    address[][] public _data = [[0x407d73d8a49eeb85d32cf465507dd71d507100c1,0x407d73d8a49eeb85d32cf465507dd71d507100c2],[0x407d73d8a49eeb85d32cf465507dd71d507100c3,0x407d73d8a49eeb85d32cf465507dd71d507100c4]];
    
    function data() external view returns (address[][]) {
        return _data;
    }

}
```

nivida:
This should be fixed since I've implemented the new ABICoder. Could you test it and drop a comment here if not?

PR #1591, opened 2018-04-27
#1570 - updated incorrect tests for abi encoding and decoding 

biboran:
As per #1570 
* Commented out incorrect decoding test and updated its content
* Updated content for an already commented out encoding test


nivida:
I've closed this PR because I'm currently rewriting all the tests as you can see in the PR #2000

PR #1601, opened 2018-05-02
remove duplicate example

kevinwucodes:
The 2 examples shown for `web3.eth.accounts.privateKeyToAccount` are the same so we should remove the duplicate example.


nivida:
This fix got already merged :)

PR #1624, opened 2018-05-11
Update package.json

etaletai13:
fix a problem with systems that don't recognize the git+https protocol.


frozeman:
And this works? since which node version?

etaletai13:
@frozeman to be entirely honest I had no idea it ever worked the other way.

frozeman:
Looking at this, it looks wrong your link: https://stackoverflow.com/a/35892637

etaletai13:
@frozeman And yet, it works. See how travis succeeded with it? Try it yourself if you don't believe me. I know it isn't documented, but you can even run `npm install` or `yarn add` (yarn is far more useful and not broken) with a github address. For instance, in Musicoin, we used the command `yarn add github:ethereum/web3.js` and have only seen success. Plus, it works on all systems and not just linux. I would also recommend using yarn and requiring node 8.

```{
```"engines": {
```  "node": "8.9.0",
```  "yarn": "1.5.0"
```  }
```}

and then commit yarn.lock. This will make builds perfectly reproduced, every time. I know what I commit is often not in following with expected ways, but I also know that this does work, does solve a problem, and does pass tests. Is there any cause for not merging?

etaletai13:
Let me know if anybody figures out why this actually works? I didn't mean to even figure it out in the first place, but for a typo. But it really should have a document if it doesn't.

Issue #550, opened 2017-01-22
error with Events using indexed string


iurimatias:
steps:

contract:
```
pragma solidity ^0.4.7

contract Test {

  event TestEvent {
    string indexed _foo,
    string bar
  }

  function Test() {
  }

  function triggerEvent() {
     testEvent("hello", "world");
  }
}
```

javascript code:
```javascript
   var contract = web3.eth.contract(abi).at(address);
   contract.TestEvent(function() {
      console.log("event triggered!");
   });
   contract.triggerEvent();
```

expected: prints "event triggered!"
actual: Uncaught Error: new BigNumber() not a base 16 number: 

Changing the indexed to another type such as uint256 (and calling testEvent(123, "world")) works as expected it seems to only happen when it's a (indexed) string.

ygnr:
You cannot add indexed to string, as string can have arbitrary length. Event arguments that are indexed end up in topics and topics can have only fixed types. Indexed strings are hashed and you will get hash of the string in topics. This is the reason for your error.

MichalZalecki:
You can index by string and bytes but what is going to be stored as topic is Keccak-256 hash. The problem is that web3 just fails to parse it.

PR #1626, opened 2018-05-12
Fix error when decoding events containing indexed strings as parameters

### LINKED ISSUE #1566: Events containing indexed string causes ERROR: The returned value is not a convertible string
**I know indexed strings aren't stored in the logs**

But it is impossible to subscribe to events that contain **indexed string** parameters at all. Any attempt to do so just crashes.

Tested with web3-eth 1.0.0-beta.33.

**Test.sol:**
```
pragma solidity ^0.4.21;
contract Test {
    event Updated(address from, string indexed data);
    function test(string _data) public {
        emit Updated(msg.sender, _data);
    }
}
```

**test.js:**
```
const Web3Eth = require('web3-eth');
const eth = new Web3Eth('ws://127.0.0.1:8546');

const addr = "0x79b79127fcb21120bce34a41d89dd52bd826b552"; // update to match deployed contract
const abi = [{"constant":false,"inputs":[{"name":"_data","type":"string"}],"name":"test","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"anonymous":false,"inputs":[{"indexed":false,"name":"from","type":"address"},{"indexed":true,"name":"data","type":"string"}],"name":"Updated","type":"event"}];
const contract = new eth.Contract(abi, addr);
contract.events.Updated().subscribe((err, res) => {
  console.log("Updated", err, res);
});
```

**Crashes with:**
```
Error: ERROR: The returned value is not a convertible string:
    at SolidityTypeString.formatOutputString [as _outputFormatter] (node_modules\web3-eth-abi\src\formatters.js:249:15)
    at node_modules\web3-eth-abi\src\type.js:246:25
    at SolidityTypeString.SolidityType.decode (node_modules\web3-eth-abi\src\type.js:247:11)
    at node_modules\web3-eth-abi\src\index.js:327:49
    at Array.forEach (<anonymous>)
    at ABICoder.decodeParameters (node_modules\web3-eth-abi\src\index.js:326:13)
    at ABICoder.decodeLog (node_modules\web3-eth-abi\src\index.js:371:30)
    at Object.Contract._decodeEventABI (node_modules\web3-eth-contract\src\index.js:371:31)
    at Subscription._formatOutput (node_modules\web3-core-subscriptions\src\subscription.js:123:84)
    at node_modules\web3-core-subscriptions\src\subscription.js:263:44
```

**Subscribing to events containing indexed strings should still work even if the *value* of the indexed strings isn't available in the callback.**

jchen86:
Fixes #1566 & #550.

Subscribing to events containing indexed strings causes the application to crash. This is because [indexed string parameters are stored as hash values in the topics](https://ethereum.stackexchange.com/a/7170/36266).


frozeman:
Basically all dynamic types are strored as hashes and shouldn't be decoded. The soon to be merged #1751 branch should fix that. See https://github.com/ethereum/web3.js/blob/newAbiLib/packages/web3-eth-abi/src/index.js#L272-L274

Thanks for the contribution anyway.

PR #1627, opened 2018-05-13
Fix incorrect event bool returnValues

### LINKED ISSUE #1403: event's returnValue is null instead of false
I'm not totally sure if this is a solidity issue or a web3 issue. I'm using 1.0.0-beta.29 (got same result on beta.30) over testrpc.

In my solidity contract, I set an event like this to check if some non-constant successfully run or failed:
`event opDone(bool indexed _isSet);`
Then in some function I used it like this:
```
function setRecord(address _target, string _record) public {
        if(!rights[_target][msg.sender]) {
            opDone(false);
            return;
        }
        User storage _usr = userList[_target];
        _usr.recordIndices.push(recordCount);
        records[recordCount] = _record;
        recordCount++;
        rights[_target][msg.sender] = false;
        opDone(true);
    }
```
On the JavaScript side, I called the function and captured the event like this:
```
contractInstance.methods.setRecord(target, encrypted.cipher).send({
      from: accountAddress,
      gas: '1500000'
    }).on('transactionHash', (hash) => {
      console.log('hash: ' + hash);
    }).on('confirmation', (confirmationNumber, receipt) => {
      console.log(res.json(receipt.events));
    }).on('error', console.error);
```
When the event's returnValue is true, everything seems to be fine, but if it's false (or when it should be), I got the event like this:
```
{
    "opDone": {
        "logIndex": 0,
        "transactionIndex": 0,
        "transactionHash": "0xbce295310aacd114b36474aa9a8dff199460c8af5bb2c6c8ce6f8ffd85006ba5",
        "blockHash": "0x71ce77af33ed7f601c6035f432200c8de75d8f6818fd42870541ae220920d6fa",
        "blockNumber": 8,
        "address": "0x307dcF18755ED7746CA33E70d1B8656fBe2DcA5A",
        "type": "mined",
        "id": "log_5fd6bfb4",
        "returnValues": {
            "0": null,
            "_isSet": null
        },
        "event": "opDone",
        "signature": "0xd02ea5da3b63612ae7b8f9b8708ebd07566a19a447b98061c09b4b6a9ac7de9b",
        "raw": {
            "data": "0x0",
            "topics": [
                "0xd02ea5da3b63612ae7b8f9b8708ebd07566a19a447b98061c09b4b6a9ac7de9b",
                "0x0000000000000000000000000000000000000000000000000000000000000000"
            ]
        }
    }
}
```
You can see that not only my return value (_isSet) itself looks weird, the index become null, too.
I'm not sure if it's a bug or some environment specific problem or maybe just my own fault. I know event subscription doesn't work while using http provider but I think it's okay just getting return values from an event.
BTW can anyone tell me how to test event subscription without connecting to a real chain? For example is there some "testipc" stuff? Thanks a lot if you can provide any help!

jchen86:
Fixes #1403 - False values in events are returned as nulls.

Also, the initial value for solidity bool type is false, so a bool return value should be false even when it's not set explicity.


jchen86:
@frozeman would you mind taking a look at this? This fixes #1403 . Thank you!

nivida:
This will be solved with the new ABI library.

Issue #505, opened 2016-10-18
Parity Trace module integration


diego-G:
Hi everyone,

I would like to ask if you have plans to introduce the Parity trace module into web3.js.

Thanks  in advance

rachyandco:
hi here is a fork with the trace function implemented. https://github.com/BitySA/web3.js

PR #1632, opened 2018-05-15
support parity/trace

hackmod:
these commits are imported from https://github.com/gobitfly/web3.js. the author of the etherchain-light.
- cleanup some whitespaces, (dos2unix, trim trailing spaces etc.)
- rebased onto `develop` branch

See also #505

----

- https://wiki.parity.io/JSONRPC-trace-module
- https://wiki.parity.io/JSONRPC-parity-module

### DIFF
--- a/lib/web3.js
+++ b/lib/web3.js
@@ -39,6 +39,7 @@ var Shh = require('./web3/methods/shh');
 var Net = require('./web3/methods/net');
 var Personal = require('./web3/methods/personal');
 var Swarm = require('./web3/methods/swarm');
+var Trace = require('./web3/methods/trace');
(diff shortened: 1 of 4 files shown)


frozeman:
This is great, but as it is a non generic functionality only specific to parity, im hesitant to add this to web3. I would suggest you write a module, which can extend existing web3 and publish that separate.
You can use the `web3._extend` functionality, as described here: 

Its described in the 1.0 here: http://web3js.readthedocs.io/en/1.0/web3.html#extend

And the old one has an example here:  https://github.com/ethereum/go-ethereum/blob/master/internal/web3ext/web3ext.go#L36-L63

PR #1638, opened 2018-05-17
Fix TypeScript definitions for dependencies

wbobeirne:
Adds the `@types/*` to dependencies for packages/web3 so that users don't have to add them themselves, and fixes an incorrect import of `bn.js` when this project uses `bignumber.js`


frozeman:
Thanks

PR #1650, opened 2018-05-22
Fixes #1559: Remove `Sec-WebSocket-Protocol: undefined` …

### LINKED ISSUE #1559: Invalid Websocket headers in 1.0.0-beta.34
So the issue came when updating to 1.0.0-beta.34 version, the websocket connection fails to a parity node with the following error. 

`WebSocket connection to 'wss://kovan.decenter.com:4443/' failed: Error during WebSocket handshake: Unexpected response code: 403`

The issue is that in the request headers we have value of `Sec-WebSocket-Protocol: undefined` which gets handled badly on the node since it's undefined. Reverting the version to 1.0.0-beta.33 fixes the issue.

`
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9,sr-RS;q=0.8,sr;q=0.7
Cache-Control: no-cache
Connection: Upgrade
DNT: 1
Host: kovan.decenter.com:4443
Origin: http://127.0.0.1:8080
Pragma: no-cache
Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits
Sec-WebSocket-Key: l+ZG48eY3Nt11jph4pXlYA==
Sec-WebSocket-Protocol: undefined
Sec-WebSocket-Version: 13
Upgrade: websocket
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/65.0.3325.181 Safari/537.36
`

masonforest:
...from WebSocket headers

The WebSocket constructor in a browser based environment [only takes two
arguments](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket).
Before we were passing it extra arguments which was causing it to send
invalid headers. This commit ignores all arguments except for the first two.

### DIFF
--- a/packages/web3-providers-ws/src/index.js
+++ b/packages/web3-providers-ws/src/index.js
@@ -29,7 +29,9 @@ var Ws = null;
 var _btoa = null;
 var parseURL = null;
 if (typeof window !== 'undefined' && typeof window.WebSocket !== 'undefined') {
-    Ws = window.WebSocket;
+    Ws = function(url, protocols) {
+      return new window.WebSocket(url, protocols);
+    };
     _btoa = btoa;
     parseURL = function(url) {
         return new URL(url);


Levino:
Would it be very difficult to add a test for this?

masonforest:
Hmmm, I'm not sure. If we really wanted to make a proper test for this I think we should boot up a WebSocket server and insure the provider is connecting. Are there existing tests for  WebSocketProvider? If not I think that will probably be a significant amount of work and might make sense to break out into a second issue.

Mischi:
Hi. 

I'm running into this problem atm and are wondering what the status of this PR is?
Let me know if I can help out.

Cheers,
Fabian

Mischi:
I put together a little test case that demonstrated the problem with the help of [puppeteer](). See https://github.com/Mischi/chrome-websockets-test . 
In summary I'm now convinced that this PR should fix the problem and should therefore be merged.

Let me know if I can further help to land this PR.

Cheers,
Fabian

Mischi:
@frozeman, @nivida do you know when you could merge this PR? It gets kinda painful for us to keep it as a local patch ...

I'm still interested in helping with whatever is necessary to land this PR. Just let me know 😃

frozeman:
Thanks for the test

PR #1675, opened 2018-06-02
[1.0] Add Contract.destroy()

ovx:
Add `.destroy()` method to objects created via `packageInit`. It allows for proper destruction (and proper garbage collection) of an object. This is very useful when working with a large number of contracts where it isn't necessary to keep the instances in memory.

Calling `.destroy()` unsets the internal properties, removes event listeners and unlinks all underlying objects. It also defines a new getter `(bool) destroyed` which can be used to check whether the contract is destroyed.

```js
contract.destroy();
console.log(contract.destroyed) // prints `true`
```

The main motivation for this arises from the issue https://github.com/Giveth/giveth-dapp/issues/297. Because each instance of a contract binds new event listeners to the global provider, we encounter the warning 'possible EventEmitter memory leak detected'. In fact, these contracts and instances of `RequestManager` cannot be properly garbage collected due to the references between contracts and the global provider leading to the increased memory usage.

### DIFF
--- a/packages/web3-core-requestmanager/src/index.js
+++ b/packages/web3-core-requestmanager/src/index.js
@@ -95,14 +95,16 @@ RequestManager.prototype.setProvider = function (p, net) {
 
     // listen to incoming notifications
     if(this.provider && this.provider.on) {
-        this.provider.on('data', function requestManagerNotification(result, deprecatedResult){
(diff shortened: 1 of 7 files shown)


MattOakley:
We are having the same issue. What's the state of this pr?

nivida:
This should be solved with the PR #2000 please create a new PR or issue if not :-)

PR #1684, opened 2018-06-06
prevent crash if browser is blocking 3rd party content

xtools-at:
prevent crash if loaded in iframe and browser is blocking 3rd party content, which would give you an `Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.` if the localStorage is there, but not allowed to access.

Would be glad if you could merge my PR, so I don't need to frankenstein my own web3 module anymore and can get back to using npm :)


xtools-at:
wtf Node 5? 😅

nivida:
@xtools-at Can you please do these changes in the ethereumProvider branch because the code architecture of web3.js has changed a lot there :)

PR #1701, opened 2018-06-13
fix TransactionReceipt.status

nerdDan:
`TransactionReceipt.status` is supposed to be boolean according to the document


nerdDan:
@coveralls very strange. why would changing type annotation decreases coverage?

nerdDan:
The log of Travis CI does not blame my change. What should I do?

PR #1725, opened 2018-06-21
Fix detection BigNumber's objects

mik-laj:
Hello, 

The library does not need to implement this function from scratch, but more importantly the implementation of the function in a way makes it impossible to use it together with UglifyJS. 

Cheers


frozeman:
Thanks!

Issue #1423, opened 2018-03-05
web3.js  v1.0.0-beta.30 send() - Error: Failed to check for transaction receipt:


netkiller:
my version is 1.0.0-beta.30

> console.log(web3.version)
1.0.0-beta.30

my program is the follow code.

`
fs = require('fs');
var net = require('net');
var Web3 = require('web3');
var web3 = new Web3('/Users/neo/Library/Ethereum/geth.ipc', net);  // to connect roosten network

console.log(web3.version)
const abi = fs.readFileSync('output/TokenERC20.abi', 'utf-8');

const contractAddress = "0x70682386d0dE84B1e549DC3c4305CCB2D261b2a8";
const coinbase = "0xB94054c174995AE2A9E7fcf6c7924635FBa8ECF7";
const toAddress = "0xf56b81a2bcb964D2806071e9Be4289A5559BB0fA";

balanceWei = web3.eth.getBalance(coinbase);
console.log(balanceWei);

const contract = new web3.eth.Contract(JSON.parse(abi), contractAddress, { from: coinbase , gas: 100000});

web3.eth.personal.unlockAccount(coinbase, "netkiller").then(function(result){
console.log(result)
contract.methods.balanceOf(coinbase).call().then(console.log).catch(console.error);
contract.methods.balanceOf(toAddress).call().then(console.log).catch(console.error);
});

contract.methods.transfer(toAddress, 10000).send().then(function(receipt){

console.log(receipt);

}).catch(console.error);

contract.methods.balanceOf(coinbase).call().then(console.log).catch(console.error);
contract.methods.balanceOf(toAddress).call().then(console.log).catch(console.error);

`

when I execute send()  " Error: Failed to check for transaction receipt: "

`
> contract.methods.transfer(toAddress, 10000.00).send().then(function(receipt){
... 
... console.log(receipt);
... 
... }).catch(console.error);
Promise {
  _bitField: 0,
  _fulfillmentHandler0: undefined,
  _rejectionHandler0: undefined,
  _promise0: undefined,
  _receiver0: undefined }
> Error: Failed to check for transaction receipt:
{}
    at Object._fireError (/Users/neo/ethereum/web3/node_modules/web3-utils/src/index.js:56:17)
    at /Users/neo/ethereum/web3/node_modules/web3-core-method/src/index.js:259:23
    at bound (domain.js:375:14)
    at runBound (domain.js:388:12)
[...]

PR #1735, opened 2018-06-27
fix issue #1423

leon740727:
don't unsubscribe the newBlockHeaders listener when can't get receipt, let it try until TIMEOUTBLOCK exceed.

### DIFF
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -253,16 +253,10 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
             }
             // if we have a valid receipt we don't need to send a request
             return (existingReceipt ? promiEvent.resolve(existingReceipt) : _ethereumCall.getTransactionReceipt(result))
-            // catch error from requesting receipt
-            .catch(function (err) {
-                sub.unsubscribe();
-                promiseResolved = true;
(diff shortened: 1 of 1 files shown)


packages/web3-core-method/src/index.js
```diff
@@ -253,16 +253,10 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
             }
```
nivida: Can you please do this changes in a new branch an submit this PR against 1.0? Because it has nothing to do with the new ABI lib. Thanks :)

leon740727:
Sure! my bad. Should I just merge this commit into 1.0 branch or create a new branch against 1.0 and cherry pick this commit onto it? I am a newbie on github, please help me, thanks :-)

nivida:
@leon740727 i would create a new branch with 1.0 as base and cherry pick the commits :)

leon740727:
@nivida thanks! :)

frozeman:
Can you add tests, i dont see how your fix is related to timeout blocks?

leon740727:
sorry, I am very busy recently :-(
and that situation won't happen every time. If need some test cases, it will need mock the block chain.
I haven't mock like that before. So, I will try to add some test case when I get some time...

Issue #1724, opened 2018-06-21
Out of gas exception is thrown when receipt.status is false


MattOakley:
From looking at the code, it looks like the exception should be `Transaction has been reverted by the EVM`.

But there's a bug in `web3-core-method` starting a line `362` where the receipt is converted to a string before the status is checked, meaning it always goes to the else block and throws an out of gas exception.

```
if(receipt) {
    receipt = JSON.stringify(receipt, null, 2);
}
if (receipt.status === false || receipt.status === '0x0') {
    utils._fireError(new Error("Transaction has been reverted by the EVM:\n" + receipt),
        defer.eventEmitter, defer.reject);
} else {
    utils._fireError(
        new Error("Transaction ran out of gas. Please provide more gas:\n" + receipt),
        defer.eventEmitter, defer.reject);
}
```

Caoxuyang:
Yeah, I got this error too. Here is my receipt:
`{
  "blockHash": "0xebbbfa844bffb844082a03052c544019a0b11591df28f4b4ad6edb067c1f0404",
  "blockNumber": 83,
  "contractAddress": null,
  "cumulativeGasUsed": 21000,
  "gasUsed": 21000,
  "logs": [],
  "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  "root": null,
  "status": false,
  "transactionHash": "0xe34185bdc0cdc09bb4fefa39ea59c725ca2e5854b86e31f7b93da6bdf3abffb3",
  "transactionIndex": 0
}`

And I goes to the Error: Transaction ran out of gas directly.

PR #1743, opened 2018-06-29
Bug when displaying error from receipt.

GrandVizierOlaf:
The existing code would never reach line 366 because the receipt variable had already been overwritten with a string.


GrandVizierOlaf:
@frozeman Can I get you to take a look at this for approval? It fixes #1724.

packages/web3-core-method/src/index.js
```diff
@@ -360,14 +360,14 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
 
                     } else {
                         if(receipt) {
-                            receipt = JSON.stringify(receipt, null, 2);
+                            var receiptString = JSON.stringify(receipt, null, 2);
```
nivida: When I run the tests i got this JSHint error at the end: 
```
packages/web3-core-method/src/index.js: line 366, col 104, 'receiptString' used out of scope.
packages/web3-core-method/src/index.js: line 370, col 102, 'receiptString' used out of scope.
```
Could you please move the ```var``` declaration out of the ```if``` block?

GrandVizierOlaf:
@nivida, does that fix things for you?

PR #1744, opened 2018-06-30
Propagate output formatter errors to batch callbacks

raulk:
Output formatters can throw errors. This PR wraps the call to the output formatter when processing batch responses in a try-catch block, propagating the error back to the callback.


raulk:
Not sure why coverage is decreasing here. The catch path is exercised by the new test I added, and the success path is already covered by previous tests.

PR #1751, opened 2018-07-03
New abi lib

nivida:


pors:
@nivida awesome, is this already part of a release? I still have the problem from #1626 with beta 35

PR #1753, opened 2018-07-03
[web3-providers-http] Re-use existing http connections.

ARitz-Cracker:
Sometimes, you have to make tens of thousands of http calls to your local full node to scrape through buttloads of data, such as finding the entire transaction history for an account. However `web3-providers-http` doesn't re-use existing connections! Resulting in the OS freaking out and running out of ports, which leads to an `EADDRINUSE` error every time you use any function.

This PR fixes this issue and (probably) increases performance by _not_ creating a new connection every time you do something!

Works with both `xhr2-cookies` and `xhr2`.


packages/web3-providers-http/src/index.js
```diff
@@ -24,21 +24,33 @@
 
 var errors = require('web3-core-helpers').errors;
 var XHR2 = require('xhr2-cookies').XMLHttpRequest // jshint ignore: line
+const http = require('http');
```
nivida: Could you change ```const``` to ```var``` because 1.0 is not yet es6 compatible.

packages/web3-providers-http/src/index.js
```diff
@@ -24,21 +24,33 @@
...
+const https = require('https');
+
 
 /**
  * HttpProvider should be used to send rpc calls over http
  */
 var HttpProvider = function HttpProvider(host, options) {
     options = options || {};
     this.host = host || 'http://localhost:8545';
+    if (this.host.substring(0,5) == "https"){
```
nivida: Please change ```==``` to ```===``` and use single quotes.

frozeman:
Thats a great additions, thanks!

PR #1759, opened 2018-07-06
ENS support for web3 1.0

tetratorus:
Adding ens support i.e. web3.eth.ens, and fixed broken tests

It may make sense to add it as a top-level package instead i.e. web3.ens


nivida:
@tetratorus maybe you could copy my code from here: https://github.com/nivida/web3.js/tree/ens-integration
I've added all the missing functions that @Arachnid had mentioned in his PR as you can see here https://github.com/ethereum/web3.js/pull/886 with the exception of "Add support for ENS names anywhere an address is accepted (the big one)."

nivida:
closed because of this PR: https://github.com/ethereum/web3.js/pull/1784

PR #1771, opened 2018-07-12
Remove unnecessary variable assign

### LINKED ISSUE #1770: Unnecessary variable assign in web3-core-subscription
https://github.com/ethereum/web3.js/blob/3d74aa674b951b6bcbae1faf646c1bd4d21d916b/packages/web3-core-subscriptions/src/index.js#L43

This line is unnecessary.

I think the cause of it is from copy & paste action from web3-core-method, `index.js`

There are same line 
```
 func.call = this.call; 
```
However, `this.call` is declared in web3-core-method, `index.js` file, in contrast.

nujabes403:
fixes #1770


frozeman:
Already merged somewhere in #1836

PR #1772, opened 2018-07-12
fix: websocket authn with legacy URL parser

Freyert:
Came across an issue on Stack Overflow where someone was having issues connecting to Kaleido using basic auth. Turns out their Node.js client was falling back to the legacy URL parser and we were missing it.

Seems 1.0.0-beta.34 is missing the check for the new parser, but when it's delivered this PR will be made obsolete on April 19, 2019 when Node 6 goes out of maintenance.


Freyert:
Related commits:

https://github.com/ethereum/web3.js/commit/e0f2de12433ac9c2b17111a047502c4f5ea1944c#diff-215dff4ef1684c2278bb08b065ba5037

PR #1784, opened 2018-07-17
ENS

nivida:
Implemented the API and all the wishes @Arachnid  mentioned here: https://github.com/ethereum/web3.js/pull/886 
With the exception of "Add support for ENS names anywhere an address is accepted"


packages/web3-eth-ens/src/ENS.js
```diff
@@ -0,0 +1,162 @@
...
+ * @date 2018
+ */
+
+"use strict";
+
+var config = require('./config');
+var Registry = require('./contracts/Registry');
+
+/**
+ * varructs a new instance of ENS
```
Arachnid: This is a new one on me. Victim of search-and-replace?

packages/web3-eth-ens/src/ENS.js
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * @method ENS
+ * @param {Object} eth
+ * @varructor
```
Arachnid: And here

packages/web3-eth-ens/src/ENS.js
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Sets the new content
+ *
+ * @param {string} name
+ * @param {string} hash
+ * @param {string} from
+ * @returns {Promise<Transaction>}
+ */
+ENS.prototype.setContent = function(name, hash, from) {
```
Arachnid: Could you add support for multihash/setMultihash too, please?

packages/web3-eth-ens/src/config.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
+
+var config = {
+    addresses: {
+        main: "0x314159265dD8dbb310642f98f50C066173C1259b",
+        ropsten: "0x112234455c3a32fd11230c42e7bccd4a84e02010"
+    },
```
Arachnid: It's also on Rinkeby at 0xe7410170f87102df0055eb195163a03b7f2bff4a.

packages/web3-eth-ens/src/contracts/Registry.js
```diff
@@ -0,0 +1,81 @@
...
+ * Returns the address of the owner of an ENS name.
+ *
+ * @method owner
+ * @param {string} name
+ * @return {Promise<any>}
+ */
+Registry.prototype.owner = function (name) {
+    return this.contract.then(function (contract) {
+        return contract.methods.owner(namehash.hash(name)).call();
+    }).catch(function (error) {
```
Arachnid: Does this do anything?

frozeman: How so? It calls the owner function?

packages/web3-eth-ens/src/ENS.js
frozeman: What are the respective function names? Its not in the current ABI

Arachnid: https://eips.ethereum.org/EIPS/eip-1062

packages/web3-eth-ens/src/contracts/Registry.js
Arachnid: I mean the catch, which immediately throws.

PR #1785, opened 2018-07-17
Update web3-eth.rst

ZeframLou:
Added the fact that for sign(), the input data will be prefixed with "\x19Ethereum Signed Message:\n" + message.length before signed. This omission has wasted far too many hours for far too many developers, please merge this thanks.


frozeman:
The text is not fully correct, as web3.js is not pre hashing the input, but geth, or the node does. The prefix is also added by the node.

nivida:
I've closed this due to inactivity.

PR #1799, opened 2018-07-26
Add debug_accountRangeAt/debug_storageRangeAt methods

axic:
These are needed for the ewasm testnet blockexplorer.

cc @cdetrio

### DIFF
--- a/lib/web3.js
+++ b/lib/web3.js
@@ -39,6 +39,7 @@ var Shh = require('./web3/methods/shh');
 var Net = require('./web3/methods/net');
 var Personal = require('./web3/methods/personal');
 var Swarm = require('./web3/methods/swarm');
+var Debug = require('./web3/methods/debug');
 var Settings = require('./web3/settings');
 var version = require('./version.json');
 var utils = require('./utils/utils');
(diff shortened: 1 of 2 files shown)


frozeman:
I dont think they should be part of the main web3.js lib, as they are node specific, but you can simply attach them later by using the extend function in your code: https://web3js.readthedocs.io/en/1.0/web3.html#extend

We can have a chat about that in skype if you like.

axic:
Actually I don't think these are designed to be client specific (I agree the parity ones are).

These supposed to be standard tracing methods, used by Remix and all the frameworks for debugging.

Btw, can you quickly explain the rules of inclusion? Would it be possible to have these here as a separate module (optionally loaded)? Or what is the best way to supply a distribution of these?

frozeman:
If they are supported in all/most nodes we can surely include them. 
You could always make a separate package that runs `web3.extend()..` before and adds it to the running web3.js. Ill merge this if its widely used. We should discuss adding this to 1.0 as well.

axic:
Does `extend` only works on the instance and not on the package?

axic:
@frozeman can you have a look at https://github.com/axic/web3-extended.js if this is the way you've suggested?

PR #1801, opened 2018-07-26
Add Parity trace_*/parity_* methods

axic:
These are commits from https://github.com/gobitfly/web3.js made by @ppratscher and @digithinksa.

These are used by https://github.com/gobitfly/etherchain-light and we are trying to use it for the ewasm testnet (fork: https://github.com/ewasm/etherchain-light).

cc @cdetrio

### DIFF
--- a/lib/web3.js
+++ b/lib/web3.js
@@ -39,6 +39,7 @@ var Shh = require('./web3/methods/shh');
 var Net = require('./web3/methods/net');
 var Personal = require('./web3/methods/personal');
 var Swarm = require('./web3/methods/swarm');
+var Trace = require('./web3/methods/trace');
 var Settings = require('./web3/settings');
 var version = require('./version.json');
 var utils = require('./utils/utils');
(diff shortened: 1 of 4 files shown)


frozeman:
I dont think they should be part of the main web3.js lib, as they are node specific, but you can simply attach them later by using the extend function in your code: https://web3js.readthedocs.io/en/1.0/web3.html#extend

We can have a chat about that in skype if you like.

nivida:
Awesome thanks! Sorry for bothering you after that long time but could you add these methods to the ethereumProvider branch? It should be really easy to add them (see web3-core-method/models/methods). I think it would be good to add these methods as a debug module, writing a simple documentation about and to declare them as node specific (Parity). 

The readme of web3-core & web3-core-method will help you to implement a new module.

nivida:
0.20.x will not get any new features because of the maintenance mode. Didn't see that this PR belongs to v0.20.x.
If you would like to create a module for the debug methods then please check out the new possibilities the 1.0.0-beta.38 release will give you for creating such a module: 
https://github.com/ethereum/web3-examples/tree/development/modules 

Or create a feature request issue and I will implement it later. :-)

Because of that I will close this PR.

axic:
See my comment here: https://github.com/ethereum/web3.js/pull/1799#issuecomment-414059971

I have actually moved both my PRs into a module.

PR #1813, opened 2018-07-29
[1.0] Adding default value to the subscribe callback

leonprou:
Fixes https://github.com/ethereum/web3.js/issues/1797


frozeman:
Thanks, yeah can be tricky in some cases, please change your code to ES5, as web3.js code base is not yet babel precomplied and shouldnt be expecting ES6 yet. There is a separate branch to convert the lub to ES6

leonprou:
@frozeman Done, just changed the implementation to use `_.identity`.

frozeman (approved):

PR #1821, opened 2018-07-31
Issue and PR template added

nivida:


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+Please delete options that are not relevant.
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
+
+## Checklist:
+
+- [ ] I have selected the correct base branch.
+- [ ] I have fixed the tests and extended if necessary.
```
frozeman: not sure if we should suggest fixing tests, its note like fix your code to pass tests :)

frozeman: and adding tests is a must, if new things are added :)

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+- [ ] I have performed a self-review of my own code.
```
frozeman: I also should test my code on the live network. 

And we should add, i ran `npm run build` and tested the resulting file from the `dist` folder in a browser.

frozeman (changes requested):
Looks good, just add the few things i noted

frozeman (approved):

PR #1829, opened 2018-08-02
Fix #1828 Wrongly ordered JSON.stringify function

### LINKED ISSUE #1828: Wrongly ordered JSON.stringify function on web3-core-method `index.js`
When it failed on checking transaction receipt, it goes to this code below:

```
if(receipt) {
    receipt = JSON.stringify(receipt, null, 2);
}
if (receipt.status === false || receipt.status === '0x0') {
    utils._fireError(new Error("Transaction has been reverted by the EVM:\n" + receipt),
        defer.eventEmitter, defer.reject);
} else {
    utils._fireError(
        new Error("Transaction ran out of gas. Please provide more gas:\n" + receipt),
        defer.eventEmitter, defer.reject);
}
```

1) Unnecessary checking whether receipt is truthy or falsy value.
`if (receipt) { ... }`
- There is no need to check receipt since it already resolved from previous .then block. So there can't be the case receipt is falsy value.

2) Wrongly ordered JSON.stringify function.

`receipt = JSON.stringify(receipt, null, 2)` code changes the receipt type from object to JSON string.

Then we can't get a property of the object.

ex)
var obj = { a: 1, b: 2 }
obj.a // 1
obj = JSON.stringify(obj)
obj.a // undefined  

So we need to fix this statement.

Whole code: 
```
// CHECK for normal tx check for receipt only
  .then(function(receipt) {

      if (!isContractDeployment && !promiseResolved) {

          if(!receipt.outOfGas &&
              (!gasProvided || gasProvided !== receipt.gasUsed) &&
              (receipt.status === true || receipt.status === '0x1' || typeof receipt.status === 'undefined')) {
              defer.eventEmitter.emit('receipt', receipt);
              defer.resolve(receipt);

              // need to remove listeners, as they aren't removed automatically when succesfull
              if (canUnsubscribe) {
                  defer.eventEmitter.removeAllListeners();
              }

          } else {
              if(receipt) {
                  receipt = JSON.stringify(receipt, null, 2);
              }
              if (receipt.status === false || receipt.status === '0x0') {
                  utils._fireError(new Error("Transaction has been reverted by the EVM:\n" + receipt),
                      defer.eventEmitter, defer.reject);
              } else {
                  utils._fireError(
                      new Error("Transaction ran out of gas. Please provide more gas:\n" + receipt),
                      defer.eventEmitter, defer.reject);
              }
          }

          if (canUnsubscribe) {
              sub.unsubscribe();
          }
          promiseResolved = true;
      }

  })
```

nujabes403:
Fixed wrongly ordered JSON.stringify function. (#1828 )


nujabes403:
@frozeman Please check my request

nivida (approved):

PR #1831, opened 2018-08-02
add prettier to format code follow .editorconfig

pierreneter:


frozeman:
Thanks for this, but sadly it effects to many lines of code, to be able to be reviewable. Major refactors like this need to be done step by step, to ensure no malicious code is inserted. I will take a look at prettier code, but not sure if we will apply it. Thanks anyway.

PR #1836, opened 2018-08-02
delete a code, not meaningful

RuleTheMoments:
there is no other call, so that assignment meaningless


pierreneter:
duplicate of #1771

pierreneter:
Duplicate of #1771

frozeman (approved):
Thanks for the find

PR #1845, opened 2018-08-05
Fix #1844 rpc call 'getGasPrice' is called even when tx.gasPrice is set to 0.

### LINKED ISSUE #1844: Call 'getGasPrice' rpc when tx.gasPrice is set to 0 intentionally.
Sometimes we need to send a transaction of which gasPrice is 0.
Then we should send a transaction like below:
```
contractInstance.methods.myMethod().send({
  from: ...,
  to: ...,
  value: ...,
  gasPrice: 0,
})
```

However, this code can't work as we want. When we want to send a transaction with 0 gasPrice, we need to wrap it as a string type.

ex)
```
contractInstance.methods.myMethod().send({
  from: ...,
  to: ...,
  value: ...,
  gasPrice: '0', // string type
})
```

This is because of the code below: (in 'packages/web3-core-method/src/index.js' L564)
```
if(isSendTx && _.isObject(payload.params[0]) && !payload.params[0].gasPrice) {
...getGasPrice (request getGasPrice rpc call to node.)
}
```

This code helps when we don't want to specify gasPrice for transaction(it requests rpc call 'getGasPrice' manually when we didn't set a gasPrice for transaction.)

However, the expression `!payload.params[0].gasPrice` is passed even when gasPrice is set to 0.

So we need to change the code like this: `payload.params[0].gasPrice === undefined`

It can covers when gasPrice is set 0 intentionally.

Of course, in ethereum mainnet, there can't be a transaction of which gasPrice is 0, but it needs when we send a transaction  in our private network.

whole code: 
```
if(isSendTx && _.isObject(payload.params[0]) && !payload.params[0].gasPrice) {

    var getGasPrice = (new Method({
        name: 'getGasPrice',
        call: 'eth_gasPrice',
        params: 0
    })).createFunction(method.requestManager);

    getGasPrice(function (err, gasPrice) {

        if (gasPrice) {
            payload.params[0].gasPrice = gasPrice;
        }
        sendRequest(payload, method);
    });

} else {
```

nujabes403:
Fixed wrongly calling 'getGasPrice' even when tx.gasPrice is set to 0. (#1844 )

### DIFF
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -561,7 +561,7 @@ Method.prototype.buildCall = function() {
         };
 
         // Send the actual transaction
-        if(isSendTx && _.isObject(payload.params[0]) && !payload.params[0].gasPrice) {
+        if(isSendTx && _.isObject(payload.params[0]) && payload.params[0].gasPrice === undefined) {
 
             var getGasPrice = (new Method({
                 name: 'getGasPrice',
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -561,7 +561,7 @@ Method.prototype.buildCall = function() {
         };
 
         // Send the actual transaction
-        if(isSendTx && _.isObject(payload.params[0]) && payload.params[0].gasPrice === undefined) {
+        if(isSendTx && _.isObject(payload.params[0]) && typeof payload.params[0].gasPrice === 'undefined') {
 
             var getGasPrice = (new Method({
                 name: 'getGasPrice',


nujabes403:
Please check my request @frozeman @nivida :)

packages/web3-core-method/src/index.js
```diff
@@ -561,7 +561,7 @@ Method.prototype.buildCall = function() {
         };
 
         // Send the actual transaction
-        if(isSendTx && _.isObject(payload.params[0]) && !payload.params[0].gasPrice) {
+        if(isSendTx && _.isObject(payload.params[0]) && payload.params[0].gasPrice === undefined) {
```
frozeman: better is `typeof payload.params[0].gasPrice === 'undefined'`

nujabes403: I agree!

nujabes403: @frozeman I agree. I changed comparison logic through `typeof` operator. https://github.com/ethereum/web3.js/pull/1845/commits/d034858a6d59ca2a73cb4ccd6cbfd0cf8dc13372

nujabes403:
**Description**
Fixed wrongly calling 'getGasPrice' even when tx.gasPrice is set to 0. 

**Type of change**
- [x]  Bug fix (non-breaking change which fixes an issue)
- [ ]  New feature (non-breaking change which adds functionality)
- [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)

**Checklist:**
- [x]  I have selected the correct base branch.
- [x]  I have performed a self-review of my own code.
- [x]  I have commented my code, particularly in hard-to-understand areas.
- [ ]  I have made corresponding changes to the documentation.
- [x]  My changes generate no new warnings.
- [ ]  Any dependent changes have been merged and published in downstream modules.
- [x]  I ran npm run test with success and extended the tests if necessary.
- [x]  I ran npm run build and tested the resulting file from dist folder in a browser.
- [x]  I have tested my code on the live network.

@nivida Changed PR formatting.

frozeman (approved):

PR #1851, opened 2018-08-07
Support websocket reconnect on web3-providers-ws.

### LINKED ISSUE #1558: Reconnect on Websocket Drop
Hello, 

I am making a websocket connection to Infura Ropsten
`wss://ropsten.infura.io/ws`

and then making contract event subscription like so
` contract.events.MyEvent()
        .on('data', function(event) {
            console.log('Event Generated', data)
        })
        .on('error', function(error) {
            console.log('Event Error', error)
        })`

Infuras websockets seem to be unstable and drop randomly. Unfortunately the websocket drops silently in the background and my event never receives the error. My app just continues to work and stops receiving event data. 

I can tell that websocket drops by running `setInterval` on
`web3.eth.net.isListening()`

Is there a graceful way to catch the websocket drop and reconnect immediately? If not what is a safe interval to run isListening on?

Thanks!

nujabes403:
## Description

Support websocket reconnect, when connection is closed.

## Type of change

## Checklist:

#fix #1558 #1852


nujabes403:
Please check my request @frozeman @nivida 🙂

packages/web3-providers-ws/src/index.js
```diff
@@ -76,6 +76,7 @@ var WebsocketProvider = function WebsocketProvider(url, options)  {
     }
 
     this.connection = new Ws(url, protocol, undefined, headers);
+    this.reconnect = () => new Ws(url, protocol, undefined, headers);
```
levino, line 79: I would think you must bind the result of `this.reconnect()` to `this.connection`. If not then you are awkwardly relying on some side-effect from the Websockets constructor call. My bet is: This does not work at all. Have you tested it?

My suggestion: Add an automated test for this.

packages/web3-providers-ws/src/index.js
```diff
@@ -263,7 +264,9 @@ WebsocketProvider.prototype.send = function (payload, callback) {
         } else {
             console.error('no error callback');
         }
-        callback(new Error('connection not open'));
+        // try reconnect, when connection is gone
+        this.reconnect();
+        callback(new Error('connection not open. try reconnecting...'));
```
levino, line 269: What is the point of throwing an error when at the same trying to solve it. You should silently retry until you are sure there is no way to recover. Then you should stop trying to reconnect and throw an error. This here does not make sense.

levino (changes requested):
I do not think that this code works. Adding a test that shows that it works would help me in my confidence. Also I think https://github.com/ethereum/web3.js/issues/1852 is not an issue that should be accepted. It is okay to throw an error when the websocket connection fails and let the user handle the problem upstream (by recreating the provider for example)

packages/web3-providers-ws/src/index.js
frozeman: I agree

nujabes403: Yes `binding` will be good. Thanks :)

packages/web3-providers-ws/src/index.js
nujabes403: I agree

frozeman: There is a need to inform web3.js and the developer, as subscription are socket bound, if the socket breaks and auto-reconnects, developers will wonder why they never get their subscriptions again, as it was basically canceled without them knowing. Also web3.js has some logic of re-subscribing. Please check if that still works! https://github.com/ethereum/web3.js/blob/1.0/packages/web3-core-subscriptions/src/subscription.js#L278-L289

nujabes403: @frozeman Thanks! I'll check it :)

nivida:
I've closed this PR because of #1966

PR #1853, opened 2018-08-07
Fix basic auth connection in README

hshar7:
I spent a really long time debugging this until I found out that the documented way on this page is outdated.


frozeman (approved):
This should work, but the real order is `host, timeout, user, password, headers`. Its different for 1.0, though.

Issue #1863, opened 2018-08-10
Code does not follow .editorconfig definitions.


levino:
Currently the code does not follow the `.editorconfig` definitions. For example: https://github.com/ethereum/web3.js/blob/9bf580321868430554b00c75a650bbdc856f9804/package.json has an indentation of 2 instead of 4. `jshint` looks like a bad tool to me. What about using `eslint` instead?

frozeman:
That should actually be `lerna` and `npm --save` that auto updates package.json and reformats them to 2 spaces.

levino:
You are right, these tools format the `package.json` files. But they do not lint them. We should have a linter in place to make sure the format stays correct. Then one can add `npm --save` or so to the `format` script.

nivida:
This got fixed with adding of eslint.

PR #1864, opened 2018-08-10
Use eslint as linting tool.

levino:
eslint throws 7501 errors with a basic configuration. Most of them are low hanging fruits like unused variables and stuff. The code quality is disturbingly bad!

Obviously this PR cannot be merged before all errors are fixed or the severity for these error types is turned to warning. I would immediatly start fixing these errors before doing anything else @nivida


frozeman:
Many of these "errors" are a question of taste. What is wrong with `jshint`, which we use?

levino:
https://www.sitepoint.com/comparison-javascript-linting-tools/

For example "unused variables" are a big problem for the code quality. As JSHint happily ignores these things I would prefer a stricter linter. As I was saying this originates from my observation that the `package.json` files are not correctly formatted. See #1863 I think adding eslint does not immediatly fix that issue but it should be possible to use it to lint the `package.json` files.

leonprou:
I would suggest to use [standartjs](https://standardjs.com/) - npm, github and bitcoinjs are using this coding style.

levino:
Also okay with me.

levino:
BTW: standardjs is basically a specific eslint config, right? The thing with standardjs and `web3` is that `web3` so far uses semicolons at the end of lines. This is why I think the change is a rather hard one. I would vote to act more lightweight and just apply eslint with the same rules as before and fix obvious errors like unused variables.

decanus (approved):

nivida:
@Levino Hay I've added eslint in the ethereumProvider branch and because of this I will close it but thanks for submitting this PR!

PR #1873, opened 2018-08-14
fixed bytes[] decode issue

LeapM:
for return type bytes[], the current version dynamic offset doesn't add the base offset.


nivida:

PR #1879, opened 2018-08-15
removed compile methods

nivida:
solves https://github.com/ethereum/web3.js/issues/493


frozeman (approved):

xareelee:
Hi, how can I compile solidity code via web3 v1.0 after removing this function?

PR #1895, opened 2018-08-22
docs: Add License explicitly, and edit the README

RichardLitt:
This PR adds the LGPL-3.0 license into the License file, so that GitHub will
pick up the license and add it to the repository metadata, and so that users
don't need to go to an external site to get the license, which should probably
be bundled with the code anyway.

I also removed the Migration section and put it in Usage, as it doesn't need
to be at the top of the README at the moment; it was added there three years
ago, according to `git blame`. I also added a Table of Contents and edited the
file for readability.


frozeman (approved):

frozeman:
Thanks, we could do that for the 1.0 branch too

PR #1897, opened 2018-08-23
Specify HTTPS protocol for github link to WebSocket-Node

mikeseese:
There seems to be an issue with `git://` or `git+ssh://` URLs when executing `npm install` on a module which requires a package which specifies a dependency with one of these protocols.

This issue seems to be mitigated in NPM versions 6+, but in v5 NPM will fail to retrieve the package (in this case, specifically `websocket`). This is due to the fact that `git://` is the same as `git+ssh://` which is only accessible if you have ssh access to that repository. NPM 6 must have a convenience fallback to the https URL when the SSH one is fixed.

However, we should still specify the `git+https://` in our repositories so that users who have not upgraded to NPM 6 yet can still properly install the repository.

This PR fixes the link for `websocket` within the `web3-providers-ws` package.

CC @frozeman


frozeman:
So we hat just a PR before who changed it back, as he ad issues retrieving a package and changed it to "github": https://github.com/ethereum/web3.js/pull/1624/files, What is better?

mikeseese:
The absolute best way is for you to publish an NPM package to something like `@web3/websocket` and use that as the dependency.

I haven't heard of anyone saying `git+https://` wouldn't work, but I'm having issues reproducing the error now. NPM does seem to have varying issues with git urls. I'm not sure which one is better personally! Either one is probably better than `git://` or `git+ssh://`.

However an NPM package would definitely be best.

nivida:
Thanks for submitting this issue I'll talk to @frozeman for publishing this as an NPM module and because of that I'll close this PR.

PR #1898, opened 2018-08-23
Add sha256 util

komninoschatzipapas:


nivida:

PR #1900, opened 2018-08-23
Add sha256 & soliditySha256 utils

komninoschatzipapas:


frozeman:
Cloud you describe why you added this and how it differs to the sha3 function?

komninoschatzipapas:
@frozeman Hello! Yes, I added it because I need to use the sha256 functionality in one of my projects. It is a different hash function than sha3 and both sha3 & sha256 are available in Solidity. Sha3 as an EVM opcode and sha256 as a precompiled contract at address 2.

nivida:
Could you add these methods with a new PR? The Web3.js architecture did change a lot.

PR #1912, opened 2018-08-31
Fix wallet.remove when address checksum is all lowercase

fvictorio:
If the checksum of an address is all lowercase, removing it throws an error. This happens because first the checksumed address is removed and then the private key of the lowercased address is accessed. If they are the same value, an error is thrown.

I added an address with a lowercase checksum to the tests, and a check before deleting the lowercased address.


nivida:
I've closed this PR because the architecture did change a lot and it would simplify this PR if a new one gets opened.

fvictorio:
This problem doesn't seem to be happening in the latest version, so no need to open a new PR.

PR #1929, opened 2018-09-09
Add helper function that build ethereum links according to eip-681 and eip-831

artall64:
EIP: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-831.md


artall64:
Closed to redesign, and improve coverage

PR #1940, opened 2018-09-13
wrong order fix

greg-flexidao:
Using: 
* parity 1.9.4 - 2.0.4 
* web3-beta.34-36 

I encountered multiple times this issue, under specific circumstances. The transaction is correctly send to the client, it's mined yet, the function fails to provide the tx hash and receipt and `.send()` never returns the promise/ triggers callback. Change of order of this functions fixes the issue.

### DIFF
--- a/packages/web3-providers-ipc/src/index.js
+++ b/packages/web3-providers-ipc/src/index.js
@@ -209,9 +209,8 @@ IpcProvider.prototype.send = function (payload, callback) {
     if(!this.connection.writable)
         this.connection.connect({path: this.path});
 
-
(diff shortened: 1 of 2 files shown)


nivida:
Could you do these changes again the ethereumProvider branch? @bytniak

greg-flexidao:
Sure, but let me test it again as I'm not 100% sure anymore if this was the sole reason for that bug.

nivida:
I'm closing this PR because I changed the providers a lot in the PR #2000 and this should no longer be the case.

Issue #1941, opened 2018-09-13
[web3@1.x] sendTransaction's error event doesn't return receipt as the second parameter


Anaphase:
When calling `web3.eth.sendTransaction()` or `web3.eth.sendSignedTransaction()`, the PromiEvent returned does not provide `receipt` as the second parameter to it's `error` event [like the docs suggest here](https://web3js.readthedocs.io/en/1.0/web3-eth.html#eth-sendtransaction-return):

```javascript
// using the event emitter
web3.eth.sendTransaction({
    from: '0xde0B295669a9FD93d5F28D9Ec85E40f4cb697BAe',
    to: '0x11f4d0A3c12e86B4b5F39B213F7E19D048276DAe',
    value: '1000000000000000'
})
.on('transactionHash', function(hash){
    ...
})
.on('receipt', function(receipt){
    ...
})
.on('confirmation', function(confirmationNumber, receipt){ ... })
.on('error', console.error); // If a out of gas error, the second parameter is the receipt.

```

Instead, it appears to be `JSON.stringify`-ing the receipt and just appending it to the error message, which isn't super useful if you need to read some data off the receipt object. As a workaround, I can just call `web3.eth.getTransactionReceipt()` in the error handler (or like, `JSON.parse()` everything after the first line break in `error.message`?), but that's not ideal since the receipt _should_ be available to me already.

After digging around a bit, I found that [this line](https://github.com/ethereum/web3.js/blob/1.0/packages/web3-core-method/src/index.js#L362) is indeed `JSON.stringify`-ing the receipt and appending it to the error message.

You can test this with some trivial example code:

```javascript
web3.eth.sendTransaction(transactionData) // specify a gasLimit you know will be too low for the contract method you're calling

  .on('error', (error, receipt) => {
    // receipt is undefined and error.message is:
    //
    // Transaction has been reverted by the EVM:
    // {
    //   "transactionHash": "0xa2ac91c7affd59a761dd088230008082a0bcd3c43e46dd70a70e79db9c5df993",
    //   "transactionIndex": 0,
    //   "blockHash": "0x8efa4a788fd29ccc0d25a893635c4c37ebe9ddc452df83f7c249181451f8c0e0",
    //   "blockNumber": 267,
    //   "gasUsed": 240358,
    //   "cumulativeGasUsed": 240358,
    //   "contractAddress": null,
    //   "logs": [],
    //   "status": false,
[...]

PR #1943, opened 2018-09-13
Updates _confirmTransaction to make receipt available in error handlers if a transaction fails

Anaphase:
This PR provides a fix for issue #1941. If a transaction fails, the receipt will now be available on `error.data`. As a bonus, any other method that calls `utils._fireError()` with the `{ message, data }` object will also have the `data` value preserved on `error.data` when passed to the callback / event.

Some example code:

```javascript
// specify a gasLimit you know will be too low for the contract method you're
//  calling so that the transaction will fail
web3.eth.sendTransaction(transactionData)
  .on('error', (error) => {

    // error.data is the raw receipt object, i.e error.data.gasUsed === 240358

    // error.message is still:
    //
    // Transaction has been reverted by the EVM:
    // {
    //   "transactionHash": "0xa2ac91c7affd59a761dd088230008082a0bcd3c43e46dd70a70e79db9c5df993",
    //   "transactionIndex": 0,
    //   "blockHash": "0x8efa4a788fd29ccc0d25a893635c4c37ebe9ddc452df83f7c249181451f8c0e0",
    //   "blockNumber": 267,
    //   "gasUsed": 240358,
    //   "cumulativeGasUsed": 240358,
    //   "contractAddress": null,
    //   "logs": [],
    //   "status": false,
    //   "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
    // }

  })
```

### DIFF
--- a/docs/web3-eth-contract.rst
+++ b/docs/web3-eth-contract.rst
@@ -545,7 +545,7 @@ The **callback** will return the 32 bytes transaction hash.
 - ``"transactionHash"`` returns ``String``: is fired right after the transaction is sent and a transaction hash is available.
 - ``"receipt"`` returns ``Object``: is fired when the transaction *receipt* is available. Receipts from contracts will have no ``logs`` property, but instead an ``events`` property with event names as keys and events as properties. See :ref:`getPastEvents return values <contract-events-return>` for details about the returned event object.
 - ``"confirmation"`` returns ``Number``, ``Object``: is fired for every confirmation up to the 24th confirmation. Receives the confirmation number as the first and the receipt as the second argument. Fired from confirmation 1 on, which is the block where it's minded.
-- ``"error"`` returns ``Error``: is fired if an error occurs during sending. If a out of gas error, the second parameter is the receipt.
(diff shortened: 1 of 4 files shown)


Anaphase:
Hmm, I'm not sure why coveralls is failing on that test for Node 7. When I switch nvm to use Node 7 and run `npm run test` and `npm run test-coveralls` locally, all tests pass for both ¯\\\_(ツ)_/¯

nivida:
Thanks for your pull request!
Could you update this PR later when the ethereumProvider branch got merged or do these changes against the ethereumProvider branch?

Anaphase:
Well it turns out the ethereumProvider branch is vastly different than the 1.0 branch so it's not a quick & easy port. This bug may not even exist in the new branch but I don't have the time to investigate at the moment.

PR #1956, opened 2018-09-18
allow web3 connections to localhost without credentials

therightstuff:
allow web3 connections to localhost without credentials for sample code and development


therightstuff:
Travis is failing because of a Promise support issue between ES5 and ES6, going to fix that now

PR #1964, opened 2018-09-22
update web3-eth-accounts

captPython:
under https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html#signtransaction
Parameter section 
gasPrice - String: (optional) The gas price set by this transaction, if empty, it will use web3.eth.gasPrice()

Change ->>> web3.gasPrice() should be replaced by web3.getGasPrice()


captPython:
under https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html#signtransaction
Parameter section
gasPrice - String: (optional) The gas price set by this transaction, if empty, it will use web3.eth.gasPrice()

Change ->>> web3.gasPrice() should be replaced by web3.getGasPrice()

PR #1966, opened 2018-09-22
Full support for websocket reconnection/resubscription 

### LINKED ISSUE #1085: Provider does not reconnect after failure
How to reproduce:
1. instantiate a web3 instance using a websockets provider (in my case - I connected it to a local parity client)
2. create a subscription to an event (in my case - newBlockHeaders)
3. kill parity 
Result: The process crashes with the following output:
[..]node_modules/web3-core-subscriptions/src/subscription.js:279
                            _this.options.requestManager.provider.reconnect();
                                                                  ^
TypeError: _this.options.requestManager.provider.reconnect is not a function
    at Timeout._onTimeout ([..]node_modules/web3-core-subscriptions/src/subscription.js:279:67)
    at ontimeout (timers.js:488:11)
    at tryOnTimeout (timers.js:323:5)
    at Timer.listOnTimeout (timers.js:283:5)

### LINKED ISSUE #1391: Question: Keepalive for ws subscriptions?
Current implementation of subscriptions using the websocket provider show the following behavior:

1. Create a subscription i.e. for new blocks. Do nothing else.
2. Ensure new blocks are being received.
3. Take the net connectivity down.
4. As there are no other packets to send through the provider, the subscription never errores out.
5. Take the connectivity up again.
6. The subscription never reconnects as it never detected the error condition.

Is there any plan or any idea on how to implement some kind of keepalive packages so the disconnection is detected by the subscription manager and timely notify the error to the lib user?

### LINKED ISSUE #1558: Reconnect on Websocket Drop

### LINKED ISSUE #1852: Reconnect feature is missing when websocket connection is closed.
**Expected behavior**
After websocket connection is closed, it should try reconnecting.

**Actual behavior**
After websocket connection is closed, it doesn't do anything except notifying "connection is closed".

**Steps to reproduce the behavior**
Try offline mode on chrome developer tool.

**Versions**
npm 5.6.0
node 8.11.3
web3js 1.0.0.beta-35
OS X high sierra

### LINKED ISSUE #1933: Can't assign more than one handler to connection events
I'd like to catch `end` event in one place and do reconnect and resubscribe to `.allEvents` in another place, but I can't assign more than function to `connect` `end` and `error` events of websocket provider. Here is part of code explaining why it happens
```
WebsocketProvider.prototype.on = function (type, callback) {

    if(typeof callback !== 'function')
        throw new Error('The second parameter callback must be a function.');

    switch(type){
        case 'data':
            this.notificationCallbacks.push(callback);
            break;

        case 'connect':
            this.connection.onopen = callback;
            break;

        case 'end':
            this.connection.onclose = callback;
            break;

        case 'error':
            this.connection.onerror = callback;
            break;
    }
```

gabmontes:
## Description

Websocket disconnection is a long standing issue in the 1.0 branch of this library and is the source of many issues when developing applications that for any reason use a not-so-reliable connection to the nodes.

Solving the disconnection and reconnection means basically resolving several major problems:
- Reconnecting the transport layer, this is the websocket connection itself.
- Reattaching client's event listeners that handle the different websocket connection states.
- Resubscribing to higher-level events like incoming blocks, events.

This PR addresses these problems by:
- Updating the events management functions of the websocket provider, adding `once` and updating `on` to use `addEventListener` instead of just overwriting the default event listener. That was needed to properly manage the error events that needed to be sent to every `RequestManager`s using the provider and later allow them to listen for the new `connect` events.
- Update the `RequestManager`'s error handling logic so it does the same steps when the connection was already established and an error occurs or when the connection cannot be established on start due to a recoverable cause. This was needed to resubscribe in all scenarios. In addition there was two small bugs that was fixed so the resubscriptions can actually be processed and issues are avoided in the case several resubscriptions are tried simultaneously (no function reentry).
- Once errors are sent by the websocker provider, those have to be properly broadcasted and handled by all subscriptions to trigger unsubscription/resubscription logic.
- In order to avoid extensive changes to the websocket provider, the websocket object used was -optionally- replaced by a wrapper object exposing the W3C Websocket API that automatically handles the reconnection on recoverable error conditions.
- On Node.js environments, the websocket package used has the ability to send keepalive packets and detect a zombie connection (connection was lost but since no packages are sent, no errors are triggered). In this scenarios, the reconnection is also triggered. Unfortunately, browser implementations do not have keepalive options so client's code will need to detect zombie connections and force a restart.

The changes were extensively tested and work as expected, reconnecting and resubscribing under several different circumstances.

The commits of this PR have extensive comments explaining the rationale behind each one.

Fixes #1085
Fixes #1391
Fixes #1558
Fixes #1852 
Fixes #1933

And possibly addresses others issues as well.

@nivida @frozeman Please review and evaluate this PR. I will be happy to address any comment you might have on it.

Thanks!

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-core-requestmanager/src/index.js
+++ b/packages/web3-core-requestmanager/src/index.js
@@ -205,17 +205,20 @@ RequestManager.prototype.addSubscription = function (id, name, type, callback) {
  * @param {Function} callback   fired once the subscription is removed
  */
 RequestManager.prototype.removeSubscription = function (id, callback) {
-    var _this = this;
-
     if(this.subscriptions[id]) {
+        var type = this.subscriptions[id].type;
 
+        // remove subscription first to avoid reentry
+        delete this.subscriptions[id];
+
+        // then, try to actually unsubscribe
         this.send({
-            method: this.subscriptions[id].type + '_unsubscribe',
+            method: type + '_unsubscribe',
             params: [id]
         }, callback);
-
-        // remove subscription
-        delete _this.subscriptions[id];
+    } else if (typeof callback === 'function') {
+        // call the callback if the subscription was already removed
+        callback(null);
     }
 };
 
--- a/packages/web3-core-requestmanager/src/index.js
+++ b/packages/web3-core-requestmanager/src/index.js
@@ -103,13 +103,16 @@ RequestManager.prototype.setProvider = function (p, net) {
                 _this.subscriptions[result.params.subscription].callback(null, result.params.result);
             }
         });
-        // TODO add error, end, timeout, connect??
-        // this.provider.on('error', function requestManagerNotification(result){
-        //     Object.keys(_this.subscriptions).forEach(function(id){
-        //         if(_this.subscriptions[id].callback)
-        //             _this.subscriptions[id].callback(err);
-        //     });
-        // }
+
+        // notify all subscriptions about the error condition
+        this.provider.on('error', function (event) {
+            Object.keys(_this.subscriptions).forEach(function(id){
+                if(_this.subscriptions[id] && _this.subscriptions[id].callback)
+                _this.subscriptions[id].callback(event.code || new Error('Provider error'));
+            });
+        });
+
+        // TODO add end, timeout, connect??
     }
 };
 
--- a/packages/web3-core-subscriptions/src/subscription.js
+++ b/packages/web3-core-subscriptions/src/subscription.js
@@ -271,37 +271,44 @@ Subscription.prototype.subscribe = function() {
                         _this.callback(null, output, _this);
                     });
                 } else {
-                    // unsubscribe, but keep listeners
-                    _this.options.requestManager.removeSubscription(_this.id);
+                    _this._resubscribe(err);
+                }
+            });
+        } else {
+            _this._resubscribe(err);
+        }
+    });
+
+    // return an object to cancel the subscription
+    return this;
+};
+
+Subscription.prototype._resubscribe = function (err) {
+    var _this = this;
+
+    // unsubscribe
+    this.options.requestManager.removeSubscription(this.id);
 
                     // re-subscribe, if connection fails
-                    if(_this.options.requestManager.provider.once) {
-                        _this._reconnectIntervalId = setInterval(function () {
+    if(this.options.requestManager.provider.once && !_this._reconnectIntervalId) {
+        this._reconnectIntervalId = setInterval(function () {
                             // TODO check if that makes sense!
                             if (_this.options.requestManager.provider.reconnect) {
                                 _this.options.requestManager.provider.reconnect();
                             }
                         }, 500);
 
-                        _this.options.requestManager.provider.once('connect', function () {
+        this.options.requestManager.provider.once('connect', function () {
                             clearInterval(_this._reconnectIntervalId);
+            _this._reconnectIntervalId = null;
                             _this.subscribe(_this.callback);
                         });
                     }
-                    _this.emit('error', err);
+
+    this.emit('error', err);
 
                      // call the callback, last so that unsubscribe there won't affect the emit above
-                    _this.callback(err, null, _this);
-                }
-            });
-        } else {
-          _this.callback(err, null, _this);
-          _this.emit('error', err);
-        }
-    });
-
-    // return an object to cancel the subscription
-    return this;
+    this.callback(err, null, this);
 };
 
 module.exports = Subscription;
--- a/packages/web3-core-subscriptions/src/subscription.js
+++ b/packages/web3-core-subscriptions/src/subscription.js
@@ -289,25 +289,29 @@ Subscription.prototype._resubscribe = function (err) {
     // unsubscribe
     this.options.requestManager.removeSubscription(this.id);
 
-                    // re-subscribe, if connection fails
+    // re-subscribe, if connection fails
     if(this.options.requestManager.provider.once && !_this._reconnectIntervalId) {
         this._reconnectIntervalId = setInterval(function () {
-                            // TODO check if that makes sense!
-                            if (_this.options.requestManager.provider.reconnect) {
-                                _this.options.requestManager.provider.reconnect();
-                            }
-                        }, 500);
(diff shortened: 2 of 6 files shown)


frozeman:
Looks very interesting! Thanks for describing your PR so well. 
Be aware that reconnecting needs to warms the developer properly as  subscriptions are based on sockets. So silent reconnects would lead to canceled subscriptions (or better those are sent to an non existing socket now) so people would wonder why they don’t get logs etc anymoe. 

@nivida you probably need to manually incorporate this into your work on the ethereum provider.

gabmontes:
@frozeman thanks for taking a look at it! You made a very key point here. Most of the complexity behind this PR was to ensure the lib resubscribes after websocket reconnects.

This PR handles both automatic websocket reconnection **and** automatic unsubscription/resubscription on websocket reconnection so i.e. new header events and logs are received again after reconnection.

My rationale was not to notify the client/developer if some error condition is being automatically handled. That worked for me but to be honest I also thought on the possibility of adding some additional events so the client/developer is aware of the reconnection/resubscription taking place in the background.

By the way, Travis is failing only in Node.js 5 due to a rest argument in the websocket reconnection library. I will work on that so all tests pass.

monitz87:
Does this also reattach SolidityEvent listeners? Is that a possibility if not?

gabmontes:
@monitz87 yes it does!

Once the underlying websocket reconnects, all subscriptions (i.e. `web3.eth.subscribe()`, `myContract.events.MyEvent()`) are resend to the Ethereum node over the new websocket connection so events are received again.

monitz87:
Maybe this is covered, but better safe than sorry. When I was toying with this concept, reattaching event listeners caused the queue to grow uncontrollably until the memory ran out.

Just a heads up. I hope this pans out and your solution gets merged soon.

nivida:
@gabmontes First of all thanks for submitting this PR! This was on my roadmap for the next release! The problem is that I've refactored close to the entire code architecture in the ethereumProvider branch because the current code is not as good as it could be ;) (see here: https://github.com/ethereum/web3.js/pull/2000)
Could you do these changes against the ethereumProvider branch? I think it shouldn't be a big thing to change it to ES6 etc. sorry :)

nivida:
This got reimplemented and improved in the ethereumProvider branch and because of this I will close your PR. Thanks for the inspiration and your efforts it speeded up the process a lot.

gabmontes:
Awesome!

7-of-9:
@nivida Any word on when the ethereumProvider implementation of this will be available in an npm package version?

I just tried a server disconnect/reconnect test with npm v1 beta 37 but there wasn't any reconnect on the client when the server was put back online. Is the fix present in that release, or any special incantations required to get it to operate?

thanks in advance!

kielabokkie:
@nivida Is this going to be part of beta.38 or will it be a later version?

7-of-9:
+1 -- would be great to get this in v38 !

7-of-9:
Hi @nivida , fyi -- still can't get this to work intuitively (or at all) in beta 38 or 41 (Chrome, Windows 10).

Network right click disable... 
  provider.on("error") fires (expected), subscription callbacks (.on('data')) stop firing (expected).

Network, right click, enable...
  subscription(s) on('data') {} don't fire; also web3.eth.subscribe(.... () => {}) callbacks also don't fire.

I can see another websocket is setup and is streaming -- issue is that the subscription callbacks aren't being called (expected sections marked below /*###*/).  Is there any "reconnect" or similar event that needs to be trapped and/or subscriptions recreated?

`
web3s[x] = new Web3(new Web3.providers.WebsocketProvider(web3_pendingTx_ws_config[x]))
                const web3 = web3s[x]
                const provider = web3.currentProvider

                // set disconnect/error handlers
                provider.on("error", err => { 
                    console.warn(`## web3 ${x} socket error, err=`, err)
                    diag_networkConnected[x] = false
                    NetworkStatusChanged(x)
                })
                provider.on("end", err => {
                    console.warn(`## web3 ${x} socket end, err=`, err)
                    diag_networkConnected[x] = false
                    NetworkStatusChanged(x)
                })
                
                provider.on("connect", data => {
                    console.log(`DIAG WS ${x} - web3 - socket connect, data=`, data)
                    diag_networkConnected[x] = true //web3.eth.net.isListening() // always true, even when no connection

                        var sub1 = web3.eth.subscribe('pendingTransactions', { /*options*/ }, (err, txid) => { /*###*/ })
                        sub1.on('data', (txid) => {
                           /*###*/

                            console.log(`DIAG WS ${x} - web3 - pendingTransactions`, txid) 
                            NetworkStatusChanged(x, txid)
                        })

                        var sub2 = web3.eth.subscribe('newBlockHeaders', { /*POTIONS!*/ }, (err, evt) => {  /*###*/ })
                        sub2.on('data', (evt) => {
                            console.log(`DIAG WS ${x} - web3 - newBlockHeaders`, evt) 
                             /*###*/
                            }
                        })
                    }
`

gabmontes:
@nivida what do you guys think about revisiting this PR but targeting branch `1.x`?

It was based on `1.0.0-beta.37` so I think it is still relevant to that branch. Of course is not compatible with `2.x` due to the re-architecture.

gabmontes:
@alcuadrado please consider adding this PR to #3070 list.

cgewecke:
@gabmontes This topic *is* currently being tracked there and is a high priority fwiw. I make sure this PR is ref'd though. Thanks for pinging.

PR #2007, opened 2018-10-19
fixed decodeLog

touuki:
Fixed the bug that the decodeLog function don't work when all parameters of the event are indexed.


nivida:
Thanks for the submission of this PR but could you do this improvement in the ethereumProvider branch because the architecture of web3.js has changed a lot there :) @TangHHH

PR #2016, opened 2018-10-23
Add new example to web3.eth.accounts.privateKeyToAccount

adityanalge:
The web3.eth.accounts.privateKeyToAccount function displayed duplicate examples. Replace second example with new private address and public key pair.

### DIFF
--- a/docs/web3-eth-accounts.rst
+++ b/docs/web3-eth-accounts.rst
@@ -127,10 +127,10 @@ Example
         encrypt: function(password){...}
     }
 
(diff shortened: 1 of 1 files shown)


nivida:
I think it would be the better approach to remove the second example instead of just changing the properties but thanks for submitting this PR. It would be really cool if you could do this for me here in this PR :) @adityanalge

adityanalge:
@nividia I created a new pull request #2099 and removed the duplicate function as you asked

joshstevens19:
👍

PR #2021, opened 2018-10-26
Handle Revert code and throw containing message

termosa:
Hi @frozeman 
I'm not an expert at all about web3, but I find out that it doesn't handle require/revert messages (https://solidity.readthedocs.io/en/v0.4.24/control-structures.html?highlight=revert#error-handling-assert-require-revert-and-exceptions).

I don't know what is a good way to do it, and I don't even know how to parse Ethereum response correctly. But I had to implement this feature for my product somehow and I've decided to share this code with you so then you can decide how to do it better.
Or maybe I can finish this if you give me some tips on where to look for the decoder and how to better organize the code.

### DIFF
--- a/packages/web3-eth-contract/src/index.js
+++ b/packages/web3-eth-contract/src/index.js
@@ -41,6 +41,29 @@ var errors = require('web3-core-helpers').errors;
 var promiEvent = require('web3-core-promievent');
 var abi = require('web3-eth-abi');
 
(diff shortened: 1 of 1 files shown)


nivida:
Thanks for the submission of this PR but could you do this improvement in the ethereumProvider branch because the architecture of web3.js has changed a lot there :) @termosa

nivida:
I've closed this PR because the creator of it didn't answer the open question. Feel free to submit a new PR.

PR #2031, opened 2018-11-01
replaced indexOf with startsWith where required to improve performace

clintonyeb:
The source code makes a lot of checks to make sure a string starts with *0x". If this is true, the string is treated as a hash value or a number otherwise.

However using *indexOf* to test for this is inefficient as compared to its equivalent *startsWith*.

Here is a small benchmark test:

 Here is the GitHub Gist: [web3_indexof_starts_with.js](https://gist.github.com/clintonyeb/b9b84ac66447a7d9c9b3ff3655db4695)

To verify the results, run:
 
```bash
$ node web3_indexof_starts_with.js

Testing startsWith:::
Benchmark took 1834165 nanoseconds
Completed startsWith.

Testing indexOf:::
Benchmark took 2679461 nanoseconds
Completed indexOf.
```

Thanks.

### DIFF
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -277,9 +277,9 @@ var toHex = function (val) {
 
     // if its a negative number, pass it through fromDecimal
     if (isString(val)) {
-        if (val.indexOf('-0x') === 0)
+        if (val.startsWith('-0x'))
             return fromDecimal(val);
-        else if(val.indexOf('0x') === 0)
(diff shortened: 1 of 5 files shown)


nivida:
It would be cool if you could submit also a PR for the 1.0 version. If you will do this then please to it in the ethereumProvider branch because the architecture of web3.js has changed a lot there :)

nivida:

PR #2051, opened 2018-11-16
Update Personal (method) properties syntax

themondays:
**What**
Update properties method assignment for future extensions

**Where**
Personal method

**Test**
Local tests has been passed


nivida:

PR #2056, opened 2018-11-18
Declaring types within the repos packages itself

joshstevens19:
Hi, i work with TypeScipt all the time and do use web3 constantly, i always find the types are well out of date and it just never keeps in sync. As per TypeScripts suggestion they state if possible putting your `d.ts` files in the library itself gives a better user ability for the developers, as it will instantly expose the types. These are still all versioned with updates etc so you are getting a better developer experience. 

`BigNumber.js` do this as well as most big solutions which i think we should follow - https://github.com/MikeMcl/bignumber.js/blob/master/bignumber.d.ts

What i have done is created the first `utils.d.ts` types for us in the `web3-utils` package - i am planning to go through all the packages to expose these types so it's all in one repo if you approve of this PR. TypeScript only ever suggest using the `@types` if you are not in-control of the repo - which we are. 

This solution will mean updating methods in the future and updating it's typing will be super easy for all the maintainers as it's in 1 repo. We can also add this to the contribution guidelines and pull request checklists meaning it will never be missed. I know a lot of people may just use `js` but as a regular `TS` developer using these library i think this would be a very good way to start improving this. Like i said i am happy to go through all of them and do a PR for each `package`. 

The plan then would be once all types are in then we eventually remove type `@types` package and everything installed with 1 command, or depreciate it on a certain version. 

It won’t make any difference to the way js developers work, it will still be 

`import * as Utils from 'web3-utils';`

but when doing that in a `.ts` file it will now expose all the typings allowing people to know the parameters types, return types etc! 

I have done this within the active `ethereumProvider` branch, but if you think it’s better suited for another branch let me know! 👍 

I see @nivida as the most active community member recently - take a look and let me know :) 

Thanks
Josh


joshstevens19:
can someone look at this please @nivida ... not sure best person to tag

nivida:
Thank you for submitting this PR! I will have a close look on it later but I will definitely merge it into the ethereumProvider branch. But yes as you said it would be good to have typings for each package. Could you please add the missing last line to the type declaration of utils? and it would be also good to add this files to the ```files``` property in the package.json of the utils package otherwise it want be published on npm :)

joshstevens19:
Hey Nivida, thanks for the response i am glad you agree. Like i said once this one is merged in i will do a PR for every package :). This will be a huge step forward. So firstly there is no missing last line i forgot to remove the white space which i have done. Also by default npm brings all the files over unless they are in the `.npmignore` so i don't think we need to create `file` property in our package.json as everything in the package needs bringing through. 

Lastly i just seen a really nice way to export these types, as we do not have classes we can do what `fs` do and `fs-extra` which is a lovely way, so quickly changing that now. This will then allow them to only import what they need i.e. `import { isBigNumber } from 'web3-utils'`. Will do my updates now :)

joshstevens19:
Hey just pushed my latest changes have a look and if happy merged in it and il get the other PRs to you when i get a chance. 1 PR per package 👍 cheers

nivida:
@joshstevens19 Great thanks! I thought it would be the better approach to define the files property in the package.json instead of ignoring all. It's much easier to just say what I would like to have in the NPM package instead of ignore anything I don't want to have in my NPM package. I've added the files property in all package.json files. 

I will resolve the conflicts and add your typings to the files property in the package.json. 

As you can see here: https://github.com/ethereum/web3.js/blob/ethereumProvider/packages/web3-utils/src/index.js#L33 I'm exporting the methods from the Utils file and all others too. It should be already possible to have imports like: ```import { isBigNumber } from 'web3-utils'```

packages/web3-utils/index.d.ts
```diff
@@ -0,0 +1,137 @@
+import BigNumber from "bn.js";
+import us from "underscore";
```
nivida: I moved from the underscore package to the underscore-es package otherwise I can't import just single methods :)

packages/web3-utils/README.md
```diff
@@ -39,6 +39,10 @@ console.log(Utils);
 }
 ```
 
+## Types 
+
+If you are using TypeScript all the types are defined in the `index.d.ts` - please use them :)
```
nivida: I would remove the smile ;-)
``` If you are using TypeScript all the types are defined in the `index.d.ts` file ```

nivida (changes requested):

packages/web3-utils/README.md
joshstevens19: No problem, will do that once on a computer

joshstevens19:
Ok cool, I will resolve your comments when on a computer. In terms of what you said about only exporting what you need, yes I know we could already do it but in the TS world we would do that and we have no typing, so my point was whooop TypeScript devs can now import 1 function and have typing for it 👍

In terms of files in the NPM package - ok makes sense! 

Thanks il sort last comments a little later 👍

nivida:
@joshstevens19 oh then I got you wrong I thought you were talking about the index.js. Yes, I agree with you it is better to export the types like this. Thanks for being so active!

packages/web3-utils/index.d.ts
joshstevens19: if that's the case we can probably remove this completely from the types now.

joshstevens19:
Pushed those changes up for you 👍 let me know if that's all good.. once this is merged in il go onto the next package and so on until all our complete. I may branch off after this merge so i can do many PRs at the same time (as we wont have any merge conflicts as they are different packages)

Issue #2062, opened 2018-11-22
Unhandled Promise Rejection on 'sendRequest'


gabmontes:
When `.send()`'ing a contract method, web3 creates the handling logic, returns a `promiEvent` and finally executes the call at:

https://github.com/ethereum/web3.js/blob/0d54d94e637832abf97834e4fc444fbff87f874b/packages/web3-core-method/src/index.js#L581

Inside `sendRequest`, there is a call to `signTransaction`:

https://github.com/ethereum/web3.js/blob/0d54d94e637832abf97834e4fc444fbff87f874b/packages/web3-core-method/src/index.js#L536

If for any reason the signing fails, i.e. the `net` API is not enabled in the underlying node, `signTransaction` returns a failed promise that will be lost as the original `sendRequest` call does handle it (does not act on `defer` or `defer.eventEmitter`).

The result is an `Unhandled Promise Rejection` error and the web3 user code client hangs given the `promiEvent` does never emit even an `error` event.

@nivida I'm happy to contribute with a fix/PR. Just wondering if you or @frozeman have any thoughts on the better way to handle this.

Issue #1220 is possibly related or similar to this.

nivida:
This will be fixed with the PR #2000

nivida:
This is fixed since beta.38.

PR #2000, opened 2018-10-15
EthereumProvider and the related refactoring

nivida:
- [x] Port Web3 to es6.
- [x] Refactor the architecture to more SOLID & clean code.
- [x] Improve provider handling (Web3 should use internally always the same interface).
- [x] Each RPC method should be implemented once. 
- [x] Method calls should be delegated with the JS Proxy object instead of adding them on construction.
- [x] Create AbstractWeb3Module class and remove ```core.packageInit()```.
- [x] Simplify the internal subscription handling.
- [x] Improve the documentation.
- [x] Fix the log subscription.
- [x] Write unit tests with Jest and move them to the related module folders.
- [x] defaultBlock, defaultAccount, timeoutBlock and confirmationBlock should be configurable for each module over the constructor.
- [x] Refactor ENS module (dependency inversion and not injection, EIP 1577)
- [x] Set up rollup for each module
- [x] Set up eslint & prettier
- [x] Update travis CI 
- [x] Improve web3-utils & web3-core-helpers exports

PR #2063, opened 2018-11-23
Bubble up tx signing errors

gabmontes:
This is a proposal to solve the `Unhandled Promise Rejection` reported in #2062.

If for any reason the call to `signTransaction` fails, the `promiEvent` created during the contract method call will emit an `error` message -if working as an event emitter- and will also be rejected.


nivida:
Hay @gabmontes could you do your changes again the ethereumProvider branch? But I think this issue should already be solved there :)

nivida:
This issue got solved in the PR #2000

PR #2073, opened 2018-11-28
Web3-bzz types

joshstevens19:
As already discussed, we are bringing typing into the web3 packages itself. 

Here is the web3-bzz types :)

### DIFF
--- a/packages/web3-utils/index.d.ts
+++ b/packages/web3-utils/index.d.ts
@@ -1,3 +1,16 @@
+/*
+    This file is part of web3.js.
+    web3.js is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Lesser General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    web3.js is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License for more details.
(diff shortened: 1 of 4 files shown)


packages/web3-utils/index.d.ts
```diff
@@ -1,3 +1,9 @@
+/**
+ * @file index.d.ts
+ * @author Josh Stevens <joshstevens19@hotmail.co.uk>
+ * @date 2018
+ */
```
nivida, line 18: Thanks for writing these typings!

Could you please add the file doc as we have it in all other files: 
```

/*
    This file is part of web3.js.
    web3.js is free software: you can redistribute it and/or modify
    it under the terms of the GNU Lesser General Public License as published by
    the Free Software Foundation, either version 3 of the License, or
    (at your option) any later version.
    web3.js is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU Lesser General Public License for more details.
    You should have received a copy of the GNU Lesser General Public License
*/
/**
 * @file index.d.ts
 * @author Josh Stevens <joshstevens19@hotmail.co.uk>
 * @date 2018
 */
```

I think I forgott to mention this in the other PR sorry.

nivida (changes requested):

packages/web3-utils/index.d.ts
joshstevens19: i have done that now 👍 will make sure i carry it on with the next typing's.

joshstevens19:
All ready to be merged :)

PR #2082, opened 2018-11-30
Web3 core promievent types

joshstevens19:
Types for `web3-core-promievent` 👍

### DIFF
--- a/packages/web3-core-promievent/index.d.ts
+++ b/packages/web3-core-promievent/index.d.ts
@@ -1,0 +1,23 @@
+/*
+    This file is part of web3.js.
+    web3.js is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Lesser General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    web3.js is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License for more details.
+    You should have received a copy of the GNU Lesser General Public License
+    along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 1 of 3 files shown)


joshstevens19:
need to write some tests for this which will come in another PR. I am now going back to web3-utils and doing full tests then sorting web3-bzz with tests then onto this 1 to do test. Can be merged in will come back and sort tests in the next few days.

packages/web3-core-promievent/index.d.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * @file index.d.ts
+ * @author Josh Stevens <joshstevens19@hotmail.co.uk>
+ * @date 2018
+ */
+
+export declare class PromiEvent {
+    proxyHandler(target: PromiEvent, name: string): Function;
+ }
+
```
nivida, line 23: This type is not correct because I'm using the JS ES6 proxy to combine all methods of the EventEmitter and Promise. The proxy handler itself can't be called because I'm returning the proxy object in the constructor and the ES6 proxy delegates the method calls.

nivida (changes requested):

packages/web3-core-promievent/index.d.ts
joshstevens19: ahh ok so we probably don't need types for this then?

nivida: We definitely need types for this :)

ES6 Proxy: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Proxy

joshstevens19:
You can close this as this is not part of the public API anyway so we do not need types for it 👍 going from https://web3js.readthedocs.io/en/1.0/index.html 😄

nivida:
This is not true because the public API returns PromiEvent objects and I think it would be cool when they are typed. That means it is indirectly a part of the public API

joshstevens19:
Okay 😁

PR #2088, opened 2018-12-01
[Documentation] Fix return values of generateSymKeyFromPassword function

arku:
Fixes https://github.com/ethereum/web3.js/issues/1924

### DIFF
--- a/docs/web3-shh.rst
+++ b/docs/web3-shh.rst
@@ -587,7 +587,7 @@ Returns
 -------
 
 
-``String`` - Key ID on success and an error on failure.
+``Promise<String|Error>``: The Key ID on success and an error on failure.
 
 
 -------


docs/web3-shh.rst
```diff
@@ -587,7 +587,8 @@ Returns
 -------
 
 
-``String`` - Key ID on success and an error on failure.
+``Promise`` returns ``String`` : The Key ID, if no callback is provided
+``undefined`` - If a callback is provided
```
nivida: First of all thanks for submitting this PR. I'm currently in contact with Gitcoin because of the bounties and hopefully, they are online next week.
I would write: 
```
Promise<String|Error>: The Key ID on success and an error on failure.
```

Currently this method will return always undefined. In the next release it will always return a Promise and if the callback is defined then it will additionally call the callback method. If the method throws an error and a callback is given then it will not throw the error for the Promise because otherwise, you would get the ```Unhandled Promise rejection error```. I've fixed this in the PR #2000 to be align with the documentation and to have consistency over the complete library. 

Maybe you are a super great person (😄) and you would change this also here: 
https://web3js.readthedocs.io/en/1.0/web3-eth.html#
https://web3js.readthedocs.io/en/1.0/web3-eth-personal.html

I think it would be done really fast.

Sorry for the confusion.

nivida (changes requested):

docs/web3-shh.rst
arku: @nivida Done!

Btw, I am not sure what to change in `web3-eth.rst` and `web3-eth-personal.rst` 

Can you please guide me with it?

Thanks.

nivida: We have there also promises defined like this:
```
String - Description of this value
```
instead of
```
Promise<String|Error> - Description of this value
```
:-)

nivida (approved):

PR #2090, opened 2018-12-02
Web3 eth iban types

joshstevens19:
Here are the `web3-eth-iban` types + tests.

All the tests are done and because it was all for 1 class it has stayed in 1 file as it made me sense. 

Also looking at this package i think we should delete `leftPad` method in `Iban.js` and use the `Utils` `leftPad` method 😄 

👍

### DIFF
--- a/packages/web3-eth-iban/package.json
+++ b/packages/web3-eth-iban/package.json
@@ -20,7 +20,8 @@
         "web3-utils": "1.0.0-beta.36"
     },
     "files": [
-        "dist"
+        "dist",
+        "types/index.d.ts"
     ],
     "devDependencies": {
         "dtslint": "^0.3.0"
--- a/packages/web3-eth-iban/package.json
+++ b/packages/web3-eth-iban/package.json
@@ -11,13 +11,18 @@
     "scripts": {
         "build": "rollup -c",
         "dev": "rollup -c -w",
-        "test": "jest"
+        "test": "npm run-script dtslint && jest",
+        "dtslint": "dtslint types"
     },
+    "types": "types",
     "dependencies": {
         "bn.js": "4.11.8",
         "web3-utils": "1.0.0-beta.36"
     },
     "files": [
         "dist"
-    ]
+    ],
+    "devDependencies": {
+        "dtslint": "^0.3.0"
+    }
 }
--- a/packages/web3-eth-iban/types/tests/iban-tests.ts
+++ b/packages/web3-eth-iban/types/tests/iban-tests.ts
@@ -17,7 +17,7 @@
  * @date 2018
  */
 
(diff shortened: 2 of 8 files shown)


packages/web3-eth-iban/package.json
```diff
@@ -11,13 +11,19 @@
...
         "web3-utils": "1.0.0-beta.36"
     },
     "files": [
-        "dist"
-    ]
+        "dist",
+        "types/index.d.ts"
+    ],
+    "devDependencies": {
+        "dtslint": "^0.3.0"
```
nivida, line 27: This dependency should only be on the root folder. (web3.js/package.json)

packages/web3-eth-iban/types/tests/iban-tests.ts
```diff
@@ -0,0 +1,139 @@
...
+    (at your option) any later version.
+    web3.js is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License for more details.
+    You should have received a copy of the GNU Lesser General Public License
+    along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/**
+ * @file iban-tests.ts
```
nivida, line 15: I'm using pascal case in Web3.js. Check this also on the Utils PR.

nivida (changes requested):

packages/web3-eth-iban/package.json
joshstevens19: dtslint will not run if its in the parent package.. it seems dtslint can only run if it's in the node_modules of the current package. 

You get `'dtslint' is not recognized as an internal or external command,` if you try.

packages/web3-eth-iban/types/tests/iban-tests.ts
joshstevens19: issue with this is that setting `tslint` to `"pascal-case"` `dslint` then wants `index.d.ts` to be `pascal-case` as well which does not work, as windows is not case sensitive but Linux is and this will cause huge issues on different systems (i.e. Index.D.ts on windows will work on Linux will not). This is why for testings and types i have used `kebab-case` + is what `ts` promote for typing file naming, if you look at DT you see basically every single type package is `kebab-case` 👍 let me know your thoughts :)

joshstevens19:
Think this now can be merged in :)

nivida:
I'll test it tomorrow and when all tests run as expected I'll merge it. 
Thanks 👍

nivida (approved):

PR #2094, opened 2018-12-03
Fix sha3 to take BN type

### LINKED ISSUE #2093: web3.utils.sha3 doesn't take a BN type argument.
There is a example which takes an BN type argument in https://web3js.readthedocs.io/en/1.0/web3-utils.html#sha3.

However, it throws an error when take an argument with BN type. like below:
`TypeError: str.slice is not a function`

I think example is wrong, we should fix it.

nujabes403:
Tryting to fix 
#2093 issue.

Currently, sha3 can't take an BN type argument. However, there is an example in https://web3js.readthedocs.io/en/1.0/web3-utils.html#sha3 which takes an BN instance as an argument.

@joshstevens19 Thank you :)

### DIFF
--- a/packages/web3-utils/src/utils.js
+++ b/packages/web3-utils/src/utils.js
@@ -425,12 +425,12 @@ var isTopic = function (topic) {
 var SHA3_NULL_S = '0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470';
 
 var sha3 = function (value) {
-    if (isHexStrict(value) && /^0x/i.test((value).toString())) {
-        value = hexToBytes(value);
+    if (isBN(value)) {
+        value = value.toString();
     }
 
-    if (isBN(value)) {
-        value = value.toString(10);
+    if (isHexStrict(value) && /^0x/i.test((value).toString())) {
+        value = hexToBytes(value);
     }
 
     var returnValue = Hash.keccak256(value); // jshint ignore:line
--- a/packages/web3-utils/src/utils.js
+++ b/packages/web3-utils/src/utils.js
@@ -429,6 +429,10 @@ var sha3 = function (value) {
         value = hexToBytes(value);
     }
 
+    if (isBN(value)) {
+        value = value.toString(10);
+    }
+
     var returnValue = Hash.keccak256(value); // jshint ignore:line
 
     if(returnValue === SHA3_NULL_S) {


packages/web3-utils/src/utils.js
```diff
@@ -429,6 +429,10 @@ var sha3 = function (value) {
         value = hexToBytes(value);
     }
 
+    if (isBN(value)) {
+        value = value.toString(10);
```
joshstevens19: do 

```
if (isBN(value)) {
   value = value.toString()
}
```
as `BN` has a `toString` method so no need for the `10` it should always return what you want

joshstevens19: Also i would probably do that at the start so if it needs to go into the first if it can 👍

nujabes403: @joshstevens19 Since the facts that passed isBN means the value is an instance of BN. So value.toString(10) is calling of BN method.

nujabes403: @joshstevens19 Since the facts that passed isBN means the value is an instance of BN. So value.toString(10) is calling of BN method.

joshstevens19:
Yep but you do not need the (10) part 😄

nujabes403:
@joshstevens19 Yep, you're right. Since there is no argument for it, it automatically set `base` value to `10` 

```
  BN.prototype.toString = function toString (base, padding) {
    base = base || 10;
    padding = padding | 0 || 1;
...
```
ref) https://github.com/indutny/bn.js/blob/master/lib/bn.js#L447~L448

However, for clarity, how about representing `(10)`? 

Thx :)

joshstevens19:
@nujabes403 think for now if you can remove the `10` to keep it consistent then i can approve it 👍 thanks for PR

nujabes403:
@joshstevens19 
Thank you for comment! :) 
Then do you think should I change it also for consistency?
There is one more function to expose `(10)` in same file.
If you agree, I'll change it also.

```
/**
 * Converts value to it's decimal representation in string
 *
 * @method hexToNumberString
 * @param {String|Number|BN} value
 * @return {String}
 */
var hexToNumberString = function (value) {
    if (!value) return value;

    return toBN(value).toString(10);
};
```

joshstevens19:
@nujabes403 Looking at the code we use `toString(10)` in most instances of converting a `BN` to `string` (pointless i know) but probably needs another task to clean that up (its doing no harm) so yeah leave it how it is `toString(10)`. Unless you want to clean all the `toString(10)` on `BN` to `toString()`, but don't expect you to do that 👍 - Sorry for the confusion.

packages/web3-utils/src/utils.js
joshstevens19: Read comments on thread if you want to know the conclusion 👍

joshstevens19:
@nujabes403 Can you move the check so it's the first thing it does please.. i.e. 

```
var sha3 = function (value) {
    if (isBN(value)) {
        value = value.toString(10);
    }

    if (isHexStrict(value) && /^0x/i.test((value).toString())) {
        value = hexToBytes(value);
    }
    var returnValue = Hash.keccak256(value); // jshint ignore:line

    if(returnValue === SHA3_NULL_S) {
        return null;
    } else {
        return returnValue;
    }
};
```

so if it needs to go into the next `if` it can

nujabes403:
@joshstevens19 Please take a look again, I changed condition checking order. :)

joshstevens19 (approved):

joshstevens19:
Thanks @nujabes403 that looks great! We need 2 approvals to merge so @nivida can you take a look and merge it 👍 

Cheers again

nivida (approved):

PR #2095, opened 2018-12-04
Add method signature to invalid number of solidity args error

angus-hamill:
It can be tough as a developer to determine the source of this error when executing a number of transactions (eg. a truffle test)

It may also be worth including the arguments themselves?


joshstevens19:
Yeah can't see how this can hurt, if your going to do it for `InvalidNumberOfSolidityArgs` you may as well do the same for `InvalidNumberOfRPCParams` as that would be pretty useful as well :)

Why don't you do the same for that and i will then relook. 

Thanks for the PR 👍

joshstevens19:
Looks good to me 👍 We need 2 approvals to get this in @nivida can you approve and merge when you get a second 

Thanks again @angus-hamill

joshstevens19 (approved):

nivida (approved):

yvettep321 (commented):
thank you

PR #2102, opened 2018-12-07
Fix comment

jtakalai:
should be 8 to f, not 9 to f, judging from the code immediately below

### DIFF
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -463,7 +463,7 @@ var toChecksumAddress = function (address) {
     var checksumAddress = '0x';
 
     for (var i = 0; i < address.length; i++ ) {
-        // If ith character is 9 to f then make it uppercase
+        // If ith character is 8, 9, ..., f then make it uppercase
         if (parseInt(addressHash[i], 16) > 7) {
           checksumAddress += address[i].toUpperCase();
         } else {


joshstevens19:
Hey thanks for the PR, yes you are correct - in my view though this comment doesn't even need to exist. I think everyone can work out that `parseInt(x, 16)` when x = 8 or higher will return true in: 

```if (parseInt(addressHash[i], 16) > 7)```

I vote to just remove the comment completely - @nivida thoughts?

nikitasius:
> I vote to just remove the comment completely - @nivida thoughts?

Please keep the comments, they help to understand faster the code, especially if you code on another language. But just fix a spell mistake.

joshstevens19:
@nikitasius OK if people think this comment is useful we shall keep it, i did not really think about people coming from a different language may find that comment useful! 

I will approve this as the comment fix is needed 👍 cheers

joshstevens19 (approved):

nikitasius:
> i did not really think about people coming from a different language may find that comment useful

When syntax is different, it helps (no need to go deep into the documentation). Personally i had issue with this function (due Keccak256 SHA3 and not classic SHA3-256). So i was reading code and comments [here](https://github.com/ethereum/web3.js/blob/a80d2a0232f3228c68e1a8e233efaae130306f0d/lib/utils/utils.js#L441).
It was hard to get in what in [this](https://github.com/ethereum/web3.js/blob/a80d2a0232f3228c68e1a8e233efaae130306f0d/lib/utils/sha3.js) file you use Keccak.

Thanks to @jtakalai , becase we started to dig. So the comment:
> the nth letter should be uppercase if the nth digit of casemap is 1

for non-native english people is much more worst ("nth?!") than:

> If ith character is 9 to f then make it uppercase

To be sure what in my case ported logic is correct and thats a hash who was wrong.

*  It would be nice if you will mention what as `sha3` you use `Keccak256` in a newline inside this comment.

P.S. i ported this logic with checksum and formatting ETH addresses on Java in my tool.

nivida:
@nikitasius Thanks for your effort to submit this PR. Could you please do this change against the develop branch?

nivida (changes requested):
Wrong base branch.

nikitasius:
> Could you please do this change against the develop branch?

This PR made here by @jtakalai , guess it will be betetr if he will send this PR in correct branch.

PR #2105, opened 2018-12-08
Allow keepAlive option when instantiating HttpProvider

gnidan:
This PR proposes to make `keepAlive` an option for the HTTP Provider.

As mentioned in trufflesuite/truffle#1504, Truffle is not prepared to handle `keepAlive: true` in all cases.

Please also see @elenadimitrova's paper-trail at trufflesuite/truffle#1461, which identified the issue on Truffle's end.

Thanks to @eggplantzzz for pursuing and implementing the fix.

### DIFF
--- a/packages/web3-providers-http/src/index.js
+++ b/packages/web3-providers-http/src/index.js
@@ -31,16 +31,18 @@ var https = require('https');
 /**
  * HttpProvider should be used to send rpc calls over http
  */
-var HttpProvider = function HttpProvider(host, options = {}) {
-    const keepAlive =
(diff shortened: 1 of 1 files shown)


gnidan:
(The build failure appears to be a problem in `ethereum:1.0`)

frozeman (approved):

nivida (approved):

nivida:
Hay @gnidan, 
First of all thanks for submitting this PR. The build fails because we run the process also on a node runtime of version 5. I already did this in the PR #2000 but this will be merged and released later.

Issue #2065, opened 2018-11-23
new web3-eth-abi decodeLog throw error when the event's input are all indexed


libotony:
When the input of an event are all indexed, go-ethereum's JSON-RPC will response with data '0x', it will throw an error cause by the following code

``` javascript
ABICoder.prototype.decodeParameters = function (outputs, bytes) {
    if (!bytes || bytes === '0x' || bytes === '0X') {
        throw new Error('Returned values aren\'t valid, did it run Out of Gas?');
    }
    //...
}
```

I found the following test code, I assume `Line 82` should be `0x` not empty string

https://github.com/ethereum/web3.js/blob/0d54d94e637832abf97834e4fc444fbff87f874b/test/eth.abi.decodeLog.js#L70-L91

sawyna:
I'd like to work on it.

dileepfrog:
Confirmed same with 1.0.0-beta.36, appears to be a regression from unaffected 1.0.0-beta.35

krzkaczor:
Just in case you're wondering, `ganache` in this case returns: `0x00` which works.

krzkaczor:
btw. i think this one is related: https://github.com/ethereum/web3.js/issues/1916

conscott:
Any simple workaround for this? My contract event filters are now broken.

krzkaczor:
@conscott I didn't manage to prepare a PR fixing it since current develop branch is much different so I used [patch-package](https://www.npmjs.com/package/patch-package) to fix it locally and create git patch applied on every `yarn install`. A dirty solution, but it works 🤷‍♂️ 

Here's patch:
```
patch-package
# FIX FOR: https://github.com/ethereum/web3.js/issues/2065
--- a/node_modules/web3-eth-abi/src/index.js
+++ b/node_modules/web3-eth-abi/src/index.js
@@ -222,7 +222,7 @@ ABICoder.prototype.decodeParameter = function (type, bytes) {
  * @return {Array} array of plain params
  */
 ABICoder.prototype.decodeParameters = function (outputs, bytes) {
-    if (!bytes || bytes === '0x' || bytes === '0X') {
+    if (!bytes) {
         throw new Error('Returned values aren\'t valid, did it run Out of Gas?');
     }
```

conscott:
thanks @krzkaczor !

rmujica:
the behavior still occurs on 1.0.0-beta.37 even though is reported on the changelog to be fixed. how can I be of more help?

[...]

PR #2106, opened 2018-12-08
Fix overly strict function validation requirements for decoding parameters

### LINKED ISSUE #1916: getPastEvents failing to decode empty data
I'm trying out web3 `v1.0.0-beta.36` and having some trouble with pulling events from a contract.

The events do not return any data, and `getPastEvents` is throwing this error:
```
Error: Returned values aren't valid, did it run Out of Gas?
```

The same action works in web3 `0.20.7` no problem.

Is it possible a bug in decoding of empty event data `0x` may have slipped through the cracks?

gnidan:
This should fix #1916 and #2065. 
Arises from trufflesuite/truffle#1487. Also being tracked on Truffle's end at trufflesuite/truffle#1504.

There is a bit of history here, in that empty bytes arrays have been variously represented as:
- `"0x"` (correct value, others malformed)
- `"0X"`
- `"0x0"`
- `""`
- who knows what else

Anyway, Ganache was passing `"0x0"` for awhile, and now that's stopped, exposing this conditional that asserts `bytes` must contain data. In the case where `outputs` equals `[]`, an empty value for `bytes` is really all that makes sense.

This PR thus loosens the conditional a bit, to throw the exception only if `bytes` is empty and `outputs` is non-empty.

### DIFF
--- a/packages/web3-eth-abi/src/index.js
+++ b/packages/web3-eth-abi/src/index.js
@@ -222,7 +222,7 @@ ABICoder.prototype.decodeParameter = function (type, bytes) {
  * @return {Array} array of plain params
  */
 ABICoder.prototype.decodeParameters = function (outputs, bytes) {
-    if (!bytes || bytes === '0x' || bytes === '0X') {
+    if (outputs.length > 0 && (!bytes || bytes === '0x' || bytes === '0X')) {
         throw new Error('Returned values aren\'t valid, did it run Out of Gas?');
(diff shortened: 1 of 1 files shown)


frozeman (approved):

nivida (approved):

nivida:
I reference here the PR #2000 just to be sure this fix will not be overwritten on merging.

PR #2110, opened 2018-12-08
compacting `var isAddress = function()` code

nikitasius:
Merged 2 regex into 1 + moved 1st condition into 3rd (which became 2nd).

I.e. :
* we check if address matching in lo/up case only and return `true` when it's true :)
* if not - we return if address matching basic requirements AND if checksum if correct.

If inside the 2-nd condition address isn't matching basic requirements, checksum function will not be executed and `false` will be returned.

Here is large regex:
image

### DIFF
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -413,15 +413,12 @@ var isStrictAddress = function (address) {
  * @return {Boolean}
 */
 var isAddress = function (address) {
-    if (!/^(0x)?[0-9a-f]{40}$/i.test(address)) {
-        // check if it has the basic requirements of an address
-        return false;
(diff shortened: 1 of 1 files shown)


nikitasius:
## Clarifications

Before:
* compared address with case insensitive regex if not matched you return false [what address have correct format `in case of mixed letter cases`]
* else then compared with low OR up case if matched you return true [when all low or up case AND matching format `in case of low/up case`]
* else then checked the checksum and returned as result [correct or not checksum]

Now:
* you compare low/up case and return true if match [all low or up case and matching the format]
* else you return comparation case-insentive regex && checksum checking [matching the format for mixed up/low case and if matched we check the checksum)

I know what short!=good in some cases and perfomance we win will be small (cause there are 1 `if` less and 1 text operation less due merged 2in1 regex).

But new code doesn't add a mess, just a pretify & remove/merge some operations to get the same result in less steps.

nivida:

PR #2114, opened 2018-12-09
Web3 core types

joshstevens19:
Created web3-core types 

@nivida if you can confirm what the types are for any of the `any` types that would be great (if they can be strongly typed). 

This was the main one i would like to sort.. 

```
// not sure what net type is meant to be
setProvider(provider: provider | string, net: any): boolean;
```

Also due to `web3-core` using `web3-core-methods` even though `web3-core` is not dependant on `web3-core-methods`, makes it super hard to write any types for those. I have left them as `any` for now as we can't really do much until we sort the dependency issue out. 

cheers 👍


packages/web3-core/types/index.d.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+import {AbstractProviderAdapter, ProvidersModuleFactory, provider} from 'web3-providers';
+
+export class AbstractWeb3Module {
+    constructor(
+        provider: AbstractProviderAdapter | provider | string,
+        providersModuleFactory: ProvidersModuleFactory,
+        // not sure what the below object structure is
+        providers: {},
```
nivida: it is:
```js
{
    HttpProvider: HttpProvider,
    WebsocketProvider: WebsocketProvider,
    IpcProvider: IpcProvider
}
```

as you can see here: https://github.com/ethereum/web3.js/blob/ethereumProvider/packages/web3-providers/src/index.js#L35-L39

packages/web3-core/types/index.d.ts
```diff
@@ -0,0 +1,62 @@
...
+        // dont have type yet
+        // as this is in web3-core-method
+        // can be sorted later once dependencies
+        // are cleaned up
+        methodController: any,
```
nivida: The methodController dependency no longer exists because of the latest improvements.

The new constructor of the AbstractWeb3Module is: https://github.com/ethereum/web3.js/blob/ethereumProvider/packages/web3-core/src/AbstractWeb3Module.js#L26-L43

packages/web3-core/types/index.d.ts
```diff
@@ -0,0 +1,62 @@
...
+        // dont have type yet
+        // as this is in web3-core-method
+        // can be sorted later once dependencies
+        // are cleaned up
+        AbstractMethodModelFactory?: any,
```
nivida: The AbstractMethodFactory is defined here: https://github.com/ethereum/web3.js/blob/ethereumProvider/packages/web3-core-method/lib/factories/AbstractMethodFactory.js

packages/web3-core/types/index.d.ts
```diff
@@ -0,0 +1,62 @@
...
+    readonly defaultGasPrice: string;
+    readonly defaultGas: number;
+    readonly transactionPollingTimeout: number;
+    readonly transactionConfirmationBlocks: number;
+    readonly transactionBlockTimeout: number;
+    readonly defaultBlock: string | number;
+    readonly defaultAccount: string | null;
+    readonly currentProvider: AbstractProviderAdapter | provider;
+    // if we can get a strongly typed object for `net` that would be great
+    setProvider(provider: AbstractProviderAdapter | provider | string, net: any): boolean;
```
nivida: the net argument is the node net package: https://nodejs.org/api/net.html

packages/web3-core/types/tests/abstract-web3-module-test.ts
```diff
@@ -0,0 +1,86 @@
...
+// $ExpectType number
+abstractWeb3Module.transactionBlockTimeout;
+
+// $ExpectType string | number
+abstractWeb3Module.defaultBlock;
+
+// $ExpectType string | null
+abstractWeb3Module.defaultAccount;
+
+// $ExpectType AbstractProviderAdapter | HttpProvider | IpcProvider | WebsocketProvider
```
nivida: Will just be: AbstractProviderAdapter

packages/web3-core/types/index.d.ts
```diff
@@ -0,0 +1,62 @@
...
+        options?: Web3ModuleOptions
+    );
+    readonly defaultGasPrice: string;
+    readonly defaultGas: number;
+    readonly transactionPollingTimeout: number;
+    readonly transactionConfirmationBlocks: number;
+    readonly transactionBlockTimeout: number;
+    readonly defaultBlock: string | number;
+    readonly defaultAccount: string | null;
+    readonly currentProvider: AbstractProviderAdapter | provider;
```
nivida: Will just be: AbstractProviderAdapter

packages/web3-core/types/tests/abstract-web3-module-test.ts
```diff
@@ -0,0 +1,86 @@
...
+abstractWeb3Module.currentProvider;
+
+// $ExpectType boolean
+abstractWeb3Module.setProvider(httpProvider, {});
```
nivida: The second parameter will only be used with the IpcProvider because this provider requires the Net package of node :)

nivida (changes requested):

packages/web3-core/types/index.d.ts
joshstevens19: Yes i know i was planning to leave this as any for now as `AbstractMethodModelFactory` will mean creating 3 interfaces in the wrong place. Hence my comment above.. if you don't agree i will create them 👍

joshstevens19:
@nivida addressed all comments please look again

packages/web3-core/types/index.d.ts
nivida: Yes, but the AbstractMethodModelFactory does no longer exist because it's called AbstractMethodModel now. :-)

joshstevens19: cool addressed it now

joshstevens19:
@nivida can you merge now?

PR #2119, opened 2018-12-11
Web3 eth contract types

joshstevens19:
web3-eth-contract types.

### DIFF
--- a/packages/web3-eth-contract/package.json
+++ b/packages/web3-eth-contract/package.json
@@ -8,10 +8,12 @@
     "main": "dist/web3-eth-contract.cjs.js",
     "module": "dist/web3-eth-contract.esm.js",
     "browser": "dist/web3-eth-contract.umd.js",
+    "types": "types",
     "scripts": {
         "build": "rollup -c",
         "dev": "rollup -c -w",
-        "test": "jest"
+        "test": "dtslint types && jest",
+        "dtslint": "dtslint types"
     },
     "dependencies": {
         "underscore-es": "^1.9.8",
@@ -26,7 +28,11 @@
         "web3-providers": "1.0.0-beta.36",
         "web3-utils": "1.0.0-beta.36"
     },
+    "devDependencies": {
+        "dtslint": "^0.3.0"
+    },
     "files": [
-        "dist"
+        "dist",
+        "types/index.d.ts"
     ]
 }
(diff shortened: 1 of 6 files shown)


packages/web3-eth-contract/package.json
```diff
@@ -8,10 +8,12 @@
     "main": "dist/web3-eth-contract.cjs.js",
     "module": "dist/web3-eth-contract.esm.js",
     "browser": "dist/web3-eth-contract.umd.js",
+    "types": "types",
     "scripts": {
         "build": "rollup -c",
         "dev": "rollup -c -w",
-        "test": "jest"
+        "test": "dtslint types && jest",
```
nivida: Could you do the same change here as described in the accounts PR?

packages/web3-eth-contract/types/index.d.ts
```diff
@@ -0,0 +1,93 @@
...
+    constructor(
+        provider: AbstractProviderAdapter | provider,
+        jsonInterface: {},
+        address?: string,
+        options?: ContractOptions
+    )
+    options: Options;
+    clone(): Contract;
+    deploy(options: DeployOptions): DeployTransactionResponse;
+    methods: any;
```
nivida, line 32: The correct definition would be: ```methods: Function[]```

packages/web3-eth-contract/types/index.d.ts
```diff
@@ -0,0 +1,93 @@
...
+    once(event: string, callback: (error: Error, event: EventData) => void): void;
+    once(event: string, options: EventOptions, callback: (error: Error, event: EventData) => void): void;
+    events: any;
```
nivida, line 35: The correct definition would be: ```events: Function[]```

nivida (changes requested):

packages/web3-eth-contract/types/index.d.ts
joshstevens19: As discussed offline has to stay `any` due to not knowing what the method is being executed i.e. methods.josh() in theory `josh` in compile time will be `any` not `Function[]` so will not work 👍

packages/web3-eth-contract/types/index.d.ts
joshstevens19: As discussed offline has to stay `any` due to not knowing what the method is being executed i.e. events.josh() in theory `josh` in compile time will be `any` not `Function[]` so will not work 👍

joshstevens19:
fixed the `test` script @nivida 👍

PR #2121, opened 2018-12-11
Web3 net types

joshstevens19:
web3-net types

### DIFF
--- a/packages/web3-net/types/index.d.ts
+++ b/packages/web3-net/types/index.d.ts
@@ -1,0 +1,31 @@
+/*
+    This file is part of web3.js.
+    web3.js is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Lesser General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    web3.js is distributed in the hope that it will be useful,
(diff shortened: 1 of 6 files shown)


packages/web3-net/types/index.d.ts
```diff
@@ -0,0 +1,31 @@
...
+import {AbstractWeb3Module, Web3ModuleOptions} from 'web3-core'
+
+export class Network extends AbstractWeb3Module {
+    constructor(
+        provider: AbstractProviderAdapter | provider,
+        options?: Web3ModuleOptions
+    );
+    getId(callback?: (error: Error, id: number) => void): Promise<number>;
+    isListening(callback?: (error: Error, listening: boolean) => void): Promise<boolean>;
+    getPeerCount(callback?: (error: Error, peerCount: number) => void): Promise<number>;
```
nivida, line 31: getBlock and getNetworkType methods are missing.

nivida (changes requested):

packages/web3-net/types/index.d.ts
joshstevens19: added `getNetworkType` but `getBlock` will be exposed in `web3-eth` so types will go in there for that 👍

nivida (approved):

PR #2122, opened 2018-12-14
web3-eth typings

joshstevens19:
web3-eth types 👍


packages/web3-eth/types/index.d.ts
```diff
@@ -0,0 +1,268 @@
...
+    subscribe(type: 'logs', callback?: (error: Error, result: Subscribe<Log>) => void): Promise<Subscribe<Log>>
+    subscribe(type: 'logs', options?: Logs, callback?: (error: Error, result: Subscribe<Log>) => void): Promise<Subscribe<Log>>;
+    subscribe(type: 'syncing', callback?: (error: Error, result: Subscribe<any>) => void): Promise<Subscribe<any>>
+    subscribe(type: 'newBlockHeaders', callback?: (error: Error, result: Subscribe<BlockHeader>) => void): Promise<Subscribe<BlockHeader>>
+    subscribe(type: 'pendingTransactions', callback?: (error: Error, result: Subscribe<Transaction>) => void): Promise<Subscribe<Transaction>>
+    setProvider(provider: AbstractProviderAdapter | provider, net?: net.Server): boolean;
+    readonly providers: Providers;
+    readonly givenProvider: AbstractProviderAdapter | provider | null;
+    BatchRequest(): BatchRequest;
+    extend(methods: Methods): any;
```
nivida: the method extend does no longer exists.

packages/web3-eth/types/index.d.ts
```diff
@@ -0,0 +1,268 @@
...
+    blockHash: string;
+    blockNumber: number;
+}
+
+export interface Subscribe<T> {
+    subscription: {
+        id: string
+        subscribe(callback?: (error: Error, result: Subscribe<T>) => void): Subscribe<T>
+        unsubscribe(callback?: (error: Error, result: boolean) => void): void | boolean
+        arguments: object
```
nivida: It's called options and not arguments :)
You could create an AbstractSubscription type: 
https://github.com/ethereum/web3.js/blob/ethereumProvider/packages/web3-core-subscriptions/lib/subscriptions/AbstractSubscription.js

packages/web3-eth/types/index.d.ts
```diff
@@ -0,0 +1,268 @@
...
+
+export interface Block extends BlockHeader {
+    transactions: Transaction[];
+    size: number
+    difficulty: number
+    totalDifficulty: number
+    uncles: string[];
+}
+
+export interface SignedTransaction {
```
nivida: SignedTransaction interface is defined twice in this file.

packages/web3-eth/types/index.d.ts
```diff
@@ -0,0 +1,268 @@
...
+    messageHash?: string;
+    r: string;
+    s: string;
+    v: string;
+    rawTransaction?: string;
+}
+
+export interface Sign extends SignedTransaction {
```
nivida: It does not really extend from SignedTransaction it's just an encrypted string

packages/web3-eth/types/tests/eth.tests.ts
```diff
@@ -0,0 +1,414 @@
...
+
+// $ExpectType Promise<Subscribe<BlockHeader>>
+eth.subscribe('pendingTransactions');
+// $ExpectType Promise<Subscribe<BlockHeader>>
+eth.subscribe('pendingTransactions', (error: Error, result: Subscribe<Transaction>) => {});
+
+// $ExpectType Providers
+eth.providers;
+
+// $ExpectType AbstractProviderAdapter | provider | null
```
nivida: givenProvider will never return AbstractProviderAdapter just provider or null. :-)

packages/web3-eth/types/tests/eth.tests.ts
```diff
@@ -0,0 +1,414 @@
...
+eth.givenProvider;
+
+// $ExpectType BatchRequest
+eth.BatchRequest();
```
nivida, line 81: Have you defined the BatchRequest typings in web3-providers?

packages/web3-eth/types/tests/eth.tests.ts
```diff
@@ -0,0 +1,414 @@
...
+// $ExpectType Promise<SignedTransaction>
+eth.signTransaction({
+    from: '0xEB014f8c8B418Db6b45774c326A0E64C78914dC0',
+    gasPrice: '20000000000',
+    gas: '21000',
+    to: '0x3535353535353535353535353535353535353535',
+    value: '1000000000000000000',
+    data: ''
+});
+// $ExpectType Promise<SignedTransaction>
```
nivida, line 293: missing break

packages/web3-eth/types/tests/eth.tests.ts
```diff
@@ -0,0 +1,414 @@
...
+eth.signTransaction({
+    from: '0xEB014f8c8B418Db6b45774c326A0E64C78914dC0',
+    gasPrice: '20000000000',
+    gas: '21000',
+    to: '0x3535353535353535353535353535353535353535',
+    value: '1000000000000000000',
+    data: ''
+}, '0xEB014f8c8B418Db6b45774c326A0E64C78914dC0');
+// $ExpectType Promise<SignedTransaction>
+eth.signTransaction({
```
nivida, line 303: missing break

packages/web3-eth/types/tests/eth.tests.ts
```diff
@@ -0,0 +1,414 @@
...
+    from: '0xEB014f8c8B418Db6b45774c326A0E64C78914dC0',
+    gasPrice: '20000000000',
+    gas: '21000',
+    to: '0x3535353535353535353535353535353535353535',
+    value: '1000000000000000000',
+    data: ''
+}, (error: Error, signedTransaction: SignedTransaction) => {});
+// $ExpectType Promise<SignedTransaction>
```
nivida, line 311: missing break

packages/web3-eth/types/tests/eth.tests.ts
```diff
@@ -0,0 +1,414 @@
...
+    "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef",
+    "0xD1FE5700000000000000000000000000D1FE5700000000000000000000000000"
+]);
+
+// $ExpectType Promise<boolean>
+eth.submitWork([
+    "0x0000000000000001",
+    "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef",
+    "0xD1FE5700000000000000000000000000D1FE5700000000000000000000000000"
+], (error: Error, result: boolean) => {});
```
nivida, line 397: Idk, but maybe it would be better to split that up in some files. It's easier to fix something and to keep the overview if this would be splitted up. :)

nivida (changes requested):

packages/web3-eth/types/tests/eth.tests.ts
joshstevens19: it is all contained within the same class so in my eyes it should be in 1. You have to generate the class and test the context of it, splitting up feels unnecessary. I could split it up and import them into the the 1 test file for example 

in eth.tests.ts
```
import { SignedTests }  from './signed-transaction-test.ts'

const eth = new Eth('http://localhost:8545');

SignedTests(eth)
..... rest to follow
```

would you like that more?

packages/web3-eth/types/tests/eth.tests.ts
joshstevens19: i only break after the section of tests are done, so not for every test a new break for every bunch of tests on the same method. Same style has been throughout

packages/web3-eth/types/tests/eth.tests.ts
joshstevens19: i only break after the section of tests are done, so not for every test a new break for every bunch of tests on the same method. Same style has been throughout

packages/web3-eth/types/tests/eth.tests.ts
joshstevens19: i only break after the section of tests are done, so not for every test a new break for every bunch of tests on the same method. Same style has been throughout

packages/web3-eth/types/tests/eth.tests.ts
joshstevens19: yep 

```
export class BatchRequest {
    constructor(provider: provider, jsonRpcMapper: JsonRpcMapper, jsonRpcResponseValidator: JsonRpcResponseValidator);
    add(request: any): void;
    execute(): void;
    hasOutputFormatter(request: any): boolean;
}
```

packages/web3-eth/types/index.d.ts
joshstevens19: removed 👍

joshstevens19: removed and removed docs as well

packages/web3-eth/types/tests/eth.tests.ts
nivida: Yeah that's true let us keep it in one file :)

packages/web3-eth/types/tests/eth.tests.ts
joshstevens19: 👍

packages/web3-eth/types/index.d.ts
joshstevens19: was not even used so remove completely.

packages/web3-eth/types/index.d.ts
joshstevens19: oops sorted 👍

joshstevens19:
All PR comments are resolved 👍 Please merge  @nivida

nivida (approved):

PR #2123, opened 2018-12-14
Web3 types

joshstevens19:
web3 types with some fixing to some other types

### DIFF
--- a/packages/web3-providers/types/index.d.ts
+++ b/packages/web3-providers/types/index.d.ts
@@ -55,6 +55,7 @@ export class SocketProviderAdapter extends AbstractProviderAdapter {
 }
 
 export class BatchRequest {
+    constructor();
     constructor(provider: provider, jsonRpcMapper: JsonRpcMapper, jsonRpcResponseValidator: JsonRpcResponseValidator);
     add(request: any): void;
     execute(): void;
--- a/packages/web3/types/index.d.ts
+++ b/packages/web3/types/index.d.ts
@@ -21,8 +21,8 @@ import {AbstractWeb3Module, Web3ModuleOptions, Providers} from 'web3-core';
 import {Utils} from 'web3-utils';
 import * as net from 'net';
 import {provider, AbstractProviderAdapter, BatchRequest} from 'web3-providers';
-import {Eth} from "web3-eth";
-import {Network} from "web3-net";
+import {Eth} from 'web3-eth';
+import {Network} from 'web3-net';
 
 export class Web3 extends AbstractWeb3Module {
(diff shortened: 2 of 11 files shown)


packages/web3-providers/types/index.d.ts
```diff
@@ -55,6 +55,7 @@ export class SocketProviderAdapter extends AbstractProviderAdapter {
 }
 
 export class BatchRequest {
+    constructor();
     constructor(provider: provider, jsonRpcMapper: JsonRpcMapper, jsonRpcResponseValidator: JsonRpcResponseValidator);
```
nivida, line 59: That's not the current constructor but I will change it by my self after I wrote all the providers tests.

packages/web3/types/index.d.ts
```diff
@@ -0,0 +1,53 @@
...
+*/
+/**
+ * @file index.d.ts
+ * @author Josh Stevens <joshstevens19@hotmail.co.uk>
+ * @date 2018
+ */
+
+import {AbstractWeb3Module, Web3ModuleOptions, Providers} from 'web3-core';
+import {Utils} from 'web3-utils';
+import * as net from 'net';
```
nivida, line 22: I think ```import * as net from 'net'``` is not needed this is just for the nodejs runtime and you will have there the types.

packages/web3/types/index.d.ts
```diff
@@ -0,0 +1,53 @@
...
+import {provider, AbstractProviderAdapter, BatchRequest} from 'web3-providers';
+import {Eth} from "web3-eth";
+import {Network} from "web3-net";
```
nivida: Could you run eslint for the typing files to change the double quotes to single quotes and other stuff? :)

nivida (changes requested):

packages/web3-providers/types/index.d.ts
joshstevens19: that constructor was what it as before, yeah if you can change it what would be great 👍

packages/web3/types/index.d.ts
joshstevens19: oops normally do will sort now

packages/web3/types/index.d.ts
joshstevens19: this is a types file, i am not sure that would work? I think you need to `import * as net from 'net'`. For example even though fs is built into node you still need to import it to use it. Think it has to stay as it is and i think should as its more explicit of what namespace its coming from.. thoughts?

nivida (approved):

PR #2125, opened 2018-12-14
Web3 bzz types

joshstevens19:
web3-bzz types - as they have no dependencies on shared web3-core + web3-provider i have kept them loosely typed. 

This should really have a dependency on `web3-providers` if we could put that dependency on we could strongly type most of this without repeating typings? I think it should also inherit `AbstractWeb3Module` as well on the Bzz class. 

This is why i decided to keep it loosely typed, if we can have a discussion on the above to bring these dependencies in then that would be great.

👍

### DIFF
--- a/packages/web3-bzz/types/tests/bzz-test.ts
+++ b/packages/web3-bzz/types/tests/bzz-test.ts
@@ -63,3 +63,6 @@ bzz.pick.directory();
 
 // $ExpectType Promise<any>
 bzz.pick.data();
+
+// $ExpectType any
+new bzz.BatchRequest();
--- a/packages/web3-bzz/types/tests/bzz-test.ts
+++ b/packages/web3-bzz/types/tests/bzz-test.ts
@@ -1,0 +1,62 @@
+/*
+    This file is part of web3.js.
+    web3.js is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Lesser General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    web3.js is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License for more details.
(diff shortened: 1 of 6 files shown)


acud:
hello earthlings, greetings from the swarm team :)

could someone please get in touch over gitter (with me or on `ethersphere/orange-lounge`) so we could sync up a bit on what's happening with web3 and swarm?

thank you <3

nivida:
@justelad Yes, what's your username in the EF discord?

nivida:
@joshstevens19 
Didn't refactor this module just ported it to ES6. Yes, there is a lot we could improve in this module. The imported swarm-js isn't really maintained and because of this web3 does not support all swarm features. I've planned to include another existing implementation of swarm or will write one by my own for a better support of it.

The bzz provider has nothing to do with all the other providers. That's why it is correct to not have web3-providers as a dependency. It's currently also no Web3 module because it is just a swarm-js wrapper.

packages/web3-bzz/types/tests/bzz-test.ts
```diff
@@ -0,0 +1,68 @@
...
+    kind: 'directory',
+    defaultFile: '/index.html'
+});
+
+// $ExpectType Promise<string | {} | Buffer>
+bzz.download('a5c10851ef054c268a2438f10a21f6efe3dc3dcdcc2ea0e6a1a7a38bf8c91e23');
+// $ExpectType Promise<string | {} | Buffer>
+bzz.download('a5c1', '/target/dir');
+
+// $ExpectType Promise<any>
```
nivida, line 58: Isn't it possible to define these types more explicit then ```Promise<any>```?

nivida (changes requested):

packages/web3-bzz/types/tests/bzz-test.ts
joshstevens19: This is a swarm type and from investigating i think it can be any on these cases. Let me know if you find a better typing.

acud:
> @justelad Yes, what's your username in the EF discord?

We are on Gitter not Discord - you can find us at `ethersphere/hq` and myself under the same git handle (@justelad)

nivida (approved):

PR #2126, opened 2018-12-15
remove Swarm from package exports

acud:
Currently, Swarm implementation in `web3.js` is broken. This PR removes the problematic references from the main `Web3` object until we sort this issue out.

If someone could point out how and where `web3-bzz` is maintained since the npm repo link is broken.

### DIFF
--- a/lib/web3.js
+++ b/lib/web3.js
@@ -38,6 +38,7 @@ var DB = require('./web3/methods/db');
 var Shh = require('./web3/methods/shh');
 var Net = require('./web3/methods/net');
 var Personal = require('./web3/methods/personal');
+var Swarm = require('./web3/methods/swarm');
 var Debug = require('./web3/methods/debug');
 var Settings = require('./web3/settings');
 var version = require('./version.json');
@@ -61,6 +62,7 @@ function Web3 (provider) {
     this.net = new Net(this);
     this.personal = new Personal(this);
     this.debug = new Debug(this);
+    this.bzz = new Swarm(this);
     this.settings = new Settings();
     this.version = {
         api: version.version
--- a/lib/web3.js
+++ b/lib/web3.js
@@ -38,7 +38,6 @@ var DB = require('./web3/methods/db');
 var Shh = require('./web3/methods/shh');
 var Net = require('./web3/methods/net');
 var Personal = require('./web3/methods/personal');
-var Swarm = require('./web3/methods/swarm');
 var Debug = require('./web3/methods/debug');
 var Settings = require('./web3/settings');
 var version = require('./version.json');
@@ -62,7 +61,6 @@ function Web3 (provider) {
     this.net = new Net(this);
     this.personal = new Personal(this);
     this.debug = new Debug(this);
-    this.bzz = new Swarm(this);
     this.settings = new Settings();
     this.version = {
         api: version.version
--- a/lib/web3/methods/swarm.js
+++ b/lib/web3/methods/swarm.js
@@ -24,7 +24,6 @@
 
 "use strict";
 
-var Method = require('../method');
 var Property = require('../property');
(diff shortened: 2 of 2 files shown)


nivida:
This is supported here by this monorepo but it's just a wrapper of swarm-js. 
(https://github.com/MaiaVictor/swarm-js)

I'm currently refactoring the entire code base as you can see here #2000.
I would like to implement the swarm API the correct way until the stable release of v1.0 but will first finish my ongoing refactoring. 

We could remove this package in version 0.20.x and keep it in the 1.0 branch because I will fix it there.

acud:
Whoops, my bad - I wanted to keep the existing properties on `web3.bzz`.

We had a bit of a discussion over this topic on the swarm team and:
1. we don't want to work with content (i.e. upload/download) through JSON-RPC (that's the current (broken) implementation in web3)
2. we don't want to have two different packages for working with swarm (`web3` && `web3-bzz`). 

I am, however, aware of the fact that the latter may be troublesome with current design due to the fact that the providers you support OOB are `Ipc` and `Ws`; that being said most of the Swarm API lives on the HTTP proxy that the node exposes (that's how `swarm-js` works). We could, however, solve this with providing a JSON-RPC function that publishes relevant node info (e.g. on which address/port the HTTP proxy listens on), this way `new Swarm` can still be initialised with an `Ipc` provider, query for the relevant info then switch-over under the hood to normal HTTP operations for manipulating content.

As a first step I'd be happy to see these broken JSON-RPC methods go away, they're broken and cause nothing but discomfort for everyone involved.

After that we can work jointly to finding a neat solution with a long term maintenance outlook. Currently there are a few libraries out there that support working with [Swarm](https://swarm-guide.readthedocs.io/en/latest/introduction.html#available-apis), I'd advise that if anything, we should aim to coalesce the different implementations into one maintainable repository (and eliminating `web3-bzz` would be the next thing - it should probably be ported directly into `web3`).

nivida:
> we don't want to have two different packages for working with swarm (web3 && web3-bzz).

This is not really the case. If someone wants to install the complete lib they can install the web3 module with NPM and will get all the modules. But there is also the possibility to just install a part of the lib and that's why several packages on NPM exists. (e.g. ``web3-bzz``)
But this should change in a later refactoring. Because of the tree-shaking it isn't longer required to publish smaller packages to save some kilobytes. 
There is an open discussion (https://github.com/ethereum/web3.js/issues/2109) about it.
That's how it will be after the folderstructure improvement:

```
import {Swarm} from 'web3';
const swarm = new Swarm(provider);
```  
---

>I am, however, aware of the fact that the latter may be troublesome with current design due to the fact that the providers you support OOB are Ipc and Ws; that being said most of the Swarm API lives on the HTTP proxy that the node exposes (that's how swarm-js works). We could, however, solve this with providing a JSON-RPC function that publishes relevant node info (e.g. on which address/port the HTTP proxy listens on), this way new Swarm can still be initialised with an Ipc provider, query for the relevant info then switch-over under the hood to normal HTTP operations for manipulating content.

Will ask you some questions about it on gitter. Because I can't see any JSON-RPC method I should have implemented during the refactoring in web3 for swarm and also in swarm-js.

---

> After that we can work jointly to finding a neat solution with a long term maintenance outlook. Currently there are a few libraries out there that support working with Swarm, I'd advise that if anything, we should aim to coalesce the different implementations into one maintainable repository (and eliminating web3-bzz would be the next thing - it should probably be ported directly into web3).

That would be cool! Yes, having one swarm implementation that everyone is using instead of N different implementations would be great. But I think it will end up that we will implement swarm by our own just to be sure we can release hot-fixes and patches fast without to wait on a third party dependency.

acud:
@nivida should we merge this?

PR #2127, opened 2018-12-15
remove `web3.bzz` from docs

acud:
Currently, Swarm implementation in `web3.js` is broken and the documentation is very much misleading. 
This PR removes the `web3.bzz` documentation from the TOC in order to reduce clutter and confusion among users till we sort this out.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -38,7 +38,6 @@ Contents:
     web3-eth-iban
     web3-eth-abi
     web3-net
-    web3-bzz
     web3-shh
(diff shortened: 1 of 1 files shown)


nivida:
Yes, I think we could remove the documentation until swarm is correctly implemented in web3. But I would like to discuss first all the things.

acud:
👍+referencing https://github.com/ethereum/web3.js/pull/2126 to keep track of things

acud:
@nivida should we merge this?

PR #2128, opened 2018-12-15
Web3 shh typings

joshstevens19:
web3-shh full typing's. 

All typing's have been written, i have brought `web3-core` into the dependencies as it was a child dependency anyway, as was brought it by other packages so this is *not* a breaking change. This should of really been a parent dependency from the start. 

Cheers

### DIFF
--- a/docs/web3-shh.rst
+++ b/docs/web3-shh.rst
@@ -63,7 +63,7 @@ Returns
 -------
 
 
-``String`` - The version of the current whisper running.
+``Promise`` returns ``String`` - The version of the current whisper running.
 
 
 -------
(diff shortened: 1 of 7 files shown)


docs/web3-shh.rst
```diff
@@ -63,7 +63,7 @@ Returns
 -------
 
 
-``String`` - The version of the current whisper running.
+``Promise`` returns ``String`` - The version of the current whisper running.
```
nivida, line 66: I would like this syntax more: ```Promise<String>```  🤔

nivida (changes requested):

docs/web3-shh.rst
joshstevens19: I would as well but the web3-eth describes it like this so i wanted to keep it consistent 

https://web3js.readthedocs.io/en/1.0/web3-eth.html#getcoinbase

nivida: Okay then I'll change that later everywhere :)

nivida (approved):

Issue #1267, opened 2017-12-31
web3.accounts.create() generating wrong public key


agilani:
Which Web3.js version?: 1.0.0-beta.27
Which operating system?: Windows / Chrome
How installed?: NPM
Using it with Angular-CLI project

# actual behavior
An account generated by web3.accounts.create() is creating an account with public/private key. Which is desired behavior, But when i try to access the account using the private key on myetherwallet.com, it displayes a different public key than the one given by web3.

# expected behavior
The public key should be same.

# steps to reproduce
initialize web3 and create a new account. I created 10 accounts and 8 out of them had the issue.

DanielMReed:
A minimum working example so I can be reproduced would help, is MEW giving you the public key or the address?

phoax:
Same problem, I try to get a public key from a private key, and
`web3.eth.accounts.privateKeyToAccount('c7864d1ad07dd6ee3e1ec554e36bfe50cb95fe660bd873fd96dfe32f8908a6d0');`
generates 
`address: "0x69E36154eba1D933aa1820195a54a5068622902f"`
It should be `0x48d48309670e7E64CD277778DE02689F5268081d`

phoax:
I have a similar problem, I try to get a public key from a private key, and
`web3.eth.accounts.privateKeyToAccount('c7864d1ad07dd6ee3e1ec554e36bfe50cb95fe660bd873fd96dfe32f8908a6d0');`
generates 
`address: "0x69E36154eba1D933aa1820195a54a5068622902f"`
It should be `0x48d48309670e7E64CD277778DE02689F5268081d`

kevinwucodes:
@phoax the private key should always be prefixed with `0x` because it is expecting a hexadecimal string

ARitz-Cracker:
Lots of confusion could be solved by allowing a buffer to be passed as well, not just a string that'll have its first 2 bytes stripped cuz it's expecting a "0x" prefix.

samjam48:
I had the same error with an incorrect account address being generated from passing in a privateKey variable using:
`web3.eth.accounts.privateKeyToAccount(privateKey)`

Simply fixed by using:
`web3.eth.accounts.privateKeyToAccount("0x"` + privateKey)`

Thanks @kevinwucodes


Of course I upgrade my web3 and went in day's worth of debugging different errors to arrive back at this problem and before actually finding this solution....

tcrowe:
[...]

PR #2129, opened 2018-12-15
[documentation] Clarify privateKeyToAccount private string is hex

tcrowe:
## What does it do?

#1267 brought up a good point that `web3.eth.accounts.privateKeyToAccount` `privateKey` does not specify that it's a hex string. This PR fixes that in the documentation.

## Does it work?

screen shot 2018-12-15 at 4 50 09 pm

### DIFF
--- a/docs/web3-eth-accounts.rst
+++ b/docs/web3-eth-accounts.rst
@@ -104,7 +104,7 @@ Creates an account object from a private key.
 Parameters
 ----------
 
-1. ``privateKey`` - ``String``: The private key to convert.
+1. ``privateKey`` - ``String``: The private key hex string beginning with ``0x``.
 
 -------
 Returns


joshstevens19 (approved):

joshstevens19:
Thanks @tcrowe, looks fine to me, we need 2 approvals before we can merge so @nivida can you merge when you get a second. 

Just a note @tcrowe  if your doing doc updates can you please do it against ethereumProvider branch as that will be the new 1.0.x and default branch once its deployed.

nivida:
@joshstevens19 That's wrong they don't have to be against the ethereumProvider because I didn't change the documentation there and I will merge the ethereumProvider branch into the 1.0 branch.

@tcrowe Thanks for submitting this PR! :)

joshstevens19:
Surely you want to keep your `ethereumProvider` inline with the latest code so would of assumed you would want it in the `ethereumProvider`, sorry if that was incorrect 👍

Issue #1440, opened 2018-03-09
(Documentation flawed) `receiptsRoot` missing from example output


drandreaskrueger:
How do I get the `receiptsRoot` ?

> Tries in Ethereum  
> All of the merkle tries in Ethereum use a Merkle Patricia Trie.  
> From a block header there are 3 roots from 3 of these tries.  
> 1 stateRoot
> 2 transactionsRoot
> 3 receiptsRoot
> [...]
> Receipts Trie  
> Every block has its own Receipts trie. A path here is: rlp(transactionIndex).  
> transactionIndex is its index within the block it's mined.  
> Never updates.  

See [tries-in-ethereum](https://github.com/ethereum/wiki/wiki/Patricia-Tree#tries-in-ethereum).

I would expect `getBlock` to answer my question, but it doesn't.

The [example](https://github.com/ethereum/wiki/wiki/JavaScript-API#web3ethgetblock) answer is:

```
{
  "number": 3,
  "hash": "0xef95f2f1ed3ca60b048b4bf67cde2195961e0bba6f70bcbea9a2c4e133e34b46",
  "parentHash": "0x2302e1c0b972d00932deb5dab9eb2982f570597d9d42504c05d9c2147eaf9c88",
  "nonce": "0xfb6e1a62d119228b",
  "sha3Uncles": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
  "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  "transactionsRoot": "0x3a1b03875115b79539e5bd33fb00d8f7b7cd61929d5a3c574f507b8acf415bee",
  "stateRoot": "0xf1133199d44695dfa8fd1bcfe424d82854b5cebef75bddd7e40ea94cda515bcb",
  "miner": "0x8888f1f195afa192cfee860698584c030f4c9db1",
  "difficulty": BigNumber,
  "totalDifficulty": BigNumber,
  "size": 616,
  "extraData": "0x",
  "gasLimit": 3141592,
  "gasUsed": 21662,
[...]

PR #2147, opened 2018-12-24
[documentation] Adding receiptsRoot in example's output

nanspro:
receiptsRoot: The Keccak 256-bit hash of the root node of the trie structure populated with the receipts of each transaction in the transactions list portion of the block.
It should be present inside a block and getBlock() should function returns a recieiptsRoot along itself.
Fixes https://github.com/ethereum/web3.js/issues/1440


drandreaskrueger:
Thanks a lot. I have closed that issue #1440 now. Well done.

PR #2148, opened 2018-12-24
[documentation] Changing parameters of methods.myMethod.call()

### LINKED ISSUE #1287: web3 1.0: Specifying block number in contract method calls.
From the 1.0 documentation it looks as if you can't specify the block number when calling contract methods querying state, ie. `myContract.myMethod().call({defaultBlock: 'latest' })`. Is this a feature that will be added, since it was in web3 <1.0: https://github.com/ethereum/wiki/wiki/JavaScript-API#contract-methods (third parameter)

nanspro:
Right now as https://github.com/ethereum/web3.js/issues/1287 stated,
blockNumber is not defined in the docs and methods.myMethod.call() function's input parameters are poorly written.
This PR fixes https://github.com/ethereum/web3.js/issues/1287 !!

### DIFF
--- a/docs/web3-eth-contract.rst
+++ b/docs/web3-eth-contract.rst
@@ -432,7 +432,7 @@ methods.myMethod.call
 
 .. code-block:: javascript
 
-    myContract.methods.myMethod([param1[, param2[, ...]]]).call(options[, callback])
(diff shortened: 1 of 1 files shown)


nanspro:
@nivida could you review and tell me if there is any issue with changes proposed.

einaralex (approved):
LGTM

nivida:
Thanks for fixing this documentation issue!

PR #2159, opened 2019-01-04
Typo fix.

MoMannn:
Both `leftPad` and `rightPad` methods have the same return description.


nivida:
I've fixed this typo in the PR #2000 because the architecture did change a lot there. Thanks for pointing this out!

PR #2192, opened 2019-01-15
add getId to the umbrella module

### LINKED ISSUE #1353: getNetwork not available
using the tag: ^1.0.0-beta.29

screen shot 2018-02-03 at 10 42 50 am

The version property on the `web3` object is a string, not an object. So attempting to call properties like `node`, `network`, or `getNetwork` all fail. Any guidance is appreciated.

OFRBG:
Fixes #1353
In order to expose what web3 0.2x showed via `web3.version` for the network ID, it now exposes `web3.getId` instead. This is a shortcut for `web3.eth.net.getId`. `web3.version` remains a string. Docs are updated to include the info for `web3.getId`.

There was a pattern mismatch at `/packages/web3-eth/src/index.js`. The first entry of the methods didn't have the `params: 0` option.


nivida:
I don't think we really have to provide such an alias and the mentioned issue is not a issue of the lib. The problem is that a user is trying to use methods from the v0.20.x with the v1.0. Because of this I will close your PR. Thanks for your efforts!

PR #2194, opened 2019-01-15
make the tx immutable

### LINKED ISSUE #2190: Function 'sendSignedTransaction' changes the value of the 'to' field from 'null' to '0x'
```
let options = {
    gas : gas
};

    return transactionReceipt;
}
```

```
```

Thanks

OFRBG:
Fixes #2190
The transaction object can be deep cloned by parsing to a string and back. I do coincide that muting an object by passing the reference, without telling the user, is a bad idea.


nivida:
Could you do these changes against the ethereumProvider. I've changed the code architecture a lot there and will merge it in the next days.

OFRBG:
Moved to #2199.

PR #2202, opened 2019-01-21
WIP: web3-core-types submodule

OFRBG:
A common problem with web3 usage is transaction object creation. Passing around the ad hoc transaction object is confusing and error-prone.

In order to address this issue, `web3-core-types` includes a class to create transactions that need to take all required parameters `{from, to, value, gas, gasPrice, data, nonce}` or the default markers, `auto` and `none`. All need to be explicitly set to default or provided. Furthermore, the error messages should provide enough feedback for the developer to understand why a parameter was rejected.

This submodule should also include type checks and parsing for:
* Hex
* Address
* Transaction
* TransactionReceipt
* BigNumber
* ABI
* ABIItem
* Ascii
* Ether
* Topic
* Bloom
* Sha3
* Block
* Event
* RpcPayload
* SyncEvent

This should reduce confusion regarding object building, increase type consistency, and improve cohesion.

 ## Type of change
 * [x]  Feature (non-breaking)
 
## Functionality
At the very least, the `web3-core-types` module should:
* [x] always build a valid, signable transaction.
* [x] throw on invalid parameters and provide sufficient information to fix them.
* [x] work as a standalone + `web-utils` + `web3-core`.
* [ ] build a transaction with all numeric parameters set in `BN` and addresses in valid strings.
* [ ] export a hex-encoded version of the transaction.
* [ ] export a human-readable version of the transaction.

 ## Checklist:
 * [x]  I have selected the correct base branch. [`ethereumProvider`]
 * [ ]  I have performed a self-review of my own code.
 * [ ]  I have commented my code, particularly in hard-to-understand areas.
 * [ ]  I have made corresponding changes to the documentation.
 * [ ]  My changes generate no new warnings.
 * [ ]  Any dependent changes have been merged and published in downstream modules.
 * [ ]  I ran `npm run test` with success and extended the tests if necessary.
 * [ ]  I ran `npm run build` and tested the resulting file from `dist` folder in a browser.
 * [ ]  I have tested my code on the live network.

### DIFF
--- a/packages/web3-tx/README.md
+++ b/packages/web3-tx/README.md
@@ -1,27 +1,7 @@
-# web3-eth-abi
+# web3-tx
 
 This is a sub package of [web3.js][repo]
 
-This is the abi package will be used in the `web3-eth` package.
+This is the transaction package that allows building of web3-valid transactions.
+.
 Please read the [documentation][docs] for more.
-
-## Installation
-
-```bash
-npm install web3-eth-abi
-```
-
-## Usage
-
-```js
-import {AbiCoder} from 'web3-eth-abi';
-
-new AbiCoder();
-```
-
-## Types 
-
-All the typescript typings are placed in the types folder. 
-
-[docs]: http://web3js.readthedocs.io/en/1.0/
-[repo]: https://github.com/ethereum/web3.js
--- a/packages/web3-tx/README.md
+++ b/packages/web3-tx/README.md
@@ -1,0 +1,27 @@
+# web3-eth-abi
+
+This is a sub package of [web3.js][repo]
+
+This is the abi package will be used in the `web3-eth` package.
+Please read the [documentation][docs] for more.
+
+## Installation
+
+```bash
+npm install web3-eth-abi
+```
+
+## Usage
+
+```js
+import {AbiCoder} from 'web3-eth-abi';
+
+new AbiCoder();
+```
+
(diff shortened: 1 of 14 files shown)


nivida:
Hay,

This is a great idea but not the way we will implement it in the web3.js lib. The goal is to simplify the folder structure and to reduce the amount of provided packages on NPM. 
I had the idea to create value objects for each web3.js related type and they should be defined in a folder e.g.: "core/types/...".

If you would like to start working on something like this we could create a web3-core-types module for now until I get the time to simplify the folder structure.

nivida:
PS: I wrote down some weeks ago a "draft" of such a module:
https://gist.github.com/nivida/d75c26daea14607f0c57b42582b34f78

OFRBG:
@nivida Gotcha. I thought making it a standalone module would work because often times devs only want to build the transaction and do signing elsewhere. Unifying `utils` and `type-utils` sounds good to me too. I'll adapt the branch to host `Transaction` as well as the listed types.

Should I keep the checklist idea of this package being "strict" and standalone?

nivida:
Yep, this module would remove the web3-utils and web3-core-helpers module. But this will not be merged and released that fast because it needs a deprecation announcement etc. :-)

OFRBG:
That's perfect by me. This is definitely not an overnight module to build anyway.

PR #2206, opened 2019-01-21
Added safer localStorage detection so that web3 does not break the JS…

asselstine:
… on the page when turned off.

Original PR was https://github.com/ethereum/web3.js/pull/1923

### DIFF
--- a/packages/web3-eth-accounts/src/Accounts.js
+++ b/packages/web3-eth-accounts/src/Accounts.js
@@ -753,18 +753,18 @@ function storageAvailable(type) {
         storage.removeItem(x);
         return true;
     }
-    catch(e) {
-      var isDOMException = typeof DOMException !== 'undefined' && e instanceof DOMException
+    catch(error) {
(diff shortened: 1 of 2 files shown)


joshstevens19:
Hey i think the PR https://github.com/ethereum/web3.js/pull/2203 fixes this already? If localStorage fails i think it puts out code 18 which is covered in all regards. This PR went in today so let me know if you think we need what you have defined if not we can close it.

Thanks again for your pull request 👍

nivida:
@asselstine Thanks for submitting this PR. I'll test it and merge it after.

asselstine:
@nivida I've fixed the build- looks like Unicorn was unhappy with a variable name and an unused import.

Thanks for taking some time for this one, having the project work without localStorage would really help us out.

nivida:
This will get fixed with PR https://github.com/ethereum/web3.js/pull/2367 because it just gets items from the ``localStorage`` when you explicit call the ``accounts.wallet.load`` or ``accounts.wallet.save`` method.

PR #2208, opened 2019-01-22
[WIP] web3-core-types submodule

OFRBG:
# `web3-core-types`

The `web3-core-types` submodule is a possible replacement for `web3-utils` and `web3-core-utils`. Passing around types, objects, and references is highly error prone.

Example: A common problem with web3 usage is transaction object creation. Passing around the ad hoc transaction object is confusing and error-prone.

In order to address this issue, `web3-core-types` includes a class to create transactions that need to take all required parameters `{from, to, value, gas, gasPrice, data, nonce}` or the default markers, `auto` and `none`. All need to be explicitly set to default or provided. Furthermore, the error messages should provide enough feedback for the developer to understand why a parameter was rejected.

## Features

This submodule should include type checks, parsing, and object creation for:
* [ ] ABI
* [ ] ABIItem
* [x] Address
* [ ] Ascii
* [ ] BigNumber
* [ ] Block
* [ ] Bloom
* [x] Ether
* [ ] Event
* [x] Hex
* [ ] RpcPayload
* [ ] Sha3
* [ ] SyncEvent
* [ ] Topic
* [x] Transaction
* [ ] TransactionReceipt

 
## Functionality

At the very least, the `web3-core-types` module should:
* [ ] work as a standalone.
* [ ] always build a valid object.
* [ ] throw on invalid parameters and provide *more than enough* information to fix them.
* [ ] build a objects that take a set of types but always outputs a standard.
* [ ] export a hex-encoded version of the object.
* [ ] export a human-readable version of the transaction.

## Contributing

Pick a type from the features and implement its methods and types. Be as self-contained as possible, not relying on external packages or other modules.

---

### Proposed Type File Dataflow
```js
        /* -) params are the values given to the constructor
         * -) this.props are the params fed via the constructor
         *      after being filtered.
         * -) this.props start assigned to undefined via initParams */

        /* 1) Set the errors */
        this.error = error; 

        /* 2) Set the inital values */
        this.initParams = initParams; 
        
        /* 3) Initialize the parameters */
        this.props = cloneDeep(initParams); 
      
        /* 4) Reshape params to emulate different constructor overrides */
        if(!isObject(params)) {
            params = {
                p1: params,
                p2: /* default safe value */
            }
        }
        
        /* 5) Check for type and format validity */
        this.props.p = /* condition check */
                ? /* parameter standarization */
                : undefined;


        /* 6) Check for default, auto, none, etc. key values */
        if (params.p === 'auto') this.props.p = /* default/auto/empty value */

        /* 7) Throw if any parameter is still undefined */
        Object.keys(this.props).forEach((key) => {
            typeof this.props[key] === 'undefined' && this._throw(this.error[key]);
        });
        
        /* 8) Make the params immutable */
        Object.freeze(this.props);
```


OFRBG:
@nivida Are you looking to move all of the integer data to `bignumber.js` from `bn.js`?

nivida:
I've updated my draft which I will release this week: https://gist.github.com/nivida/d75c26daea14607f0c57b42582b34f78 

Try to create these objects in a similar way as described in this gist. Feel free to drop a comment there if you have anything to improve in the specification.

OFRBG:
Hey @nivida. Up to now I've been working on `Address`, `Transaction`, `Hex`, and `Ether`. Would you mind checking around the files to see if the implementation matches what you need?

I'm keeping the interface basically intact, but there are some inner implementation aspects that will need work e.g. standardization of the hex values.

The structure looks like this for Address:
```
Address/           (root for type)
├── Address.js     (address state and inner logic)
├── Iban.js        (helpers)
├── factories      
│   └── AddressFactory.js  (injects error description, params, and some preprocessing)
└── index.js       (exports the address creation and exposes static calls)
```

Let me know if this is up to standards and if it needs refactoring.

OFRBG:
@frozeman I would appreciate it a lot if you looked around to check if the module structure has the required quality.

packages/web3-core-types/src/Hex/Hex.js
```diff
@@ -0,0 +1,378 @@
...
+
+export default class Hex {
+    /**
+     * @dev Wrap as object
+     * @param {String} hex
+     *
+     * @constructor
+     */
+    constructor(params) {
+        const requires = ['hex'];
```
nivida: What exactly is the idea behind this array?

packages/web3-core-types/src/Hex/Hex.js
```diff
@@ -0,0 +1,378 @@
...
+        }
+
+        requires.forEach((propName) => {
+            if (typeof this.props[propName] === 'undefined') {
+                this._throw(propName, params[propName]);
+            }
+        });
+
+        /* Make the props immutable */
+        Object.freeze(this.props);
```
nivida: We should wrap this object with an es6 proxy and with a strict getter and setter handler.

packages/web3-core-types/src/Hex/Hex.js
```diff
@@ -0,0 +1,378 @@
...
+     * Build an object of Hex from a base 10 number
+     *
+     * @method fromNumber
+     *
+     * @param {Number} value
+     *
+     * @returns {Hex}
+     */
+    static fromNumber(value) {
+        if (!isNumber(value)) throw new Error(`The given value ${value} is not number type.`);
```
nivida: I'm not a big fan of shorthands because the bundler will improve it anyways later. :-)

packages/web3-core-types/src/Hex/Hex.js
```diff
@@ -0,0 +1,378 @@
...
+
+        const params = {
```
nivida: It's not necessary to define a var if the value is only used once.

packages/web3-core-types/src/Hex/Hex.js
```diff
@@ -0,0 +1,378 @@
...
+
+    /* Instance accessors */
+    /**
+     * Check if the supplied string is hex with 0x prefix
+     *
+     * @method isStrict
+     *
+     * @return {boolean}
+     */
+    isStrict() {
```
nivida: I think the static method alone is enough :)

packages/web3-core-types/src/Hex/Hex.js
OFRBG: For each type, the `requires` array holds the minimum internal properties required to have a valid Type. If by the end of the constructor any of the required properties is still undefined, an error is thrown. It "self tests".

packages/web3-core-types/src/Hex/Hex.js
OFRBG: Since `this.props` are meant to be the immutable, shouldn't it be enough with getters? The core idea was to use `props` are the internal database, and any modifications to the props would require creating a new object. Is this too strict?

nivida:
This PR got closed because of the reorganisation of the branches. Feel free to open a new PR for the 2.x branch.

PR #2215, opened 2019-01-26
types: fixed utils.toWei() to match docs and js implementation

zlumer:
## Description

Migrated code from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32416 by @alexkvak and https://github.com/DefinitelyTyped/DefinitelyTyped/pull/27528 by @zlumer

## Type of change

- [x] Bug fix 

## Checklist:

- [x] I ran ```npm run build``` and tested it in the browser and with node.
- [x] I ran ```npm run-script dtslint``` and tested that all my types are correct


joshstevens19:
Thanks for the PR can you update the method docs as well

`* @param {Number|String|BN} number can be a number, number string or a HEX of a decimal` > `* @param {String|BN} number can be a number, number string or a HEX of a decimal`

then i can approve 👍

zlumer:
This is insufficient for all use cases, see #2214

joshstevens19:
Hey, 

I have put the fixes in https://github.com/ethereum/web3.js/pull/2218 for the ones you mentioned in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32493 

The reason we have a interface is its used in other typing's in different packages like:

https://github.com/ethereum/web3.js/blob/1.0/packages/web3-core-method/types/index.d.ts#L27
https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth-ens/types/index.d.ts#L76
https://github.com/ethereum/web3.js/blob/1.0/packages/web3/types/index.d.ts#L39

So obviously we needed a interface to define the type of for example `web3.Utils` but we also wanted to introduce ability to only import what you need within the `Utils` lib i.e.

```
import { asciiToHex } from 'web3-utils';
```

this was the most cleanest way to do this so developers can use partial imports in their code as well as a nice interface on the `Utils` so other packages can reuse it. This is the only place it repeats typing's like this. 

You are correct it seems we don't have any tests for that interface which i will add to my checklist (just double checked and this is the only place not tested, I will jump on that when I have a second). 

So yeah at least this will be fixed on next release. If you do spot anything else or any other PR which went into DT web3 let me know or do a PR and we get it merged in. 

Cheers

PR #2220, opened 2019-01-27
fix: incorrect removal of tests + static methods in types

joshstevens19:
## Description

A mistake happened in the web3 lib and somehow static methods had been removed from the type definitions which has caused a break of some logic. This should fix all the issues which got undone before. 

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 

## Checklist:

### DIFF
--- a/packages/web3-core/types/index.d.ts
+++ b/packages/web3-core/types/index.d.ts
@@ -25,7 +25,9 @@ import {
     provider,
     ProvidersModuleFactory,
     WebsocketProvider,
-    BatchRequest
+    BatchRequest,
+    HttpProviderOptions,
+    WebsocketProviderOptions
 } from 'web3-providers';
 
 export class AbstractWeb3Module {
@@ -44,10 +46,13 @@ export class AbstractWeb3Module {
     transactionPollingTimeout: number;
     defaultGasPrice: string;
     defaultGas: number;
+    static readonly providers: Providers;
     readonly providers: Providers;
     defaultAccount: string | null;
     readonly currentProvider: EthereumProvider | HttpProvider | IpcProvider | WebsocketProvider;
+    static readonly currentProvider: EthereumProvider | HttpProvider | IpcProvider | WebsocketProvider;
     readonly givenProvider: provider | null;
+    static readonly givenProvider: provider | null;
 
     setProvider(provider: provider, net?: net.Socket): boolean;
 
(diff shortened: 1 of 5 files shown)


packages/web3-core/types/index.d.ts
```diff
@@ -44,10 +46,13 @@ export class AbstractWeb3Module {
     transactionPollingTimeout: number;
     defaultGasPrice: string;
     defaultGas: number;
+    static readonly providers: Providers;
     readonly providers: Providers;
     defaultAccount: string | null;
     readonly currentProvider: EthereumProvider | HttpProvider | IpcProvider | WebsocketProvider;
+    static readonly currentProvider: EthereumProvider | HttpProvider | IpcProvider | WebsocketProvider;
```
nivida: the currentProvider property isn't static.

packages/web3-core/types/index.d.ts
```diff
@@ -44,10 +46,13 @@ export class AbstractWeb3Module {
     transactionPollingTimeout: number;
     defaultGasPrice: string;
     defaultGas: number;
+    static readonly providers: Providers;
     readonly providers: Providers;
```
nivida: the providers property is only a static property. I've changed this some minutes ago. I can do these changes also by myself. https://github.com/ethereum/web3.js/tree/fix/abstract-web3-module-providers

packages/web3-core/types/index.d.ts
```diff
@@ -67,31 +72,37 @@ export interface Web3ModuleOptions {
 }
 
 export interface Providers {
-    HttpProvider: HttpProvider;
-    WebsocketProvider: WebsocketProvider;
-    IpcProvider: IpcProvider;
+    HttpProvider: new (host: string, options?: HttpProviderOptions) => HttpProvider;
```
nivida: This does not have to get changed because I removed the non-static property. As said before I can do these changes also by myself.

packages/web3-core/types/tests/abstract-web3-module-test.ts
```diff
@@ -67,13 +68,22 @@ abstractWeb3Module.defaultGasPrice;
 abstractWeb3Module.defaultGas;
 
 // $ExpectType Providers
-abstractWeb3Module.providers
+abstractWeb3Module.providers;
```
nivida: Only the static property will exist.

packages/web3-core/types/tests/abstract-web3-module-test.ts
```diff
@@ -67,13 +68,22 @@ abstractWeb3Module.defaultGasPrice;
...
+// $ExpectType Providers
+AbstractWeb3Module.providers;
 
 // $ExpectType string | null
 abstractWeb3Module.defaultAccount;
 
 // $ExpectType HttpProvider | IpcProvider | WebsocketProvider | EthereumProvider
 abstractWeb3Module.currentProvider;
+// $ExpectType HttpProvider | IpcProvider | WebsocketProvider | EthereumProvider
+AbstractWeb3Module.currentProvider;
```
nivida, line 82: currentProvider is not static

packages/web3-eth/types/tests/eth.tests.ts
```diff
@@ -71,6 +71,24 @@ eth.subscribe('pendingTransactions');
 // $ExpectType Promise<Subscribe<Transaction>>
 eth.subscribe('pendingTransactions', (error: Error, result: Subscribe<Transaction>) => {});
 
+// $ExpectType Providers
+eth.providers;
```
nivida, line 75: Same here with the static and non-static properties. I think you don't have to test it here too. It's enough if you test it on the AbstractWeb3Module because the Eth module and all others are extending from this class.

nivida (changes requested):

packages/web3-core/types/index.d.ts
joshstevens19: Ok if you fix can remove the readonly typings in this branch as not on a computer!

packages/web3-core/types/index.d.ts
joshstevens19: Yeah your correct remove it 👍

packages/web3-core/types/index.d.ts
joshstevens19: This NEEDS to be like this or the typings will not work! web3.providers is a static thing and is stated in the docs

‘const web3 = new Web3(new Web3.providers.HttpProvider('http://localhost:8545'));’ 

If you don’t have these typings on all of them to this then nobody will ever be able to use the providers static method as you create a new context of the provider hence the update.

packages/web3-core/types/tests/abstract-web3-module-test.ts
joshstevens19: Ok can you fix in this branch as not around on a computer today!

packages/web3-core/types/tests/abstract-web3-module-test.ts
joshstevens19: Same as above

packages/web3-eth/types/tests/eth.tests.ts
joshstevens19: I disagree this is testing the return type of the Eth classes properties! If something changes and eth inherits is removed or something we need to tests to pick it up that this would break the types hence why you would test it on this class as well. Just my view.

nivida: >  If something changes and eth inherits is removed or something we need to tests

This will be achieved with the unit tests I wrote. They test if the module is an instance of the AbstractWeb3Module.

joshstevens19: 👍ok

nivida (approved):

Issue #2275, opened 2019-01-30
TypeScript and Web3 constructor


wnz99:
#### Expected behavior

Initialize Web3 instance without passing an initial provider.

#### Actual behavior

I am trying to initialize Web3 in a TS project as follows:

````javascript
const web3 = new Web3()
web3.setProvider(someProvider)
````
however the compiler outputs the following error:

`An argument for 'provider' was not provided`

Is this expected?

#### Steps to reproduce the behavior

````javascript
const web3 = new Web3()
web3.setProvider(someProvider)
````

#### Error Logs

`An argument for 'provider' was not provided`

#### Versions

Web3 1.0.0-beta.41

**Comment by wnz99** [comment]
It looks like this is something that was introduced recently, `web3@1.0.0-beta.37` allows to initiate a Web3 instance without providing a provider. 

Just for confirmation, should I consider this pattern no longer possible?

```javascript
const web3 = new Web3()
web3.setProvider(someProvider)
```
`const web3 = new Web3()` is something that I came across quite a lot on available tutorials.

andy0130tw:
Also encountered this issue when updating dependencies. I am worried about that If this breaking change is intentional, web3 instances cannot even be initialized in browsers without a injected web3 provider and requires much code for pre-`1.0.0-beta.37` Dapps to adapt to.

For a quick workaround, pass a dummy (but invalid) URI that would be recognized as an HTTP RPC endpoint: `new Web3('http://dummy')`

nivida:
@wnz99  
Thanks for submitting this issue I'll fix this on Monday. I'm currently sick in the bed.

wnz99:
sure, thank you.

nivida:
@wnz99  or @andy0130tw can you explain me the exact use case for initiating a Web3 object without a provider?

nivida:
I'm closing this issue because there was no answer from the creator.

PR #2321, opened 2019-02-04
Simpler provider parameter handling on construction of a Web3 Module

nivida:
## Description

The goal of this PR is to simplify the provider handling. Currently, there are too many different possibilities we provide from the Web3.js library. In the future, there should be one clear path of how Web3.js should get initiated with the provider.

Related issues:
#2275 

## Type of change

- [x] Enhancement

## Checklist:


packages/web3-utils/package-lock.json
```diff
@@ -1,8 +1,6 @@
 {
-	"name": "web3-utils",
-	"version": "1.0.0-beta.41",
```
zachlysobey, line 3: Its a little concerning that the lockfile would change like it does here without any corresponding changes to package.json, etc... Any idea what'd going on here?

nivida: This is because I've done a clean install and deleted all the package-lock.json files before. Will finish this work and release it ~beta.45

PR #2360, opened 2019-02-09
Resolve the PromiEvent that is returned when sending a transaction

ahoppen:
#### TL;DR:
- The `PromiEvent` returned from sending a method to the Ethereum chain is never resolved
- The tests just wait for the `transactionHash` event to be fired and succeed without waiting for the `PromiEvent` to be resolved. Thus the issue was never covered in tests.
- Fixing the second bug uncovers some failing tests that were previously not executed, which I fixed.

#### Long story

I noticed that the `PromiEvents` returned from sending a method to a smart contract is never resolved. Trying to write tests that catch that behaviour, I noticed that you can assert `expect(true).toBe(false)` at the end of `'calls execute with wallets defined'` in `SendTransactionMethodTest.js` without any test failing. It appears the root cause of this is that the test finishes executing when `done` is called from within the callback of the `transactionHash` event. The `await` for the `PromiEvent` is never fulfilled because of the above issue and thus the code after the `await` is never executed. 

## Resolution

The first issue can easily been addressed by calling `promiEvent.resolve(response);` in `AbstractSendMethod.js`.

For the second issue, I think the `done` callback shouldn't be used in combination with an `async` test body. Wrapping the `'transactionHash'` callback inside a promise and awaiting it makes the test body wait for both the `'transactionHash'` event to be fired and the `PromiEvent` to be resolved.

I eliminated all cases where a `done` callback was used in an async test body. That uncovered some test code that was previously not executed. I was able to fix these test cases by just modifying the test code. I'll add comments on GitHub to lines where I'm not 100% sure what the correct behaviour should be.

### DIFF
--- a/packages/web3-core-method/tests/src/methods/transaction/SendTransactionMethodTest.js
+++ b/packages/web3-core-method/tests/src/methods/transaction/SendTransactionMethodTest.js
@@ -70,7 +70,7 @@ describe('SendTransactionMethodTest', () => {
         expect(method.transactionSigner).toEqual(transactionSignerMock);
     });
 
-    it('calls execute with wallets defined', async (done) => {
+    it('calls execute with wallets defined', async () => {
         accountsMock.wallet[0] = {privateKey: '0x0'};
 
         transactionSignerMock.sign = jest.fn(() => {
@@ -83,33 +83,36 @@ describe('SendTransactionMethodTest', () => {
 
         moduleInstanceMock.currentProvider = providerMock;
 
-        promiEvent.on('transactionHash', (response) => {
-            expect(response).toEqual('0x0');
+        const transactionHashEventFired = new Promise((resolve) => {
+            promiEvent.on('transactionHash', (response) => {
+                expect(response).toEqual('0x0');
 
-            expect(transactionConfirmationWorkflowMock.execute).toHaveBeenCalledWith(
-                method,
-                moduleInstanceMock,
-                '0x0',
-                promiEvent
-            );
+                expect(transactionConfirmationWorkflowMock.execute).toHaveBeenCalledWith(
+                    method,
+                    moduleInstanceMock,
+                    '0x0',
+                    promiEvent
+                );
 
-            expect(providerMock.send).toHaveBeenCalledWith(method.rpcMethod, method.parameters);
+                expect(providerMock.send).toHaveBeenCalledWith(method.rpcMethod, method.parameters);
 
-            expect(method.rpcMethod).toEqual('eth_sendRawTransaction');
+                expect(method.rpcMethod).toEqual('eth_sendRawTransaction');
 
-            expect(transactionSignerMock.sign).toHaveBeenCalledWith({gasPrice: '0x0'});
+                expect(transactionSignerMock.sign).toHaveBeenCalledWith({gasPrice: '0x0'});
 
-            done();
+                resolve();
+            });
         });
 
         const response = await method.execute(moduleInstanceMock, promiEvent);
+        await transactionHashEventFired;
 
         expect(response).toEqual('0x0');
 
         expect(method.callback).toHaveBeenCalledWith(false, '0x0');
     });
 
-    it('calls execute with wallets defined and uses the module default gas properties', async (done) => {
+    it('calls execute with wallets defined and uses the module default gas properties', async () => {
         accountsMock.wallet[0] = {privateKey: '0x0'};
 
         transactionSignerMock.sign = jest.fn(() => {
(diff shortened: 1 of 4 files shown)


packages/web3-core-method/tests/src/methods/transaction/SendTransactionMethodTest.js
```diff
@@ -124,37 +127,36 @@ describe('SendTransactionMethodTest', () => {
...
         const response = await method.execute(moduleInstanceMock, promiEvent);
+        await transactionHashEventFired;
 
         expect(response).toEqual('0x0');
 
         expect(method.callback).toHaveBeenCalledWith(false, '0x0');
-
-        expect(method.parameters[0].gas).toEqual(100);
-
-        expect(method.parameters[0].gasPrice).toEqual(100);
```
ahoppen, line 154: These checks are already performed in line 145 so we can delete them.

packages/web3-core-method/tests/src/workflows/TransactionConfirmationWorkflowTest.js
```diff
@@ -124,7 +124,7 @@ describe('TransactionConfirmationWorkflowTest', () => {
 
             expect(methodMock.afterExecution).toHaveBeenCalledWith({blockHash: '0x0'});
 
-            expect(transactionReceiptValidatorMock.validate).toHaveBeenCalledWith({blockHash: '0x0'});
+            expect(transactionReceiptValidatorMock.validate).toHaveBeenCalledWith({blockHash: '0x0'}, undefined);
```
ahoppen, line 127: I'm not 100% sure whether the second parameter is supposed to be `undefined`.

packages/web3-core-method/tests/src/workflows/TransactionConfirmationWorkflowTest.js
```diff
@@ -158,13 +158,15 @@ describe('TransactionConfirmationWorkflowTest', () => {
 
             expect(contractDeployMethodMock.afterExecution).toHaveBeenCalledWith({blockHash: '0x0'});
 
-            expect(transactionReceiptValidatorMock.validate).toHaveBeenCalledWith({blockHash: '0x0'});
+            expect(transactionReceiptValidatorMock.validate).toHaveBeenCalledWith({blockHash: '0x0'}, undefined);
```
ahoppen, line 161: I'm not 100% sure whether the second parameter is supposed to be `undefined`.

packages/web3-core-method/tests/src/workflows/TransactionConfirmationWorkflowTest.js
```diff
@@ -195,34 +197,42 @@ describe('TransactionConfirmationWorkflowTest', () => {
...
+                expect(methodMock.callback).toHaveBeenCalledWith(false, {blockHash: '0x0'});
 
-            expect(newHeadsWatcherMock.stop).toHaveBeenCalled();
+                expect(newHeadsWatcherMock.stop).toHaveBeenCalled();
 
-            expect(methodMock.afterExecution).toHaveBeenCalledWith({blockHash: '0x0'});
+                expect(methodMock.afterExecution).toHaveBeenCalledWith({blockHash: '0x0'});
 
-            expect(transactionReceiptValidatorMock.validate).toHaveBeenCalledWith({blockHash: '0x0'});
+                expect(transactionReceiptValidatorMock.validate).toHaveBeenCalledWith({blockHash: '0x0'}, undefined);
```
ahoppen, line 224: I'm not 100% sure whether the second parameter is supposed to be `undefined`.

nivida:
> I noticed that the `PromiEvents` returned from sending a method to a smart contract is never resolved.

I've tested it and it resolves with the transaction receipt. The ``PromiEvent`` resolve method will be called in the ``TransactionConfirmationWorkflow`` when the transaction got enough confirmations. I get the transaction hash in the ``AbstractSendMethod`` and call the confirmation workflow with this hash. The ``TransactionConfirmationWorkflow`` is watching for the transaction receipt and waits until it got enough confirmations for resolving with the receipt.


>  The tests just wait for the `transactionHash` event to be fired and succeed without waiting for the `PromiEvent` to be resolved. Thus the issue was never covered in tests.
Fixing the second bug uncovers some failing tests that were previously not executed, which I fixed.

I've just tested it the tests are working they are failing when I change the expected value to a wrong one. I only had to move the ``await promiEvent`` in ``TransactionConfirmationWorkflowTest`` from line 198 to 222 as fix. 

I will update the ``TransactionConfirmationWorkflowTest`` in a new PR.

ahoppen:
> I've tested it and it resolves with the transaction receipt. The `PromiEvent` resolve method will be called in the `TransactionConfirmationWorkflow` when the transaction got enough confirmations.

Sorry, I somehow assumed that the promise would resolve in the same way the callback was called in the 0.x releases, i.e. when the transaction hash is available.

One issue still stands though: You can add `expect(true).toBe(false)` at the end of `'calls execute with wallets defined'` without breaking the test (see https://github.com/ahoppen/web3.js/blob/unexpected-test-pass/packages/web3-core-method/tests/src/methods/transaction/SendTransactionMethodTest.js#L111). That shouldn't be the case, right?

PR #2421, opened 2019-02-25
Check if protocol is defined.

### LINKED ISSUE #1965: websocket issues: connection not open on send()
I am running `web3@1.0.0-beta.36` on Kovan + Infura. 
The mention issue happens only sometimes and my feeling tells me it happens more often if you didn't run the application for a longer time and then start everything.

```
[nodemon] 1.18.3
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: *.*
[nodemon] starting `node API.js`
Listening on port 4000...
connection not open on send()
connection not open on send()
connection not open on send()
connection not open on send()
```

Here is my nodejs code:

```
const contractFunction = myContract.methods.updateLevel(Number(_myVal));
const functionBytes = contractFunction.encodeABI();

//construct the Tx data
const rawTx = {
	chainId: 42,
	gasLimit: web3.utils.toHex(5000000),
	to: contractAddress,
	from: addressFrom,
	data: functionBytes
	};

//sign & send Tx
web3.eth.accounts.signTransaction(rawTx, privateKey)
	.then(RLPencodedTx => {
		
	let executed = false;    //to avoid double callback
	myContractWebSocket.once('Events1',
		function(error, event) {
			if (!executed) {
				executed = true;
				if (!error) callback(event, null); 
				else callback(null, error);
			}
		}
	);
			
	myContractWebSocket.once('Event2',
		function(error, event) {
			if (!executed) {
				executed = true;
				if (error) callback(null, error);
				//If no error wait 1 sec in case Events1 is available, so it can go first
				else setTimeout(function() {callback(event, null);}, 1000);
			}
		}
	);

	//sending signed Tx
	web3.eth.sendSignedTransaction(RLPencodedTx['rawTransaction'])
	.on('error', error => { callback(null, error); });		
})
.catch('error', error => callback(null, error));
```

passabilities:
## Description

Without this check, if an undefined value is passed within Firefox, it fails to connect.

Fixes #1965 

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 

## Checklist:

- [ ] My changes generate no warnings.
- [ ] I ran ```npm run test``` in the root folder with success and extended the tests if necessary.

### DIFF
--- a/packages/web3-providers/src/factories/ProvidersModuleFactory.js
+++ b/packages/web3-providers/src/factories/ProvidersModuleFactory.js
@@ -145,7 +145,11 @@ export default class ProvidersModuleFactory {
 
             connection = new W3CWebsocket(url, options.protocol, null, headers, null, options.clientConfig);
         } else {
-            connection = new window.WebSocket(url, options.protocol);
+            const args = [url];
+            if (options.protocol) {
+                args.push(options.protocol);
+            }
+            connection = new window.WebSocket(...args);
         }
 
         return new WebsocketProvider(connection, options.timeout);


nivida:
First of all, thanks for opening this PR! I just checked the [WebSocket documentation](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket).

> Either a single protocol string or an array of protocol strings. These strings are used to indicate sub-protocols, so that a single server can implement multiple WebSocket sub-protocols (for example, you might want one server to be able to handle different types of interactions depending on the specified protocol). If you don't specify a protocol string, an empty string is assumed.

This means if nothing got passed e.g.: ``undefined`` it should use a empty string. I will test it but I think this would be the better solution:

``` javascript 
if (typeof options.protocol === 'undefined') {
  options.protocol = '';
}
connection = new window.WebSocket(url, options.protocol);
```

Can you tell me with which version of FF this issue has? latest?

passabilities:
Yes, this is happening on the latest FF `65.0.1`. It's especially weird because After doing some manual debugging, it appears to actually be defaulting to an empty string when it is undefined but still throws this error. However, when testing by not passing anything in the second argument it works perfectly. I will test if it works with an empty string.

nivida:
@passabilities Did you find the time to test it?:)

nivida:
This PR got closed because of the reorganisation of the branches. Feel free to open a new PR for the 2.x branch.

PR #2466, opened 2019-03-06
Transaction confirmation workflow improvements

### LINKED ISSUE #2418: `PromiEvent` event listener `on` `confirmations` not returning correct confirmation information
#### Expected behavior

When i hook onto `PromiEvent` event listener on confirmations i get the correct confirmation number at the correct time not just random count of confirmations number every second.

#### Actual behavior

When i hook onto `PromiEvent` event listener on confirmations it seems its not inline with when the next block has been mined, on Rinkeby it's every 15 seconds but on web3 it returns the confirmation every second which is not correct.

```
web3.eth.sendSignedTransaction(rawTransaction)
.on(
            'confirmation',
            (confirmationNumber: number, receipt: TransactionReceipt) => {
              // fires new confirmation number up to 24 every second
            },
          )
```

It also states in the documentation that it stops the event listener firing after 12 confirmations yet it goes up to 24. 

This should be in line with the proper confirmations number and fire correctly and not just return a fake confirmation number every second.

#### Versions
1.0.0-beta.46

nivida:
## Description

This PR will fix the confirmation workflow which does not correctly return the confirmations count over the ``HttpProvider``. I will also do some other improvements regarding the transaction method handling.

Fixes #2418

## Type of change

- [x] Bug fix 
- [x] Enhancement

## Checklist:


packages/web3-eth-accounts/src/models/Account.js
```diff
@@ -102,7 +102,7 @@ export default class Account {
      * @returns {EncryptedKeystoreV3Json | {version, id, address, crypto}}
      */
     encrypt(password, options) {
-        return Account.fromPrivateKey(this.privateKey, this.accounts.transactionSinger).toV3Keystore(password, options);
+        return Account.fromPrivateKey(this.privateKey, this.accounts.transactionSigner).toV3Keystore(password, options);
```
bwasty, line 105: This method still fails for me when I fix the typo, because `this.accounts` is `undefined` when created via `web3.eth.accounts.create()` or `web3.eth.accounts.wallet.create(1)`. I solved it by defaulting `this.accounts` to `{}` in the constructor.

packages/web3-core-method/lib/methods/AbstractMethod.js
```diff
@@ -23,21 +23,21 @@
...
+     * @param {AbstractWeb3Module} moduleInstance
      *
      * @constructor
      */
-    constructor(rpcMethod, parametersAmount, utils, formatters) {
+    constructor(rpcMethod, parametersAmount, utils, formatters, moduleInstance) {
         this.utils = utils;
         this.formatters = formatters;
-        this.promiEvent = new PromiEvent();
+        this.moduleInstance = moduleInstance;
```
OFRBG, line 40: Perhaps it's not my place to ask, but wouldn't this add a copy of the module to all methods? I haven't tested this branch, but instead of copying the whole module instance, perhaps modules could have a concrete `moduleInstance.config`? (I don't remember the full structure of the package, but this might also avoid circular references somewhere unexpected.)

 Maybe this is preemptive optimization, but I thought I'd put it out there.

nivida: The garbage collector of JS will collect and remove the ``AbstractMethod`` after the execution of the method. The ``AbstractMethod`` classes will be created in the ``MethodProxy`` when a declared method gets called. It will not copy the ``moduleInstance`` each time. It will just set the reference of the pointer. This means ``this.moduleInstance`` is just a "symlink" to the actual object.

OFRBG: I see. I wasn't sure if the `moduleInstance` would end up as a pointer or as a copy. It only takes one random line referencing the object to end up allocated till unmounting.

OFRBG: Bildschirmfoto 2019-03-11 um 22 05 04

Maybe it works directly with the ES6 structure, but the `.umd.js` files don't seem to be correctly collected. Maybe for this release working on the `.umd.js` transpiler settings would be beneficial. The image above is taken from the (semi-)production build of the `getBlock` example with webpack. https://codesandbox.io/s/0ykmkv2380

OFRBG: This is the actual production build heap snapshot after some digging. I've tested the `.umd.js` every now and then and its behavior strays quite a bit from its `.esm.js` counterpart.

Bildschirmfoto 2019-03-11 um 22 43 14

nivida: There are still some points I could optimize the memory usage and other stuff. But I will do this later because I've already broken the heap snapshot down from ~550mb to ~280mb. Thanks for your tests! :-)

OFRBG:
Sorry if I came across as pedantic. What worries me overall is that the
transpiling to browser JS might end up having unexpected behavior. Every
now and then it happens to me that the ES6 or CJS code works but the
browser code doesn't. Probably not a pressing issue, but something I'd
avoid neglecting since the package is widely used client-side, but the
tests don't strictly cover it.

Samuel Furter <notifications@github.com> schrieb am mar, 12 de marzo de
2019 um 10:32:

packages/web3-core-method/lib/methods/AbstractMethod.js
OFRBG: Sorry if I came across as pedantic. What worries me overall is that the
transpiling to browser JS might end up having unexpected behavior. Every
now and then it happens to me that the ES6 or CJS code works but the
browser code doesn't. Probably not a pressing issue, but something I'd
avoid neglecting since the package is widely used client-side, but the
tests don't strictly cover it.

nivida:
Merged with a failed build because ``dtslint`` is just failing on travis and it works locally with the NodeJS versions we have configured on travis.

PR #2478, opened 2019-03-08
AbiInput and AbiOutput interface update

teef:
AbiInput and AbiOutput interface update

## Description
AbiInput and AbiOutput interfaces supplemented with optional components property.
AbiOutput typo fixed.

## Type of change

- [x] Bug fix
- [x] Enhancement

## Checklist:


joshstevens19 (approved):
Good spot on the bad spelling my mistake ha. Thanks for this PR looks good to me.

joshstevens19 (changes requested):
oops sorry no it is not - can you fix the tests for the types please. @teef

teef:
Hi @joshstevens19 . I have fixed typings tests and also extended AbiItem in `json-interface-method-to-string-test.ts` file to match current definition.
Static property from Web3 class was removed. To implement utils as a static property it is required to modify Web3 class in Web3.js file. I will provide my proposal in next PR.

nivida:
@teef The static property got removed because there is also a non-static property ``utils`` and I'm recommending to use the web3-utils module directly if you just want to use the utility methods.

teef:
Hi @nivida.
Thanks for explanation regarding utils.

At the moment I see that utilities originate from `web3-utils`. This module aggregates all of them and expose as named exports. Furthermore utilities are imported in `node_modules\web3\src\Web3.js` file and used in Web3 class constructor. This means that each and every Web3 instance will have entire `utils` package attached. Apart of a fact that this may not be super efficient in terms of resource utilization, I thought that having this accessible as static property is closer to OOP paradigm.

Regardless of the decision I saw some inconsistency between Utils TS definition, documentation and implementation. There are some methods missing in current `utils` module. Please find more details below. Description originates from this PR: https://github.com/Meelabs/web3.js/pull/1

Changes: https://github.com/Meelabs/web3.js/pull/1/commits/ae012bb384d7290b6cec3ee2956d4623cde3834c

Missing utilities added to utils property of Web3 class:

`isBigNumber()`: added
`leftPad()`: added for compatibility with TS Interface
`rightPad()`: added for compatibility with TS Interface
`unitMap()`: added

Known issues:

`testAddress()`: defined but implementation is missing
`testTopic()`: defined but implementation is missing
both methods do not exist in documentation of version 1.0.

I would grateful if you can put some comment on that.

Additionally let me know if the PR can be merged?

nivida:
@teef Thanks for your feedback! 

Yes, that's true the ``utils`` were available as static property over the Web3 class. I'm not a big fan of static properties because it's always hard to figure out in a bigger architecture where exactly the static methods and properties are used. The only reason I will use them is for factory methods of a VO (e.g.: ``Transaction.from(...)``) 
With the removing of the static property ``utils`` do I enforce the developers to use the utils module directly which will be better for the bundle size. The long-term goal is to create the [``Web3 Namespace``](https://gist.github.com/nivida/c7c24723353c243e295dbff954d06b50) and the [``Web3 Types Module``](https://gist.github.com/nivida/d75c26daea14607f0c57b42582b34f78) which will remove the umbrella class and replace the utils module with clear VO's. 

Feel free to give your feedback in the gist files!

nivida (approved):

PR #2474, opened 2019-03-08
add input to transaction type

bradennapier:
## Description

Adds `input` to `Transaction` type for the Typescript types.

Fixes #2473 

## Type of change

- [x] Bug fix 

## Checklist:

- [x ] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no warnings.
- [x] I have updated or added types for all modules I've changed
- [ ] Any dependent changes have been merged and published in downstream modules.
- [x] I ran ```npm run test``` in the root folder with success and extended the tests if necessary.
- [x] I ran ```npm run build``` in the root folder and tested it in the browser and with node.
- [x] I ran ```npm run dtslint``` in the root folder and tested that all my types are correct
- [ ] I have tested my code on an ethereum test network.

PR #2479, opened 2019-03-08
 Separate Transaction and TransactionConfig in typings

vshab:
## Description

Currently Transaction has all field optional what is incorrect. This causes senseless checks when using transaction:
```ts
eth.getTransaction(transactionHash).then((transaction: Transaction) => {
  if (transaction.hash) {
    transaction.hash //...
  }

  if (transaction.nonce) {
    transaction.nonce //...
  }
});
```

I think the Transaction and TransactionConfig should be different types, as it has been done before: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/web3/eth/index.d.ts (See Transaction and Tx) I change names though and corrected types a bit. This also should fix #2474

I also updated RLPEncodedTransaction type with respect to https://web3js.readthedocs.io/en/1.0/web3-eth-personal.html#signtransaction but I'm not 100% sure about this. @nivida could you take a closer look please?

## Type of change

- [x] Bug fix 
- [x] Breaking change
- [x] Enhancement

## Checklist:


nivida:
Great idea! I'll check it if we probably forgot something somewhere and will merge it later.

joshstevens19:
@vshab awesome idea! types look spot on to me. Thanks for the PR! great work.

joshstevens19 (approved):

PR #2529, opened 2019-03-19
Use genuine BN type

### LINKED ISSUE #2350: [types][web3-utils] Use bn.js directly for class BN instead of wrapper
#### Expected behavior
Let `BN` denote the class as exported from `web3-utils`, not `bn.js`.
Assigning results of `BN` methods, like `add()`, to variables of type `BN` should work.

#### Actual behavior
Doesn't compile, throws typescript errors.

#### Steps to reproduce the behavior with error logs
The code 
```ts
import { BN } from "web3-utils";
let a = toBN(0); // or new BN(0)
let b: BN;
b = a.add(toBN(1));
```
gives error
```
error TS2741: Property 'super' is missing in type 'BN' but required in type 'BN'.

71 b = a.add(toBN(1));
   ~

  node_modules/web3-utils/types/index.d.ts:73:5
    73     super(number: number | string | number[] | Buffer | BN, base?: number | 'hex', endian?: 'le' | 'be'): BigNumber;
           ~~~~~
    'super' is declared here.
```
So right now I do arithmetics with `a.add(toBN(1)) as BN` hacks.

Now I could instead do `import BN = require('bn.js');` but then I get another error when passing those `BN`s into, e.g., `web3.utils.soliditySha3`:
```
error TS2345: Argument of type 'BN' is not assignable to parameter of type 'Mixed'.
  Property 'super' is missing in type 'BN' but required in type 'BN'.
```

#### Background
The web3-utils type declarations internally first do
```ts
import BigNumber = require('bn.js');
```
(which in itself is confusing, because `bn.js` is usually imported as `BN` and `BigNumber` is used for `bignumber.js`) and then declares and exports a wrapper class
```ts
export class BN extends BigNumber {
    constructor(number: number | string | number[] | Buffer | BN, base?: number | 'hex', endian?: 'le' | 'be');
    super(number: number | string | number[] | Buffer | BN, base?: number | 'hex', endian?: 'le' | 'be'): BigNumber;
}
```
Why is that? Why doesn't web3 just use `bn.js` directly?

This whole `BN` situation is a bit confusing, see also  #2171
 
#### Versions
Tested with node `v11.9.0` web3 version `1.0.0-beta.43` and typescript `3.3.1` in a truffle `v5.0.2` project.

vshab:
## Description

Don't wrap BN but use type from BN library to improve compatibility with BN operations. Without this change this code is invalid:
```ts
const bigNumber = new BN(3);

fromWei(bigNumber.add(new BN(100)));
```

Fixes #2350

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/packages/web3-core/types/index.d.ts
+++ b/packages/web3-core/types/index.d.ts
@@ -18,6 +18,7 @@
  */
 
 import * as net from 'net';
+import BN = require('bn.js');
 import {AbstractMethodFactory} from 'web3-core-method';
 import {
     BatchRequest,
@@ -29,7 +30,6 @@ import {
     WebsocketProvider,
     WebsocketProviderOptions
 } from 'web3-providers';
-import {BN} from 'web3-utils';
 
 export class AbstractWeb3Module {
     constructor(
--- a/packages/web3-utils/types/index.d.ts
+++ b/packages/web3-utils/types/index.d.ts
@@ -214,11 +214,11 @@ export interface AbiInput {
     name: string;
     type: string;
     indexed?: boolean;
-    components?: AbiInput[];
+	components?: AbiInput[];
 }
 
 export interface AbiOutput {
     name: string;
     type: string;
-    components?: AbiOutput[];
+	components?: AbiOutput[];
 }
--- a/packages/web3-utils/types/index.d.ts
+++ b/packages/web3-utils/types/index.d.ts
@@ -20,7 +20,7 @@
  * @date 2018
  */
 
-import BigNumber = require('bn.js');
+import BN = require('bn.js');
 import {Buffer} from 'buffer';
 
 export type Unit =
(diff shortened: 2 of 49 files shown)


packages/web3-core/types/index.d.ts
```diff
@@ -18,6 +18,7 @@
  */
 
 import * as net from 'net';
+import BN = require('bn.js');
```
nivida, line 21: Could you change it to: ``import BN from 'bn.js'``

packages/web3-utils/types/index.d.ts
```diff
@@ -20,7 +20,7 @@
  * @date 2018
  */
 
-import BigNumber = require('bn.js');
+import BN = require('bn.js');
```
nivida, line 23: Could you change it to: ``import BN from 'bn.js'``

nivida: Could you do this in all files? :)

nivida (changes requested):

packages/web3-core/types/index.d.ts
vshab: Hello!

I can do it, no problem. But according to TypeScript [handbook](https://www.typescriptlang.org/docs/handbook/modules.html) 
>When exporting a module using export =, TypeScript-specific import module = require("module") must be used to import the module.

BN is exported using `export =`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bn.js/index.d.ts#L584

What do you think?

nivida: Great, thanks for referencing the documentation! Yes, then we will use your way ;-)

nivida (approved):

vshab:
@nivida CI fixed

PR #2548, opened 2019-03-21
Fix issue with boolean typed receipt status

angus-hamill:
## Description
The `status` field of the transaction receipt is sometimes returned as a boolean, not hex. This was causing web3 to think the transaction reverted when it had actually succeeded.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change
- [ ] Enhancement

## Checklist:

### DIFF
--- a/packages/web3-core-method/lib/methods/transaction/AbstractObservedTransactionMethod.js
+++ b/packages/web3-core-method/lib/methods/transaction/AbstractObservedTransactionMethod.js
@@ -172,6 +172,9 @@ export default class AbstractObservedTransactionMethod extends AbstractMethod {
      * @returns {Boolean}
      */
     hasRevertReceiptStatus(receipt) {
+        if (typeof receipt.status === 'boolean') {
+            return !receipt.status;
+        }
         return Boolean(parseInt(receipt.status)) === false && receipt.status !== undefined && receipt.status !== null;
     }
 }
--- a/packages/web3-core-method/tests/lib/methods/transaction/AbstractObservedTransactionMethodTest.js
+++ b/packages/web3-core-method/tests/lib/methods/transaction/AbstractObservedTransactionMethodTest.js
@@ -134,6 +134,24 @@ describe('AbstractObservedTransactionMethodTest', () => {
         expect(afterExecutionMock).toHaveBeenCalledWith({status: '0x1'});
     });
 
+    it('calls execute and returns with the expected resolved Promise when the status is a boolean', async () => {
+        providerMock.send.mockReturnValueOnce(Promise.resolve('transactionHash'));
+
+        observableMock.subscribe = jest.fn((next, error, complete) => {
+            next({count: 0, receipt: {status: true}});
+
+            complete();
+        });
+
+        await expect(method.execute()).resolves.toEqual({status: true});
+
+        expect(providerMock.send).toHaveBeenCalledWith('rpcMethod', []);
+
+        expect(beforeExecutionMock).toHaveBeenCalledWith(moduleInstanceMock);
+
+        expect(afterExecutionMock).toHaveBeenCalledWith({status: true});
+    });
+
     it('calls execute and returns with the expected rejected Promise', async () => {
         providerMock.send.mockReturnValueOnce(Promise.resolve('transactionHash'));
 


chris-shyft:
awesome, some of my team was having this issue in certain system configurations

nivida:
This isn't an issue of Web3.js. Ganache has to be aligned with the JSON-RPC spec.

angus-hamill:
> This isn't an issue of Web3.js. Ganache has to be aligned with the JSON-RPC spec.

@nivida 
https://github.com/ethereum/web3.js/blob/1.0/packages/web3-core-helpers/src/Formatters.js#L282
I believe the ganache issues are due to the web3 formatter parsing the result to a bool, here. Should this be changed?

angus-hamill:
@nivida 
I was mistaken in thinking this was a ganache only issue. While ganache does emit that status field as a boolean sometimes (which should be fixed) there is also a web3js issue.



As you see [here](https://github.com/ethereum/web3.js/blob/1.0/packages/web3-core-method/src/observers/TransactionObserver.js#L149) the observer is calling the `GetTransactionReceiptMethod` which calls a formatter as part of its execution [here](https://github.com/ethereum/web3.js/blob/1.0/packages/web3-core-method/src/methods/transaction/GetTransactionReceiptMethod.js#L48).

This casts the status field to a boolean [here](https://github.com/ethereum/web3.js/blob/1.0/packages/web3-core-helpers/src/Formatters.js#L282).

I guess my fix is still too hacky for this though, what do you propose? Should the transaction receipt formatter still be casting to a bool like this?

dileepfrog:
I wouldn't consider this hacky, but rather robustifying the receipt status check to handle varying upstream formatting with consistent semantics. Just my $0.02

SwissArmyBud:
@nivida This error is thrown when using Web3 against Geth 1.8 - no Ganache anywhere in the ecosystem. Here's my receipt from a Geth IPC connection in Web3:
```
{ blockHash:
   '0x7edf444c9920e6bcdbbc6597d8089d0f454a162870e4e204f1e20db5b8fcc0d1',
  blockNumber: 4,
  contractAddress: '0x88EbAB3EFe955D427D817D50a1C9444aC9DdB7ab',
  cumulativeGasUsed: 569523,
  from: '0xf6f63d9596ed1d10cf6191fc448ad4af980e0222',
  gasUsed: 569523,
  logs: [],
  logsBloom:
   '0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000',
  status: true,
  to: null,
  transactionHash:
   '0x7081793c182edd6100bc67b4f51941c7847cdeb26bf6cc53a1b57a7fc0305c0f',
  transactionIndex: 0 }
```

And a Node.js REPL can demonstrate the error:
```
> var rc = new Object({status: true})
{ status: true }
> rc.status
true
> parseInt(rc.status)
NaN
> Boolean(parseInt(rc.status))
false
```

The parsing scheme is located in [two places in Web3.js](https://github.com/search?l=&q=Boolean+parseInt+receipt.status+repo%3Aethereum%2Fweb3.js&type=Code) and prevents legitimately completed transactions being parsed without a large error handling block:
```
let receipt = await web3Contract.deploy({ data: contract.bytecode })
                      .send( contract.options )
                      .on('transactionHash', (tx) => {
                        console.log("[DEPLOY] -> Tx Hash: " + tx)
                      }).catch(e => {
                        if( e.toString().indexOf("reverted by the EVM:") == -1 ){
                          return undefined;
                        } else {
                          var er = e.toString();
                          try{
                            var jsonIndex = {
                              start: er.indexOf( "{", er.indexOf("EVM:")),
                              end: er.indexOf( "}", er.indexOf("logsBloom")) + 1
                            }
                            er = JSON.parse( er.substring( jsonIndex.start, jsonIndex.end ) );
                            return er;
                          } catch (error) {
                            console.log("[DEPLOY] -> WARNING - WEB3 REPORTS EVM ROLLBACK, AND BAD RECEIPT PARSE...");
                            console.log(er);
                            console.log(error);
                            // Our error parsing has failed - no guaranteed state, return undefined
                            return undefined;
                          }
                        }
                      });
```

angus-hamill:
@SwissArmyBud 
The status field was always (or at least for a long time) returned as a boolean when you queried it directly to the node with web3js.
The problem is it's now using this get transaction receipt method with the formatter internally as of #2466 (in the place I highlighted in my comment above)

Issue #2539, opened 2019-03-20
Subscribe to pendingTransactions fails while using IPC Provider


MaxFriedman17:
### Description 
Subscribing to pending transactions gives an error and exits the program while **using IPC connection.** 

#### Expected behavior
Subscribing to new pending transaction (web3.eth.subscribe('pendingTransactions')) is expected to return a json response with a txhash of new transaction.

#### Actual behavior
It returns a json response but it terminates the execution of the program with an error (see in Error logs section the actual error message).

I actually tried to debug the code and it seems that it fails here:

```
if (!isObject(response)) {
        response = JSON.parse(response);
      }
```

while trying to parse the JSON response that comes from the IPC pipe. It seems that in the variable response it appears not only the JSON response of the pending transaction but also the first response that gives the subscription ID when starting the subscription to the node so it fails to parse that object to json.
(This is my interpretation, it can be another thing but hope this helps to fix the problem but if it is this, maybe other subscription types -like block headers- will fail as well).

#### Steps to reproduce the behavior

1. Import web3
2. Set IPC Provider
3. Subscribe to new pending transactions.

#### Error Logs 
SyntaxError: Unexpected token { in JSON at position 71
    at JSON.parse (<anonymous>)
    at IpcProvider.onMessage (/node_modules/web3-providers/dist/web3-providers.cjs.js:285:25)
    at IpcProvider.onMessage (/node_modules/web3-providers/dist/web3-providers.cjs.js:551:71)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:250:11)
    at Socket.Readable.push (_stream_readable.js:208:10)
    at Pipe.onread (net.js:607:20)

### Versions 
* web3.js: 1.0.0-beta.50
* nodejs: v 8.10.0
* browser: -
* ethereum node: geth v1.8.23-stable-c9427004

Regards!

nivida:
Thanks for opening this issue! I will fix and release it asap.

PR #2549, opened 2019-03-21
Fix issue decoding log with only indexed parameters

angus-hamill:
## Description
If a log has only indexed parameters, the data field may still be present as `0x`. The `AbiCoder` was attempting to decode this and erroring.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change
- [ ] Enhancement

## Checklist:

### DIFF
--- a/packages/web3-eth-contract/src/decoders/EventLogDecoder.js
+++ b/packages/web3-eth-contract/src/decoders/EventLogDecoder.js
@@ -47,6 +47,10 @@ export default class EventLogDecoder {
             argTopics = response.topics.slice(1);
         }
 
+        if (response.data === '0x') {
+            response.data = null;
+        }
+
         response.returnValues = this.abiCoder.decodeLog(abiItemModel.getInputs(), response.data, argTopics);
         response.event = abiItemModel.name;
         response.signature = abiItemModel.signature;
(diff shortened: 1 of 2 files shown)


packages/web3-eth-abi/src/AbiCoder.js
```diff
@@ -229,7 +229,7 @@ export default class AbiCoder {
             nonIndexedInputItems.push(input);
         });
 
-        if (data) {
+        if (data && data !== '0x') {
```
nivida: Could you handle the '0x' case here: https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth-contract/src/decoders/EventLogDecoder.js#L50 

Thanks for the PR!

nivida (changes requested):

packages/web3-eth-abi/src/AbiCoder.js
angus-hamill: Would something like
`if (response.data === '0x') delete response.data` suffice?
Is that what you meant?

nivida: I would do it this way:
``` javascript 
if (response.data === '0x') {
reponse.data = null;
}
```

angus-hamill: Done 👍

nivida (approved):

nivida:
@angus-hamill FYI: I'll release this fix when I've closed the issue #2539

PR #2585, opened 2019-03-26
eth.subscribe types improvement

### LINKED ISSUE #2583: web3.eth.subscribe("logs", ..) types issue
### Description 
I cannot subscribe to logs, because of typescript errors.
If I comment out the web3-eth/types' index.d.ts:58-60 rows, it works. Should I set some attribute in the tslint.json?
Also I cannot pass array for the address, but the documentation says It's possible.

#### Expected behavior

```
this.web3.eth.subscribe("logs", {
    address: addressesToWatch
}, (error, event: Log) => {
    ...
});
```
And I've got the event about the specified addresses.
#### Actual behavior

cannot pass array to the address parameter, only string
also tslint prints: "Argument of type "'logs'" is not assignable to parameter of type '"pendigTransactions'"."

#### Error Logs 

"Argument of type "'logs'" is not assignable to parameter of type '"pendigTransactions'"."
Type 'string[]' is not assignable to type 'string'

### Versions 
* web3.js: 1.0.0-beta.50
* nodejs: 10.11.0
* ethereum node: geth 1.8.22-stable

nivida:
## Description

Fixes #2583 

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/packages/web3-eth/types/index.d.ts
+++ b/packages/web3-eth/types/index.d.ts
@@ -58,7 +58,11 @@ export class Eth extends AbstractWeb3Module {
     subscribe(type: 'syncing', options?: null, callback?: (error: Error, result: Syncing) => void): Subscription<Syncing>;
     subscribe(type: 'newBlockHeaders', options?: null, callback?: (error: Error, blockHeader: BlockHeader) => void): Subscription<BlockHeader>;
     subscribe(type: 'pendingTransactions', options?: null, callback?: (error: Error, transactionHash: string) => void): Subscription<string>;
-    subscribe(type: 'pendingTransactions' | 'logs' | 'syncing' | 'newBlockHeaders', options?: any, callback?: (error: Error, result: any) => void): Subscription<Log | BlockHeader | Syncing | string>;
+    subscribe(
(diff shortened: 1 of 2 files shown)


packages/web3-eth/types/index.d.ts
```diff
@@ -54,10 +54,11 @@ export class Eth extends AbstractWeb3Module {
 
     clearSubscriptions(): Promise<boolean>;
 
-    subscribe(type: 'logs', options?: Logs, callback?: (error: Error, log: Log) => void): Subscription<Log>;
-    subscribe(type: 'syncing', options?: null, callback?: (error: Error, result: any) => void): Subscription<any>;
-    subscribe(type: 'newBlockHeaders', options?: null, callback?: (error: Error, blockHeader: BlockHeader) => void): Subscription<BlockHeader>;
-    subscribe(type: 'pendingTransactions', options?: null, callback?: (error: Error, transactionHash: string) => void): Subscription<string>;
+    subscribe(
```
joshstevens19, line 61: Making this generic actually loses a lot of good typings for these methods. It before use to map a proper typed single response depending on your type string you supply which is the best kinds of types., the less generic in typings like this the better it is. This will make calling this method super hard in TypeScript world, near impossible to get proper compile time errors.

Issue #2499, opened 2019-03-13
Not possible to call methods on existing contract


russanto:
### Description
It is not possible to call function on already deployed contracts.

#### Expected behavior
Creating a contract instance with jsonInterface and address I should be able to call the function on that contract

#### Actual behavior
Contract instance isn't binded to the contract at that address. Instead if I deploy the contract in the same execution just before calling methods, it works perfectly.

#### Steps to reproduce the behavior

    const Web3 = require('web3');
    const web3 = new Web3("ws://127.0.0.1:8546");
    var simpleContract = new web3.eth.Contract(jsonInterface, address);
    simpleContract.methods.myFunction(params).call().then( do something );

#### Error Logs 
Error: Invalid bytes string given: 0x
    at AbiCoder.decodeParameters (./node_modules/web3-eth-abi/dist/web3-eth-abi.cjs.js:64:15)
    at AbiCoder.decodeParameter (.node_modules/web3-eth-abi/dist/web3-eth-abi.cjs.js:55:19)
    at CallContractMethod.afterExecution (./node_modules/web3-eth-contract/dist/web3-eth-contract.cjs.js:743:28)
    at CallContractMethod._callee$ (./node_modules/web3-core-method/dist/web3-core-method.cjs.js:354:35)
    at tryCatch (./node_modules/regenerator-runtime/runtime.js:62:40)
    at Generator.invoke [as _invoke] (./node_modules/regenerator-runtime/runtime.js:288:22)
    at Generator.prototype.(anonymous function) [as next] (./node_modules/regenerator-runtime/runtime.js:114:21)
    at asyncGeneratorStep (./node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
    at _next (./node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)

### Versions 
* web3.js: 1.0.0-beta.48
* nodejs: v8.9.4
* ethereum node: geth 1.8.22

nivida:
Could you add the ABI and the bytecode of this contract for testing it?

russanto:
ABI
[...]

Issue #2444, opened 2019-03-03
web3.eth.Contract can't parse ABI correctly [web.1.0.0-beta.47]


lebed2045:
Simple nodejs program which calls totalSupply() from Dai smartcontract
```
const Web3 = require('web3');
const rpcURL = 'https://mainnet.infura.io/v3/<your_key>';
const web3 = new Web3(rpcURL);


const daiSmartcontractAddress = "0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359";
[...]

PR #2608, opened 2019-03-29
Invalid bytes string error fixed 

### LINKED ISSUE #2527: Getting Error: Invalid bytes string given: 0x [web3@1.0.0-beta.48] on simple Solidity view function
*I've looked into the other issues related to this error, but this is caused by something different

### Description

Using `web3@1.0.0-beta.48` I have the following contract function: 

```
pragma solidity ^0.5.2;

contract Versioned {
    string[] public data; // when deployed, data already has 1 item

    function getVersionIndex() public view returns (uint count) {
        return data.length - 1;
    }
}
```

The Web3 code is: 

```
  const versionIndex = await contract.getVersionIndex();

  // ...
  public getVersionIndex(): Promise<any> {
    return this.contract.methods.getVersionIndex().call(); 
  }
```

Contract class already has an address and the method does not require any params. 

I've tried adding the `{ from: "0xadress" }`, but it makes no difference. 

#### Expected behavior

Return the value of the Solidity `data.length` as `uint count`. 

As a note, I've debugged the contract using https://remix.ethereum.org/ and the `getVersionIndex()` function works as expected.

#### Error Logs 

The error is originating at `AbiCoder.decodeParameters` of `web3-eth-abi.umd.js:65`: 

```
key: "decodeParameters",
      value: function decodeParameters(outputs, bytes) {

        if (!bytes || bytes === '0x' || bytes === '0X') {
          throw new Error("Invalid bytes string given: ".concat(bytes));
        }
```

### Versions 
* web3.js: web3@1.0.0-beta.48
* nodejs: v10.15.3
* browser: Chrome Version 72.0.3626.121 (Official Build) (64-bit) MaxOSX
* ethereum node: Ropsten test network with Metamask

nivida:
## Description

If the response is '0x' then it will return 'null'.

Fixes #2527 #2499 #2444

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/packages/web3-eth-contract/src/methods/CallContractMethod.js
+++ b/packages/web3-eth-contract/src/methods/CallContractMethod.js
@@ -49,7 +49,7 @@ export default class CallContractMethod extends CallMethod {
      * @returns {Array|String}
      */
     afterExecution(response) {
-        if (!response) {
+        if (!response || response === '0x') {
             return null;
         }
 
--- a/packages/web3-eth-contract/tests/src/methods/CallContractMethodTest.js
+++ b/packages/web3-eth-contract/tests/src/methods/CallContractMethodTest.js
@@ -61,4 +61,12 @@ describe('CallContractMethodTest', () => {
 
         expect(abiCoderMock.decodeParameter).toHaveBeenCalledWith('bytes', '0x0');
     });
+
+    it('calls afterExecution and response is empty', () => {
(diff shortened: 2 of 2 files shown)


MacdaraD:
I am still getting this error running on a local node with the most up to date versions of geth, nodejs and web3. I understand that if there's any hope of you solving my problem you'll need more information than that so let me know what you may need and I'll supply it (provided you have time to answer me).

PR #2655, opened 2019-04-05
getPendingTransactions implemented in web3-core-methods

princesinha19:
## Description
Inside transaction in web3-core-methods pending transactions was missing. So, I implemented that method.
eth.getPendingTransactions will give pending transaction array.

## Type of change
- [x] New feature 

## Checklist:


packages/web3-core-method/src/methods/transaction/GetPendingTransactionMethod.js
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * @method afterExecution
+     *
+     * @param {Object} response
+     *
+     * @returns {Object}
+     */
+    afterExecution(response) {
+        if (response.length === 0) {
+            return 'No Pending Transactions !!';
```
OFRBG: I don't think adding a message for 0 pending transactions is efficient. The return object should just be the response itself.

princesinha19: Yes, It could be, I thought that if there will be 0 pending transaction then, it should return a message. Otherwise, it will return an empty array **[ ]**.
@nivida can you comment on this.

joshstevens19: Returning a empty array is the correct response here. If there is 0 pending transactions then the result is [], it doesn’t matter if it’s empty or not that is the result. Also making other apps have to compare the response against a string is never a good idea.

princesinha19: Agreed @OFRBG & @joshstevens19, Done.

packages/web3-core-method/src/methods/transaction/GetPendingTransactionMethod.js
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * This method will be executed after the RPC request.
+     *
+     * @method afterExecution
+     *
+     * @param {Object} response
+     *
+     * @returns {Object}
+     */
+    afterExecution(response) {
```
nivida: You are able to remove the ``afterExecution`` because the ``AbstractMethod`` class will pass the response through.

nivida (changes requested):

packages/web3-core-method/src/methods/transaction/GetPendingTransactionMethod.js
princesinha19: okay, thanks. I will remove it.

princesinha19: Done @nivida.

packages/web3-core-method/src/methods/transaction/GetPendingTransactionMethod.js
```diff
@@ -0,0 +1,36 @@
...
+export default class GetPendingTransactionMethod extends AbstractMethod {
+    /**
+     * @param {Utils} utils
+     * @param {Object} formatters
+     * @param {AbstractWeb3Module} moduleInstance
+     *
+     * @constructor
+     */
+    constructor(utils, formatters, moduleInstance) {
+        super('eth_pendingTransactions', 0, utils, formatters, moduleInstance);
```
nivida: Where did you see this JSON-RPC method documented? I could only find the [parity](https://wiki.parity.io/JSONRPC-parity-module#parity_pendingtransactions) version of it. @princesinha19

princesinha19: @nivida There is no any documentation available for eth_pendingTransactions. But, I tested on Geth, and it was working.
I think we should ask the Geth team to add documentation for it.

princesinha19: @nivida It seems that this method is not supported by Ganache. So, I created an issue in their repo.
https://github.com/trufflesuite/ganache-cli/issues/638

princesinha19: I also updated the JSON_RPC wiki.
https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_pendingtransactions

packages/web3-core-method/src/methods/transaction/GetPendingTransactionMethod.js
```diff
@@ -0,0 +1,36 @@
...
+*/
+/**
+ * @file GetPendingTransactionMethod.js
+ * @author Prince Sinha <sinhaprince013@gmail.com>
+ * @date 2018
+ */
+
+import AbstractMethod from '../../../lib/methods/AbstractMethod';
+
+export default class GetPendingTransactionMethod extends AbstractMethod {
```
nivida: I would rename this method to ``PendingTransactionsMethod`` because the RPC method is called ``eth_pendingTransactions``.

nivida: I will merge it after the renaming. Great work! thanks! :)

Edit: 
Could you extend the Eth types too?

nivida (changes requested):

packages/web3-core-method/src/methods/transaction/GetPendingTransactionMethod.js
princesinha19: okay, cool @nivida. I will do that.

nivida (approved):

PR #2660, opened 2019-04-07
Geth "Miner and DAG" management API's implemented (web3-eth-miner)

### LINKED ISSUE #2637: Add all management API methods of Geth
### Description 

Implement all methods which are listed [here](https://github.com/ethereum/go-ethereum/wiki/Management-APIs).

#### Expected behavior

I can, for example, call the debug methods.

#### Actual behavior

Not all methods of the management API are supported.

### Versions 
* web3.js: latest

princesinha19:
## Description

I implemented all miner management API's provided by geth.
https://github.com/ethereum/go-ethereum/wiki/Management-APIs#list-of-management-apis

Fixes #2637 Partially

## Type of change

- [x] New feature 

## Checklist:

- [ ] I ran ```npm run test``` in the root folder with success and extended the tests if necessary.


packages/web3-eth-miner/types/index.d.ts
```diff
@@ -0,0 +1,53 @@
...
+        miningThread: Hex,
+        callback?: (error: Error, result: boolean) => void
+    ): Promise<boolean>;
+
+    stopMining(
+        callback?: (error: Error, result: boolean) => void
+    ): Promise<boolean>;
+}
+
+export type Hex = string | number;
```
joshstevens19: if your going to reuse this `hex` all the time like you have please put it in `web3-core`, personally i think just do the type as `string | number`. I think its overkill creating a custom type and the rest of lib except for utils doesn't do it like this. Just my view :+1:

joshstevens19:
If this has to get to through web3.eth you have to add this to the Eth typings class or it wont be exposed in the global scope when ts users try to use it.

https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth/types/index.d.ts

Please add it

packages/web3-eth-miner/types/index.d.ts
princesinha19: cool.

princesinha19: Done @joshstevens19.

princesinha19:
> If this has to get to through web3.eth you have to add this to the Eth typings class or it wont be exposed in the global scope when ts users try to use it.
> 
> https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth/types/index.d.ts
> 
> Please add it

Why it is so, web3-eth-miner is an independent module and it has it's own **typing** class. So, what's the need to check type in web3-eth. This is same for web3-eth-personal and all other modules. I think this dependency should be removed. Correct me if I am wrong @joshstevens19.
@nivida

joshstevens19:
@princesinha19 @nivida that's not true. If you want to get to `web3.eth.personal` through the main web3 class - https://web3js.readthedocs.io/en/1.0/web3-eth.html#personal

In code you have to write `web3.eth.personal` if `personal` is not on the `web3.eth` typing's you can never get to it https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth/types/index.d.ts#L50

I am not saying check types in `web3.eth` just exposing them as the above link shows ^^ BUT I think all your PRs are standalone modules the user should only be able to get to it if they import the module and not through the public interface. If that's the case then great and shouldn't add them to `web3.eth`.

If you remove this it will break and not compile in typescript. You need to expose this on the web3-eth typing's interface if you want people to be able to get to it through there like personal. If you don't want `web3-eth-miner` to ever be able to be accessed through `web3.eth` and import only then that's fine. 

We can not remove `web3.eth.personal` as it will break what the docs say which people support and then break TS apps.

princesinha19:
@joshstevens19 can you please review this too?

packages/web3-eth/package.json
```diff
@@ -29,6 +29,7 @@
         "web3-eth-ens": "1.0.0-beta.52",
         "web3-eth-iban": "1.0.0-beta.52",
         "web3-eth-personal": "1.0.0-beta.52",
+        "web3-eth-miner": "1.0.0-beta.52",
```
joshstevens19: I thought we said we were going to keep web3-eth-miner in its own module?

princesinha19: Yes, sorry I forgot to remove it.

princesinha19: Done @joshstevens19.

packages/web3-eth-miner/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "module": "dist/web3-eth-miner.esm.js",
+    "browser": "dist/web3-eth-miner.umd.js",
+    "scripts": {
+        "build": "rollup -c",
+        "dev": "rollup -c -w",
+        "test": "jest",
+        "dtslint": "dtslint types --onlyTestTsNext"
+    },
+    "types": "types/index.d.ts",
+    "dependencies": {
```
nivida, line 18: The module eth-accounts isn't used here. I will update the dependencies on all new modules after the merge happened but feel free to do it for me.:)

princesinha19: yes sure.

packages/web3-eth/package.json
nivida: @joshstevens19 @princesinha19 Miner should be a standalone module. This means it doesn't have to be a dependency of the web3-eth module.

nivida (approved):

packages/web3-eth/package.json
princesinha19: Yes, right. @nivida

packages/web3-eth-miner/package.json
princesinha19: `constructor(provider: provider, net?: net.Socket|null, options?: Web3ModuleOptions, accounts?: Accounts|null);`

@nivida inside typings web3-eth-accounts is used. So, for now, I will keep web3-eth-accounts.

nivida (approved):

packages/web3-eth-miner/package.json
```diff
@@ -0,0 +1,37 @@
...
+        "web3-core-helpers": "1.0.0-beta.52",
+        "web3-core-method": "1.0.0-beta.52",
+        "web3-net": "1.0.0-beta.52",
+        "web3-providers": "1.0.0-beta.52",
+        "web3-utils": "1.0.0-beta.52"
+    },
+    "devDependencies": {
+        "definitelytyped-header-parser": "^1.0.1",
+        "dtslint": "0.4.2",
+        "web3-eth-accounts": "1.0.0-beta.52"
```
joshstevens19, line 30: you cant add this as a `dev` dependency if you got it in the typing's, it has to be a a proper dependency as NPM wont publish that on install. If i was you in your typing's i would just make it a `any` to avoid adding a dependency.

packages/web3-eth-miner/types/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+    You should have received a copy of the GNU Lesser General Public License
+    along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/**
+ * @file index.d.ts
+ * @author Prince Sinha <sinhaprince013@gmail.com>
+ * @date 2019
+ */
+
+import {Accounts} from 'web3-eth-accounts';
```
joshstevens19, line 20: kill this due to bringing in unneeded dependencies.

packages/web3-eth-miner/types/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+import {provider} from 'web3-providers';
+import {AbstractWeb3Module, Web3ModuleOptions, Hex} from 'web3-core';
+import * as net from 'net';
+
+export class Miner extends AbstractWeb3Module {
+    constructor(provider: provider, net?: net.Socket|null, options?: Web3ModuleOptions, accounts?: Accounts|null);
```
joshstevens19, line 26: when you kill the above just do `accounts?: any` or repeat the Account type in here. Until we flatten the dependencies and make it cleaner we have to do this or we have to add dependencies for everything. Avoid adding new dependencies to any module if its just for types and repeat with a comment above why. :+1:

princesinha19: okay sounds great.

joshstevens19 (approved):

joshstevens19:
You have some conflicts to resolve @princesinha19 👍

PR #2693, opened 2019-04-16
Add integration tests.

### LINKED ISSUE #2682: Integration tests
### Description 

The Web3.js library QA should get improved with integration tests. The goal is to test it against several node types (Geth, Parity, and Ganache).
These integration tests should run in a docker image for later adding it to the [GitHub actions](https://github.com/features/actions).

#### Expected behavior

I can execute ``npm run e2e`` and it will test the whole library with several node types.

#### Actual behavior

There are no integration tests.

### LINKED ISSUE #2683: ENS reverse lookup
### Description 

Implement the ENS reverse lookup.

Docs: https://docs.ens.domains/contract-api-reference/reverseregistrar

levino:
## Description

I only added a simple integration test against a local ganache.

Fixes #2682


nivida:
We have currently a similar PR open here: https://github.com/ethereum/web3.js/pull/2688

levino:
~~Mine is **much** simpler.~~ Not so simple any more.

nivida:
Yep, just looked at the other and it looks way more complicated. We should probably merge this PR and use your structure as a base. @micahriggan @szerintedmi Would that be okay for both you?

.travis.yml
```diff
@@ -1,26 +1,24 @@
...
-  apt:
-    sources:
-    - ubuntu-toolchain-r-test
-    packages:
-    - g++-4.8
+    apt:
+        sources:
+            - ubuntu-toolchain-r-test
+        packages:
+            - g++-4.8
```
levino, line 14: I think 4 space indent just follows `eslint.config`.

docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    nm15:
+    nm16:
+    nm17:
+    nm18:
+    nm19:
+    nm20:
+    nm21:
+    nm22:
+    nm23:
+    nm24:
```
levino: I will add a newline once the tests in ci have passed.

szerintedmi:
> Yep, just looked at the other and it looks way more complicated. We should probably merge this PR and use your structure as a base. @micahriggan @szerintedmi Would that be okay for both you?

I don't have a strong opinion, I'm just happy it's happening :)  I need to try both locally to tell which approach would I prefer.

docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+            - nm15:/home/node/packages/web3-eth-iban/node_modules
+            - nm16:/home/node/packages/web3-eth-personal/node_modules
+            - nm17:/home/node/packages/web3-eth-txpool/node_modules
+            - nm18:/home/node/packages/web3-net/node_modules
+            - nm19:/home/node/packages/web3-providers/node_modules
+            - nm20:/home/node/packages/web3-providers-http/node_modules
+            - nm21:/home/node/packages/web3-providers-ipc/node_modules
+            - nm22:/home/node/packages/web3-providers-ws/node_modules
+            - nm23:/home/node/packages/web3-shh/node_modules
+            - nm24:/home/node/packages/web3-utils/node_modules
```
levino: These are necessary so that the `npm install` in the docker world does not write to the host and writes deps as root and also possibly for a different operating system.

levino:
To run locally it should suffice to:
```bash
npm install
npm run test:integration
```

Requirement: docker-compose installed on local machine.

levino:
The whole `lerna` crazyness is only slowing things down in this repo (e.g. one has to rollup all packages before one can run a test... WTF?). It should be undone asap.

levino:
Oh this part from the CI output is golden:

```bash
Installing to /home/travis/build/ethereum/web3.js/packages/web3-eth-iban/node_modules/dtslint/typescript-installs/next...
Installed!
lerna info run Ran npm script 'dtslint' in 'web3-shh' in 9.6s:
> web3-shh@1.0.0-beta.52 dtslint /home/travis/build/ethereum/web3.js/packages/web3-shh
> dtslint types --onlyTestTsNext
Installing to /home/travis/build/ethereum/web3.js/packages/web3-shh/node_modules/dtslint/typescript-installs/next...
Installed!
lerna info run Ran npm script 'dtslint' in 'web3-eth-ens' in 9.5s:
> web3-eth-ens@1.0.0-beta.52 dtslint /home/travis/build/ethereum/web3.js/packages/web3-eth-ens
> dtslint types --onlyTestTsNext
Installing to /home/travis/build/ethereum/web3.js/packages/web3-eth-ens/node_modules/dtslint/typescript-installs/next...
Installed!
lerna info run Ran npm script 'dtslint' in 'web3-eth' in 9.8s:
> web3-eth@1.0.0-beta.52 dtslint /home/travis/build/ethereum/web3.js/packages/web3-eth
> dtslint types --onlyTestTsNext
Installing to /home/travis/build/ethereum/web3.js/packages/web3-eth/node_modules/dtslint/typescript-installs/next...
Installed!
lerna info run Ran npm script 'dtslint' in 'web3-core-helpers' in 9.1s:
> web3-core-helpers@1.0.0-beta.52 dtslint /home/travis/build/ethereum/web3.js/packages/web3-core-helpers
> dtslint types --onlyTestTsNext
Installing to /home/travis/build/ethereum/web3.js/packages/web3-core-helpers/node_modules/dtslint/typescript-installs/next...
Installed!
lerna info run Ran npm script 'dtslint' in 'web3-core-method' in 11.4s:
> web3-core-method@1.0.0-beta.52 dtslint /home/travis/build/ethereum/web3.js/packages/web3-core-method
> dtslint types --onlyTestTsNext
Installing to /home/travis/build/ethereum/web3.js/packages/web3-core-method/node_modules/dtslint/typescript-installs/next...
Installed!
lerna info run Ran npm script 'dtslint' in 'web3-providers' in 22.3s:
> web3-providers@1.0.0-beta.52 dtslint /home/travis/build/ethereum/web3.js/packages/web3-providers
> dtslint types --onlyTestTsNext
Installing to /home/travis/build/ethereum/web3.js/packages/web3-providers/node_modules/dtslint/typescript-installs/next...
Installed!
lerna info run Ran npm script 'dtslint' in 'web3-core' in 15.3s:
```

packages/web3/package.json
```diff
@@ -52,6 +52,9 @@
         "build": "rollup -c",
         "dev": "rollup -c -w",
         "test": "jest",
+        "docker": "docker-compose run --workdir='/home/node/packages/web3' web3 npm run",
+        "predocker": "docker-compose run web3 npm install --unsafe-perm",
+        "test:integration": "npm run docker test --testMatch='/**/*.Itest.js'",
```
szerintedmi: do we need to introduce a new file extension? Can it be done with `--scope` flag?

levino:
Why are linting, coverage and tests not run in parallel? There is no need to wait for linting to run the tests. Is TravisCI so bad?

packages/web3/package.json
levino: I f****g hate jest and do not know what kind of magic it can do. It is very probable that there is a more elegant way.

nivida:
> The whole lerna crazyness is only slowing things down in this repo (e.g. one has to rollup all packages before one can run a test... WTF?). It should be undone asap.

Yes, it will be possible to remove lerna if we move the modules together to one web3 module.

> Why are the tests not run in parallel? There is no need to wait for linting to run the tests. Is TravisCI so bad?

Didn't look at it closer. But yes we could parallelize it. Hopefully, GitHub will soon release the GitHub actions also for public repositories.

levino:
Boom, tests pass.

docker-compose.yml
levino:
```suggestion
    nm24:
    
```

packages/web3/integrationTests/SendTransaction.Itest.js
```diff
@@ -0,0 +1,15 @@
...
+    it('Should successfully resolve on sending ether', async () => {
+        const web3 = new Web3('http://ganache:8545', undefined, {transactionConfirmationBlocks: 1});
+        const [sender, recipient] = await web3.eth.getAccounts();
+        await expect(
+            web3.eth.sendTransaction({
+                value: '1',
+                from: sender,
+                to: recipient
+            })
+        ).resolves;
```
levino: I am not sure that 
```js
await expect(myPromise).resolve
```
does exactly what I intend to do: Wait for the promise to resolve and ignore the returned value but fail when the promise is rejected. Does anyone know more?

nivida:
``` javascript 
await expect(
            web3.eth.sendTransaction({
                value: '1',
                from: sender,
                to: recipient
            })
        ).resolves.toEqual(...);
``` 

Would be the correct way to test the Promise case and the EventEmitter case has to get added to. 

``` javascript 
it('Should successfully resolve on sending ether', (done) => {
        const web3 = new Web3('http://ganache:8545', undefined, {transactionConfirmationBlocks: 1});
        const [sender, recipient] = await web3.eth.getAccounts();
        web3.eth.sendTransaction({
            value: '1',
            from: sender,
            to: recipient
        })
        .on('confirmation', (confirmation) => { .... });
        .on('receipt', (receipt) => { expect(receipt).toEqual(...); done();  });
});
```

nivida: https://jestjs.io/docs/en/expect

levino: I wanted to add one simple test. You can later add more tests if you want.

Also I do not want to verify the value returned by the promise. I only want to make sure the promise resolves. Right now I include just a very very simple test to show how it works. Please do not make the PR more complex by asking for more tests. They can be added easily later.

levino: I think it works as is btw because I tested with a non-working url for the ganache server and the test failed.

levino: Maybe there is something like
```js
await expect(
            web3.eth.sendTransaction({
                value: '1',
                from: sender,
                to: recipient
            })
        ).resolves.toEqual(iDontGiveAF***);
```

nivida:
``` javascript
 await web3.eth.sendTransaction({
                value: '1',
                from: sender,
                to: recipient
            })
```

Is enough because the test will fail if the Promise takes to long or if it rejects with an error.

levino: That is not a test. If this fails you will never get proper information on exactly which test failed.

nivida: It will show you the description string and the failed state. We could also just expect the properties from the receipt which aren't generic. 

``` javascript 
await expect(web3.eth.sendTransaction({
  value: '1',
  from: sender,
  to: recipient
})).resolves.toHaveProperty('status', true);
```

But I think it would be better if we could expect the complete receipt object with ``toEqual()``.

packages/web3/package.json
szerintedmi: tell me.. I spent 10 minutes writing the actual tests / docker stuff and hours in Lerna and Jest swamp  `</rant>`

packages/web3/integrationTests/SendTransaction.Itest.js
levino: Please make this a suggestion next time like this:
```suggestion
        ).resolves.toHaveProperty('status', true);;
```

packages/web3/package.json
```diff
@@ -52,6 +52,9 @@
         "build": "rollup -c",
         "dev": "rollup -c -w",
         "test": "jest",
+        "docker": "docker-compose run --workdir='/home/node/packages/web3' web3 npm run",
+        "predocker": "docker-compose run web3 npm install --unsafe-perm",
```
levino, line 56: This will go into the docker container for service `web3` and install all deps. Because I added a `prepare` step to the root `package.json` scripts after `npm install` lerna will bootstrap and also build all packages (all in the docker container / the mounted volumes). Then the integration tests will be run in the docker container.

The only problem I just realised is that the `build` output will be written to the host. One must also create volumes for all the `/build` folders. Damn lerna!

szerintedmi: Ah, I think i got it, thanks. (I still getting lost in this spaghetti build process, I can't see the reason of all of this complication in the package)

levino: I do not find it so spaghetti to be honest. I mean lerna is annoying and makes no sense here but other than that is usualy that one npm script triggers the other...

levino: I made virtual `/dist` folders for the docker container.

nivida:
@Levino I would rename the folder ``integrationTests`` to ``e2e``.

szerintedmi:
> To run locally it should suffice to:
> 
> ```shell
> npm install
> npm run test:integration
> ```
> 
> Requirement: docker-compose installed on local machine.

it hangs for me on this (after successful `npm install`, pulled PR branch just now)
```
$ npm run test:integration

> web3@1.0.0-beta.52 test:integration /Users/petro/ether/web3.js
> lerna run test:integration

lerna notice cli v3.13.1
lerna info Executing command in 1 package: "npm run test:integration"
```

micahriggan:
> Yep, just looked at the other and it looks way more complicated. We should probably merge this PR and use your structure as a base. @micahriggan @szerintedmi Would that be okay for both you?

That's fine with me, mine was based off of the wrong branch anyway.

There are some aspects to mine that I think are important, and are missing here

* Testing websockets
* Testing geth and parity

Other than that, nice work, glad this is happening

levino:
> > To run locally it should suffice to:
> > ```shell
> > npm install
> > npm run test:integration
> > ```
> > 
> > 
> > Requirement: docker-compose installed on local machine.
> 
> it hangs for me on this (after successful `npm install`, pulled PR branch just now)
> 
> ```
> $ npm run test:integration
> 
> > web3@1.0.0-beta.52 test:integration /Users/petro/ether/web3.js
> > lerna run test:integration
> 
> lerna notice cli v3.13.1
> lerna info Executing command in 1 package: "npm run test:integration"
> ```

It just takes very long. You can go to the package (`cd package/web3`) and then run the tests there `npm run test:integration` then you can see the progress.

levino:
> * Testing websockets
> * Testing geth and parity
> 

Please be more specific.

Websockets should be testable against `ganache`, right? Can geth and parity be put into any state? Especially empty state? Or do you want to run integration tests against Kovan etc? I think that is a very bad idea, because of the statefulness.

micahriggan:
Parity and Geth both have a dev mode, which start empty chains, I was attempting to utilize that mode in my MR. 

As for the websockets on ganache, definitely, those should be tested as well 

But I think it'd be smart to have the tests run against each node, for each protocol. So that way you know if there's a breaking change that affects one node, or if there's some missing functionality.

I think what you have is good to merge for a starting point. The other nodes and protocol stuff can be done later

micahriggan:
I think what you have is good to merge for a starting point. The other nodes and protocol stuff can be done later for sure

levino:
Remarks should have been adressed. Anything else?

packages/web3-eth-contract/tests/src/methods/SendContractMethodTest.js
```diff
@@ -23,7 +23,16 @@ describe('SendContractMethodTest', () => {
...
-        sendContractMethod = new SendContractMethod({}, formatters, {}, {}, {}, {}, allEventsLogDecoderMock, abiModelMock);
+        sendContractMethod = new SendContractMethod(
+            {},
+            formatters,
+            {},
+            {},
+            {},
+            {},
+            allEventsLogDecoderMock,
+            abiModelMock
```
levino, line 34: These changes came from `npm run travis`.

packages/web3-eth-accounts/src/models/Account.js
```diff
@@ -169,7 +169,10 @@ export default class Account {
             cipher.final()
         ]);
 
-        const mac = keccak256(Buffer.concat([derivedKey.slice(16, 32), Buffer.from(ciphertext, 'hex')])).replace('0x', '');
+        const mac = keccak256(Buffer.concat([derivedKey.slice(16, 32), Buffer.from(ciphertext, 'hex')])).replace(
+            '0x',
+            ''
+        );
```
levino, line 175: These changes came from npm run travis.

levino:
The build fails because the type defs do not pass linting for TsNext any more. I think that setup is badly configured. Branch `1.0` should also not build atm.

levino:
I can make this much simpler. I would only do it if there is interest in actually merging this. Is there @nivida ?

nivida:
Thanks for asking @Levino. I've renamed all the branches for having a better overview and to prepare the coming stable release. I will use this PR as base for the integration tests in the 2.x branch.

PR #2200, opened 2019-01-18
Add tx hash for sign transaction es6

OFRBG:
Implements PR #2151 for the ES6 branch and includes #2199.

OG by @nujabes403.

PR #2694, opened 2019-04-16
Add 'txHash' field to the result of 'accounts.signTransaction'

nujabes403:
Sometimes, an user wants to know transaction hash before sending a `eth_sendRawTransaction`.

Of course, if the user send a transaction with `eth_sendTransaction` rpc call, user should wait for the response of the rpc call to receive transaction hash from Ethereum node. (It assumes the user doesn't use private key for signing, he uses unlockedAccount instead.)

On the otherhand, if the user send a transaction with `eth_sendRawTransaction` rpc call, user can know the transaction hash before sending the rpc call. (It assumes the user uses private key for signing directly.)

According to go-ethereum/src/internal/api.go, there are 2 steps to return `txHash` for a response of the rpc call.

Step 1: call `SendRawTransaction` function.

It accepts rlp-encoded transaction as an argument which is the the value `rawTransaction` of the result of calling `web3.eth.accounts.signTransaction(tx, privateKey)`.

After getting the encoded `rawTransaction`, this function decodes the encoded one. (rlp decode)
Then stores the decoded value to `tx` variable.

At last, call `submitTransaction()` function with `tx` variable which is decoded raw transaction.

```go
func (s *PublicTransactionPoolAPI) SendRawTransaction(ctx context.Context, encodedTx hexutil.Bytes) (common.Hash, error) {
	tx := new(types.Transaction)
	if err := rlp.DecodeBytes(encodedTx, tx); err != nil {
		return common.Hash{}, err
	}
	return submitTransaction(ctx, s.b, tx)
}
```

Step 2: call `submitTransaction`, return `tx.Hash()`

It returns `tx.Hash()` which is the payload for the response of `sendRawTransaction`.

```go
func submitTransaction(ctx context.Context, b Backend, tx *types.Transaction) (common.Hash, error) {
	if err := b.SendTx(ctx, tx); err != nil {
		return common.Hash{}, err
	}
	if tx.To() == nil {
		signer := types.MakeSigner(b.ChainConfig(), b.CurrentBlock().Number())
		from, err := types.Sender(signer, tx)
		if err != nil {
			return common.Hash{}, err
		}
		addr := crypto.CreateAddress(from, tx.Nonce())
		log.Info("Submitted contract creation", "fullhash", tx.Hash().Hex(), "contract", addr.Hex())
	} else {
		log.Info("Submitted transaction", "fullhash", tx.Hash().Hex(), "recipient", tx.To())
	}
	return tx.Hash(), nil
}
```

How `tx.Hash()` works?

Following code is the internal code of `tx.Hash`.

`go-ethereum/blockchain/types/transaction.go`
```go
func (tx *Transaction) Hash() common.Hash {
	if hash := tx.hash.Load(); hash != nil {
		return hash.(common.Hash)
	}
	v := rlpHash(tx)
	tx.hash.Store(v)
	return v
}
```

`go-ethereum/blockchain/types/block.go`
```go
func rlpHash(x interface{}) (h common.Hash) {
	hw := sha3.NewKeccak256()
	rlp.Encode(hw, x)
	hw.Sum(h[:0])
	return h
}
```

What it does is simple: do `keccak256` hash for the encoded transaction(rawTransaction.)

That's it. 

It means that if you already know encoded raw transaction (with signature, of course), you can achieve `txHash` before sending `eth_sendRawTransaction` rpc call, by `keccak256(rawTransaction)`.

It would be better if we add `txHash` field for the result of `web3.eth.accounts.signTransaction()`.

@nivida @frozeman @joshstevens19  Please take a look :)

-----------------------------------------
This PR is originated from #2151(@nujabes403) and #2200(@OFRBG).
Types, documentation, and test code are also added in this PR.


docs/web3-eth-accounts.rst
```diff
@@ -163,7 +163,7 @@ Returns
     - ``s`` - ``String``: Next 32 bytes of the signature
     - ``v`` - ``String``: Recovery value + 27
     - ``rawTransaction`` - ``String``: The RLP encoded transaction, ready to be send using :ref:`web3.eth.sendSignedTransaction <eth-sendsignedtransaction>`.
-
+    - ``txHash`` - ``String``: The transaction hash for the RLP encoded transaction.
```
princesinha19: I think we should use **transactionHash** instead of **txHash** to make module consistent. what do you say @nivida?

docs/web3-eth-accounts.rst
```diff
@@ -182,7 +182,8 @@ Example
         v: '0x25',
         r: '0xc9cf86333bcb065d140032ecaab5d9281bde80f21b9687b3e94161de42d51895',
         s: '0x727a108a0b8d101465414033c3f705a9c7b826e596766046ee1183dbc8aeaa68',
-        rawTransaction: '0xf869808504e3b29200831e848094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca008025a0c9cf86333bcb065d140032ecaab5d9281bde80f21b9687b3e94161de42d51895a0727a108a0b8d101465414033c3f705a9c7b826e596766046ee1183dbc8aeaa68'
+        rawTransaction: '0xf869808504e3b29200831e848094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca008025a0c9cf86333bcb065d140032ecaab5d9281bde80f21b9687b3e94161de42d51895a0727a108a0b8d101465414033c3f705a9c7b826e596766046ee1183dbc8aeaa68',
+        txHash: '0xde8db924885b0803d2edc335f745b2b8750c8848744905684c20b987443a9593',
```
princesinha19: remove "**,**" from this line.

docs/web3-eth-accounts.rst
```diff
@@ -199,7 +200,8 @@ Example
         r: '0x9ebb6ca057a0535d6186462bc0b465b561c94a295bdb0621fc19208ab149a9c',
         s: '0x440ffd775ce91a833ab410777204d5341a6f9fa91216a6f3ee2c051fea6a0428',
         v: '0x25',
-        rawTransaction: '0xf86a8086d55698372431831e848094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca008025a009ebb6ca057a0535d6186462bc0b465b561c94a295bdb0621fc19208ab149a9ca0440ffd775ce91a833ab410777204d5341a6f9fa91216a6f3ee2c051fea6a0428'
+        rawTransaction: '0xf86a8086d55698372431831e848094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca008025a009ebb6ca057a0535d6186462bc0b465b561c94a295bdb0621fc19208ab149a9ca0440ffd775ce91a833ab410777204d5341a6f9fa91216a6f3ee2c051fea6a0428',
+        txHash: '0xd8f64a42b57be0d565f385378db2f6bf324ce14a594afc05de90436e9ce01f60',
```
princesinha19: same here.

packages/web3-eth/src/signers/TransactionSigner.js
```diff
@@ -70,13 +71,15 @@ export default class TransactionSigner {
...
+        const txHash = Hash.keccak256(rawTransaction)
 
         return {
             messageHash: Buffer.from(ethTx.hash(false)).toString('hex'),
             v: '0x' + Buffer.from(ethTx.v).toString('hex'),
             r: '0x' + Buffer.from(ethTx.r).toString('hex'),
             s: '0x' + Buffer.from(ethTx.s).toString('hex'),
-            rawTransaction
+            rawTransaction,
+            txHash,
```
princesinha19: same here.

docs/web3-eth-accounts.rst
nujabes403: I agree. Let's use `transactionHash` then.

docs/web3-eth-accounts.rst
nujabes403: Thanks. I removed it.

docs/web3-eth-accounts.rst
nujabes403: Thanks. I removed it. :)

packages/web3-eth/src/signers/TransactionSigner.js
nujabes403: Thanks. I removed it.

princesinha19: You missed removing comma at this line.

nujabes403: Please check again :)

packages/web3-eth/package.json
```diff
@@ -31,7 +31,8 @@
         "web3-eth-personal": "1.0.0-beta.52",
         "web3-net": "1.0.0-beta.52",
         "web3-providers": "1.0.0-beta.52",
-        "web3-utils": "1.0.0-beta.52"
+        "web3-utils": "1.0.0-beta.52",
+        "eth-lib": "0.2.8"
```
joshstevens19: Why do you add a dependency here of eth-lib you can use the sha3 method on web3-utils and not bring in that dependency.

packages/web3-eth/src/signers/TransactionSigner.js
```diff
@@ -70,13 +71,15 @@ export default class TransactionSigner {
 
         const rlpEncoded = ethTx.serialize().toString('hex');
         const rawTransaction = '0x' + rlpEncoded;
+        const transactionHash = Hash.keccak256(rawTransaction)
```
joshstevens19: Use web3-utils sha3 here instead of bringing in eth-lib dependency 😁

packages/web3-eth/src/signers/TransactionSigner.js
joshstevens19: Ha I think this is depending on your eslint settings and preference BUT saying that I prefer the trailing comma as when you next add a new property to that object it’s only 1 git line that changes and not 2, which I may view is how it should be. Trailing commas FTW 😁!

packages/web3-eth/src/signers/TransactionSigner.js
nujabes403: Fixed it! please take a look again 😄

packages/web3-eth/package.json
nujabes403: I agree, I changed it to use `this.utils` instead.

packages/web3-eth/src/signers/TransactionSigner.js
nujabes403: @joshstevens19 
I prefer dangling comma as well, maybe we should have another PR for adding dangling comma 🙂 . In this PR, I don't want to add trailing comma, since the objective of this PR is not that. :D Thanks!

nivida (changes requested):
Thanks for submitting this PR! Could you update the type definitions too?

nujabes403:
@nivida I thought I've updated all type definitions for this PR in following link. 
I updated a type definition in this file, "packages/web3-core/types/index.d.ts"
https://github.com/ethereum/web3.js/pull/2694/files#diff-aac01cbafcc20bbb250c097650f50c8f

May I know which type definition do you want to update? 
(Which file should I change?)

joshstevens19 (approved):
Nice work 👍

princesinha19 (approved):

nivida (approved):

PR #2699, opened 2019-04-17
Minor changes in some packages

princesinha19:
## Description

Changed readme section os some module. Also, changed some typings and package dependency.

## Type of change
- [x] Enhancement

## Checklist:

### DIFF
--- a/docs/web3-eth-txpool.rst
+++ b/docs/web3-eth-txpool.rst
@@ -2,9 +2,9 @@
 
 .. include:: include_announcement.rst
 
-===============
+==================
 Web3 Txpool Module
-===============
+==================
 
 
 The ``web3-eth-txpool`` package gives you access to several non-standard RPC methods to inspect the contents of the transaction pool containing all the currently pending transactions as well as the ones queued for future processing.
--- a/docs/web3-eth-txpool.rst
+++ b/docs/web3-eth-txpool.rst
@@ -3,7 +3,7 @@
 .. include:: include_announcement.rst
 
 ===============
-web3.eth.txpool
+Web3 Txpool Module
 ===============
 
 
--- a/packages/web3-eth-miner/types/index.d.ts
+++ b/packages/web3-eth-miner/types/index.d.ts
@@ -17,13 +17,12 @@
  * @date 2019
  */
 
-import {Accounts} from 'web3-eth-accounts';
 import {provider} from 'web3-providers';
-import {AbstractWeb3Module, Web3ModuleOptions, Hex} from 'web3-core';
+import {AbstractWeb3Module, Web3ModuleOptions} from 'web3-core';
 import * as net from 'net';
 
 export class Miner extends AbstractWeb3Module {
-    constructor(provider: provider, net?: net.Socket|null, options?: Web3ModuleOptions, accounts?: Accounts|null);
+    constructor(provider: provider, net?: net.Socket|null, options?: Web3ModuleOptions, accounts?: any);
 
     setEtherbase(
         address: string,
@@ -49,3 +48,5 @@ export class Miner extends AbstractWeb3Module {
         callback?: (error: Error, result: boolean) => void
     ): Promise<boolean>;
 }
+
+export type Hex = string | number;
--- a/packages/web3-eth-miner/types/index.d.ts
+++ b/packages/web3-eth-miner/types/index.d.ts
@@ -19,10 +19,11 @@
 
 import {provider} from 'web3-providers';
 import {AbstractWeb3Module, Web3ModuleOptions} from 'web3-core';
+import {Accounts} from 'web3-eth-accounts';
 import * as net from 'net';
 
 export class Miner extends AbstractWeb3Module {
-    constructor(provider: provider, net?: net.Socket|null, options?: Web3ModuleOptions, accounts?: any);
+    constructor(provider: provider, net?: net.Socket|null, options?: Web3ModuleOptions, accounts?: Accounts|null);
 
     setEtherbase(
         address: string,
--- a/packages/web3-eth-personal/package.json
+++ b/packages/web3-eth-personal/package.json
@@ -22,12 +22,12 @@
         "web3-core-method": "1.0.0-beta.52",
         "web3-net": "1.0.0-beta.52",
         "web3-providers": "1.0.0-beta.52",
-        "web3-utils": "1.0.0-beta.52"
+        "web3-utils": "1.0.0-beta.52",
+        "web3-eth-accounts": "1.0.0-beta.52"
     },
     "devDependencies": {
         "definitelytyped-header-parser": "^1.0.1",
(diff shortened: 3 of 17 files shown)


packages/web3-eth-miner/types/index.d.ts
```diff
@@ -17,13 +17,12 @@
  * @date 2019
  */
 
-import {Accounts} from 'web3-eth-accounts';
 import {provider} from 'web3-providers';
-import {AbstractWeb3Module, Web3ModuleOptions, Hex} from 'web3-core';
+import {AbstractWeb3Module, Web3ModuleOptions} from 'web3-core';
 import * as net from 'net';
```
princesinha19, line 23: Here we are making net as a dependency of this module. Only for typing. I think we can remove this. What do you say @joshstevens19, @nivida?

docs/web3-eth-txpool.rst
```diff
@@ -3,7 +3,7 @@
 .. include:: include_announcement.rst
 
 ===============
-web3.eth.txpool
+Web3 Txpool Module
 ===============
```
nivida: The title lines (``====``) are too short.

packages/web3-eth-miner/types/index.d.ts
nivida: It's the node net module and not the web3 net module that's why I think this would be okay.

packages/web3-eth-miner/types/index.d.ts
```diff
@@ -17,13 +17,12 @@
...
 
 export class Miner extends AbstractWeb3Module {
-    constructor(provider: provider, net?: net.Socket|null, options?: Web3ModuleOptions, accounts?: Accounts|null);
+    constructor(provider: provider, net?: net.Socket|null, options?: Web3ModuleOptions, accounts?: any);
```
nivida: this should be ``Accounts|null`` and not any.

packages/web3-eth-personal/package.json
```diff
@@ -22,12 +22,12 @@
         "web3-core-method": "1.0.0-beta.52",
         "web3-net": "1.0.0-beta.52",
         "web3-providers": "1.0.0-beta.52",
-        "web3-utils": "1.0.0-beta.52"
+        "web3-utils": "1.0.0-beta.52",
+        "web3-eth-accounts": "1.0.0-beta.52"
```
nivida, line 26: ``eth-accounts`` should be a devDependency.

packages/web3-eth-txpool/types/index.d.ts
```diff
@@ -17,13 +17,12 @@
...
  */
 
-import {Accounts} from 'web3-eth-accounts';
 import {provider} from 'web3-providers';
 import {AbstractWeb3Module, Web3ModuleOptions, Content} from 'web3-core';
 import * as net from 'net';
 
 export class Txpool extends AbstractWeb3Module {
-    constructor(provider: provider, net?: net.Socket|null, options?: Web3ModuleOptions, accounts?: Accounts|null);
+    constructor(provider: provider, net?: net.Socket|null, options?: Web3ModuleOptions, accounts?: any);
```
nivida: This has to be ``Accounts|null``

nivida (changes requested):

packages/web3-eth-miner/types/index.d.ts
princesinha19: Ya, you are correct @nivida. But we had a discussion about it. @joshstevens19 suggested that if it's only used in typing then we shouldn't make accounts as a dependency of this module. That's why I changed it to any.

docs/web3-eth-txpool.rst
princesinha19: Ya, sorry I forgot this.

packages/web3-eth-miner/types/index.d.ts
princesinha19: okay cool.

packages/web3-eth-miner/types/index.d.ts
princesinha19: @nivida Please check this [comment](https://github.com/ethereum/web3.js/pull/2660#discussion_r275371902). 

This too: https://github.com/ethereum/web3.js/pull/2660#discussion_r275373009

packages/web3-eth-personal/package.json
princesinha19: https://github.com/ethereum/web3.js/pull/2660#discussion_r275371902

packages/web3-eth-miner/types/index.d.ts
joshstevens19: Thinking about this more we may as well move `Accounts` interface into `web3-core` typing's so it can be shared without adding any dependencies or not supplying solid typing's... thoughts?

packages/web3-eth-personal/package.json
joshstevens19: Yes you can't add this as a dev dependency in production it would fail when you compile in TS. 

Thinking about this more we may as well move `Accounts` interface into `web3-core` typing's so it can be shared here without adding any dependencies or not supplying solid typing's... thoughts?

packages/web3-eth-miner/types/index.d.ts
princesinha19: Yes, I was also thinking about the same @joshstevens19.

packages/web3-eth-personal/package.json
princesinha19: yes right.

packages/web3-eth-miner/types/index.d.ts
princesinha19: I think the `provider` interface should also be added to `web-core` typing to reduce dependency.

joshstevens19: Have you only brought provider in for typings? If so yes I agree. We should never bring a dependency in just for a type, if this is the case we have a common place for it in web3-core 👍

princesinha19: No, I think we have to put web3-providers. It got used in multiple places. But we can export `Account` type into `web3-core`.

packages/web3-eth-personal/package.json
nivida: Yes, we should probably move it to the web3-core module. This issue will solve this problem later in a more clean way: https://github.com/ethereum/web3.js/issues/2109

princesinha19: Yes right. @nivida.

princesinha19: @nivida I think we should keep web3-eth-accounts as a dependency for now as if I move Accounts in core-module for example.
https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth-accounts/README.md#usage

We should think for a good solution to solve this.

nivida (approved):

PR #2713, opened 2019-04-20
eth_getProof method implementation

### LINKED ISSUE #2712: Implementing EIP-1186.
## Description
According to [EIP-1186](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1186.md), We should support RPC-Method to get Merkle Proofs, which is **eth_getProof**.

https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getproof

## Expected behavior
Possible implementation:
web3.eth.getProof(...)

Versions
web3.js: latest

princesinha19:
## Description

Implemented eth_getProof based on EIP-1186.

Fixes #2712 

## Type of change
- [x] New feature 

## Checklist:


packages/web3-eth/types/index.d.ts
```diff
@@ -220,3 +222,17 @@ export interface Subscription<T> {
...
+    jsonrpc: string;
+    id: number;
+    result: {
+      address: string;
+      accountProof: string[];
+      balance: number | string;
+      codeHash: string;
+      nonce: number | string;
+      storageHash: string;
+      storageProof: [];
```
nivida: A small detail but I would define it with: ``any[]``

nivida (changes requested):

packages/web3-eth/types/index.d.ts
joshstevens19:
```
storageProof: ARRAY - Array of storage-entries as requested. Each entry is a object with these properties:

key: QUANTITY - the requested storage key
value: QUANTITY - the storage value
proof: ARRAY - Array of rlp-serialized MerkleTree-Nodes, starting with the storageHash-Node, following the path of the SHA3 (key) as path.
```

example:

```js
"storageProof": [
      {
        "key": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
        "proof": [
          "0xf90211a...0701bc80",
          "0xf90211a...0d832380"
        ],
        "value": "0x1"
      }
    ]
```

so the type should be 

```
{
  key: string;
  proof: string[];
  value: string;
}[];
```

Lets keep up the high standard typing's when possible please.

packages/web3-eth/types/index.d.ts
```diff
@@ -220,3 +222,17 @@ export interface Subscription<T> {
...
+
+export interface GetProof {
+    jsonrpc: string;
+    id: number;
+    result: {
+      address: string;
+      accountProof: string[];
+      balance: number | string;
+      codeHash: string;
+      nonce: number | string;
```
joshstevens19: i think this will always come back as a hex string `0x23` so i think string is fine.

packages/web3-eth/types/index.d.ts
```diff
@@ -220,3 +222,17 @@ export interface Subscription<T> {
...
     on(type: 'error', handler: (data: Error) => void): void
 }
+
+export interface GetProof {
+    jsonrpc: string;
+    id: number;
+    result: {
+      address: string;
+      accountProof: string[];
+      balance: number | string;
```
joshstevens19: i think this will always come back as a hex string i.e. `0x23` so i think string is fine here.

packages/web3-eth/types/index.d.ts
```diff
@@ -145,6 +145,8 @@ export class Eth extends AbstractWeb3Module {
     submitWork(data: [string, string, string], callback?: (error: Error, result: boolean) => void): Promise<boolean>;
 
     pendingTransactions(callback?: (error: Error, result: []) => void): Promise<[]>;
+
+    getProof(address: string, storageKey: string[], blockNumber: number | string, callback?: (error: Error, result: GetProof) => void): Promise<GetProof>;
```
joshstevens19: ```blockNumber: number | string | "latest" | "earliest"``` will make this even more clearer to people.

princesinha19: Ya think so.

packages/web3-eth/types/index.d.ts
princesinha19: Yes, right @joshstevens19. Thanks.

princesinha19:
Thanks, @nivida, @joshstevens19 for your suggestions. I made the requested changes.

joshstevens19 (approved):

packages/web3-eth/types/index.d.ts
princesinha19: @joshstevens19 T[ ] is deprecated. linter says that we should use `Array<T>` instead.

joshstevens19: Both do the same thing, which linter is saying that? `Array<T>` and `T[]` should both be supported. Keep it the same `T[]` for now if that is a thing from typescript then we can change it everywhere.  It is still in all the ts docs - http://www.typescriptlang.org/docs/handbook/generics.html#working-with-generic-type-variables

princesinha19: could we use:-
```
export interface GetProof {
    jsonrpc: string;
    id: number;
    result: {
      address: string;
      accountProof: string[];
      balance: string;
      codeHash: string;
      nonce: string;
      storageHash: string;
      storageProof: Array<StorageProof>;
    };
}

export interface StorageProof {
    key: string;
    proof: string[];
    value: string;
}
```

joshstevens19: `Array<StorageProof>` and `StorageProof[]` are exactly the same things. To be consistent i would use `StorageProof[]`

princesinha19: Yes, you are right. Typescript supports both,

> which linter is saying that?

can you please check the build pipeline?

princesinha19: okay so, for now, I will remain it as `StorageProof[ ].` In future, we can change.

joshstevens19: TypeScript do not state you should use 1 or the other its up to you. What your seeing is a linter config set up. This error will go away when you turn it into a interface as you shown above. 

We should use `T[]` throughout as it follows consistency and TypeScript say you can use both and `T[]` is defo not `deprecated`. We should always create a `interface` for it and never do one on the fly - my example was merely showing the type it should be. The error is saying `create a interface for it` basically so its doing its job.

princesinha19: okay cool. Thanks.

princesinha19: This was the build pipeline error @joshstevens19 
`array-type  Array type using 'T[]' is forbidden for non-simple types. Use 'Array<T>' instead.`

joshstevens19: yep its a bad error its basically saying `create a interface for it` and not do a dynamic object like you did before (and my example shown). That is coming from dtslint tslint - its not a ts `deprecated` warning.

princesinha19: Cool. Thanks for clarification @joshstevens19.

packages/web3-core-method/src/methods/node/GetProofMethod.js
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * This method will be executed after the RPC request.
+     *
+     * @method afterExecution
+     *
+     * @param {Object} response
+     *
+     * @returns {Number}
+     */
+    afterExecution(response) {
```
nivida, line 46: Thanks for adding the ``afterExecution`` method! 
Could you extend the ``GetProofMethodTest`` for having the ``afterExecution`` method also covered?

nivida (changes requested):

packages/web3-core-method/src/methods/node/GetProofMethod.js
princesinha19: Ya sure, I will add.

princesinha19: @nivida After some research, I added test cases for after afterExecution.

packages/web3-core-method/tests/src/methods/node/GetProofMethodTest.js
```diff
@@ -0,0 +1,69 @@
...
+        expect(method.utils).toEqual(Utils);
+
+        expect(method.formatters).toEqual(null);
+    });
+
+    it('afterExecution should map the response', () => {
+        const response = {
+            jsonrpc: '2.0',
+            id: 1,
+            result: {
```
nivida: the ``afterExecution`` method just get the result object and not the complete JSON-RPC response.

docs/web3-eth.rst
```diff
@@ -1512,3 +1512,82 @@ Example
...
+
+    web3.eth.getProof(
+        "0x1234567890123456789012345678901234567890",
+        ["0x0000000000000000000000000000000000000000000000000000000000000000","0x0000000000000000000000000000000000000000000000000000000000000001"],
+        "latest"
+    ).then(console.log);
+    > {
+        "jsonrpc": "2.0",
+        "id": 1,
+        "result": {
```
nivida: It should just return the mapped result object from the JSON-RPC response

nivida (changes requested):

packages/web3-core-method/tests/src/methods/node/GetProofMethodTest.js
princesinha19: Sorry my mistake.

docs/web3-eth.rst
princesinha19: Yes.

packages/web3-core-method/tests/src/methods/node/GetProofMethodTest.js
princesinha19:
```
> eth.getProof("0x651c77c61b4a53a3c1d2e062796770af5feb9380", INFO [04-24|15:59:57.894] New local node record                    seq=198 id=920485397a8116e7 ip=14.143.40.34 udp=30303 tcp=30303
> eth.getProof("0x651c77c61b4a53a3c1d2e062796770af5feb9380", ["0x0000000000000000000000000000000000000000000000000000000000000000","0x0000000000000000000000000000000000000000000000000000000000000001"],"latest")
{
  accountProof: ["0xf8518080a0517337efc973654f74640b6ef0529e88635e69b6e3731fc051281e33719c23a8a0f5482f368e10755e10c917e5370419a455885fc66901a33cdfbd7a510d722f5280808080808080808080808080", "0xf872a035e21f3f1ec4126dd0768297052fc8c04c64b8073461ca1ba04b7504824e2341b84ff84d0189013f306a2409fc0000a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a0c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"],
  address: "0x651c77c61b4a53a3c1d2e062796770af5feb9380",
  balance: "0x13f306a2409fc0000",
  codeHash: "0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470",
  nonce: "0x1",
  storageHash: "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
  storageProof: [{
      key: "0x0000000000000000000000000000000000000000000000000000000000000000",
      proof: [],
      value: "0x0"
  }, {
      key: "0x0000000000000000000000000000000000000000000000000000000000000001",
      proof: [],
      value: "0x0"
  }]
}
```
Balance can also be a big number. I guess balance should also be converted to BN. @nivida.
I tested this in my local geth.

nivida: Yes, balance should be a BN too. You could also transform the nonce because it's possible that the nonce will exceed the max integer value of JS.

princesinha19: Also, should I convert balance from Wei to ether or just pass as it is?

nivida (approved):

PR #2720, opened 2019-04-23
Types of the management modules improved

nivida:
## Description

@princesinha19 created a lot of modules which will provide the missing methods from the management JSON-RPC API. For not annoy him too much did I decide to go through the types and documentation later when all the modules are merged.

## Type of change

- [x] Enhancement

## Checklist:


packages/web3-core/types/index.d.ts
```diff
@@ -201,8 +201,30 @@ export interface Log {
...
 export interface Content {
-    pending: any;
-    queued: any;
+    pending: TxPoolContent;
+    queued: TxPoolContent;
+}
+
+export interface TxPoolContent {
+    [address: string]: {
+        [nonce: number]: Array<Transaction>
```
princesinha19: @nivida Before some time, I and @joshstevens19 had a discussion that for consistency we should use Transaction[ ] not Array<Transaction>.

nivida: Funny I had it first defined with ``Transaction[]`` and saw above that we were using ``Array<type>``. I'll go through all types and will change it to ``type[]``. Thanks! 💪

packages/web3-core/types/index.d.ts
```diff
@@ -200,9 +200,25 @@ export interface Log {
...
+}
+
+export interface TxPoolInspect {
+    pending: TxPool;
+    queued: TxPool;
+}
+
+export interface TxPool {
+    [address: string]: {
+        [nonce: number]: string[] | Transaction[];
```
princesinha19, line 215: this should be only Transaction[ ].

packages/web3-eth-admin/types/index.d.ts
```diff
@@ -40,7 +39,7 @@ export class Admin extends AbstractWeb3Module {
 
     getPeers(
         callback?: (error: Error, result: any[]) => void
```
princesinha19: Sorry @nivida, I think you forgot to change this. The result should be `PeerInfo[ ]` instead of `any[ ]`. Same in test.👍

packages/web3-core/types/index.d.ts
nivida: Not in all cases. I've combined two interfaces to one.

PR #2753, opened 2019-04-29
Transaction.to can be null

alexkvak:
According to [docs](https://web3js.readthedocs.io/en/1.0/web3-eth.html#gettransaction)
`to` - `String`: Address of the receiver. `null` when its a contract creation transaction.

### DIFF
--- a/packages/web3-core/types/index.d.ts
+++ b/packages/web3-core/types/index.d.ts
@@ -125,7 +125,7 @@ export interface Transaction {
     blockNumber: number | null;
     transactionIndex: number | null;
     from: string;
-    to?: string;
+    to: string | null;
     value: string;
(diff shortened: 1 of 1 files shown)


princesinha19 (commented):
I think we shouldn't make it as optional. RPC returns null if it's a contract creation so, we shouldn't make it as an optional parameter. In all case `to` parameter will exist.

alexkvak:
@princesinha19 agree, fixed

joshstevens19 (approved):
Thanks for the PR @alexkvak :+1: @nivida can you merge.

princesinha19 (approved):
Thanks for the PR @alexkvak.

nivida (approved):

PR #2144, opened 2018-12-22
fix quorum timestamp

dmytronasyrov:
Quorum timestamp has nanoseconds. Make outputBlockFormatter possible to work with it.

PR #2768, opened 2019-05-02
bugfix: issue 1351 Quourm block timestamp bigNumber conversion issue

### LINKED ISSUE #1215: Block timestamp bigNumber coversion issue (WebSocket connection)
Hello,

I am using "web3": "~1.0.0-beta.26" to make connection to JPMorgan Quorum (https://github.com/jpmorganchase/quorum). Quorum network is configured to use Raft consensus algorithm. Raft algorithm is making changes to block timestamp value, so it is larger (Unix nanoseconds - 1506355254821575394 e.g.) then default - https://github.com/jpmorganchase/quorum/issues/190. Due to this, when I am trying to deploy contract, using WebSocket provider, I receive error from bigNumber.js library: 
**Error: Number can only safely store up to 53 bits**

This error is caused by https://github.com/ethereum/web3.js/blob/develop/lib/web3/formatters.js#L171; looks like Raft timestamp is too big for converting it to the number. 

Timestamp value from failing block object:
```json
{ 
"timestamp": "0x14fe06494f652193"
}
```

As I understood outputBlockFormatter is only used during WebSocket connection (ws provider), Rpc connection (http provider) is deploying fine. It was OK to use only RPC connection, but, with recent updates, subscription is possible only using WebSocket. 

I understand, that nonstandard ethereum networks is not Your main field; however, would be cool to hear some thoughts about this issue.

ThomasRalee:
## Description
Resolves #1215

Replica of PR #2144 by @dmytronasyrov  with conflicts resolved


nivida (changes requested):
Could you check the length of the given string with a BN object instead of letting it thrown an error? 

Btw.: Could you also extend the test cases?

packages/web3-core-helpers/tests/src/Formatters/OutputBlockFormatterTest.js
```diff
@@ -96,4 +96,49 @@ describe('OutputBlockFormatterTest', () => {
...
+                    gas: 100,
+                    gasPrice: '100',
+                    nonce: 1,
+                    value: '100',
+                    to: '0x03C9A938fF7f54090d0d99e2c6f80380510Ea078',
+                    from: '0x03C9A938fF7f54090d0d99e2c6f80380510Ea078'
+                }
+            ],
+            miner: '0x03C9A938fF7f54090d0d99e2c6f80380510Ea078'
+        });
```
princesinha19, line 191: Could you add end `});` for the test.

packages/web3-core-helpers/tests/src/Formatters/OutputBlockFormatterTest.js
```diff
@@ -96,4 +96,49 @@ describe('OutputBlockFormatterTest', () => {
             miner: '0x03C9A938fF7f54090d0d99e2c6f80380510Ea078'
         });
     });
+    it('[Quorum] call outputBlockFormatter with a valid block with nano format timestamp', () => {
+        const block = {
+            gasLimit: 0x0,
+            gasUsed: 0x0,
+            size: 0x0,
+            timestamp: '0x159adc720a96d665',
```
princesinha19, line 152: You have not checked for the timestamp in the test.

princesinha19 (changes requested):
Could you please do these changes. @ThomasRalee

packages/web3-core-helpers/src/Formatters.js
```diff
@@ -302,8 +302,19 @@ export const outputBlockFormatter = (block) => {
     block.size = Utils.hexToNumber(block.size);
 
     // Support Quorum 2.2.0 - timestamp is not present in the Quorum getBlock response
-    if (block.timestamp !== null) {
+    if (block.timestamp) {
+      if (block.timestamp.length < 16) {
```
princesinha19: You can merge line no. 305 with this line by making:-
`if(block.timestamp && block.timestamp.length < 16)`.

Also, why you are checking for `block.timestamp.length < 16`?

princesinha19 (changes requested):
@ThomasRalee, Can you please update the test accordingly so, that we could achieve 100% coverage. Btw thanks for this PR.

packages/web3-core-helpers/src/Formatters.js
ThomasRalee: Updated: Currently using BN Object to determine `block.timestamp` format. 

Base on @nivida comment: `Could you check the length of the given string with a BN object instead of letting it thrown an error?`

Also added additional test to increase test coverage.

ThomasRalee:
@nivida can I please have some feedback? Would be really awesome if this can be merged

nivida:
Closed because of this PR: https://github.com/ethereum/web3.js/pull/2791

PR #2815, opened 2019-05-13
Fix error on contract method proxy when method parameter is null.

Pedro-vk:
## Description

After upgrade to Web3 beta 55, the method proxy was failing because some parameters were `null` instead of `undefined`. This change allows receiving `null` parameters.

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/packages/web3-eth-contract/src/proxies/MethodsProxy.js
+++ b/packages/web3-eth-contract/src/proxies/MethodsProxy.js
@@ -191,7 +191,7 @@ export default class MethodsProxy {
         method.setArguments(methodArguments);
 
         // If no parameters are given for the eth_call or eth_send* methods then it will set a empty options object.
-        if (typeof method.parameters[0] === 'undefined') {
+        if (!method.parameters[0]) {
             method.parameters[0] = {};
         }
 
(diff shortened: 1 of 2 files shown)


packages/web3-eth-contract/src/proxies/MethodsProxy.js
```diff
@@ -191,7 +191,7 @@ export default class MethodsProxy {
         method.setArguments(methodArguments);
 
         // If no parameters are given for the eth_call or eth_send* methods then it will set a empty options object.
-        if (typeof method.parameters[0] === 'undefined') {
+        if (typeof method.parameters[0] === 'undefined' || method.parameters[0] === null) {
```
nivida: Could you extend the test with the new case? :)

nivida (changes requested):

packages/web3-eth-contract/src/proxies/MethodsProxy.js
Pedro-vk: @nivida Done!

nivida: Great! Thanks for your changes:) 

Sorry if I annoy you but a test on ``falsy`` is probably the better solution. This will allow us to use ``false``, ``null`` and ``undefined`` without checking of the type.

``` javascript
if (method.parameters[0]) {
    method.parameters[0] = {};
}
```

Pedro-vk: Don't worry :P This change is better than the initial ;)
Done!

nivida (approved):

PR #2847, opened 2019-05-26
Resolve parity nodes invalid "eth_getBlockByNumber" JSONRPC call

joshstevens19:
## Description

If you send a transaction with a parity node web3 does not take into account that parity nodes can return the transaction receipt even if the transaction has not been mined. Unlike geth who return it once it has been mined. This causes a invalid JSONRPC call to "eth_getBlockByNumber" passing in a null block number causing the nodes to throw bad requests. This change just makes sure it does not call "eth_getBlockByNumber" until it has been mined. 

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change
- [ ] Enhancement

## Checklist:

- [ ] My changes generate no warnings.


packages/web3-core-method/src/observers/TransactionObserver.js
```diff
@@ -158,8 +159,15 @@ export default class TransactionObserver {
                             this.emitNext(receipt, observer);
                         }
                     } else {
-                        this.lastBlock = await this.getBlockByNumber(receipt.blockNumber);
-                        this.confirmations++;
+                        // on parity nodes you can get the receipt without it being mined
+                        // so the receipt may not have a block number at this point.
+                        // we should check that the blockNumber is defined before we do this call
+                        // geth nodes only return the receipt once mined as the spec states.
+                        if (receipt.blockNumber) {
```
nivida: the ``receipt.blockNumber`` check can be moved on line 152 and this big inline comment can be changed to: ``// blockNumber check for parity``.

nivida (changes requested):

packages/web3-core-method/src/observers/TransactionObserver.js
joshstevens19: yeah good shout we should only emit the receipt once mined i think else Dapps could get super confused. Will change

joshstevens19: Ok Updated i have made the comment smaller but i have still explained the point. Think its critical to explain why as people will forget.

nivida (approved):

PR #2856, opened 2019-05-29
EIP-1193 updates

JunichiSugiura:
## Description

Currently, if we instantiate Web3 with EIP1193 compatible provider, the provider resolver resolves CustomProvider for Eth and Shh module. This causes that the provider's send method receives payload as the first argument and callback function as the second argument when a dapp calls `web3.eth.getAccounts()` for example.

### Code to reproduce
```
const Web3 = require('web3')
const EventEmitter = require('events')

class Provider extends EventEmitter {
  constructor() {
    super()
    this.isEIP1193 = true
  }
}

const provider = new Provider()
const web3 = new Web3(provider)

console.log(web3.currentProvider.constructor.name)
// -> Web3EthereumProvider

console.log(web3.eth._currentProvider.constructor.name)
// -> CustomProvider
```

I added another condition for the resolver to check if the provider is Web3EthereumProvider then return the provider itself.

I hope this helps.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change
- [ ] Enhancement

## Checklist:

### DIFF
--- a/packages/web3-providers/src/resolvers/ProviderResolver.js
+++ b/packages/web3-providers/src/resolvers/ProviderResolver.js
@@ -75,7 +75,7 @@ export default class ProviderResolver {
             return provider;
         }
 
-        if (provider instanceof Web3EthereumProvider) {
+        if (provider.host === 'Web3EthereumProvider') {
             return provider;
         }
 
--- a/packages/web3-providers/src/resolvers/ProviderResolver.js
+++ b/packages/web3-providers/src/resolvers/ProviderResolver.js
@@ -22,7 +22,6 @@
 
 import isFunction from 'lodash/isFunction';
 import isObject from 'lodash/isObject';
-import Web3EthereumProvider from '../providers/Web3EthereumProvider';
 
 const global =
     (function() {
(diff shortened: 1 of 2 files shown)


JunichiSugiura:
Another solution is to add `isEIP1193' property to Web3EthereumProvider class. @nivida What do you think?

nivida:
The problem we have currently is the undefined state of the [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193.). The current state is containing accidental merged proposed changes from fabian and I would first like to finish this discussion.

packages/web3-providers/src/resolvers/ProviderResolver.js
```diff
@@ -74,6 +75,10 @@ export default class ProviderResolver {
             return provider;
         }
 
+        if (provider instanceof Web3EthereumProvider) {
```
nivida: Could you change this check to: ``if(provider.host == 'Web3EthereumProvider')`` because this is 100% mangle save :)

nivida (changes requested):

packages/web3-providers/src/resolvers/ProviderResolver.js
JunichiSugiura: Sounds good! fixed:  https://github.com/ethereum/web3.js/pull/2856/commits/9194f4146b93621a412b8ce851eda838da88ca06

JunichiSugiura:
@nivida I applied your change request and added a test for the resolver logic. Could you review this again?

nivida (approved):

PR #2862, opened 2019-06-01
Fix JSON-RPC response validation

JunichiSugiura:
## Description

Fixed the send method on Web3EthereumProvider to return response result, not response object itself otherwise [GetAccountsMethod.js#L47](https://github.com/ethereum/web3.js/blob/1.0/packages/web3-core-method/src/methods/account/GetAccountsMethod.js#L47) throws an error.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change
- [ ] Enhancement

## Checklist:

### DIFF
--- a/packages/web3-providers/src/providers/Web3EthereumProvider.js
+++ b/packages/web3-providers/src/providers/Web3EthereumProvider.js
@@ -17,7 +17,6 @@
  * @date 2018
  */
 
-import JsonRpcResponseValidator from '../validators/JsonRpcResponseValidator';
 import AbstractSocketProvider from '../../lib/providers/AbstractSocketProvider';
 
 export default class Web3EthereumProvider extends AbstractSocketProvider {
@@ -137,7 +136,7 @@ export default class Web3EthereumProvider extends AbstractSocketProvider {
         }
 
         if (result === undefined) {
-            throw new Error(`Validation error: Undefined JSON-RPC result`);
+            throw new Error('Validation error: Undefined JSON-RPC result');
         }
 
         return result;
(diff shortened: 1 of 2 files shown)


JunichiSugiura:
I’ve been reading through the [API](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#send) described in EIP1193 and become skeptical about my [fix](https://github.com/ethereum/web3.js/pull/2862/commits/408f50e6f27314431d67abc058cd6b0009270a7a) above. The example showed [here](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#examples) seems like the provider’s send method only returns the JSON-RPC response result.

In this case, the current return value from Web3EthereumProvider.send method seems right but the validation method needs to be changed to something like below.

```
// Web3EthereumProvider.js
...
async send (method, parameters) {
    let response;
    try {
        response = await this.connection.send(method, parameters);
    } catch (error) {
        throw new Error(`Node error: ${error.message}`);
    }

    if (response === undefined) {
        throw new Error(`Validation error: Undefined JSON-RPC result`);
    }

    return response;
}
...
```

@nivida Could you verify this approach, please?

packages/web3-providers/src/providers/Web3EthereumProvider.js
```diff
@@ -129,14 +128,18 @@ export default class Web3EthereumProvider extends AbstractSocketProvider {
...
+        let result;
+        try {
+            result = await this.connection.send(method, parameters);
+        } catch (error) {
+            throw new Error(`Node error: ${error.message}`);
+        }
 
-        if (validationResult instanceof Error) {
-            throw validationResult;
+        if (result === undefined) {
```
nivida: This check isn't required :)

nivida (changes requested):

JunichiSugiura:
@nivida Undefined check is now removed!

packages/web3-providers/src/providers/Web3EthereumProvider.js
JunichiSugiura: fix: https://github.com/ethereum/web3.js/pull/2862/commits/747554bce786962eae0f04db867d2365cd153551

nivida (approved):

Issue #2839, opened 2019-05-23
An intent to build a single-file, standalone web3.js


andy0130tw:
Hi there.

Since there has been no "dist" folder for nearly half a year (see #2623), I tried to build standalone (or/and minified) web3 JS files with latest release, beta 55, and gather the materials to a repository. The job is mostly done with webpack, with some caveats I encountered written in the comments.

https://github.com/andy0130tw/web3.js-browser

For now there is no built file in this repository. My plan is to also include built files after my packing stablizes. If you are a DApp developer, hope it helps you! As I am not a webpack expert, feel free to open issues for bugs or requests.

nivida:
Thanks for providing such a repository! But I will close this issue because it's a duplication of #2623. The minified file will be added back asap. The current on-going development processes are documented in the ``Projects`` tab of this GitHub repository.

PR #2898, opened 2019-06-18
web3js dependency on webpack to run on browser

pepoospina:
This would have saved me a few days of struggling. CDN bundles likes [this](https://cdn.jsdelivr.net/npm/web3@1.0.0-beta.55/dist/web3.esm.min.js) are not browser compatible neither!


andy0130tw:
Hi! For browser usage, you can have a look at my webpacked bundle in #2839. (**Disclaimer**: I am the owner of that issue)

PR #2922, opened 2019-07-02
Update provider resolver for the latest EIP1193 spec

JunichiSugiura:
Determine whether a provider is EIP1193 compatible or not by checking "send" and "on" method.


packages/web3-providers/src/resolvers/ProviderResolver.js
```diff
@@ -82,7 +82,7 @@ export default class ProviderResolver {
             return this.providersModuleFactory.createMistEthereumProvider(provider);
         }
 
-        if (provider.isEIP1193) {
+        if (typeof provider.send === 'function' && typeof provider.on === 'function') {
```
nivida: Could you put this ``if`` statement below the metamask provider check? This because the metamask provider does have a send and on method as well. I will test it after with the ``web3-provider-engine`` and the ``truffle-hdwallet-provider``

nivida (changes requested):

packages/web3-providers/src/resolvers/ProviderResolver.js
JunichiSugiura: fixed: https://github.com/ethereum/web3.js/pull/2922/commits/1f8bb4e22bc849b9d48df314650a29438a15ab80

GregTheGreek:
This is quite outdated since `web3-providers` got broken apart.

Will close unless there is objection.

PR #2927, opened 2019-07-04
Fix removeAllSocketListeners method in Web3EthereumProvider

JunichiSugiura:
This fix prevents removing listeners which are not be added by Web3.


JunichiSugiura:
`npm run dtslint` seems failed on CI. I tried locally on my laptop and it succeeds. I'm a bit confused since the error message mentions 'web3-utils' which I don't modify anything.

```
$node -v
v10.16.0
$npm -v
6.9.0
```

nivida (approved):

PR #2929, opened 2019-07-04
Length check and formatting private key in accounts.privateKeyToAccount method

### LINKED ISSUE #2928: No bounds on key length in method privateKeyToAccount()
### Description 
`web3.eth.accounts.privateKeyToAccount()` doesn't check for the input string length. Neither does it check for 

#### Expected behavior

Calling `const my_account = await web3.eth.accounts.privateKeyToAccount(<Really Long String>)` should throw an error pointing out that the string is too long for a typical Ethereum private key.

#### Actual behavior

No error is thrown. `console.log(my_account.privateKey)` prints out the huge string.

```js
const Web3 = require('web3')

async function convertKey() {
  const web3 = new Web3("http://localhost:8545")

  const account = await web3.eth.accounts.privateKeyToAccount(<REALLY LONG HEX STRING>)

  console.log(account.privateKey)

}

convertKey()
```

Save it as `.js` file and run it using `node`.

#### Additional
The problem seems to be in the [eth-lib](https://github.com/MaiaVictor/eth-lib/blob/d959c54faa1e1ac8d474028ed1568c5dce27cc7a/lib/account.js#L23) npm package. The `fromPrivate` method does not check for the key length. Also, it does not check if the first two characters are '0x', it just slices out the first two characters.

### Versions 
* web3.js: v1.0.0-beta.55
* nodejs: v12.1.0
* browser: Not used
* ethereum node: Not used

chirag-parmar:
Fixes #2928

### DIFF
--- a/packages/web3-eth-accounts/src/models/Account.js
+++ b/packages/web3-eth-accounts/src/models/Account.js
@@ -128,7 +128,7 @@ export default class Account {
         }
 
         // 64 hex characters + hex-prefix
-        if (privateKey.length != 66) {
+        if (privateKey.length !== 66) {
             throw new Error("Private key must be 32 bytes long");
         }
 
--- a/packages/web3-eth-accounts/src/models/Account.js
+++ b/packages/web3-eth-accounts/src/models/Account.js
@@ -123,10 +123,14 @@ export default class Account {
      * @returns {Account}
      */
     static fromPrivateKey(privateKey, accounts = {}) {
-        const formatPrivKey = privateKey.startsWith('0x') ? formatPrivKey : '0x' + privateKey;
+        if (!privateKey.startsWith('0x')) {
+            privateKey = '0x' + privateKey;
+        }
 
-        if (formatPrivKey && formatPrivKey.length != 66) // 64 hex characters + 0 and x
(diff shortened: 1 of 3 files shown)


packages/web3-eth-accounts/src/models/Account.js
```diff
@@ -123,6 +123,11 @@ export default class Account {
      * @returns {Account}
      */
     static fromPrivateKey(privateKey, accounts = {}) {
+        const formatPrivKey = privateKey.startsWith('0x') ? formatPrivKey : '0x' + privateKey;
```
nivida: Could you write this out to: 

``` javascript 
if (!privateKey.startsWith('0x')) { 
  privateKey = '0x' + privateKey;
}

// 64 hex characters + hex-prefix
if (privateKey.length != 66) {
            throw new Error("Private key must be 32 bytes long.");
}

... 

```

nivida (changes requested):

packages/web3-eth-accounts/src/models/Account.js
chirag-parmar: For sure. This code makes more sense.

PR #2934, opened 2019-07-10
Fix constructor name checking when reconnect websocket with options

S3bb1:
When the websocket provider has options attached they should be reattached again when reconnection the websocket. 

This is currently implemented in the websocket provider but the check in the provider checks the wrong constructor name. The constructor of the W3CWebSocket name is "W3CWebSocket" and the check checks for "W3CWebsocket"

This pull request fixes this.


nivida:
@S3bb1 I will fix the build and merge the branch after. Thanks!

Issue #2913, opened 2019-06-26
Node 12 support


ellis2323:
web3.js uses web3-eth-accounts which uses scrypt.js which uses scrypt. Scrypt is not maintained anymore and is incompatible with node 12. I've contacted scrypt.js but i'm not sure this package is well maintained. a pull request is available on scrypt.js to fix the problem.

https://github.com/barrysteyn/node-scrypt/issues/193

nivida:
Thanks for opening this issue! We have already discussed this issue and @alcuadrado opened the PR's for the scrypt.js package. The goal would be to have this fix included in the next release.

SebastianGiro:
https://github.com/ethereum/web3.js/pull/2903

@ellis2323  @nivida

levino:
Why not remove the package altogether and use something else?

nivida:
There are several ways to solve it: 

1. Moving of the logic from the ``scrypt.js`` package to Web3.js to fix and improve it.
2. Alex (axic) the maintainer and part of the Ethereum Foundation can handover us the ownership and we will update the package. (Which would be good because it's used in many other ethereum JS related packages and projects which are depending on these modules) 
3. Usage of the ``scrypt.js`` logic and publishing of a new package with a future proof fix
4. Checking if there are alternative packages with the same outcome.

The fix for this issue will be released with ``1.0.1`` and ``2.0-alpha.1``

levino:
> The fix for this issue will be released with `1.0.1` and `2.0-alpha.1`

Really? Why? This is **the most important** thing for web3? I think it is okay not to support node 12 for the time being...

SebastianGiro:
@Levino angular 8 depends on node 12. And is one of the most used frameworks out there.

levino:
Angular apps run in the browser, so not in node. Maybe the angular build tools depend on node 12 but then just use the minified browser version of web3 for your project and you should be fine...

joshstevens19:
Just to correct that statement here, current web3 can run on angular 8 fine using the standard NPM install and not the minified browser version we have it all working with no problems. 

Angular for all versions including 8 only recommends the latest stable node release it does work and supports node 10 and will even run on node 8. 

If angular forced something like that it would be so much disruption for companies that would then need to upgrade all of their build servers with Node 12 before thinking about creating an Angular 8 project. 

@SebastianGiro

levino:
[...]

PR #2905, opened 2019-06-20
Release - 1.0

nivida:
## Description

This is the stable release of v1.0.

### Compare View

[v1.0.0-beta.37 -> v1.0.0](https://github.com/ethereum/web3.js/compare/v1.0.0-beta.37...7e90479)

## Type of change

- [X] Release

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code. 
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no warnings.
- [x] I have updated or added types for all modules I've changed
- [x] Any dependent changes have been merged and published in downstream modules.
- [x] I ran ```npm run test``` in the root folder with success and extended the tests to cover all changes.
- [x] I ran ```npm run build``` in the root folder and tested it in the browser and with node.
- [x] I have tested my code with an ethereum test network.

PR #2938, opened 2019-07-12
Node >= v8.0.0 support for web3 1.0

michaelsbradleyjr:
Helps with #2913.

~~This PR is presently incomplete because it was necessary to disable two test scripts after removing `ethereumjs-wallet` as a *development* dependency, since that package has the Node 12 incompatible `scrypt` package as a transitive dependency.~~

~~Gaining support for Node versions from 8 to 12 inclusive seems like a very good thing, so maybe there is another testing strategy that can be used to restore the test files without needing to install `ethereumjs-wallet`?~~


michaelsbradleyjr:
I'm attempting to swap `ethers` for `ethereumjs-wallet`, so as to resurrect the test suites I disabled, but have to stop for now. It's mostly going well but I've run into a few problems re: fine-points of scrypt options; I hope to get them sorted out tomorrow.

WIP branch that would be merged back into this PR's branch:

https://github.com/michaelsbradleyjr/web3.js/blob/1.x-node-12-support-WIP/test/eth.accounts.encrypt-decrypt.js

The tests that are skipped (`.skip`) need to be fixed and not-skipped. But even with that changeset there are still errors when running `npm test` in the root... something to do with scrypt option parameters, I think.

levino (changes requested):
Imo this can wait until 1.0 has been released.

michaelsbradleyjr:
~The base branch for this PR will be changed once #2905 has been merged (or it can be closed and a new PR opened, whatever is preferable). It turns out there are some subtle but non-trivial behavioral differences between the `scrypt` package and Node's built-in `crypto.scrypt()`, so these changes should definitely *not* land in web3 `1.0.0`.~

michaelsbradleyjr:
The test suite has been fully revised to use the `ethers` package instead of `ethereumjs-wallet`.

Also, a bug was fixed that prevented web3 accounts from being decrypted if the user provides their own `salt` parameter — previously, the salt was not consistently converted to a `Buffer` during encryption.

michaelsbradleyjr:
I introduced an additional fallback behavior and console warning in the case that Node's built-in scrypt experiences memory exhaustion, as was happening in the test suite with the static tests. A large `n` parameter is the culprit, but that's not the case when using the 3rd party (deprecated) `scrypt` package; also `scryptsy` can handle the large `n` but of course performance suffers considerably.

michaelsbradleyjr:
See also: https://github.com/ethereumjs/ethereumjs-wallet/pull/95.

michaelsbradleyjr:
See also: https://github.com/ethereumjs/ethereumjs-wallet/pull/91#issuecomment-512522732.

michaelsbradleyjr:
I've rebased this PR against `1.x` following the release of `1.2.0` and also changed the base of the PR to `1.x`.

nivida (approved):

test/eth.accounts.encrypt-decrypt.js
```diff
@@ -178,19 +194,24 @@ describe("eth", function () {
...
+                assert.equal(acc.privateKey, ethWall.privateKey);
+            });
         });
 
         staticTests.forEach(function (test, i) {
             it("decrypt staticTests and compare to private key", function() {
-                // increase the test timeout
-                this.timeout(6000);
+                // disable the test timeout
+                this.timeout(0);
```
levino, line 214: This should be removed not set to 0.

package.json
```diff
@@ -77,7 +77,7 @@
         "coveralls": "^3.0.4",
         "crypto-js": "^3.1.9-1",
         "del": "^4.1.1",
-        "ethereumjs-wallet": "^0.6.3",
+        "ethers": "4.0.33",
```
levino, line 80: Putting quite some trust into the devs of `ether.js`. Also why is this version pinned here? The pinning should happen via `yarn.lock` or `package-lock.json` files. I think `^4.0.33` would be more appropriate.

packages/web3-eth-accounts/src/index.js
```diff
@@ -307,7 +307,7 @@ Accounts.prototype.decrypt = function (v3Keystore, password, nonStrict) {
         kdfparams = json.crypto.kdfparams;
 
         // FIXME: support progress reporting callback
```
levino, line 309: What about this "FIXME" comment? Is it still valid?

packages/web3-eth-accounts/src/index.js
```diff
@@ -351,13 +351,13 @@ Accounts.prototype.encrypt = function (privateKey, password, options) {
     if (kdf === 'pbkdf2') {
         kdfparams.c = options.c || 262144;
         kdfparams.prf = 'hmac-sha256';
-        derivedKey = cryp.pbkdf2Sync(Buffer.from(password), salt, kdfparams.c, kdfparams.dklen, 'sha256');
+        derivedKey = cryp.pbkdf2Sync(Buffer.from(password), Buffer.from(kdfparams.salt, 'hex'), kdfparams.c, kdfparams.dklen, 'sha256');
```
levino, line 354: Quite confusing that we are not using `salt` any more but `kdfparams.salt` all of a sudden. When I read the code above it should be the same, but still confusing.

packages/web3-eth-accounts/src/scrypt.js
```diff
@@ -0,0 +1,71 @@
+var scryptsy = require('scryptsy');
+
+var scrypt;
+
+var isNode = Object.prototype.toString.call(typeof process !== 'undefined' ? process : 0) === '[object process]';
+if (isNode) {
+    var NODE_MIN_VER_WITH_BUILTIN_SCRYPT = '10.5.0';
+    var NODE_MIN_VER_INCOMPAT_SCRYPT_PKG = '12.0.0';
+    var semver = require('semver');
```
levino, line 9: That is quite painful in webpack I fear... One can have webpack and node (e.g. lambda functions).

package.json
michaelsbradleyjr: It's just a devDep, used to test that web3's wallet matches up with an independent implementation. `ethereumjs-wallet` was serving that purpose but it also depends on `scrypt.js` so `npm i` in the repo still resulted in `scrypt` package being installed and compiled.

packages/web3-eth-accounts/src/index.js
michaelsbradleyjr: I think it can be removed since there's not much point in trying to make use of the progress callback of `scryptsy` and Node's built-in doesn't have such a thing.

packages/web3-eth-accounts/src/index.js
michaelsbradleyjr: See above:

https://github.com/michaelsbradleyjr/web3.js/blob/d7adf702f815537376e7b8adb8fbdacfece17435/packages/web3-eth-accounts/src/index.js#L348

```
var kdfparams = {
    dklen: options.dklen || 32,
    salt: salt.toString('hex')
};
```

That code was already there and normalized the `salt` whether it was a string or buffer. The problem was we weren't using the normalized value and it resulted in encrypted output that couldn't be decrypted. `ethereumjs-wallet` has the same bug and I made a PR to fix it (still open).

packages/web3-eth-accounts/src/scrypt.js
michaelsbradleyjr: Can you explain a bit more?  If this broke something in a node FaaS context, I will be glad to work on a fix.

levino: I am just not sure whether or not webpack correctly supports this dynamic require. Needs to be tested.

test/eth.accounts.encrypt-decrypt.js
michaelsbradleyjr: Setting to `0` is the way to say "no timeout" in mocha: https://mochajs.org/#suite-level.

Some of the static tests have large `n` values: https://github.com/michaelsbradleyjr/web3.js/blob/d7adf702f815537376e7b8adb8fbdacfece17435/test/eth.accounts.encrypt-decrypt.js#L32.

In the case that Node.js falls back to `scryptsy`, which will be the case in CI for Node `<10.5.0` (since no `scrypt`) and also `>=10.5.0` because of maxmem errors with large `n` (until [PR #28799](https://github.com/nodejs/node/pull/28799) makes it into releases of 10.x/12.x), then those tests will be prone to timeouts (`scryptsy` is much, much slower for large `n`). So for these particular tests, it makes sense to disable the timeout.

package.json
michaelsbradleyjr: As for pinning vs. `^`, I've been bitten so many times by 3rd party patch bumps that violate semver (i.e. have breaking changes) and/or bugs that I spec exact version deps without thinking about it. However, using `^` would be more consistent with what's spec'd for the other devDeps.

packages/web3-eth-accounts/src/scrypt.js
ylv-io: It doesn't. Webpack fails at dynamic require at web3@1.2.1. Any idea how to workaround it?
image

michaelsbradleyjr: I've seen the warning too, but it didn't cause my build to fail, and my build worked as expected. Are you sure you actually got a failed build?

In any case, the `web3-eth-accounts` module should be refactored to make use of `"browser"` in `package.json` instead of using runtime `isNode` detection. I'll work on that very soon.

michaelsbradleyjr: As I said in my previous comment, this should be refactored. But in the meantime I don't think the warning will result in broken builds. Here's what I tried:

```
$ cd ~/temp
$ npx create-react-app foo
$ cd foo && yarn add web3
```
After that I edited `src/App.js` to `import Web3 from 'web3'` and I set `window.Web3 = Web3`. Then:
```
$ yarn start
```
Once the dev server was available at localhost:3000 I was able to use `window.Web3` without any problems, though I did see the webpack warning in the console output of `yarn start`.

I checked that I could use `web3.eth.accounts` to create/encrypt/decrypt — worked as expected, so while the webpack warning isn't nice, it doesn't seem to prevent browser builds (w/ webpack) from working.

I'm not 100% sure about webpack builds targeting Node.js; but shouldn't be a problem because such builds couldn't pack up the `scrypt` package in any case (it's a compiled C++ thing).

ylv-io: Hey @michaelsbradleyjr you are absolutely right. I had other issue and `critical dependency` got me triggered. Thanks for detailed answer and check. Really appreciated.

0mkara:
Lots of love that this feature got merged. :heart:

PR #2939, opened 2019-07-12
Fix TypeScript 'ContractOptions' interface

geigerzaehler:
* Make all fields optional
* Add `transactionConfirmationBlocks` field
* Document all fields

Note that the list of options is still not exhaustive but I was unable to confidently determine the types of the other options.

### DIFF
--- a/packages/web3-eth-contract/types/index.d.ts
+++ b/packages/web3-eth-contract/types/index.d.ts
@@ -95,10 +95,16 @@ export interface EstimateGasOptions {
 }
 
 export interface ContractOptions {
-    from: string;
-    gasPrice: string;
-    gas: number;
-    data: string;
+    // Sender to use for contract calls
+    from?: string;
+    // Gas price to use for contract calls
+    gasPrice?: string;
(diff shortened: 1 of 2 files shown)


packages/web3-eth-contract/types/index.d.ts
```diff
@@ -95,10 +95,16 @@ export interface EstimateGasOptions {
 }
 
 export interface ContractOptions {
```
nivida: The ``ContractOptions`` interface can inherit from the ``Web3ModuleOptions`` interface in the ``web3-core`` module. :)

nivida (changes requested):

packages/web3-eth-contract/types/index.d.ts
geigerzaehler: Cool, didn’t know. Fixed in https://github.com/ethereum/web3.js/pull/2939/commits/0a0b2e7b6cf8e697c83f942307af6c995c3b4ae2.

nivida (approved):

PR #2947, opened 2019-07-15
fix/subscription_rpc_after_get_logs

helderjnpinto:
If we make and `get_logs` rpc,
 `myContract.events.allEvents({
        fromBlock: xBlock
        toBlock: yBlock 
    },
` 

After resolving events given this ranges of blocks, web3 only remove fromBlock property and send this as subscription for latest events `{"jsonrpc":"2.0","id":2,"method":"eth_subscribe","params":["logs",{"toBlock":"yBlock","topics":[],"address":"0x123...."}]} `, 
This request in many blockchains like Pantheon will rise an Invalid filter parameter, this filters need to be removed to be ok for all.

I tested this in Rinkeby test network And Pantheon from Pegasys, Geth, Ganache.

### DIFF
--- a/packages/web3-core-subscriptions/src/subscriptions/eth/LogSubscription.js
+++ b/packages/web3-core-subscriptions/src/subscriptions/eth/LogSubscription.js
@@ -66,6 +66,7 @@ export default class LogSubscription extends AbstractSubscription {
                     });
 
                     delete this.options.fromBlock;
+                    delete this.options.toBlock;
                     super.subscribe(callback);
                 })
                 .catch((error) => {
--- a/packages/web3-core-subscriptions/src/subscriptions/eth/LogSubscription.js
+++ b/packages/web3-core-subscriptions/src/subscriptions/eth/LogSubscription.js
@@ -20,8 +20,14 @@
  * @date 2018
  */
 
-import AbstractSubscription from '../../../lib/subscriptions/AbstractSubscription';
-import isFunction from 'lodash/isFunction';
+import AbstractSubscription from '../../../lib/subscriptions/AbstractSubscription'
+import isFunction from 'lodash/isFunction'
+import reject from 'lodash/reject'
+import isUndefined from 'lodash/isUndefined'
+import isObject from 'lodash/isObject'
+import isArray from 'lodash/isArray'
+import pickBy from 'lodash/pickBy'
+import isString from 'lodash/isString'
 
 // TODO: Move the past logs logic to the eth module
 export default class LogSubscription extends AbstractSubscription {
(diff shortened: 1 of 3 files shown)


packages/web3-core-subscriptions/src/subscriptions/eth/LogSubscription.js
```diff
@@ -66,6 +66,7 @@ export default class LogSubscription extends AbstractSubscription {
                     });
 
                     delete this.options.fromBlock;
+                    delete this.options.toBlock;
```
nivida: Thanks for opening this PR! Could you extend the test case for the deletion of the ``toBlock`` property?

Btw.: The fix should get added to the ``CHANGELOG.md`` file.

nivida (changes requested):

nivida (approved):

packages/web3-core-subscriptions/src/subscriptions/eth/LogSubscription.js
helderjnpinto: sorry only saw the message now, all stuff on tests are done right?

nivida:
I've thought closer about it. The ``myContract.getPastEvents(event[, options][, callback])`` is probably the correct method for getting the past events. It's also possible to pass ``allEvents`` as the event name to get all the past events. This means it's probably smarter to start the contract events subscription and to call the ``getPastEvents`` with the event name defined as ``allEvents``. The ``toBlock`` options property doesn't really make sense for a subscription because you are able to unsubscribe if a specific block number appears.

helderjnpinto:
So eth_subscribe only wants object with the following (optional) fields
address, either an address or an array of addresses. Only logs that are created from these addresses are returned (optional)
topics, only logs which match the specified topics (optional).

Besides `blockhash` maybe needs to remove if exist.
https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getlogs
https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB#supported-subscriptions

nivida:
> So eth_subscribe only wants object with the following (optional) fields
address, either an address or an array of addresses. Only logs that are created from these addresses are returned (optional)
topics, only logs which match the specified topics (optional).

> Besides blockhash maybe needs to remove if exist.

Yes, this is correct and as documented does the ``eth.subscribe('logs', options, ()=>{});`` method have the following valid option properties: 
- ``fromBlock``: The number of the earliest block. By default null.
- ``address``: An address or a list of addresses to only get logs from particular account(s).
- ``topics``: An array of values which must each appear in the log entries.

Instead of just deleting properties that aren't a part of the documented options properties could we implement a better options validation in the ``LogSubscription`` class to validate the given options and to throw an error if invalid options are existing. ``toBlock`` and other properties which aren't a part of the list above are invalid options properties for the ``logs`` subscription.

Edit: 
The missing logic for the ``logs`` subscription we have is a "fade-out process" for not losing any blocks if someone is defining the ``fromBlock`` property.

helderjnpinto:
@nivida I haven't had much time to change the code, i make some tests but i want test manually this, how i can build all web3 to test locally ? when i make npm run build , this deps will get from node_modules and not the files that i changed in this package.

GregTheGreek:
Is this still an issue?

PR #2950, opened 2019-07-16
convert salt to Buffer instance before encrypting

michaelsbradleyjr:
Previously, if a `salt` option was supplied it would be passed to `pbkdf2Sync`/`scrypt` as a string, but during decryption it was always converted to a Buffer instance such that supplying a `salt` option resulted in output that could not be decrypted.

This commit fixes the bug, and also makes encrypted output match up with the output of other wallet libraries, e.g.  `ethers`, whenever the equivalent encryption options are used consistently across libraries.


nivida (approved):

michaelsbradleyjr:
See also: https://github.com/ethereumjs/ethereumjs-wallet/pull/95.

PR #2952, opened 2019-07-17
implement fallback from built-in scrypt to scryptsy/scrypt packages

michaelsbradleyjr:
This commit introduces an additional fallback behavior and console warning in the case that Node's built-in scrypt experiences memory exhaustion, as was discovered in the 1.x test suite with the static tests (see #2938). A large `n` parameter is the culprit, but that's not the case when using the 3rd party (deprecated) scrypt package. The `scryptsy` package can handle a large `n`, but performance does suffer considerably.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Pull request & issue templates updated
 - Supported node versions changed (#2820)
+- fsevents bumbed to v1.2.9 (#2951)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


nivida (approved):

michaelsbradleyjr:
See also: https://github.com/ethereumjs/ethereumjs-wallet/pull/91#issuecomment-512522732.

Astrantia:
@michaelsbradleyjr @nivida This PR has resulted into this compilation warning
```
[ warn ]  ./node_modules/web3-eth-accounts/dist/web3-eth-accounts.umd.js
Module not found: Can't resolve 'scrypt' in 'C:/App/node_modules/web3-eth-accounts/dist'
```

Can this PR be reverted, or this bug fixed?

PR #2980, opened 2019-07-29
Added support for custom http agent

AsyncLegs:
Thus we will able to use custom HTTP agents such as https://github.com/koichik/node-tunnel or https://www.npmjs.com/package/https-proxy-agent


AsyncLegs:
Hi guys, is there any chances to merge these enhancements to version 1.x?

packages/web3-providers-http/src/index.js
```diff
@@ -46,6 +46,7 @@ var HttpProvider = function HttpProvider(host, options) {
     }
     this.timeout = options.timeout || 0;
     this.headers = options.headers;
+    this.agent = options.agent || this.agent;
```
nivida: Thanks for opening this PR! This line doesn't get used below and the method ``nodeJsSet()`` does accept the following options object:

``` typescript
interface {
  httpAgent: Agent;
  httpsAgent: Agent;
  baseUrl: string;
}
``` 

Could you implement it in a way to have the possibility to pass all ``nodejsSet`` related option object properties with respecting of the automatic detection of the correct agent on line [42](https://github.com/ethereum/web3.js/blob/fc76d4a22f745d84c13767a326437b9a81f00044/packages/web3-providers-http/src/index.js#L42)?

nivida (changes requested):

packages/web3-providers-http/src/index.js
AsyncLegs: Thanks for the review! I will adjust the PR asap.

nivida:
@cgewecke I've applied the requested changes as described above from me. I've noticed while improving the options handling that the existing HTTP provider test cases aren't that good as they could be but this has to get addressed with a separate PR.

packages/web3-providers-http/src/index.js
```diff
@@ -34,18 +34,19 @@ var https = require('https');
...
+    this.connected = false;
+
     var keepAlive = (options.keepAlive === true || options.keepAlive !== false) ? true : false;
     this.host = host || 'http://localhost:8545';
-    if (this.host.substring(0,5) === "https") {
+    if (!this.agent && this.host.substring(0,5) === "https") {
         this.httpsAgent = new https.Agent({ keepAlive: keepAlive });
     } else {
         this.httpAgent = new http.Agent({ keepAlive: keepAlive });
     }
```
cgewecke, line 51: @nivida Could you eli5 me why http and https are treated differently here depending on whether `options.agent` is defined or not?

cgewecke (commented):
@nivida Is it a problem to add a simple test using the new option though? 

To make sure the changes work as expected and the option is described by the tests in case someone wants to see a working example.

Also WDYT about adding `withCredentials` and `agent` to the usage example in the web3-providers-http README?

packages/web3-providers-http/src/index.js
nivida: Oh, thanks! I've missed the else statement 👌

nivida:
> @nivida Is it a problem to add a simple test using the new option though?
> To make sure the changes work as expected and the option is described by the tests in case someone wants to see a working example.

I will add those missing test cases 👌 

> Also WDYT about adding withCredentials and agent to the usage example in the web3-providers-http README?

Yep, I will add those options to the README. 

Btw.: An issue is existing to add and improve the provider documentations to the web3.js documentation on readthedocs (https://github.com/ethereum/web3.js/issues/3117).

nivida (approved):

test/httpprovider.js
```diff
@@ -11,15 +13,60 @@ var HttpProvider = SandboxedModule.require('../packages/web3-providers-http', {
     singleOnly: true
 });
 
-describe('web3-providers-http', function () {
+describe.only('web3-providers-http', function () {
```
cgewecke:
```suggestion
describe('web3-providers-http', function () {
```

cgewecke (approved):
@nivida Looks good! Nice.

There's a `.only` on the new test which is stopping the complete test suite from running. (Tried to commit it but I don't have permission on this branch).

PR #2995, opened 2019-08-02
websocket dependency updated

nivida:


leviadam:
Hi @nivida,

Using dependencies from github repos is not a good practice.
If you prefer to use a fork why not also release a package?

Anaphase:
Yeah, this change broke my deployments because my EC2 instances didn't have git installed 🙄

e00dan:
Is there a reasoning behind this change? :)

PR #2997, opened 2019-08-03
If the data is '0x' change to ''

rotcivegaf:


ryanio:
hey @rotcivegaf thanks for this PR and sorry it took SO long to get to haha.

In [v1.2.8](https://github.com/ethereum/web3.js/releases/tag/v1.2.8) we actually upgraded to the new ethers abi decoder so I added your decodeLog test and it looks like it passes so the issue should be resolved. If you find otherwise please feel free to re-open.

PR #3015, opened 2019-08-08
Fix accessing event.name where event is undefined

### LINKED ISSUE #3014: Accessing property name of undefined
This line below will cause an unhandled, since event isn't defined

```
539     if (!event) {
540         throw new Error('Event "' + event.name + '" doesn\'t exist in this contract.');
541     }
```

https://github.com/ethereum/web3.js/blob/2022b17d52d31ce95559d18d5530d18c83eb4d1c/packages/web3-eth-contract/src/index.js#L540

micahriggan:
closes #3014

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+services:
+    - xvfb
 language: node_js
 node_js:
     - "8"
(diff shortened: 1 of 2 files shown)


micahriggan:
@nivida any tips for the `The command "sh -e /etc/init.d/xvfb start" failed and exited with 127 during .` issue with travis? Did I do something wrong?

nivida:
@micahriggan Yes, the fix would be to add ``xvfb`` as service as described [here](https://benlimmer.com/2019/01/14/travis-ci-xvfb/).

nivida (approved):

PR #3026, opened 2019-08-14
Avoid using IIFE to require scrypt

ppoliani:
This should fix https://github.com/ethereum/web3.js/issues/3018

### DIFF
--- a/packages/web3-eth-accounts/src/scrypt.js
+++ b/packages/web3-eth-accounts/src/scrypt.js
@@ -14,7 +14,7 @@ if (isNode) {
         return function() {
             if (scryptPkg !== undefined) { return scryptPkg; }
             try {
-                scryptPkg = (function(m) { return require(m); })('scrypt');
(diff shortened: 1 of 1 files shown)


ppoliani:
Not sure why the build fails. All tests seem to pass. Is there something that is missing?

michaelsbradleyjr:
@ppoliani the current build pipeline doesn't deal well with static `require` in reference to a package that may be supplied by the person who installed it alongside `web3`, i.e. the `scrypt` package is  no longer a dependency of `web3` itself. That's why I introduced the dynamic require. The webpack warning that introduces doesn't cause builds to fail strictly speaking; however some developers' build pipelines may be setup to fail outright even on a warning.

See: https://github.com/ethereum/web3.js/pull/2938#discussion_r311667673.

A better solution is in the works, hopefully I'll get the PR up tomorrow.

nivida:
Closed because of the newly opened PR‘s: #3036 #3035

PR #3028, opened 2019-08-15
subscription.js Emit 'connected' event with subscription.id

lbertenasco:
Would something like this be possible to add?
It's useful when testing WS subscriptions successful connections.
Instead of retrying get `subscription.id` until it's defined.


nivida:
This could be added to a provider. The new 2.0 socket-based provider does already have that.

nivida (changes requested):
Thanks for proposing this PR! Could you please update the changelog file located in the root folder of this branch? and also extend the documentation?

lbertenasco:
> Thanks for proposing this PR! Could you please update the changelog file located in the root folder of this branch? and also extend the documentation?

Changelog and docs updated 👍

nivida (approved):

PR #3031, opened 2019-08-16
Fix for when localStorage is available but access is restricted

chaitanyapotti:


miguelmota:
Great fix @chaitanyapotti! Hope this gets merged

alcuadrado:
Hey @chaitanyapotti! Thanks for this PR. 

Can you explain a little what it does, and especially, how (the error's checks)? It'd be great to have that info in the code.

chaitanyapotti:
@alcuadrado [Here](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Feature-detecting_localStorage) is the relevant documentation

packages/web3-eth-accounts/src/index.js
```diff
@@ -525,10 +525,35 @@ Wallet.prototype.load = function (password, keyName) {
...
+    var storage;
+    try {
+        storage = window[type];
+        var x = '__storage_test__';
+        storage.setItem(x, x);
+        storage.removeItem(x);
+        return true;
+    }
+    catch(e) {
+        return e && (
```
alcuadrado, line 552: The MDN code has `return e instanceof DOMException && (` in this line. Is this equivalent? How?

alcuadrado:
Thanks for that link @chaitanyapotti. Do you mind adding it as a comment in the function?

packages/web3-eth-accounts/src/index.js
chaitanyapotti:
```js
e instanceof DOMException
```
returns a boolean which represents whether `e` is an instance of `DOMException` or not.
The correct way of doing this would be to use a [polyfill for DOMException](https://github.com/inexorabletash/polyfill/blob/master/obsolete/domexception.js) but that would be too much code.

If this code is ever executed in a browser and caught, it would always be a [DOMException](https://developer.mozilla.org/en-US/docs/Web/API/DOMException) because `storage` is part of Web API. Hence, i don't really check whether it's a DOMException or not.

In case of node, window wouldn't be defined and it would be caught and returns false

nivida (approved):

chaitanyapotti:
@nivida Hey, any timeline on when this will be released?

miguelmota:
Is this feature detection available in the latest npm package release yet?

smithki:
I'm wondering the same. Is there a timeline on this release? ~(or is it already available?)~

EDIT: I confirmed for myself that the code is not yet released. This is blocking me on Brave support right now 😢

miguelmota:
I ended up using [ethers.js](https://github.com/ethers-io/ethers.js/) which is much more lightweight and doesn't have issues like this

PR #3035, opened 2019-08-16
Replace 1.x scrypt shim of web3-eth-accounts with https://github.com/web3-js/scrypt-shim

### LINKED ISSUE #3018: webpack  scrypt.js: Critical dependency: the request of a dependency is an expression
### Description 
Building on v.1.2.1 using create-react-app fails due to webpack warning `/node_modules/web3-eth-accounts/src/scrypt.js Critical dependency: the request of a dependency is an expression"` when CI=true (ie, fail on warnings)

#### Expected behavior

Building a create-react-app using webpack should succeed

#### Actual behavior

I am able to build everything locally, however the build fails in my CI environment. The warning is as follows:
```
/node_modules/web3-eth-accounts/src/scrypt.js
Critical dependency: the request of a dependency is an expression
```
I assume it is a result of this line:
https://github.com/ethereum/web3.js/blob/1.x/packages/web3-eth-accounts/src/scrypt.js#L17

#### Error Logs 
```
Creating an optimized production build...

Treating warnings as errors because process.env.CI = true.
Most CI servers set it automatically.

Failed to compile.

/home/circleci/project/node_modules/web3-eth-accounts/src/scrypt.js
Critical dependency: the request of a dependency is an expression


info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run build stderr:
error Command failed with exit code 1.
```

### Versions 
* web3.js: v1.2.1
* nodejs: v10.16.0
* browser: circleci 
* ethereum node:

michaelsbradleyjr:
Fixes #3018


ppoliani (approved):

gnidan (approved):

michaelsbradleyjr:
I'm working toward a PR that will further improve the `web3-js/scrypt-shim` package, should be done soon; hopefully today, maybe tomorrow.

Also, I need to do an experiment, but I'm fairly certain already that the present strategy in web3 to point to "rolling updates" branches on GitHub for `web3-js/WebSocket-Node` and `web3-js/scrypt-shim` is somewhat flawed. That is, I believe web3 users' `package-lock.json` or `yarn.lock` files pin to a specific commit and then later updates will require manual intervention (in the lockfiles!). Instead, I think those packages should be published on the registry, scoped as `@web3-js` or whatever is appropriate/possible, and when changes are made to them new versions should be published as usual. If we want to stick with pointing to GitHub for those packages, then I believe we should use tags in the `github:` specifiers rather than a branch name that's always the same, i.e. a new tag can be created when there's an update.

PR #3047, opened 2019-08-26
adding FUNDING.yml through Unlock

julien51:
Hello!

First of all, thanks for your (hard) work!

As an OpenSource project, I believe Web3.js should have a `FUNDING.yml` file which lets people make donations if they enjoy the software.

This pull requests adds donation thru an [Unlock](https://unlock-protocol.com/) lock. Locks are smart contract which let people become members by purchasing a key. The key itself is a non fungible token: 🗝 . (you can view the NFT on [OpenSea](https://opensea.io/assets/0x0364c5d9c268a1870f49ad2819801624d8ce0ff0/1) for example)

This specific lock is at [`0xe8846de6aaf1b50a3e300cc2ab568d544709385b`](https://etherscan.io/address/0xe8846de6aaf1b50a3e300cc2ab568d544709385b). The current key price is 1 DAI. I currently own the lock but I'll happily transfer its ownership to you if you can send me an address ;) Keys are valid for 1 month and there can be an infinite number of keys.

The pull-request adds 2 files: `FUNDING.yml` which is a github standard file to add a "donation button" with the right links when people click on it and a file which includes the configuration for the donation page.

Here is what it looks like 
image

You can try this on another project which already uses the donation button such as [Unlock](https://github.com/unlock-protocol/unlock/), [MyCrypto](https://github.com/mycryptohq/mycrypto), or [WalletConnect](https://github.com/WalletConnect/walletconnect-monorepo/)...

When/if you merge this you also need to enable donations in the settings (see https://github.com/ethereum/web3.js/settings): 

image

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+custom: https://donate.unlock-protocol.com/?r=ethereum/web3.js
--- a/.unlock-protocol.config.json
+++ b/.unlock-protocol.config.json
@@ -1,0 +1,14 @@
+{
+   "persistentCheckout": true,
+   "icon": "https://raw.githubusercontent.com/ethereum/web3.js/2.x/assets/web3js.jpg",
+   "callToAction": {
+     "default": "Please, support our work by purchasing a key to our supporters lock. You will receive a unique Web3.js Non Fungible Token!",
+     "pending": "Thank you for your support. It means a lot to us!",
+     "confirmed": "Thank you for your support. It means a lot to us!"
+   },
+   "locks": {
+     "0xe8846DE6aAF1B50A3e300cc2AB568D544709385B": {
+       "name": "Web3.js Supporters"
+     }
+   }
+ }


julien51:
Hello @nivida @frozeman  ! I'd love to get your feedback on that one. We blogged about the [motivations behind that PR](https://unlock-protocol.com/blog/unlock-donations/).

One of the next steps that we're thinking of is to create some kind of "shared" lock between many Ethereum OSS projects. The idea is that people can then chose to support a single project, like Ethlint, or the whole ethereum OSS ecosystem. This shared lock would belong to some kind of shared address (or even a DAO) to organize the funds and maybe distribute them to specific projects.

Anyway, my point is that I'd love to kick off a discussion here :)

nivida:
Thanks for opening this PR I will check it after my vacation and the Swiss Army course every swiss citizen has to attend.

julien51:
Ha! Not sure which is going to be most fun ;p

In the meantime, I wrote more about [that DAO idea](https://www.reddit.com/r/unlock_protocol/comments/d1w8ly/a_lock_owned_by_a_dao_to_fund_open_source_in/).

GregTheGreek:
Hey im not sure the status of this, but i think it would make sense to close this unless there is objection.

Issue #3053, opened 2019-08-31
Contract events are not filtered by string type value


yzevm:
### Description 

- [x] Bug

#### Expected behavior

Calling `contract.events.MyEvent({filter: {a: "string"}})` should return an event with same `string` index.

#### Actual behavior

Calling `contract.events.MyEvent({filter: {a: "string"}})` returns `null`.

#### Steps to reproduce the behavior

Open the `repl.it` link and press `run` on the top https://repl.it/@egorzaremba/web3-contract-filter-event-by-string

Also, contract address is https://rinkeby.etherscan.io/address/0x8f044afc2202ea44790f0081ea964b02f05783bd#code

### Versions 
* web3.js: 2.0.0-alpha.1
* nodejs: 10.2.0
* ethereum node: infura

nivida:
Thanks for opening this issue! I will check this closer asap.

nivida:
Closed because dynamically allocated types as for example ``string`` can't be defined as an indexed event property. ([stackexchange discussion about](https://ethereum.stackexchange.com/questions/6840/indexed-event-with-string-not-getting-logged/7170#7170))

nivida:
re-opened it: It will add the string to the ``data`` array and it creates a hash of it and adds the hash to the ``topics`` array.

Issue #344, opened 2015-11-16
Arrays and strings as event topics


chriseth:
Arrays (including `bytes` and `string`) cannot be used as event topics as they are. Solidity automatically sha3-hashes such arguments them before setting them as topic. It might be worthwhile to do this also automatically in web3.js when filtering for events.

frozeman:
Thats a restrictions of solidity index params.

frozeman:
Does solidity also hashes strings and bytes longer than 32 bytes?

PR #3061, opened 2019-09-09
fix(eth-contract): encodeParam string type to keccak256

yzevm:
Fixes: #3053 #344 

Right now only [EventFilterEncoder](https://github.com/ethereum/web3.js/blob/2.x/packages/web3-eth-contract/src/encoders/EventFilterEncoder.js#L52) calls `encodeParameter` method

I decided in this way, but if you have any suggestion, I can change this PR

I believe its better to send string message as is. Without any `utils.toHex('STRING_MESSAGE')` and etc

```js
const contract = new web3.eth.Contract(CONTRACT_ABI, CONTRACT_ADDRESS)
  contract.events.CatchMeIfYouCan({
      filter: {b: "STRING_MESSAGE"},
      fromBlock: 5000000
  }, (__, event) => {
    console.log('_____filter_____')
    console.log(event)
  })
```


nivida:
Hey! Thanks for opening this PR. I've just checked the bug closer and forgot about that dynamically allocated types as for example ``string`` can't be defined as an indexed event property. ([stackexchange discussion about](https://ethereum.stackexchange.com/questions/6840/indexed-event-with-string-not-getting-logged/7170#7170))

nivida:
re-opened it: It will add event properties that are indexed but dynamically allocated types to the ``data`` array and it creates a keccak256 hash of it and adds this hash to the ``topics`` array.

The ``encodeParameter`` method is exposed also in the public API of the ``AbiCoder`` which is accessible over ``web3.eth.abi.encodeParameter``. This means that this fix should be applied in the ``EventFilterEncoder``. I apologize for the confusion.

yzevm:
- [x] Tested with single string value on Rinkeby network
- [x] Tested with array of string values on Rinkeby network

At first, I would like to make private method, but didn't find any this type of methods in web3.js lib (JSDoc). Then I thought about static method, but only `encode` method will use it. So I decided to do "private" closure like [here](https://github.com/ethereum/web3.js/blob/2.x/packages/web3-eth-iban/src/Iban.js#L31), if I wrong, please let me know, I'll fix it

code to test

```js
const Web3 = require('web3')
const web3 = new Web3(new Web3.providers.WebsocketProvider('wss://rinkeby.infura.io/ws'))

const CONTRACT_ADDRESS = '0x8f044afc2202ea44790f0081ea964b02f05783bd'
const CONTRACT_ABI = [ { "constant": false, "inputs": [ { "name": "_str", "type": "string" } ], "name": "click", "outputs": [], "payable": false, "stateMutability": "nonpayable", "type": "function" }, { "anonymous": false, "inputs": [ { "indexed": true, "name": "a", "type": "address" }, { "indexed": true, "name": "b", "type": "string" }, { "indexed": true, "name": "c", "type": "uint256" } ], "name": "CatchMeIfYouCan", "type": "event" } ];

(() => {
  const contract = new web3.eth.Contract(CONTRACT_ABI, CONTRACT_ADDRESS)
  contract.events.CatchMeIfYouCan({
      filter: {b: ['qwerty', 'second']},
      fromBlock: 5000000
  }, (__, event) => {
    console.log('_____filter_____')
    console.log(event)
  })
})()

```

packages/web3-eth-contract/src/encoders/EventFilterEncoder.js
```diff
@@ -22,6 +22,21 @@
...
 import isArray from 'lodash/isArray';
 
+/**
+ * @param {String} type
+ * @param {String | Number} param
+ * @param {AbiCoder} abiCoder
+ *
+ * @returns {String} encoded plain param
+ */
+const encodeParameter = (type, param, abiCoder) => {
```
nivida: On the second look after my vacation is it probably better to extend the ``AbiCoder`` with an ``encodeTopic`` method and to handle all dynamically allocated types. Sorry.

Private methods do currently not exists in web3.js but will be added asap. 

Proposals for private properties and methods:
- https://github.com/tc39/proposal-private-methods 
    + https://babeljs.io/docs/en/babel-plugin-proposal-private-methods
- https://github.com/tc39/proposal-private-fields
    + https://babeljs.io/docs/en/babel-plugin-proposal-class-properties

web3.js issue: https://github.com/ethereum/web3.js/issues/2802

nivida (changes requested):

packages/web3-eth-abi/src/AbiCoder.js
```diff
@@ -268,4 +268,22 @@ export default class AbiCoder {
...
+     *
+     * @method encodeTopic
+     *
+     * @param {String} type
+     * @param {String | Number} param
+     *
+     * @returns {String} encoded plain param
+     */
+    encodeTopic(type, param) {
+        if (type === 'string' || type === 'string[]') {
```
nivida, line 283: Thanks for doing those changes! 

The cases for ``bytes``, ``structs`` and ``Array``'s (dynamically and statically) are missing. The solidity documentation does describe all the encoding rules [here](https://solidity.readthedocs.io/en/v0.5.12/abi-spec.html#encoding-of-indexed-event-parameters).

nivida (changes requested):

yzevm:
I see that 1.x.x branch went far ahead, not sure that 2.x.x branch will alive in this state, so I close this PR for clean purposes

P.S. I'll keep in touch with 1.3.x discussion

PR #3062, opened 2019-09-10
Bug fix: Fix regeneratorRuntime error in browser

eggplantzzz:
In response to https://github.com/ethereum/web3.js/issues/3058

After researching the error mentioned in the issue above, it seems `@babel/preset-env` should prevent this error. Currently the package `babel-preset-env` is being used during bundling which is out of date (deprecated). The error no longer occurs after using the current version of this package (which is now `@babel/preset-env`), adding a necessary dependency (`@babel/core`), and updating `gulp-babel`.


nivida (changes requested):
Thanks for proposing this PR! Could you update the minified file as well?:)

eggplantzzz:
Sure!

PR #3067, opened 2019-09-13
fix inputGetLogsFormatter return value

toffick:
Fix this issue https://github.com/ethereum/web3.js/issues/3066


nivida:
Thanks for proposing this PR but the last maintenance patch for 0.20.x got released a year ago and we recommend to update to the latest version of 1.0.

PR #3078, opened 2019-09-25
fixed inconsistency between signTransaction methods of web3.eth and web3.eth.accounts

akhil2015:
Fixed the bug reported in https://github.com/ethereum/web3.js/issues/3029


nivida (changes requested):
Thanks for proposing this PR! Could you keep the ``rawTransaction`` property and define a ``getter`` for it as described [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get). This getter should print a console warning and inform the developer about the deprecation of the ``rawTransaction`` property. 

``The property rawTransaction is deprecated and will be removed with version 1.3.0``

Can you also update the related [documentation](https://web3js.readthedocs.io/en/v1.2.1/web3-eth-accounts.html#signtransaction) and the ``CHANGELOG.md`` file in the root folder of this branch?

packages/web3-eth-accounts/src/index.js
```diff
@@ -171,13 +171,24 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
+                    to: transaction.to,
+                    value: transaction.value,
+                    input: transaction.data,
+                    v: '0x' + Buffer.from(ethTx.v).toString('hex'),
+                    r: '0x' + Buffer.from(ethTx.r).toString('hex'),
+                    s: '0x' + Buffer.from(ethTx.s).toString('hex'),
+                    hash:transactionHash
+                },
+                get rawTransaction(){
+                    console.warn('The property rawTransaction is deprecated and will be removed with version 1.2.3');
```
nivida, line 189:
```suggestion
                    console.warn('The property rawTransaction is deprecated and will be removed with version 1.3.0');
```

nivida (changes requested):

packages/web3-eth-accounts/src/index.js
nivida: Could you also update the related documentation?

packages/web3-eth-accounts/src/index.js
```diff
@@ -171,13 +171,24 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
             var transactionHash = utils.keccak256(rawTransaction);
 
-            return {
-                messageHash: '0x' + Buffer.from(ethTx.hash(false)).toString('hex'),
-                v: '0x' + Buffer.from(ethTx.v).toString('hex'),
-                r: '0x' + Buffer.from(ethTx.r).toString('hex'),
-                s: '0x' + Buffer.from(ethTx.s).toString('hex'),
-                rawTransaction: rawTransaction,
-                transactionHash: transactionHash
+            result = {
```
cgewecke, line 174: Isn't this a breaking change? All the keys currently enumerated on the return object would need to be preserved.

test/eth.accounts.signTransaction.js
```diff
@@ -453,10 +453,9 @@ describe("eth", function () {
                     var testAccount = ethAccounts.privateKeyToAccount(test.privateKey);
                     assert.equal(testAccount.address, test.address);
 
-                    testAccount.signTransaction(test.transaction).then(function (tx) {
-                        assert.equal(tx.messageHash, test.messageHash);
-                        assert.equal(tx.transactionHash, test.transactionHash);
-                        assert.equal(tx.rawTransaction, test.rawTransaction);
```
cgewecke, line 459: All these assertions should remain if this change is to be non-breaking.

test/eth.accounts.signTransaction.js
```diff
@@ -471,7 +470,7 @@ describe("eth", function () {
                     transaction.to = transaction.toIban;
                     delete transaction.toIban;
                     testAccount.signTransaction(transaction).then(function (tx) {
-                        assert.equal(tx.rawTransaction, test.rawTransaction);
```
cgewecke, line 474: This assertion should remain if this change is to be non-breaking.

test/eth.accounts.signTransaction.js
```diff
@@ -497,7 +496,7 @@ describe("eth", function () {
                     testAccount.signTransaction(transaction)
                     .then(function (tx) {
                         assert.isObject(tx);
-                        assert.isString(tx.rawTransaction);
```
cgewecke, line 500: This assertion should remain if this change is to be non-breaking.

test/eth.accounts.signTransaction.js
```diff
@@ -524,7 +523,7 @@ describe("eth", function () {
...
```
cgewecke, line 527: This assertion should remain if this change is to be non-breaking.

test/eth.accounts.signTransaction.js
```diff
@@ -551,7 +550,7 @@ describe("eth", function () {
...
```
cgewecke, line 554: This assertion should remain if this change is to be non-breaking...and same below...

test/contract.js
```diff
@@ -3031,7 +3031,7 @@ describe('typical usage', function() {
                 chainId: '0x1',
                 nonce: '0x1',
             }).then(function (tx) {
-                const expected = tx.rawTransaction;
```
cgewecke, line 3034: Can we also check that this one works, instead of only removing?

cgewecke (changes requested):
I think the return object's interface has to preserve the existing keys, even if lots of duplicate info is added to normalize the object across the two methods.

test/contract.js
nivida: @cgewecke rawTransaction still exists and prints a deprecation message as commented above.

nivida:
> I think the return object's interface has to preserve the existing keys, even if lots of duplicate info is added to normalize object across the two methods.

@cgewecke The property still exists and it will print a deprecation message: https://github.com/ethereum/web3.js/pull/3078/files#diff-7bb2a20126193b9ecfe4723f83429c49R188

some comments above: 
> Thanks for proposing this PR! Could you keep the rawTransaction property and define a getter for it as described here. This getter should print a console warning and inform the developer about the deprecation of the rawTransaction property.
``The property rawTransaction is deprecated and will be removed with version 1.3.0``

cgewecke:
@nivida Yes, but what about these properties?
```javascript
messageHash: '0x' + Buffer.from(ethTx.hash(false)).toString('hex'),
v: '0x' + Buffer.from(ethTx.v).toString('hex'),
r: '0x' + Buffer.from(ethTx.r).toString('hex'),
s: '0x' + Buffer.from(ethTx.s).toString('hex'),
transactionHash: transactionHash
```
They are moved to under `tx` instead of being on the outside layer. (And messageHash is missing?)

If someone has code which accesses these properties at their current level won't they be undefined?

test/eth.accounts.signTransaction.js
cgewecke: Unresolving for the moment - these test changes are an example of how the PR breaks the current return object's interface.

akhil2015:
ok i will preserve the existing properties and print a deprecation message for each one of them

PR #3086, opened 2019-10-01
Fix hexToNumber and hexToNumberString  prefix validation

### LINKED ISSUE #2184: hexToNumberString sometimes interprets string as decimal
```
> web3.utils.hexToNumberString('13')
'13'
```

cgewecke:
Fixes #2184 in the 1.x branch by back-porting #2884.

**Notes** 

+ PR fixes both `hexToNumber` and `hexToNumberString` - #2884 only touched `hexToNumberString`. Possible 2.x needs a small addition there. 

+ The existing `hexToNumber` tests look like they were mistakenly copy-pasted from `hexToNumberString` so have amended those.


packages/web3-utils/src/utils.js
```diff
@@ -231,6 +231,10 @@ var hexToNumber = function (value) {
         return value;
     }
 
+    if (_.isString(value) && !isHexStrict(value)) {
```
cgewecke, line 233: This is probably a really bad idea (for hexToNumber). There are many places in the code where it's applied to values which might be decimal strings. Current behavior is to pass along to `toBN` without complaint..

test/utils.toNumber.js
```diff
@@ -2,14 +2,22 @@ var assert = require('assert');
...
-    describe('hexToNumberString', function () {
+    describe('hexToNumber', function () {
         it('should return the correct value', function () {
 
-            assert.equal(utils.hexToNumberString("0x3e8"), 1000);
-            assert.equal(utils.hexToNumberString('0x1f0fe294a36'), 2134567897654);
+            assert.equal(utils.hexToNumber("0x3e8"), 1000);
+            assert.equal(utils.hexToNumber('0x1f0fe294a36'), 2134567897654);
             // allow compatiblity
-            assert.equal(utils.hexToNumberString(100000), 100000);
```
cgewecke, line 11:
```javascript
// allow compatibility
```
... is an important warning. Maybe these tests should be unchanged and we should clarify what intended hexToNumber behavior is here and in the [docs](https://web3js.readthedocs.io/en/v1.2.1/web3-utils.html#hextonumber).

nivida: The function ``hexToNumber`` should do what the name is saying it's doing. I would enforce to have all passed hex values prefixed with ``0x``. In the case the passed value isn't prefixed with ``0x`` should it throw an error.

packages/web3-utils/src/utils.js
nivida: We should enforce to have hex values prefixed with ``0x`` or it should throw an error. Can you reference the code parts where a decimal string gets passed to ``hexToNumber``?

cgewecke: You are right about this - I will look through and try to see if there is anything (because I'm still afraid :)

test/utils.toNumber.js
cgewecke: Yes it should. Current behavior is a little unusual.

packages/web3-utils/src/utils.js
cgewecke: @nivida Ok I looked and everything going to hexToNumber is output formatting so it should be fine. If you want to merge that's cool.

I would like to double-check this against the E2E PR before you publish a new version and add a test for events/logs there to make sure there's nothing weird coming from ganache.

cgewecke:
@nivida I checked the `hexToNumber` change on the E2E PR (#3122) and it looks ok... 

I think you're correct and this is merge-able.

Issue #2401, opened 2019-02-20
Encoding incompatibility in status transaction receipt


alepc253:
#### Expected behavior

Any transaction receipt with `0x01` as result should be considered successful. 

#### Actual behavior

Function `isValidReceiptStatus`  in `TransactionReceiptValidator.js` doesn't consider `0x01` as valid. 

#### Steps to reproduce the behavior

This is seen in an RSK node but could happen in any node that manages this type of hexa encoding. 

1. Connect to an RSK network (i.e. `https://public-node.testnet.rsk.co`)
2. Execute this command: `web3.eth.getTransactionReceipt("0x3f7c81902c008538d5c9f39115a31e14036fc6b9110899239660f074203f3cc1")`
3. Check in the receipt that status value is `0x01`

Therefore, if you apply `isValidReceiptStatus`  from `TransactionReceiptValidator.js` to that receipt, it will throw an exception with this message: "Transaction has been reverted by the EVM".

#### Versions
Truffle v5.0.4 (core: 5.0.4)
Solidity v0.5.0 (solc-js)
Node v11.10.
web3.js 1.0.0-beta.46

nivida:
Thanks for creating this issue! I will check and fix it asap.

PR #3090, opened 2019-10-02
Allow 'null' receipt.status values

### LINKED ISSUE #2183: Possible incorrect parsing of receipt status
The Byzantinum fork, IIRC, enabled adding the `status` to the transaction receipts. But for chains that did not activate it, the `status` is still `null` instead of `0x0` or `0x1`. Following is the code that parses it to boolean:

https://github.com/ethereum/web3.js/blob/1bbfc5a36f4422a79bb0abfdd4b30aba6b3dc340/packages/web3-core-helpers/src/formatters.js#L219-L221

When testing with some transactions in a chain that did not fork, we saw the `status` field was incorrectly parsed as `false` for receipts with `status` coming as `null`, which might not be correct as `status` is actually not provided. This is not enough to say the transaction failed or not.

The condition might be changed to something like the following to properly account for `null`, `0x0` and `0x1` values:

```js
    if(typeof receipt.status === 'string') {
    }
```

@frozeman & team, can you please provide your thoughts?

cgewecke:
Fixes #2183 and verifies that #2401 is fixed on 1.x.  

Back-ports 2.x `outputTransactionReceiptFormatter` tests and adding cases for statuses: 
+ `0x1`, 
+ `0x01` 
+ `undefined`
+ `null`

NB: `null` and `undefined` are passed through uninterpreted & unmodified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,5 +56,6 @@ Released with 1.0.0-beta.37 code base.
 
 ### Fixed
 
+- Fix error incorrectly thrown when receipt.status is `null` (#2183)
 - regeneratorRuntime error fixed (#3058)
 - Fix accessing event.name where event is undefined (#3014)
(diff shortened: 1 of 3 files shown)


nivida (changes requested):
Thanks! Could you update the changelog?

nivida (approved):

PR #3096, opened 2019-10-02
feat(Network): add support for net id of Goerli

JayWelsh:
2.x Support for identification of Goerli test net via getNetworkType function and updated test case


nivida (changes requested):
Thanks for proposing this PR! Could you update the changelog file in the root folder of the 2.x branch?

nivida (approved):

Issue #2378, opened 2019-02-15
Invalid chainId used to sign transactions


gabmontes:
#### Expected behavior

When signing transactions in Ethereum Classic (might be applicable to other chains as well), the signing logic in the accounts package shall use the proper `chainId` (i.e. 61/62) if it is not specified in the transaction object by calling `eth_chainId` RPC.

#### Actual behavior

In the above mentioned case, the chainId value is actually retrieved using the `net_version` RPC call, which does return an incorrect value (prior EIP-155).

#### Steps to reproduce the behavior

1. Connect web3 to an Ethereum Classic node.
2. Create a wallet/account so allow web3 to sign the txs.
3. Send tx without specifying the chain ID.
4. Check the transaction was not created using a `chainId` value 61/62.

#### Error Logs

N/A 

#### Versions
Web3.js@1.0.0-beta.37

Related to #1644.

A quick fix for beta.37 and previous versions is to apply this patch:

```diff
--- a/node_modules/web3-eth-accounts/src/index.js
+++ b/node_modules/web3-eth-accounts/src/index.js
@@ -69,7 +69,7 @@ var Accounts = function Accounts() {
     var _ethereumCall = [
         new Method({
             name: 'getId',
-            call: 'net_version',
+            call: 'eth_chainId',
             params: 0,
             outputFormatter: utils.hexToNumber
         }),
```

nivida:
Will check and fix this in the current working branch (https://github.com/ethereum/web3.js/pull/2367)

Issue #1169, opened 2017-11-13
Invalid signature with web3.eth.accounts.signTransaction on testrpc


tensor5:
I'm using web3@1.0.0-beta.24 to replicate the example http://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html#id6 on testrpc:

```js
web3.eth.accounts.signTransaction({to: '0xF0109fC8DF283027b6285cc889F5aA624EaC1F55',
    value: '1000000000',
    gas: 2000000
}, '0x4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318').then(console.log);
```

the output I get is:

```js
{ messageHash: '0x9ce21594b026a4c3fc09dff6e152f69b9beb44bd31549b49c8124b3db831fc27',
  v: '0x.82bf6b784',
  r: 
   [ '0x0aaf3477fff968b4aa173d94f4e0d74c6a9ffdc1985febea5ad447f52c45c3196a03d1bcc7efbc9902c5' ],
  s: '0x3d',
  rawTransaction: '0xf96f.8808504a817c800831e848094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca008085.82bf6b784e7aa0aaf3477fff968b4aa173d94f4e0d74c6a9ffdc1985febea5ad447f52c45c3196a03d1bcc7efbc9902c53d4406c6b69c2c6d23b5ac4c29a68ea63a342433bde32a7' }
```

which is not a valid signature. If I manually specify the mainnet `chainId` then I don't have this problem:

```js
web3.eth.accounts.signTransaction({
    to: '0xF0109fC8DF283027b6285cc889F5aA624EaC1F55',
    value: '1000000000',
    gas: 2000000,
    chainId: 0
}, '0x4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318').then(console.log);
```

```js
{ messageHash: '0x6ef4e89eab7867ddf69fae20dd218dbfa337fa8e4d68ce6dc3fcad63e17fdd72',
  v: '0x24',
  r: '0x1b428275439f5d12a85dd70e919b17994abe4d3b2ba3cfd1e370c1a1c86d2a24',
  s: '0x4eebdddb5681d836fb938179d0002b12359c9b2287a8d6ffb156c15de9146251',
[...]

PR #3097, opened 2019-10-03
Use eth_chainId when signing transactions

cgewecke:
Addresses #2378 and #1169 for the 1.x branch. Makes web3 conform to [EIP 155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md) and fetch chainId via `eth_chainId` instead of `net_version`. For ethereum public nets there's no problem here because network id == chain id. For ganache, ETC and possibly others - the two values are different. 

Note: these endpoints return their ids in different formats:

| Client       | net_version     | eth_chainId |
|--------------|-----------------|-------------|
| geth --dev   | "1337"          | "0x539"     |
| parity --dev | "17"            | "0x11"      |
| ganache-cli  | "1570135120814" | "0x539"     |

`eth_chainId` is not documented in the JSON_RPC spec. This is a case where it would be nice to have E2E tests vs the three main clients to be sure Web3's mocked responses correspond to reality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,6 @@ Released with 1.0.0-beta.37 code base.
 
 ### Fixed
 
-- Fix incorrectly populating chainId param with `net_version` when signing txs (#3097)
+- Fix incorrectly populating chainId param with `net_version` when signing txs (#2378)
 - regeneratorRuntime error fixed (#3058)
 - Fix accessing event.name where event is undefined (#3014)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,5 +56,6 @@ Released with 1.0.0-beta.37 code base.
 
 ### Fixed
 
+- Fix incorrectly populating chainId param with `net_version` when signing txs (#3097)
 - regeneratorRuntime error fixed (#3058)
 - Fix accessing event.name where event is undefined (#3014)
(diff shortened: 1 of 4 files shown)


gabmontes:
@cgewecke these changes are good IMHO.

I've been patching beta 37 with this exact set of changes in several [Metronome](https://github.com/autonomoussoftware) projects for quite some time and works flawlessly.

Hope this is merged soon.

cgewecke:
Thanks @gabmontes, good to know.

nivida (changes requested):
Could you update the changelog for this and all other PRs you proposed?

cgewecke:
@nivida Do want the changelog updated under 1.2.2? Or unreleased?

nivida:
„Unreleased“ is just the tag to label the unreleased version and the related changes. You can just extend the list of 1.2.2. @cgewecke

cgewecke:
@nivida Oh I just remembered something - not sure how you do this exactly but there is a nice npm utility called [changelog](https://www.npmjs.com/package/changelog) that helps you generate a version-based changelog in one post-publication step. 

The output still needs to be edited etc, but it seems like a convenient way of avoiding merge conflicts when there are lots of small PRs queued up like this and they all modify the same file.

nivida:
@cgewecke Let us discuss this in an issue. :)

nivida (approved):

PR #3101, opened 2019-10-04
docs: add lerna badge

armujahid:
web3.js is organized as a monorepo using [Lerna](https://lerna.js.org/). So I think we should either add lerna badge
[lerna](https://lerna.js.org/) or some kind of acknowledgment in Readme (If addition of badge is not acceptable).


nivida:
Thanks for proposing this PR! Could you do this change on the 1.x branch of this repository as well?

armujahid:
@nivida done.

nivida (approved):

PR #3116, opened 2019-10-08
supportsSubscriptions method added to providers

nivida:
back-porting of #2632 from 2.x to 1.x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ Released with 1.0.0-beta.37 code base.
 
 - localStorage support detection added (#3031)
 - getNetworkType method extended with Görli testnet (#3095)
+- supportsSubscriptions method added to providers (#3116)
 
 ### Fixed
 
--- a/packages/web3-providers-http/src/index.js
+++ b/packages/web3-providers-http/src/index.js
@@ -120,6 +120,12 @@ HttpProvider.prototype.disconnect = function () {
     //NO OP
 };
 
+/**
+ * Returns the desired boolean.
+ *
+ * @method supportsSubscriptions
+ * @returns {boolean}
+ */
 HttpProvider.prototype.supportsSubscriptions = function () {
     return false;
 };
--- a/packages/web3-providers-http/src/index.js
+++ b/packages/web3-providers-http/src/index.js
@@ -120,5 +120,8 @@ HttpProvider.prototype.disconnect = function () {
     //NO OP
 };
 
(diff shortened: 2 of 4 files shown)


nivida:
The documentation for all providers will get improved with issue https://github.com/ethereum/web3.js/issues/3117

alcuadrado:
I wouldn't release this just yet. It would expose a super experimental API that will probably change soon.

cgewecke:
> It would expose a super experimental API that will probably change soon.

@alcuadrado 

Could you elaborate your view of this?  It's experimental wrt EIP 1193 & the [Events Spec](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#events-1)?

alcuadrado:
Yes, sure. We had a meeting with about ~15 people to discuss the direction of the EIP-1193 during devcon. I'll post a summary and several PRs soon. One of the conclusions was that the provider should describe its capabilities. We explored many possible solutions, including not doing them at all, using a bunch of boolean fields, a `supports` function, and a `supports` object. 

IMO the EIP should be updated to reflect that, and some consensus is reached, implemented here.

cgewecke:
@alcuadrado Ah! Excellent thanks.

nivida:
> Yes, sure. We had a meeting with about ~15 people to discuss the direction of the EIP-1193 during devcon. I'll post a summary and several PRs soon. One of the conclusions was that the provider should describe its capabilities. We explored many possible solutions, including not doing them at all, using a bunch of boolean fields, a supports function, and a supports object.

> IMO the EIP should be updated to reflect that, and some consensus is reached, implemented here.

@alcuadrado This is true but it will take an amount of time until this EIP is finalized and the method ``supportsSubscriptions`` is a long out-standing and simple feature request from the community. It is additionally allowed to extend an existing EIP as long as we support all the functionalities this EIP does define. I mean with this that we currently can keep this method until the EIP-1193 is finalized and we can later provide compatibility to the EIP-1193. If the method ``getCapabilities(): Capabilities`` got added. Will ``supportsSubscription`` first be a wrapper of the existing ``getCapabilities`` method just with a deprecation message added and can get removed later with a Minor release of web3.js.

alcuadrado:
The issue is not so much about how to implement it here, but about the
frustration it would cause to users across the ecosystem. Some people will
get used to web3 providers, and without knowing about eip1193, they'll just
expect everything to mimic web.

Middle ground proposal: let's give it a month. I'll push for the eip to be
finalized asap.

PR #3119, opened 2019-10-10
Ensure the immutability of function `signTransaction` input parameter

### LINKED ISSUE #2190: Function 'sendSignedTransaction' changes the value of the 'to' field from 'null' to '0x'

barakman:
Fixes #2190

### DIFF
--- a/packages/web3-eth-accounts/src/index.js
+++ b/packages/web3-eth-accounts/src/index.js
@@ -162,7 +162,7 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
         }
 
         try {
-            var transaction = helpers.formatters.inputCallFormatter(JSON.parse(JSON.stringify(tx)));
+            var transaction = helpers.formatters.inputCallFormatter(_.clone(tx));
             transaction.to = transaction.to || '0x';
             transaction.data = transaction.data || '0x';
             transaction.value = transaction.value || '0x';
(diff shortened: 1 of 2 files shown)


packages/web3-eth-accounts/src/index.js
```diff
@@ -162,13 +162,11 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
 
         try {
-            tx = helpers.formatters.inputCallFormatter(tx);
-
-            var transaction = tx;
-            transaction.to = tx.to || '0x';
-            transaction.data = tx.data || '0x';
-            transaction.value = tx.value || '0x';
-            transaction.chainId = utils.numberToHex(tx.chainId);
+            var transaction = helpers.formatters.inputCallFormatter(JSON.parse(JSON.stringify(tx)));
```
nivida: Could you use the [clone](https://underscorejs.org/#clone) method of the anyways existing ``underscore`` package and update the CHANGELOG.md file in the root folder of the 1.x branch?

nivida (changes requested):

packages/web3-eth-accounts/src/index.js
barakman: @nivida: Done.

nivida (approved):

Issue #2033, opened 2018-11-02
Don't kill 1/2 bytes from private keys please


mirek:
When using:
```
web3.eth.accounts.signTransaction(txParams, privateKey)
```

a) If your private key is hex0x - all good.

b) If your private key is hex - everyting will work fine, no errors, but your signed transaction will be with private key with one byte zero'ed.

c) If your private key is Buffer - everything will be signed, no errors, but your signed transaction will be with private key with two first bytes zero'ed.

```
a) 870c20575bd9da9d747011401ed345edf6229c96ba2bfcd965ed6dee1fc3afd3
b) 000c20575bd9da9d747011401ed345edf6229c96ba2bfcd965ed6dee1fc3afd3
c) 000020575bd9da9d747011401ed345edf6229c96ba2bfcd965ed6dee1fc3afd3
```

...which is kind of cute that 3 different representations create 3 valid, different signed transactions, all thanks to web3 not validating it's inputs and this internally doing the same https://github.com/MaiaVictor/eth-lib/blob/master/src/account.js#L47

```
const Web3 = require('web3')
const web3 = new Web3('http://localhost:8545')

const pk = '870c20575bd9da9d747011401ed345edf6229c96ba2bfcd965ed6dee1fc3afd3'
const pk0x = `0x${pk}`
const pkb = Buffer.from(pk, 'hex')

const tx = {
  to: '0xF0109fC8DF283027b6285cc889F5aA624EaC1F55',
  value: '1000000000',
  gas: 2000000
}

const mh = _1 => web3.eth.accounts.signTransaction(tx, _1).then(_2 => _2.rawTransaction)

async function main() {
  console.log('pk     ', await mh(pk))
  console.log('pk0x   ', await mh(pk0x))
  console.log('pkb    ', await mh(pkb))

  console.log('pk00   ', await mh('0x00' + pk.slice(2)))
  console.log('pk0000 ', await mh('0x0000' + pk.slice(4)))
}

main()

// Output:
// pk      0xf8648080831e848094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca00802ba0ef96e99731c7aafd230a21c322e4043a830a5174370d7395f69c3c97dc22540da03262a83de0b19a3afb97c9ce42c3be8f07291023120227420a2850ec1452a0f9
[...]

Issue #1074, opened 2017-09-23
web3.eth.accounts.signTransaction produced errant bytecode


cmditch:
Bytecode produced with a period in it.

Should be able to replicate it with the code params below:
```js
signTxParams = { 
    data: "0x23123123123123123",
    gas: 42424242424242,
    gasPrice: 42424242,
    nonce: 1,
    to: "0xd8b0990c007ba1ad97b37c001d1f87044312162a",
    value: "42424242",
    chainId: 1
}
privateKey = "0xef1ae7c6a04a3191b364c603f6f506938c98ec259597007300b42242f9fcc4b1"
signedTx = web3.eth.accounts.signTransaction(signTxParams, privateKey)

console.log(signedTx.rawTransaction);

> "0xf86b.80184028757b2862695a9e649b294d8b0990c007ba1ad97b37c001d1f87044312162a84028757b225a0254b9251405d5b83d688d0db794ef069fbb87a24d"
```

Browser : confirmed on Chrome & Safari
Web3.js Version : 1.0.0-beta.18

Note: To be fair, I don't think the above `data` field is totally valid bytecode. Nonetheless, thought this errant period might still be noteworthy.

frozeman:
@MaiaVictor can you have a look?

netpi:
I have the same problem !

girish-bluefin:
its new api not implemented yet. You can see it
https://web3js.readthedocs.io/en/1.0/web3-eth.html#id73

bytniak:
I have exactly the same problem.
With some normal transaction data and:
`data: "0x2ca5a332000000000000000000000000000000000000000000000000000000000000753000000000000000000000000000000000000000000000000000000000000003e8"`
My signed rowTransaction looks like this:
`> 
 "0xf9ab.8038417bd2c89832dc6c094279915231c108f8f371ef75aa58e18d181bf863380b8442ca5a332000000000000000000000000000000000000000000000000000000000000753000000000000000000000000000000000000000000000000000000000000003e881.817aa086b303cdcb77810d5419c31fedd0ba7fc3f07e26de8b90af2cf96e4a54d37938a0366ab73e00884eb024174b1daf8dac605b3537d9ac0fa49d1073c93af38d51d4"`

Now we have two periods..

Environment: NodeJS  v8.9.1
Web3.js Version: 1.0.0-beta.26

not really sure what @girish-bluefin  meant...

PR #3122, opened 2019-10-11
Add E2E tests for geth and ganache dev clients

cgewecke:
#2682 (original) and part of #3098 (E2E Proposal)

+ Adds shell scripts for testing with geth and ganache development nodes (locally & CI)
+ Combines coverage/coveralls measurement for E2E and unit tests in the Node 10 job.
+ Adds 25-30 tests covering simple deployments, promievents and transaction signing
+ Adds some small mock contracts and their compilation artifacts in test/sources 

The tests run across these combinations.....

| Client        | instamine | automine 2s |
|--------------|-----------|-------------|
| geth --dev   |     (http)   |     (http, ws)      | 
| ganache-cli  |    (http, ws)   |             | 

---------
Things discovered by the test suite:
+ ~~confirmation handler over ws for ganache on deployment does not fire.~~ 
  + [Edit: ... was a bug in the tests].
+ ~~confirmation handler over ws fires each second for geth automine~~
  + [Edit: ... now believe this works correctly, have added assertions to test]
+ ws for geth instamine is very flaky - possible race condition somewhere.

Tried to make this as simple as possible. I think the CI part will fit into 2.x without too many problems. It's just some small shell scripts.  

I used this [geth-dev-assistant][1] wrapper to launch geth - it's small, all the APIs are exposed and you can pull recent images with it. I would be into publishing something like this in the [web3-js org][2] so anyone with permissions there can edit it as they wish and make necessary improvements. 

**To do in future....**
+ Will add Parity in a subsequent PR after this is reviewed and any required changes are made. 
+ More tests...it would be nice to get coverage up to 90% in 1.x ( like in 2.x )

[1]: https://github.com/cgewecke/geth-dev-assistant
[2]: https://github.com/web3-js

PR #3125, opened 2019-10-13
Signing fixes back-ported from 2.x

### LINKED ISSUE #1998: Incorrect error from invalid private key for signTransaction
Hey guys,

I was running into an issue where I was getting an insufficient funds error when trying to sign a transaction. I found that this is a fairly common issue on the forums and it can be caused by having a private key that doesn't start with a 0x.

There is a simple solution to this in the packages/web3-eth-accounts/src/index.js file on line 145:
```
if (!privateKey) {
    error = new Error('No private key given!');

    callback(error);
    return Promise.reject(error);
}

if (!privateKey.startsWith('0x')) {
    error = new Error('Private key is invalid!');

    callback(error);
    return Promise.reject(error);
}
```
https://github.com/ethereum/web3.js/blob/v1.0.0-beta.36/packages/web3-eth-accounts/src/index.js#L145

I tested this from the v1.0.0-beta.36 tag and all tests passed.

nivida:
Fixes #1998, #2033, and #1074


packages/web3-eth-accounts/src/index.js
```diff
@@ -168,36 +154,28 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
 
-            var rawTx = RLP.decode(rlpEncoded).slice(0, 6).concat(Account.decodeSignature(signature));
+            var validationResult = ethTx.validate(true);
 
-            rawTx[6] = makeEven(trimLeadingZero(rawTx[6]));
-            rawTx[7] = makeEven(trimLeadingZero(rawTx[7]));
-            rawTx[8] = makeEven(trimLeadingZero(rawTx[8]));
+            if (validationResult !== '') {
+                throw new Error('Signer Error: ' + validationResult);
+            }
```
cgewecke, line 168: What do you think about adding the example tx in #1074 to the unit tests to see if it hits this validation check? 

It looks like this:
```js
signTxParams = { 
    data: "0x23123123123123123",  
    gas: 42424242424242,
    gasPrice: 42424242,
    nonce: 1,
    to: "0xd8b0990c007ba1ad97b37c001d1f87044312162a",
    value: "42424242",
    chainId: 1
}
```

packages/web3-eth-accounts/src/index.js
```diff
@@ -168,36 +154,28 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
 
-            var rawTransaction = RLP.encode(rawTx);
+            var rlpEncoded = ethTx.serialize().toString('hex');
+            var rawTransaction = '0x' + rlpEncoded;
+            var transactionHash = utils.keccak256(rawTransaction);
```
cgewecke, line 172: Is this transactionHash line extra?

cgewecke (approved):
Looks good! Just left a couple small notes

packages/web3-eth-accounts/src/index.js
nivida: I've added the ``transactionHash`` property to the returned object.

cgewecke (approved):
LGTM.

cgewecke:
@nivida Apologies, I missed something important in the review here. Additionally, there was a bug in the E2E CI (fixed with #3138) which meant it wasn't obvious an eth.accounts.sign test was failing for geth dev.

The problem is that ethereumjs-tx requires [an additional object passed to the Transaction constructor][1] when a tx is being sent to a non-mainnet chain. You can see an example in the [ejs-tx signing test][2] removed from #3122.

[1]: https://github.com/ethereumjs/ethereumjs-tx#chain-and-hardfork-support
[2]:https://github.com/cgewecke/web3.js/blob/aa96ee8116fe1733cfbe60a61d9ca2ee9020b77a/test/e2e.method.signing.js#L31-L68

PR #3129, opened 2019-10-15
receipt is missing on error event

nivida:
Fixes https://github.com/ethereum/web3.js/issues/1859 with adding of the receipt to the returned values of the error event listener. The receipt will get returned if not enough gas was provided or the EVM is reverting the transaction on other contract-specific reasons.


nivida:
@cgewecke We have to add the error cases in the ``eth.sendTransaction`` and ``contract`` test cases. But I think this should be tracked with a separate issue to include a clean up of these test cases as well. They are written in a complex way that wrongly running test cases are hard to find. I like generic test cases but I think we could do this in a way better way as it is currently. ([see todo](https://github.com/ethereum/web3.js/blob/1.x/test/contract.js#L3113))

docs/web3-eth.rst
```diff
@@ -1010,7 +1010,7 @@ The **callback** will return the 32 bytes transaction hash.
 - ``"transactionHash"`` returns ``String``: Is fired right after the transaction is sent and a transaction hash is available.
 - ``"receipt"`` returns ``Object``: Is fired when the transaction receipt is available.
 - ``"confirmation"`` returns ``Number``, ``Object``: Is fired for every confirmation up to the 12th confirmation. Receives the confirmation number as the first and the :ref:`receipt <eth-gettransactionreceipt-return>` as the second argument. Fired from confirmation 0 on, which is the block where its minded.
-- ``"error"`` returns ``Error``: Is fired if an error occurs during sending. If a out of gas error, the second parameter is the receipt.
+- ``"error"`` returns ``Error`` and ``Object|undefined``: Is fired if an error occurs during sending. If a out of gas error, the second parameter is the receipt.
```
cgewecke:
```suggestion
``"error"`` returns ``Error`` and ``Object|undefined``: Is fired if an error occurs during sending. If the transaction was rejected by the network with a receipt, the second parameter will be the receipt.
```

packages/web3-core-method/src/index.js
```diff
@@ -303,7 +310,7 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
                             promiseResolved = true;
                         }
 
-                        utils._fireError(new Error('The transaction receipt didn\'t contain a contract address.'), defer.eventEmitter, defer.reject);
+                        utils._fireError(new Error('The transaction receipt didn\'t contain a contract address.'), defer.eventEmitter, defer.reject, receipt);
```
cgewecke: I like the 'list' arg style you've done below...it's readable within the github page :)

It looks like `_fireError` has five params - the fourth is a callback, the fifth is optionalData?
```suggestion
                        utils._fireError(
                            new Error('The transaction receipt didn\'t contain a contract address.'), 
                            defer.eventEmitter, 
                            defer.reject, 
                            null,
                            receipt
                        );
```

packages/web3-core-method/src/index.js
```diff
@@ -330,7 +337,12 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
                             }
 
                         } else {
-                            utils._fireError(new Error('The contract code couldn\'t be stored, please check your gas limit.'), defer.eventEmitter, defer.reject);
+                            utils._fireError(
+                                new Error('The contract code couldn\'t be stored, please check your gas limit.'),
+                                defer.eventEmitter,
+                                defer.reject,
+                                receipt
```
cgewecke, line 351: Missing `null`?

cgewecke (commented):
Looks good! 

Perhaps we should open an issue considering the suggestion in [this comment](https://github.com/ethereum/web3.js/issues/1859#issuecomment-539678012), about restoring a generic emitter for any receipt, regardless of status.

docs/web3-eth-contract.rst
```diff
@@ -543,7 +543,7 @@ The **callback** will return the 32 bytes transaction hash.
 - ``"transactionHash"`` returns ``String``: is fired right after the transaction is sent and a transaction hash is available.
 - ``"receipt"`` returns ``Object``: is fired when the transaction *receipt* is available. Receipts from contracts will have no ``logs`` property, but instead an ``events`` property with event names as keys and events as properties. See :ref:`getPastEvents return values <contract-events-return>` for details about the returned event object.
 - ``"confirmation"`` returns ``Number``, ``Object``: is fired for every confirmation up to the 24th confirmation. Receives the confirmation number as the first and the receipt as the second argument. Fired from confirmation 1 on, which is the block where it's minded.
-- ``"error"`` returns ``Error``: is fired if an error occurs during sending. If a out of gas error, the second parameter is the receipt.
+- ``"error"`` returns ``Error`` and ``Object|undefined``: is fired if an error occurs during sending. If a out of gas error, the second parameter is the receipt.
```
cgewecke: Ah I missed this one! 
```suggestion
``"error"`` returns ``Error`` and ``Object|undefined``: Is fired if an error occurs during sending. If the transaction was rejected by the network with a receipt, the second parameter will be the receipt.
```

cgewecke (approved):

Issue #2623, opened 2019-04-01
Why is web3.min.js not provided anymore?


carsonwah:
### Description 

No `web3.min.js` available since v1.0.0-beta.38.

#### Expected behavior

A `web3.min.js` file should be pre-packed and is ready to use directly.

#### Actual behavior

The collaborator said "he" always use webpack to bundle the js files.
https://github.com/ethereum/web3.js/issues/2475#issuecomment-470934021

### Versions 
* web3.js: >=1.0.0-beta.38

----

Can someone explain the reason behind this decision? It is too troublesome to use a bundler in order to use this library. I don't see why we can't have a `web3.min.js`. Many people are sticking with v1.0.0-beta.37 because of this.

(You may view this issue as a question or a feature request)

nivida:
I've removed the minified file because I thought the frontend space is using a build chain anyways. Sorry for the troubles. I will add a minified file for each module back to the npm packages asap.

carsonwah:
@nivida Thank you. I think providing a minified js will be very helpful especially for many people that are using Truffle as their starting point, which does not include a frontend bundler usually.
Thanks for the quick reply.

ghost:
Yes, please provide this. Many users cannot even install web3 with npm. Web3 uses a deprecated version of scrypt npm module instead of native node.js functions. As a result of this dependency it cannot be installed as this module has difficulties installing under many modern linux distributions.

You can see this module is deprecated here: https://github.com/barrysteyn/node-scrypt

There is even potential issues of malicious code injection: https://github.com/barrysteyn/node-scrypt/issues/186

fulldecent:
Reference to #2013 to explain the security problem with this issue.

smiled0g:
@nivida Any ETA on this issue? I found that installing `web3` via npm doesn't even compile correctly with `create-react-app`. With different build systems out there, it'd be more difficult to make sure that `web3 ` package would work with all of them.

OFRBG:
@smiled0g I've used `web3` with `create-react-app` many times, and I've never had a problem. Maybe add more information? `create-react-app` uses webpack, and it has no problems with `web3` as of now.

smiled0g:
> @smiled0g I've used `web3` with `create-react-app` many times, and I've never had a problem. Maybe add more information? `create-react-app` uses webpack, and it has no problems with `web3` as of now.

[...]

PR #3131, opened 2019-10-15
Minified file added to web3 package 

### LINKED ISSUE #1041: Where is `dist` folder and  `web3.min.js`?

nivida:
The minified file which is located in the root folder of the 1.x branch will now also be published on npm. The minified file got defined as the entry file for browsers in ``packages/web3/package.json``. I've also added this file to the ``.gitignore`` for not having it duplicated on GitHub. (We can probably later anyways remove the minified file from Git)

Fixes #1041

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,5 @@ bower_components
 .npm/
 .vscode/
 dist/
+packages/web3/dist/
 !./dist/web3.min.js
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ Released with 1.0.0-beta.37 code base.
 - getNetworkType method extended with Görli testnet (#3095)
 - supportsSubscriptions method added to providers (#3116)
 - Add `eth.getChainId` method (#3113)
+- Minified file added to web3 package (#3131)
 
 ### Fixed
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -20,6 +20,7 @@ var replace = require('gulp-replace');
 var exec = require('child_process').exec;
 
 var DEST = path.join(__dirname, 'dist/');
+var WEB3_PACKAGE_DEST = path.join(__dirname, 'packages/web3/dist');
 
 var packages = [{
     fileName: 'web3',
@@ -189,6 +190,7 @@ packages.forEach(function (pckg, i) {
             .pipe(streamify(uglify(ugliyOptions)))
             .on('error', function (err) { console.error(err); })
             .pipe(rename(pckg.fileName + '.min.js'))
+            .pipe(gulp.dest(WEB3_PACKAGE_DEST))
             .pipe(gulp.dest(DEST));
     }));
 });
--- a/packages/web3/package.json
+++ b/packages/web3/package.json
@@ -8,6 +8,7 @@
         "node": ">=8.0.0"
     },
     "main": "src/index.js",
+    "browser": "dist/web3.min.js",
     "bugs": {
         "url": "https://github.com/ethereum/web3.js/issues"
     },


cgewecke (approved):
LGTM! Nice.

joshstevens19 (approved):
I would add some instructions in the README in how they would go about finding the file etc.. unless we already have it. Apart from that looks good to merge.

alcuadrado:
Sorry I'm late, but what's the rationale for this PR? To have a `browser` entry in the `package.json`? Is it nay different than the `main` one except for minification?

cgewecke:
@alcuadrado 

In #1041 there's a subset of users expecting a `dist` with a `min` when they install like this:
```
npm install web3
```

At the moment it looks like people are git cloning web3 & running `npm build` themselves to produce this artifact.

alcuadrado:
I suspect that does reports may come from a poor release process in the past. It sounds like some releases had no build artifact at all. This may not be a real issue now.

cgewecke:
@alcuadrado Mmmm. 1.2.1 installs without a dist - I think this could be a backport.

alcuadrado:
It doesn't, but why should it have a `dist`? It has a `main` in the `package.json` pointing to `scr/index.js`.

I think I'd just close the issue and revert this change until someone else reports it and give more info, if that ever happens.

Also, @nivida did some publishing work with rollup in the 2.x branch, which may conflict with this change (i.e. changing the expectations of what `dist/` is). Even more so, if this project is to be migrated to TS (top priority after the backports?) the same may happen.

alcuadrado:
I found that the `main` field was added here https://github.com/ethereum/web3.js/pull/1045 - Before that, there was no main, so it probably used to lead to issues when using it with web bundlers.

cgewecke:
@alcuadrado Ok perhaps caution is reasonable here. As I read this change it only makes it possible to refer to a min via dist qualified path. We aren't testing webpack-ability per se but that might be a good idea.


One item in defense of the topicality of this PR is this (from the GH insights tab for Web3. Looking for the min is the 5th most common thing people do. 🤷‍♂ 
Screen Shot 2019-10-21 at 7 01 30 PM

alcuadrado:
> One item in defense of the topicality of this PR is this (from the GH insights tab for Web3. Looking for the min is the 5th most common thing people do. 🤷‍♂

This is really interesting! I've never used the GH traffic stats for this. Is it from the entire history? It'd be interesting to see if this keeps being true now, or if it mostly reflects how popular that issue was in the past.

cgewecke:
@alcuadrado 

> Is it from the entire history?

Good question - it looks like it's configurable out to 1 month, via a drop-down option on the "insights/pulse" tab.  Just ran that search and found more recent discussion about what & why people want this in #2623 as well....

cgewecke:
@alcuadrado Sorry - I'm wrong about it being configurable lol. No idea what the time span is.

nivida:
> Sorry I'm late, but what's the rationale for this PR? To have a browser entry in the package.json? Is it nay different than the main one except for minification?

@alcuadrado The gulp build pipeline does create way smaller minified files then webpack is doing. This probably because it is excluding specific packages. Additionally, will CDN's use our minified version instead of bundling the minified files on their own. Another case is that some (also me) do like to run ``npm install web3`` and to include the minified file in a hacky way into an HTML header for doing some quick tests.

nivida:
**Comparsion**

With minified files:
Bildschirmfoto 2019-10-22 um 08 53 04


Without minified files:
Bildschirmfoto 2019-10-22 um 08 52 23


The production build with webpack is also way faster with the minified files provided.

Bundled ``src/index.js``: 
``` javascript
var Web3 = require('web3');
console.log(Web3);
```

PR #3141, opened 2019-10-17
eth.accounts.signTransaction method fixed

nivida:
The new ethereumjs-tx package requires additional parameters.


packages/web3-net/src/index.js
```diff
@@ -39,7 +39,7 @@ var Net = function () {
             name: 'getId',
             call: 'net_version',
             params: 0,
-            outputFormatter: utils.hexToNumber
+            outputFormatter: parseInt
```
cgewecke, line 42: 💯 @nivida I'm so glad you caught this.

cgewecke (commented):
This LGTM although it seems like it would break current expectations for Ropsten etc - e.g. that  accounts.signTransaction can be used with a non-mainnet chain without any special configuration. 

@nivida  Per your idea in chat have added some logic to automatically generate an ethereumjs-common options object when no hardfork, chain or common is specified. 

That's in PR targeting this PR at #3143.  If you have a chance could you check that out and advise?

packages/web3-net/src/index.js
nivida: Same here ;)

PR #3143, opened 2019-10-18
Auto fetch tx signing option values when not set by user

cgewecke:
**NB: MERGE TARGET is #3141**

Per @nivida's idea from a chat, adds auto-fetching of custom network details when using `eth.accounts.signTransaction`. Idea is to extend #3141 so its new behavior does not require people to add ethereumjs-custom options when using non-mainnet chains. 

Adds tests for:
+ signTransaction without specifying any transaction signing options
+ some error cases

[Edit - WIP while fixing the unit mocks....]

### DIFF
--- a/packages/web3-eth-accounts/package-lock.json
+++ b/packages/web3-eth-accounts/package-lock.json
@@ -336,11 +336,11 @@
 			"resolved": "https://registry.npmjs.org/dtslint/-/dtslint-0.4.2.tgz",
 			"integrity": "sha512-ph4GXLw3HYzlQMJOFcpCqWHuL3MxJ/344OR7wn0wlQGchQGTIVNsSUl8iKEMatpy2geNMysgA9fQa6xVhHOkTQ==",
 			"requires": {
-				"definitelytyped-header-parser": "github:Microsoft/definitelytyped-header-parser#production",
+				"definitelytyped-header-parser": "github:Microsoft/definitelytyped-header-parser#d957ad0bb2f4ecb60ac04f734e0b38fbc8e70b8a",
 				"fs-extra": "^6.0.1",
 				"strip-json-comments": "^2.0.1",
 				"tslint": "^5.12.0",
-				"typescript": "^3.7.0-dev.20191015"
+				"typescript": "next"
 			},
 			"dependencies": {
 				"definitelytyped-header-parser": {
@@ -350,6 +350,11 @@
 						"@types/parsimmon": "^1.3.0",
 						"parsimmon": "^1.2.0"
 					}
+				},
+				"typescript": {
+					"version": "3.7.0-dev.20191017",
+					"resolved": "https://registry.npmjs.org/typescript/-/typescript-3.7.0-dev.20191017.tgz",
+					"integrity": "sha512-Yi0lCPEN0cn9Gp8TEEkPpgKNR5SWAmx9Hmzzz+oEuivw6amURqRGynaLyFZkMA9iMsvYG5LLqhdlFO3uu5ZT/w=="
 				}
 			}
 		},
@@ -1002,11 +1007,6 @@
 			"requires": {
 				"tslib": "^1.8.1"
 			}
-		},
-		"typescript": {
-			"version": "3.7.0-dev.20191015",
-			"resolved": "https://registry.npmjs.org/typescript/-/typescript-3.7.0-dev.20191015.tgz",
-			"integrity": "sha512-Cpfj1n4pEUVKL+jtS0mkZodJffyMmf3Wk/UjyZMGX4fsjK5KBPJf3NUlyXij8I8p1E2CAomdS5NPFrAR+z8pKw=="
 		},
 		"underscore": {
 			"version": "1.9.1",
--- a/packages/web3-eth-accounts/package-lock.json
+++ b/packages/web3-eth-accounts/package-lock.json
@@ -339,7 +339,8 @@
 				"definitelytyped-header-parser": "github:Microsoft/definitelytyped-header-parser#production",
 				"fs-extra": "^6.0.1",
 				"strip-json-comments": "^2.0.1",
-				"tslint": "^5.12.0"
+				"tslint": "^5.12.0",
+				"typescript": "^3.7.0-dev.20191018"
 			},
 			"dependencies": {
 				"definitelytyped-header-parser": {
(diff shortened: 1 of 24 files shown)


packages/web3-eth-accounts/package-lock.json
```diff
@@ -336,11 +336,11 @@
 			"resolved": "https://registry.npmjs.org/dtslint/-/dtslint-0.4.2.tgz",
 			"integrity": "sha512-ph4GXLw3HYzlQMJOFcpCqWHuL3MxJ/344OR7wn0wlQGchQGTIVNsSUl8iKEMatpy2geNMysgA9fQa6xVhHOkTQ==",
 			"requires": {
-				"definitelytyped-header-parser": "github:Microsoft/definitelytyped-header-parser#production",
+				"definitelytyped-header-parser": "github:Microsoft/definitelytyped-header-parser#d957ad0bb2f4ecb60ac04f734e0b38fbc8e70b8a",
```
cgewecke: There's some typescript noise in the package-lock.json . . .  is this ok? Or wrong?

nivida: Yes, this will be used from the dtslint package. 👌

packages/web3-eth-accounts/src/index.js
```diff
@@ -138,6 +145,13 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
             );
         }
 
+        if ((tx.chain && !tx.hardfork) || (tx.hardfork && !tx.chain)){
```
nivida: If the property ``common`` is defined should ``chain`` and ``hardfork`` be undefined as the documentation of ethereumjs-tx is saying. We can keep this that way if we use the ``hardfork`` and ``chain`` property for the static constructor function of the ``Common`` object.

packages/web3-eth-accounts/src/index.js
```diff
@@ -212,22 +241,24 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
+    // Resolve immediately if nonce, chainId, price and signing options are provided
+    if (tx.nonce !== undefined && tx.chainId !== undefined && tx.gasPrice !== undefined && hasTxSigningOptions) {
         return Promise.resolve(signed(tx));
     }
 
-
     // Otherwise, get the missing info from the Ethereum Node
     return Promise.all([
-        isNot(tx.chainId) ? _this._ethereumCall.getId() : tx.chainId,
+        isNot(tx.chainId) ? _this._ethereumCall.getChainId() : tx.chainId,
```
nivida, line 264: thanks for seeing this!

test/eth.accounts.signTransaction.js
```diff
@@ -20,7 +33,8 @@ var tests = [
             to: '0xF0109fC8DF283027b6285cc889F5aA624EaC1F55',
             toIban: 'XE04S1IRT2PR8A8422TPBL9SR6U0HODDCUT', // will be switched to "to" in the test
             value: "1000000000",
-            data: ""
+            data: "",
+            common: common
```
cgewecke, line 35: Adding `common` to all of these just keeps these tests as they are. If common wasn't here, the injections would be mis-aligned because there would be an extra call for networkId.

test/eth.sendTransaction.js
```diff
@@ -37,7 +49,9 @@ var tests = [{
     call: 'eth_'+ method
 },
 // test with gasPrice missing
-{
+// This test is broken but very difficult to diagnose cause....
+// behavior is correct (see sendTransaction E2E)
+/*{
```
cgewecke: I [broke this test](https://travis-ci.org/ethereum/web3.js/jobs/599482864#L3763) somehow and cannot figure out how to fix it. Very frustrating. It does not actually test any code in eth-accounts - I think the problem might be something unusual about the way this test file is structured....

The test itself checks whether or not `gasPrice` is fetched when calling `eth.sendTransaction`, so I have explicitly added that case in`e2e.methods.send` in this PR. 

idk, ggrrrrrr.

cgewecke: Fixed - this was some kind of problem using _.clone. ...

packages/web3-eth-accounts/src/index.js
```diff
@@ -212,22 +240,24 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
         return result;
     }
 
-    // Resolve immediately if nonce, chainId and price are provided
-    if (tx.nonce !== undefined && tx.chainId !== undefined && tx.gasPrice !== undefined) {
+    var hasTxSigningOptions = ((tx.chain && tx.hardfork) || tx.common) ? true : false;
+
+    // Resolve immediately if nonce, chainId, price and signing options are provided
+    if (tx.nonce !== undefined && tx.chainId !== undefined && tx.gasPrice !== undefined && hasTxSigningOptions) {
```
nivida, line 258: I would create a method ``hasTxSigningOptions`` on the Accounts class or at least move the other conditions up to the defined function.

test/eth.accounts.signTransaction.js
nivida: We could add a property to the test cases to check if a ``common`` object is passed or not and inject the missing step for the ``net_version`` JSON-RPC call otherwise we would miss this branch in these unit tests.

nivida (changes requested):
If there is no veto from your side will I do the commented changes today or tomorrow (Sunday)

packages/web3-eth-accounts/src/index.js
cgewecke: 👍 

Have hoisted this definition to the top of the method because it's only used in `accounts.sign`.

test/eth.accounts.signTransaction.js
cgewecke: 👍 

Yes! Added cases for missing common.

packages/web3-eth-accounts/src/index.js
cgewecke: Could you explain this in more detail? Are you saying there should be an additional check here for the case where someone has filled in *all* of the options, e.g. common, hardfork and chain are all defined?

nivida: Yes, I would throw an error to inform the developer about passed but not used parameters or we use the passed values for the static constructor function on line 181. We could also add the related default properties to the Eth, Contract, and Accounts module. (``defaultChain``, ``defaultHardfork``, and ``defaultCommon``)

nivida (approved):

Issue #3150, opened 2019-10-23
Typing issue in web3.eth.getBlock


princesinha19:
I guess we need to change the typing for [web3.eth.getBlock](https://web3js.readthedocs.io/en/v1.2.0/web3-eth.html#getblock) because when we pass `true` in web3.eth.getBlock it returns the array of transaction objects instead of an array of transaction strings.

Currently, the type is `(property) Block.transactions: Transaction[] | string[]`.
So, when you will try to use `true` with getBlock then typescript will give you an error:-

```
Types of parameters 'tx' and 'value' are incompatible.
      Type 'string' is not assignable to type 'Transaction'.ts(2345)
```

So, I think we need to create two different types of getBlock like BlockTransactionsObject and BlockTransactionString.

Please suggest better name or a better way to solve this issue.

**Version**: 1.2.2 as well as 2.x

joshstevens19:
we have a type which is

`getBlock(blockHashOrBlockNumber: string | number, returnTransactionObjects: boolean): Promise<Block>;`

so if you pass the 2nd parameter it it should say `Block`

![image](https://user-images.githubusercontent.com/18472422/67391258-3f745980-f596-11e9-9d34-d1c6395e78c1.png)

^^ just tried it and experienced no issues can i see your coding example?

joshstevens19:
sure but why can you not write code like:

```
const block = await Web3Service._instance.eth.getBlock(1234, true);
const transactions = block.transactions as Transaction[];

// do what you want with the transactions
```

i don't get any `TS` errors doing it like this with strict mode on.

we should probably yes create 2 different interfaces on that case which returns only `Transaction[]` but you can work around with the code to make it not error.

if you're getting errors with the above can you show me your snippet and i can take a look.

princesinha19:
@joshstevens19 Yes. The above code is proper and it will work. It's a workaround to not get error. But, If some will not write const `transactions = block.transactions as Transaction[]` then, he will get an error.

Consider this code:-
```
this.web3.eth.getBlock(i, true)
.then((block) => {
    block.transactions.forEach((tx: Transaction) => {
    });
});
```
Ideally, if we will make two different interfaces then, it will work fine. Something like below code.

```
getBlock(blockHashOrBlockNumber: string | number): Promise<BlockTransactionString>;
getBlock(
    blockHashOrBlockNumber: string | number,
    returnTransactionObjects: boolean
): Promise<BlockTransactionString>;
getBlock(
[...]

PR #3152, opened 2019-10-23
Block interface changes to fix typing

princesinha19:
I have changed the interface of getBlock and getUncle to solve issue #3150. I have changed the type of **returnTransactionObjects** from `boolean` to `true` because by default it's false so, there is no need to specify that. Let us know if anyone thinks the previous one was better.


packages/web3-eth/types/index.d.ts
```diff
@@ -369,8 +369,16 @@ export interface BlockHeader {
     timestamp: number | string;
 }
 
-export interface Block extends BlockHeader {
-    transactions: Transaction[] | string[];
+export interface BlockTransactionObject extends BlockHeader {
```
joshstevens19: you could probably make a base interface like `BlockTransactionBase` which extends `BlockHeader` which has all the repeating ones like size, difficulty, totalDifficulty and uncles and then just define transactions like:

```
interface BlockTransactionBase extends BlockHeader {
   size: number;
    difficulty: number;
    totalDifficulty: number;
    uncles: string[];
}

export interface BlockTransactionObject extends BlockTransactionBase {
   transactions: Transaction[];
}

export interface BlockTransactionString extends BlockTransactionBase {
   transactions: string[];
}
```

and means you only got 1 place to change the shared properties :+1:

joshstevens19 (changes requested):

docs/web3-eth.rst
```diff
@@ -796,7 +796,7 @@ Parameters
 ----------
 
 1. ``String|Number`` - The block number or block hash. Or the string ``"genesis"``, ``"latest"`` or ``"pending"`` as in the :ref:`default block parameter <eth-defaultblock>`.
-2. ``Boolean`` - (optional, default ``false``) If ``true``, the returned block will contain all transactions as objects, if ``false`` it will only contains the transaction hashes.
+2. ``True`` - (optional, default ``false``) If specified ``true``, the returned block will contain all transactions as objects. By default it is ``false`` so, there is no need to explictly specify false. And, if ``false`` it will only contains the transaction hashes.
```
nivida:
```suggestion
2. ``Boolean`` - (optional, default ``false``) If specified ``true``, the returned block will contain all transactions as objects. By default it is ``false`` so, there is no need to explictly specify false. And, if ``false`` it will only contains the transaction hashes.
```

docs/web3-eth.rst
```diff
@@ -915,7 +915,7 @@ Parameters
 
 1. ``String|Number`` - The block number or hash. Or the string ``"genesis"``, ``"latest"`` or ``"pending"`` as in the :ref:`default block parameter <eth-defaultblock>`.
 2. ``Number`` - The index position of the uncle.
-3. ``Boolean`` - (optional, default ``false``) If ``true``, the returned block will contain all transactions as objects, if ``false`` it will only contains the transaction hashes.
+3. ``True`` - (optional, default ``false``) If specified ``true``, the returned block will contain all transactions as objects. By default it is ``false`` so, there is no need to explictly specify false. And, if ``false`` it will only contains the transaction hashes.
```
nivida:
```suggestion
3. ``Boolean`` - (optional, default ``false``) If specified ``true``, the returned block will contain all transactions as objects. By default it is ``false`` so, there is no need to explictly specify false. And, if ``false`` it will only contains the transaction hashes.
```

nivida (approved):

nivida (changes requested):
The CI is failing with the error ``strict-export-declare-modifiers`` [here](https://travis-ci.org/ethereum/web3.js/jobs/607670594#L798-L799).

nivida (approved):

Issue #3042, opened 2019-08-20
Possible closure leak in web3.eth.Contract 1.x ?


ngutman:
#### Expected behavior

Memory usage should stay low over time when instantiating `Contract` objects.

#### Actual behavior

After investigating the heap there seems to be a closure leak which increases memory use over time until the process stops.

#### Steps to reproduce the behavior

You can run the following sample code with the gc exposed
```javascript
const Web3 = require('web3')

async function main() {
  const web3 = new Web3(new Web3.providers.HttpProvider('https://mainnet.infura.io'))

  for (let i = 0; ;i++) {
    new web3.eth.Contract([], '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2')
    if ((i % 1000) === 0) {
      // Use node --expose-gc ./testLeak.js
      if (global.gc) { global.gc() }
      const used = process.memoryUsage().heapUsed / 1024 / 1024
      console.log(`Heap used ${used}`)
    }
  }
}

main()
```

You'll see the memory rapidly increasing:
```
node --expose-gc ./testLeak.js
.....
Heap used 120.24988555908203
Heap used 124.3317642211914
Heap used 128.3788604736328
Heap used 132.4272689819336
Heap used 136.45621490478516
Heap used 140.50664520263672
Heap used 144.50611877441406
Heap used 148.57512664794922
Heap used 152.59766387939453
Heap used 156.6605682373047
Heap used 160.72467803955078
Heap used 164.76478576660156
Heap used 168.77734375
Heap used 172.78372192382812
Heap used 176.8772201538086
Heap used 180.86630249023438
....
```

This is the commit that introduced the issue https://github.com/ethereum/web3.js/commit/26a877517f65c9de650e13f940f995114d9f65ad (apologies for the direct "blame" it's just easier to direct to the actual code with the issue)

Removing the setProvider chaining solves the leak.
### Versions 
* web3.js: v1.2.1
* nodejs:  v10.16.0
* ethereum node: infura mainnet

mgsgde:
Same problem with "web3.js: v2.0.0-alpha". In this version it seems to be not due to a closure leak, but probably due to the following line: https://github.com/ethereum/web3.js/blob/2.x/packages/web3-eth/src/Eth.js#L107 

### Version
- web3.js: v2.0.0-alpha
- nodejs: v10.16.3
- ethereum node: truffle dev network
- truffle: v5.0.16

nivida:
[...]

PR #3156, opened 2019-10-24
Fixes RequestManager handling and adding of the 'data' listener

### LINKED ISSUE #1648: MaxListenersExceededWarning when sending contract (v1.0.0-beta.33)
**Situation**: When I send a deployed contract using `contract.deploy().send()`, the command line prints a warning:
```
  contracts
(node:23488) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 data listeners added. Use emitter.setMaxListeners() to increase limit
    √ gets the accounts

  6 passing (1s)
```

_Edit (14 April 2020): removed unnecessary information_

nivida:
Fixes #1648, #3042 

### Descriptions

Fixes the ``MaxEventListener`` issue with adding of the ``setRequestManager`` function to all packages. The provider will now only get resolved once and the ``data`` listener is no longer added on each initiation of a ``Module``, ``Contract``, or on executing of the ``setProvider`` method on each package.

**Old Behaviour**

1. call ``setProvider``.
2. create RequestManager and register ``data`` listener.
3. call ``setProvider`` on all children's.
4. create RequestManager and register ``data`` listener for each child package.

**New Behaviour**

1. call ``setProvider``.
2. create RequestManager and register ``data`` listener.
3. call ``setRequestManager`` on all children's.


cgewecke (changes requested):
Fixes the warning which is awesome. 💯 💯 

+ [without this PR](https://travis-ci.org/ethereum/web3.js/jobs/603601765#L849-L851)
+ [with this PR](https://travis-ci.org/ethereum/web3.js/jobs/605095785#L878-L879)

Is there anyway to get the diff into a state where only the meaningful logic changes are represented before going over for a quick second pass?

nivida:
@cgewecke Because those code style improvements are made within commits with actually fixes is it really hard to remove those changes. I will not do such changes to newly proposed PRs. Thanks for the hint!

packages/web3-eth-contract/src/index.js
```diff
@@ -56,22 +56,18 @@ var Contract = function Contract(jsonInterface, address, options) {
         args = Array.prototype.slice.call(arguments);
 
     if(!(this instanceof Contract)) {
-        throw new Error('Please use the "new" keyword to instantiate a web3.eth.contract() object!');
+        throw new Error('Please use the "new" keyword to instantiate a web3.eth.Contract() object!');
     }
 
     // sets _requestmanager
-    core.packageInit(this, [this.constructor.currentProvider]);
+    core.packageInit(this, [this.constructor]);
```
cgewecke, line 63: Flagging as logic change.

packages/web3-eth-contract/src/index.js
```diff
@@ -606,18 +613,20 @@ Contract.prototype.once = function(event, options, callback) {
...
  * @method _on
+ *
  * @param {String} event
  * @param {Object} options
  * @param {Function} callback
+ *
  * @return {Object} the event subscription
  */
-Contract.prototype._on = function(){
+Contract.prototype._on = function(event, options, callback){
```
cgewecke: Logic change

packages/web3-eth-contract/src/index.js
```diff
@@ -606,18 +613,20 @@ Contract.prototype.once = function(event, options, callback) {
...
     var subOptions = this._generateEventOptions.apply(this, arguments);
 
 
     // prevent the event "newListener" and "removeListener" from being overwritten
-    this._checkListener('newListener', subOptions.event.name, subOptions.callback);
-    this._checkListener('removeListener', subOptions.event.name, subOptions.callback);
+    this._checkListener('newListener', subOptions.event.name);
+    this._checkListener('removeListener', subOptions.event.name);
```
cgewecke, line 629: Logic change

packages/web3-eth-ens/src/contracts/Registry.js
```diff
@@ -36,13 +35,11 @@ var RESOLVER_ABI = require('../ressources/ABI/Resolver');
...
-    var self = this;
+    const self = this;
     this.ens = ens;
     this.contract = ens.checkNetwork().then(function (address) {
-        var contract = new Contract(REGISTRY_ABI, address);
-        contract.setProvider(self.ens.eth.currentProvider);
-
-        return contract;
+        return new self.ens.eth.Contract(REGISTRY_ABI, address);
+        // contract.setProvider(self.ens.eth.currentProvider);
```
cgewecke: Logic change.

@nivida Do you want to preserve this commented out line?

packages/web3-eth-contract/src/index.js
nivida: The passed parameter doesn't exist on the ``_checkListener`` method and isn't a logic change.

packages/web3-eth-ens/src/contracts/Registry.js
nivida: Thanks for seeing this! I will remove this line 👌

packages/web3-core/src/index.js
```diff
@@ -49,36 +48,40 @@ module.exports = {
...
                 pkg._requestManager.setProvider(provider, net);
                 pkg._provider = pkg._requestManager.provider;
                 return true;
             };
         }
 
+        pkg.setRequestManager = function(manager) {
+            pkg._requestManager = manager;
+            pkg._provider = manager.provider;
+        };
```
cgewecke, line 76: Logic change

nivida: This is a required logic change and also described in the description of this PR:
> Fixes the MaxEventListener issue with adding of the setRequestManager function to all packages.

cgewecke: I'm flagging these so I can see what's material - that's all - sorry.

packages/web3-eth-contract/src/index.js
cgewecke: Ah so this was always wrong?

packages/web3-eth-contract/src/index.js
nivida: This is part of the fix and is required to have the inheritance of the web3 modules working.

cgewecke (commented):
This will take some time to review, it touches some of the most sensitive logic at Web3. The original design seems a little unusual. I think we need to go through the git blame and make sure we understand why it's currently like this.

@nivida Did you ever discuss the RequestManager logic with Frozeman? Do you remember what his views about it were?

cgewecke:
@frozeman Just pinging you here in case you might remember running into any specific difficulties or design questions around how the RequestManager and setProvider are hooked together.

Do you have any ideas for things to test / double-check?

frozeman:
Without reading all the comments. One reason why I made it this way, was so that certain modules can have a storage privet from the global one. 

Eg. Swarm or whisper can (did) run on separate binaries, and Providers.

packages/web3-eth-contract/src/index.js
nivida: Yep, it sadly was.

nivida:
> This will take some time to review, it touches some of the most sensitive logic at Web3. The original design seems a little unusual. I think we need to go through the git blame and make sure we understand why it's currently like this.

I've went through anything and wrote the logic shortly down on paper to be sure I'm not breaking anything during fixing it. 

> @nivida Did you ever discuss the RequestManager logic with Frozeman? Do you remember what his views about it were?

I've discussed this logic around a year ago and the rule was that child modules do inherit the providers from their parents. This means that providers and default options a module can have do overwrite the values of the child modules when a new value/provider got set but not vice versa. 

@frozeman 
> Without reading all the comments. One reason why I made it this way, was so that certain modules can have a storage privet from the global one.
Eg. Swarm or whisper can (did) run on separate binaries, and Providers.

Yep, this is the implemented logic. 👌 


@cgewecke 
The big difference is that it does only resolve the provider once now and does set it downstream on all child modules without resolving the provider again and without creating a new RequestManager instance for each module and Contract. Because it was resolving the provider and creating the RequestManager object all the time again was it register the ``data`` listener N times.

cgewecke:
@nivida 

> Because it was resolving the provider and creating the RequestManager object all the time again was it register the data listener N times.

Yes! I understand this - it's good. I would still like to take a block of time to look through this a bit to think about it, sorry...

nivida:
Moved this fix to #3190

Issue #3098, opened 2019-10-04
Proposal to add E2E and Integration testing on 1.x branch


cgewecke:
(This proposal is continuation of #2682, a companion to #3070, and a tracking issue for all integration testing related topics for the 1.x branch) 

##  E2E
Web3 1.x's stability might benefit from following [ethereum/solidity][8]'s example and E2E testing its latest state against large projects that already use the package.  Solidity does this [nightly in CI][1] - they build a new solc, swap it into Zeppelin, Gnosis and Colony and run those projects' own tests using the new version.  

Some targets that seem promising for this purpose here are: 
- [x] [@truffle/contract][3] 
- [ ] [gnosis/safe-react][4] 

**Technical**

Web3 is a Lerna monorepo and so are two of the potential targets listed above. Because its repo structure and published formats are quite different it won't just `npm install` from a github url. 

However, there's a nice tool called [verdaccio][5] that lets you set up a local private npm registry in CI. Web3 could "publish" a candidate branch to this ephemeral registry, clone an E2E target (ex: embark) and install the new web3 via npm proxy. Then run the target's tests.

It's also possible to chain publications together. For example, openzeppelin-solidity uses truffle so you could:
+ publish a new Web3, 
+ install it in Truffle, 
+ publish a new Truffle, 
+ swap the new Truffle into Zeppelin and run Zeppelin's tests.

Additionally: 
+ could be used to verify that built assets are located correctly and usable. 
+ [ethereum-js][9] may migrate some things to monorepo in future. Any solution developed here might pay dividends there as well. 

## Integration

**Clients**

To recap #2682 - Web3 1.x's unit tests rely primarily on mocked server responses. They're super fast but some things are hard to validate without testing against a node. There's also variance in behavior between the three most widely used clients. It would be good to create a small suite of tests for stuff like:

+ simple send and call
+ events
+ subscriptions
+ errors 
+ transaction signing 

..and make sure everything executes as expected across this matrix:

| Client       | instamine | automine 2s |
|--------------|-----------|-------------|
| geth --dev   |     x     |      x      |
| parity --dev |     x     |      x      |
| ganache-cli  |     x     |             |

**Client Helpers**

To simplify things in CI and make this work portable, Web3 could also publish small Node cli wrappers for the dockerized Parity and Geth dev clients.

These would:
+ pull down any client version from docker hub 
+ abstract away most of command config flags, 
+ open all APIs, 
+ have many well-funded accounts
[...]

PR #3157, opened 2019-10-25
Add npm registry proxy in CI to test publication/installation

cgewecke:
(Continuation of #3098) 

Adds the ability to publish the monorepo to an ephemeral virtual registry in CI and install Web3's latest state in an arbitrary target project. This lets us run other projects' unit tests to verify that no breaking changes have been introduced.

Includes:
+ npm virtual registry via the verdaccio package
+ uses @truffle/contract as an E2E target

One thing to note about this is that it doesn't work very well when the target requires that installation be done with yarn. (Getting Truffle to work with this format was a little challenging.) 

Truffle's tests are run as an "allowed failure" and CI will turn green without waiting for them to finish. This is done because these tests are mostly a pre-publication sanity check which might help us track breaking changes down to a specific PR if they occur.


nivida (approved):

alcuadrado:
This is great @cgewecke 😍

Issue #3134, opened 2019-10-16
Error "Returned values aren't valid, did it run Out of Gas" not fixed in v1.2.1


barakman:
### Description 

This error is reported in details at issue #1916.

Users mention the fact that it dates back as far as v1.0.0-beta.35 or v1.0.0-beta.36.

I know for a fact that it doesn't occur at v1.0.0-beta.34.

It was supposedly fixed at PR #2608 for v2.0.0-alpha.1 and v1.0.0-beta.52.

Unfortunately, it doesn't seem to be fixed on v1.2.1.

#### Expected behavior

Calling a method which is viable in the ABI but not in the bytecode, should result with an error-message starting with ` Couldn't decode`.

For example: ` Couldn't decode uint16 from ABI: 0x`.

AFAIK, this error is routed all the way from the provider itself (rather than being generated by the web3 client code).

#### Actual behavior

Calling a method which is viable in the ABI but not in the bytecode, results with an error-message `Returned values aren't valid, did it run Out of Gas?`, which is implemented in function `ABICoder.prototype.decodeParameters` (as part of the web3 client code).

#### Steps to reproduce the behavior

Install either web3 1.0.0-beta.34 or web3 1.2.1, and run the following script via NodeJS:
```
const Web3 = require("web3");

async function test() {
    const web3 = new Web3("https://mainnet.infura.io");
    const abi = JSON.parse('[{"constant":true,"inputs":[],"name":"token","outputs":[{"name":"","type":"address"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[],"name":"connectorTokenCount","outputs":[{"name":"","type":"uint16"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[{"name":"_index","type":"uint256"}],"name":"connectorTokens","outputs":[{"name":"","type":"address"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[],"name":"reserveTokenCount","outputs":[{"name":"","type":"uint16"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[{"name":"_index","type":"uint256"}],"name":"reserveTokens","outputs":[{"name":"","type":"address"}],"payable":false,"stateMutability":"view","type":"function"}]');
    const address = "0x5894110995b8c8401bd38262ba0c8ee41d4e4658";
    const contract = new web3.eth.Contract(abi, address);
    try {
[...]

PR #3161, opened 2019-10-27
Improve eth-abi decodeParameters error message

cgewecke:
Addresses #3134

Also adding a couple small tests / e2e fixtures for 'method.call', because those are missing. 

There's discussion in #3134 about whether current error makes sense at all because it mentions OOG in a context where that's unlikely to be a cause. Am very reluctant to change error messages in 1.x because people may be grepping them to handle certain error conditions, so have just appended some more info that might be helpful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,3 +90,4 @@ Released with 1.0.0-beta.37 code base.
 ### Fixed
 
 - Fix TS types for eth.subscribe syncing, newBlockHeaders, pendingTransactions (#3159)
+- Improve web3-eth-abi decodeParameters error message (#3134)
--- a/packages/web3-eth-abi/src/index.js
+++ b/packages/web3-eth-abi/src/index.js
@@ -223,7 +223,11 @@ ABICoder.prototype.decodeParameter = function (type, bytes) {
  */
 ABICoder.prototype.decodeParameters = function (outputs, bytes) {
     if (outputs.length > 0 && (!bytes || bytes === '0x' || bytes === '0X')) {
-        throw new Error('Returned values aren\'t valid, did it run Out of Gas?');
+        throw new Error(
+            'Returned values aren\'t valid, did it run Out of Gas? ' +
+            'Are you using the correct ABI for the contract you are ' +
+            'retrieving data from?'
+        );
     }
 
     var res = ethersAbiCoder.decode(this.mapTypes(outputs), '0x' + bytes.replace(/0x/i, ''));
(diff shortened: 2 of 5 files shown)


barakman:
Just to clarify, I have initially opened #3134 because I was grepping `error.message.startsWith("Couldn't decode")` while using web3 version 1.0.0-beta.34 (a very stable and reliable version BTW).

After upgrading to version 1.2.1, I stumble across this issue, but as a **runtime error** which occurred long after I had upgraded web3. As a result, I had to conduct a rather long investigation before I was able to figure out the cause.

So I definitely agree that this would be an API-breaking change for those already using version 1.2.1, and that it should therefore not be changed at any future version 1.x.

Side note for readers stumbling across the same problem:
I have ultimately resorted to relying only on a single error-message, namely:
```
error.message.startsWith("Invalid JSON RPC response")
```
So it could be a solution (or a workaround if you will) for some of you.

nivida (changes requested):
@cgewecke We can keep the new error msg in the ABICoder as it is but we have to back-port the fix from the [CallContractMethod](https://github.com/ethereum/web3.js/blob/2.x/packages/web3-eth-contract/src/methods/CallContractMethod.js). I was following several discussions in Geth issues about the correct handling of this case and the overall opinion and also mine was to return null.

cgewecke:
> the correct handling of this case and the overall opinion and also mine was to return null.

@nivida Ok, interesting. Is this a breaking change though?

Do you remember where this was discussed at geth so it can be cross-documented here when the change is made?

nivida:
@cgewecke 

> Ok, interesting. Is this a breaking change though?

Returning ``null`` instead of throwing an error if the requested value isn't available because of a block number defined in the future or if the node isn't in sync is in my view of point the more accurate behavior instead of throwing the current confusing error on for example a contract method call. I think this is such a small breaking change and does fix a "bug" (clarifies the situation) that we actually could do this on the 1.2.3 release.

> Do you remember where this was discussed at geth so it can be cross-documented here when the change is made?

I couldn't find it but I'm pretty sure the last research I've done for the fix in 2.x brought this conclusion up.

cgewecke:
@nivida I agree the error message is not good but changing behavior from throwing an error to not throwing one could be a big difference to someone - it's very hard to anticipate how people rely on this....

To me it seems the safest improvement is to list the known conditions which trigger this. Will add the two other cases you've mentioned.

nivida (approved):

nivida:
@cgewecke I will keep issue #3134 open for later (1.3.0) implementing an improved solution for this bug. We could, for example, do further node-state-related JSON-RPC requests to determine more details and to return a more defined error msg.

PR #3135, opened 2019-10-16
Full support for websocket reconnection/resubscription

nivida:
Proposed the PR from @gabmontes again. Because the 1.0 branch is no longer existing is not possible to re-open the already existing PR #1966.

PR #3167, opened 2019-10-29
Full support for websocket reconnection/resubscription

cgewecke:
This is #3135, #1966 redux, merged to the root repo so merge conflicts and tests can be managed here. Tests are WIP, opening as draft.

### DIFF
--- a/test/eth.subscribe.ganache.js
+++ b/test/eth.subscribe.ganache.js
@@ -1,0 +1,99 @@
+const assert = require('assert');
+const ganache = require('ganache-cli');
+const pify = require('pify');
+const Basic = require('./sources/Basic');
+const utils = require('./helpers/test.utils');
+const Web3 = utils.getWeb3();
+
+describe('subscription connect/reconnect', function(){
+    let server;
+    let web3;
+    let accounts;
+    const port = 8545;
+
+    beforeEach(async function(){
+        server = ganache.server({port: port});
+        await pify(server.listen)(port);
+        web3 = new Web3('ws://localhost:' + port);
+        accounts = await web3.eth.getAccounts();
+    })
+
+    afterEach(async function(){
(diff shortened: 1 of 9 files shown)


nivida:
@cgewecke Sorry for commenting on this draft PR. But I would like to ask you what the state is of the mentioned test cases? Is it possible to commit those and I'm gonna finish them?

cgewecke:
@nivida Ah thanks for the reminder - I will get these in..

test/eth.subscribe.ganache.js
```diff
@@ -0,0 +1,99 @@
...
+                .on('error', function(err){
+                    assert(err.message.includes("CONNECTION ERROR"))
+                    this.removeAllListeners();
+                    resolve();
+                });
+        });
+    });
+
+    // This test failing....
+    it.only('auto reconnects', function(){
```
cgewecke, line 58: @nivida This is my attempt to model reconnection working. 
+ Open server
+ Attach a subscription handler
+ Fire a tx (which subscrip hears)
+ Close server
+ Re-open server
+ Fire a tx (which subscrip hears)

But the test doesn't reconnect...do you want to take a look?

cgewecke: We need to make sure this works against an actual client, not just mock some assumptions in.

cgewecke: Maybe my model of how to do this is wrong...not sure.

cgewecke: @nivida mentioned something in a chat that's important which I'm leaving here for further reference. 

The way this test models disconnection doesn't reflect the connectivity issues people typically have online e.g. nodes are not shutting down, killing their sockets and restarting...

nivida:
Moved this to #3190

PR #3160, opened 2019-10-25
New build pipeline

nivida:
Fixes #3155 #3074 #2623 #1178

### Description

Because of repeating issues (#3155) we had with the minified bundle did I back-port and extend the build pipeline of the 2.x branch. 

**Key Facts:**

- Rollup with the related plugins will be used to bundle, transform and minify the web3 bundles.
- Karma setup updated (it wasn't using the minified file from the project)
- Bundle sizes optimized
  + parcel zero-config bundle: 1.12 MB -> 1.04 MB
  + web3.min.js: 1.1MB -> 762KB
- Modern JS can now be used anywhere
- CJS, ESM, and minified UMD bundles are now existing for each web3 package
- Babel transforms the bundle for the minified file to support the last 2 browser versions 
- Babel transforms the CJS and ESM bundle to support nodejs v8+
- ``eventemitter3`` dependency updated in PromiEvent package 
  + I've checked the release announcements for breaking changes
- ``ENS.js`` moved to ``index.js`` for consistency reasons

**Scripts**

- ``build:web3`` - Creates just the CJS and ESM bundle of the web3 umbrella package
- ``build:web3:minified`` - Creates all CJS and ESM bundles of all packages and creates a web3.min after
- ``build:all`` - Creates CJS and ESM bundles for all packages
- ``build:all:cjs`` - Creates CJS bundles for all packages
- ``build:all:esm`` - Creates ESM bundles for all packages
- ``build:all:minified`` - Creates minified UMD bundles with the ``cjs`` bundles as the source for all packages
- ``build:all:release`` - Creates CJS, ESM, and minified UMD bundles for all packages without instrumenting the code for the ``istanbul`` code coverage report.

**New Package Entries**

- ``main`` 
  + Will be used in nodejs on a normal ``require(...)`` and does have the CJS module format.
- ``module``
  + Will be used for ESM-ready tools and does have the ES format.
- ``unpkg`` 
  + Will be used from the ``unpkg`` CDN and does contain the minified UMD formatted bundle.
- ``jsdelivr``
  + Will be used from the ``jsdelivr`` CDN and does contain the minified UMD formatted bundle.

**Rollup Config Setup**

The base configuration is located in the root folder of this repository and does return the configuration function which will be used in each package of this project. 

Properties of the [config function](https://github.com/ethereum/web3.js/pull/3160/files#diff-ff6e5f22a9c7e66987b19c0199636480R140): 

- ``name: string``
  + Will be used for named exports
- ``outputFileName: string``
[...]

PR #3170, opened 2019-10-31
Fix e2e lerna publish (#3)

cgewecke:
(**DO NOT MERGE** until making suggested change in the comment below)

Unfortunately #3157 was only ~%25 correct :/ 

PR fixes several problems:
+ The packages were not being published by the lerna command correctly - they used `exec` (obviously wrong). I've had to upgrade Lerna to 3.x to get this to work.
+ Unexpectedly, truffle/contract mostly uses a [Web3 from a different truffle module][1] than itself, even though Web3 is a direct dependency there.  Both modules need to be modified to get the tests to consume the virtual publication correctly

I've added a 'proof' that this works - there's an incorrect console.log line in `web3-core-helpers` which is printed out while truffle's tests run. So we know their code is exercising the published copy.
It needs to be removed before merging..

Lastly, some of Truffle's tests are failing with an error that's already been reported at [truffle 2512][2]. I don't think any recent changes here are the cause because they [have not updated to 1.2.2 yet][3]. 

[1]: https://github.com/trufflesuite/truffle/blob/develop/packages/contract/lib/contract/constructorMethods.js#L219-L221
[2]: https://github.com/trufflesuite/truffle/issues/2512
[3]: https://github.com/trufflesuite/truffle/pull/2488

### DIFF
--- a/packages/web3-core-helpers/src/formatters.js
+++ b/packages/web3-core-helpers/src/formatters.js
@@ -219,6 +219,7 @@ var outputTransactionReceiptFormatter = function (receipt){
     if(typeof receipt.status !== 'undefined' && receipt.status !== null) {
         receipt.status = Boolean(parseInt(receipt.status));
     }
+    console.log('using new web3 at to format transaction receipt')
 
     return receipt;
 };
--- a/scripts/e2e.npm.publish.sh
+++ b/scripts/e2e.npm.publish.sh
@@ -34,11 +34,30 @@ npx npm-auth-to-token \
   -e test@test.com \
   -r http://localhost:4873
 
-# `npm version patch`
-npx lerna exec -- npm version patch
+# Prep branch for Lerna's git-checks
+BRANCH=$TRAVIS_PULL_REQUEST_BRANCH
+if [ -z "$BRANCH" ]; then
 
-# `npm publish`
-npx lerna exec --concurrency 1 -- npm publish \
-  --tag e2e \
+  BRANCH=$TRAVIS_BRANCH
+
+fi
+
+git checkout $BRANCH --
+
+# Lerna version
+npx lerna version patch \
+  --force-publish=* \
+  --no-git-tag-version \
+  --no-push \
+  --allow-branch $BRANCH \
+  --yes
+
+# Commit changes because lerna checks git before
+git commit -a -m 'virtual-version-bump'
+
+# Lerna publish to e2e tag
+npx lerna publish from-package \
+  --dist-tag e2e \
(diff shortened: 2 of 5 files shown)


packages/web3-core-helpers/src/formatters.js
```diff
@@ -219,6 +219,7 @@ var outputTransactionReceiptFormatter = function (receipt){
     if(typeof receipt.status !== 'undefined' && receipt.status !== null) {
         receipt.status = Boolean(parseInt(receipt.status));
     }
+    console.log('using new web3 at to format transaction receipt')
```
cgewecke, line 222: Edit this file to remove this line. It proves that the bundle is being published/used correctly [in travis here](https://travis-ci.com/cgewecke/web3.js/jobs/251293678#L1669-L1676).

nivida (changes requested):
The truffle e2e container is failing during the bootstrapping process:

Bildschirmfoto 2019-10-31 um 12 31 41

nivida:
@cgewecke The whole ``e2e`` tests concept should probably be documented somewhere here in the repository otherwise is it hard to follow up our current thoughts in some months (I will add a build pipeline section to the contributor's section of the readme in #3160 as well).

cgewecke:
> The whole e2e tests concept should probably be documented somewhere here in the repository otherwise is it hard to follow up our current thoughts in some months (I will add a build pipeline section to the contributor's section of the readme in #3160 as well

I agree, this is a good idea.

scripts/e2e.npm.publish.sh
```diff
@@ -34,11 +34,30 @@ npx npm-auth-to-token \
...
+if [ -z "$BRANCH" ]; then
 
-# `npm publish`
-npx lerna exec --concurrency 1 -- npm publish \
-  --tag e2e \
+  BRANCH=$TRAVIS_BRANCH
+
+fi
+
+git checkout $BRANCH --
```
cgewecke, line 45: The build is failing here because 
+ Lerna [won't publish to the virtual registry][1] when the git head is in a detached state. 
+ It's necessary to explicitly checkout a branch to trick it
+ The branch isn't present when a fork's changes are proposed for the upstream

I'm going to resolve this temporarily by merging into a staging branch at web3 to make sure the CI runs as expected here. It would be nice if this ran for outside contributors, but perhaps is not critical that it does either because this kind of test is more of a pre-publication tests that's use internally anyway. 

https://github.com/lerna/lerna/issues/1595#issuecomment-422532097

PR #3171, opened 2019-10-31
Fixes 'Connection not open on send'

### LINKED ISSUE #3092: [v1.2.1] Connection not Open.
### Description 
스크린샷 2019-10-02 오후 5 54 53

#### Expected behavior

I get an error periodically.

maybe

web3.eth.getBalance or web.eth.getBlock .. etc

<!---------------------------------------------------------------------

const Web3 = require('web3');
const web3 = new Web3('ws://' + serverList[0].host + ":" + serverList[0].port);
var endBlockNumber = await web3.eth.getBlockNumber();
-> Error. Can't get BlockNumber.

Or 

		var amounts = await web3.eth.getBalance('0xFD2ACe8086F9FeF3da294Cc69b4750D59bd3C61A');

-> Error.

But sometimes Success...

Please Help me.. 

I'm crazing...

### Versions 
* web3.js: 1.2.1
* nodejs: v10.16.3
* browser: experess.

nivida:
Fixes #3092 with back-porting of the fix from 2.x

Instead of only waiting 10ms with a ``setTimeout`` and to try it again does it now register a listener on the ``open`` event of the underlying WebSocket object. I've configured the listener options with ``once: true`` to be sure the listener gets removed after it got invoked once. 

ToDo’s: 

- [ ] Add test cases for it

### DIFF
--- a/packages/web3-providers-ws/src/index.js
+++ b/packages/web3-providers-ws/src/index.js
@@ -276,7 +276,7 @@ WebsocketProvider.prototype.send = function(payload, callback) {
     //     this.connection.connect({url: this.url});
     if (this.connection.readyState !== this.connection.OPEN) {
         if (typeof this.connection.onerror === 'function') {
-            this.connection.onerror(new Error('connection not open'));
+            this.connection.onerror(new Error('connection not open on send()'));
         } else {
             console.error('no error callback');
         }
--- a/packages/web3-providers-ws/src/index.js
+++ b/packages/web3-providers-ws/src/index.js
@@ -20,7 +20,7 @@
  * @date 2017
  */
 
-"use strict";
+'use strict';
 
 var _ = require('underscore');
 var errors = require('web3-core-helpers').errors;
@@ -41,8 +41,7 @@ if (isNode) {
         parseURL = function(url) {
             return new newURL(url);
         };
-    }
-    else {
+    } else {
         // Web3 supports Node.js 5, so fall back to the legacy URL API if necessary
         parseURL = require('url').parse;
     }
@@ -55,9 +54,7 @@ if (isNode) {
 // Default connection ws://localhost:8546
 
 
-
-
-var WebsocketProvider = function WebsocketProvider(url, options)  {
+var WebsocketProvider = function WebsocketProvider(url, options) {
     if (!Ws) {
         throw new Error('websocket is not available');
     }
@@ -81,7 +78,7 @@ var WebsocketProvider = function WebsocketProvider(url, options)  {
 
     // Allow a custom client configuration
     var clientConfig = options.clientConfig || undefined;
-    
+
     // Allow a custom request options
     // https://github.com/theturtle32/WebSocket-Node/blob/master/docs/WebSocketClient.md#connectrequesturl-requestedprotocols-origin-headers-requestoptions
     var requestOptions = options.requestOptions || undefined;
@@ -101,14 +98,14 @@ var WebsocketProvider = function WebsocketProvider(url, options)  {
         /*jshint maxcomplexity: 6 */
         var data = (typeof e.data === 'string') ? e.data : '';
 
-        _this._parseResponse(data).forEach(function(result){
+        _this._parseResponse(data).forEach(function(result) {
 
             var id = null;
 
             // get the id which matches the returned id
-            if(_.isArray(result)) {
-                result.forEach(function(load){
-                    if(_this.responseCallbacks[load.id])
+            if (_.isArray(result)) {
+                result.forEach(function(load) {
+                    if (_this.responseCallbacks[load.id])
                         id = load.id;
                 });
             } else {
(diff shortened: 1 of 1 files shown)


packages/web3-providers-ws/src/index.js
```diff
@@ -250,36 +247,42 @@ WebsocketProvider.prototype._addResponseCallback = function(payload, callback) {
...
-WebsocketProvider.prototype.send = function (payload, callback) {
+WebsocketProvider.prototype.send = function(payload, callback) {
     var _this = this;
 
     if (this.connection.readyState === this.connection.CONNECTING) {
-        setTimeout(function () {
-            _this.send(payload, callback);
-        }, 10);
+        this.connection.addEventListener(
+            'open',
```
cgewecke, line 264: What happens if this is never fired because the client is not available, for example when testing against a local node the user forgot to turn on? In the current flow it looks like it will send and error, but here maybe hang?

packages/web3-providers-ws/src/index.js
```diff
@@ -250,36 +247,42 @@ WebsocketProvider.prototype._addResponseCallback = function(payload, callback) {
...
+        );
+
         return;
     }
 
     // try reconnect, when connection is gone
     // if(!this.connection.writable)
     //     this.connection.connect({url: this.url});
     if (this.connection.readyState !== this.connection.OPEN) {
-        console.error('connection not open on send()');
```
cgewecke, line 276: I wonder why this was here initially?  I think unless there's a strong reason to remove this (other than it seems like it might have been a mistake a long time ago) it should stay in.  People often run shell commands hooking into the available streams including stdout and this could easily be something that is getting grepped for connection failure.

cgewecke (changes requested):
I left two comments - are those two locations the only logic changes? Agree with your assessment that this could use tests.

packages/web3-providers-ws/src/index.js
```diff
@@ -20,7 +20,7 @@
  * @date 2017
  */
 
-"use strict";
+'use strict';
```
cgewecke, line 23: Mentioned this in #3180. What do you think about keeping the unrelated style changes out? It reduces the diff clarity quite a bit and makes it harder to see what's critical. 

How the project is linted/styled could be addressed later at some point as separate effort, with a linter, a way to enforce the standard (in CI or a pre-commit hook like [husky](https://www.npmjs.com/package/husky#husky)).  And a strategy to introduce the changes in a uniform way.

packages/web3-providers-ws/src/index.js
nivida: This is true it will with the changed lines of code just hang until the connection got established and in the case of a non-existing host will it never execute the JSON-RPC methods we have in the pipeline. The emitted error ``net::ERR_CONNECTION_REFUSED`` which would signal us the non-existing host. Can't get caught with a try/catch in the ``WebsocketProvider`` constructor because it doesn't throw an error in the normal sense. The ``WebSocket`` class does dispatch a non-listenable event because of the way it is implemented. A solution would be to create a ``connecting-timeout`` function in the constructor of the ``WebsocketProvider`` which throws an error if the connection doesn't get established in the defined time frame.

packages/web3-providers-ws/src/index.js
nivida: Thanks for the hint! I will keep those changes in this PR and will not change the code-style in all newly opened PRs.

packages/web3-providers-ws/src/index.js
nivida: I've removed this because it is redundant. The ``console.error`` function will be called if no error listener is registered which means this error message would have been printed twice in this case and the error will anyways get passed to the provided callback function.

packages/web3-providers-ws/src/index.js
```diff
@@ -250,36 +247,42 @@ WebsocketProvider.prototype._addResponseCallback = function(payload, callback) {
...
     if (this.connection.readyState !== this.connection.OPEN) {
-        console.error('connection not open on send()');
         if (typeof this.connection.onerror === 'function') {
             this.connection.onerror(new Error('connection not open'));
         } else {
             console.error('no error callback');
         }
-        callback(new Error('connection not open'));
+
+        callback(new Error('connection not open on send()'));
```
cgewecke, line 284: 👍

packages/web3-providers-ws/src/index.js
cgewecke: Ok yes LGTM.

packages/web3-providers-ws/src/index.js
cgewecke: > A solution would be to create a connecting-timeout function in the constructor of the WebsocketProvider which throws an error if the connection doesn't get established in the defined time frame.

Great analysis! To me it seems like there are two execution contexts for Web3 with opposite expectations:
+ NodeJS (testing/tools) --> fail fast and hard.
+ Network (browser/dapps) --> keep trying for as long as it takes
 
At the moment, web3 behaves mostly the way NodeJS contexts would prefer, and there are lots of complaints on the browser / dapp side. Maybe we can just acknowledge the difference in the docs/release notes and address it by toggling the different behaviors  via  the `autoReconnect: true`  (or something similar) proposed in #3167. Here you might just check that flag - stay open if true, fail fast as Web3 currently does if false?

WDYT? This change could be paired with #3167...

packages/web3-providers-ws/src/index.js
```diff
@@ -250,36 +247,42 @@ WebsocketProvider.prototype._addResponseCallback = function(payload, callback) {
...
         return;
     }
 
     // try reconnect, when connection is gone
     // if(!this.connection.writable)
     //     this.connection.connect({url: this.url});
     if (this.connection.readyState !== this.connection.OPEN) {
-        console.error('connection not open on send()');
         if (typeof this.connection.onerror === 'function') {
             this.connection.onerror(new Error('connection not open'));
```
cgewecke: Identical message maybe?
```suggestion
            this.connection.onerror(new Error('connection not open on send()'));
```

nivida:
Moved this fix to #3190

PR #3173, opened 2019-10-31
Fix lerna e2e publish

cgewecke:
Re-opens 3170 in a local context because there's [some git logic that's necessary to trick Lerna into publishing][4] which is not working when the changes are being proposed from a fork.  It would be nice if this ran normally for outside contributors, but perhaps its not critical because this kind of test is more of a pre-publication test / sanity-check. It *should* work when an outside contribution is merged - that build can always be verified if the reviewer had questions...

Unfortunately #3157 was only ~%25 correct :/ 

PR fixes several problems:
+ The packages were not being published by the lerna command correctly - they used `exec` (obviously wrong). I've had to upgrade Lerna to 3.x to get this to work.
+ Unexpectedly, truffle/contract mostly uses a [Web3 from a different truffle module][1] than itself, even though Web3 is a direct dependency there.  Both modules need to be modified to get the tests to consume the virtual publication correctly

I've added a 'proof' that this works - there's an incorrect console.log line in `web3-core-helpers` which is printed out while truffle's tests run. So we know their code is exercising the published copy.
It needs to be removed before merging..

Lastly, some of Truffle's tests are failing with an error that's already been reported at [truffle 2512][2]. ~~I don't think any recent changes here are the cause because they [have not updated to 1.2.2 yet][3].~~ 

Actually it looks like they have some tests failing as direct result of changes in 1.2.2. See review discussion below. 

[1]: https://github.com/trufflesuite/truffle/blob/develop/packages/contract/lib/contract/constructorMethods.js#L219-L221
[2]: https://github.com/trufflesuite/truffle/issues/2512
[3]: https://github.com/trufflesuite/truffle/pull/2488
[4]: https://github.com/lerna/lerna/issues/1595#issuecomment-422532097

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
         "karma-firefox-launcher": "^1.2.0",
         "karma-mocha": "^1.3.0",
         "karma-spec-reporter": "0.0.32",
-        "lerna": "^2.11.0",
+        "lerna": "^3.18.3",
         "mocha": "^6.2.1",
         "npm-auth-to-token": "^1.0.0",
         "puppeteer": "^1.20.0",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
         "test": "mocha --grep E2E --invert; jshint *.js packages",
         "test:unit": "istanbul cover _mocha -- -R spec --grep E2E --invert",
         "dtslint": "lerna run dtslint",
-        "geth": "geth-dev-assistant --accounts 5 --tag stable",
+        "geth": "geth-dev-assistant --accounts 5 --tag stable --gasLimit 7000000",
         "test:e2e:ganache": "./scripts/e2e.ganache.sh",
         "test:e2e:geth:auto": "./scripts/e2e.geth.automine.sh",
         "test:e2e:geth:insta": "./scripts/e2e.geth.instamine.sh",
--- a/packages/web3-core-helpers/src/formatters.js
+++ b/packages/web3-core-helpers/src/formatters.js
@@ -219,6 +219,7 @@ var outputTransactionReceiptFormatter = function (receipt){
     if(typeof receipt.status !== 'undefined' && receipt.status !== null) {
         receipt.status = Boolean(parseInt(receipt.status));
     }
+    console.log('using new web3 at to format transaction receipt')
 
     return receipt;
 };
--- a/packages/web3-core-helpers/src/formatters.js
+++ b/packages/web3-core-helpers/src/formatters.js
@@ -219,7 +219,6 @@ var outputTransactionReceiptFormatter = function (receipt){
     if(typeof receipt.status !== 'undefined' && receipt.status !== null) {
         receipt.status = Boolean(parseInt(receipt.status));
     }
-    console.log('using new web3 at to format transaction receipt')
 
     return receipt;
 };
--- a/scripts/e2e.truffle.sh
+++ b/scripts/e2e.truffle.sh
@@ -16,13 +16,32 @@ git clone https://github.com/trufflesuite/truffle.git
 cd truffle
 yarn bootstrap
 
-# @truffle/contract
-cd packages/contract
+yarn config set registry http://localhost:4873
 
-# Uninstall / re-install web3
+echo ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
+echo "Updating @truffle/interface-adapter"
+echo ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
+
+# @truffle/interface-adapter
+cd packages/interface-adapter
 yarn remove web3
 
-npm install web3@e2e \
+yarn add web3@e2e \
+  --registry http://localhost:4873 \
+  --force
+
+# @truffle/contract
+echo ">>>>>>>>>>>>>>>>>>>>>>>>>>"
+echo "Updating @truffle/contract"
+echo ">>>>>>>>>>>>>>>>>>>>>>>>>>"
+
+cd ../contract
+yarn remove web3
+yarn remove web3-core-promievent
+yarn remove web3-eth-abi
+yarn remove web3-utils
+
+yarn add web3@e2e \
   --registry http://localhost:4873 \
   --force
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 services:
     - xvfb
 language: node_js
+git:
+  depth: false
 matrix:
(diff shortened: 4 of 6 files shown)


packages/web3-core-helpers/src/formatters.js
```diff
@@ -219,6 +219,7 @@ var outputTransactionReceiptFormatter = function (receipt){
     if(typeof receipt.status !== 'undefined' && receipt.status !== null) {
         receipt.status = Boolean(parseInt(receipt.status));
     }
+    console.log('using new web3 at to format transaction receipt')
```
cgewecke: Edit this file to remove this line. It proves that the bundle is being published/used correctly [in travis here](https://travis-ci.com/cgewecke/web3.js/jobs/251293678#L1669-L1676).

cgewecke:
@nivida Apologies for the earlier PR, this is now working as expected for both PR and Push builds. The truffle tests fail for reasons given in the PR description above, but the publication / installation part of this finally seems correct:

+ [Push Build Proof](https://travis-ci.org/ethereum/web3.js/jobs/605691127#L1809-L1820)
+ [PR Build Proof](https://travis-ci.org/ethereum/web3.js/jobs/605691149#L1791-L1801)

nivida (changes requested):
@cgewecke Thanks for referencing those builds but unfortunately both of them do fail and we didn't introduce a breaking change.

package.json
```diff
@@ -24,7 +24,7 @@
         "test": "mocha --grep E2E --invert; jshint *.js packages",
         "test:unit": "istanbul cover _mocha -- -R spec --grep E2E --invert",
         "dtslint": "lerna run dtslint",
-        "geth": "geth-dev-assistant --accounts 5 --tag stable",
+        "geth": "geth-dev-assistant --accounts 5 --tag stable --gasLimit 7000000",
```
nivida, line 27: Why exactly is it required to set ``gasLimit`` here to ``7000000``?

scripts/e2e.truffle.sh
```diff
@@ -16,13 +16,32 @@ git clone https://github.com/trufflesuite/truffle.git
...
 yarn bootstrap
 
+yarn config set registry http://localhost:4873
+
+echo ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
+echo "Updating @truffle/interface-adapter"
+echo ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
+
+# @truffle/interface-adapter
+cd packages/interface-adapter
```
nivida, line 26: Is there a way to have the uninstalling and installing process for ``@truffle/interface-adapter`` and ``@truffle/contract`` managed here in a more generic way? 

It will probably be the case that also other modules will get dependencies added to web3 packages and it would be great to have this for example in a well-structured and easy to read JSON object.

cgewecke: The best way would be upgrade web3 uniformly across all packages at Truffle. At truffle they do this using a 3rd party lerna utility called lerna-update-wizard which looks like it [does not support](https://github.com/Anifacted/lerna-update-wizard/issues/36) specifying a private registry for the new dependency when using yarn.

But I agree this is tightly coupled to a specific configuration of dependencies truffle/contract currently has and could need maintenance in the future.

Perhaps this question can be addressed a later PR. Honestly not sure how to resolve at the moment and the current commands achieve the near term goal... need to think about this more though.

package.json
cgewecke: Geth dev launches with a relatively low block gas limit (~5,900,000). This is just to make sure the client is [using the settings](https://github.com/trufflesuite/truffle/blob/develop/scripts/geth.sh#L23) Truffle uses in it's own tests with Geth, which is higher. In practice most of the Truffle contract tests use gas estimation so it might not be necessary, but setting this removes one variable in case things change over there.

cgewecke:
@nivida 

> Thanks for referencing those builds but unfortunately both of them do fail and we didn't introduce a breaking change.

In fact Truffle is not able to upgrade to 1.2.2 because of several problems including #3175. I have opened [an issue over there](https://github.com/trufflesuite/truffle/issues/2539) notifying them. 

There is also [a failing test for confirmations over http](https://travis-ci.org/ethereum/web3.js/jobs/605691127#L1947-L1950) which *could* be related to the changes in #3142. Those changes should get some additional E2E tests here to verify their behavior is what we're expecting. 

So overall I think this E2E is working as expected now (but was broken before when I thought everything was ok).

scripts/e2e.truffle.sh
cgewecke: > it would be great to have this for example in a well-structured and easy to read JSON object.

Oh, just coming back here for a second look - this is a really good suggestion. Would need to add logic for processing the JSON in shell, maybe even with a node script. 

Can we investigate systematically updating all the web3 deps at Truffle in a second PR and if that's not possible or too complicated - implementing this idea?

nivida: Sounds good 👌

nivida:
> In fact Truffle is not able to upgrade to 1.2.2 because of several problems including #3175. I have opened an issue over there notifying them.

Great! Thanks! I was in contact some days ago with Nicholas from truffle as well. 👌  

> There is also a failing test for confirmations over http which could be related to the changes in #3142. Those changes should get some additional E2E tests here to verify their behavior is what we're expecting.

Hmm, the behaviour didn't change. I will get in contact with the right people at truffle to do further investigations. 

> So overall I think this E2E is working as expected now (but was broken before when I thought everything was ok).

Yep, looks good! Thanks! 💪 

Feel free to resolve the conflicts and merge this PR.

nivida (approved):

packages/web3-core-helpers/src/formatters.js
cgewecke: Removed in fd8f3a7acc360d9d8251f0efb6558f644032c93f

cgewecke:
@nivida I am not allowed to merge without 2 approving reviews (which is good! Safer.) 

This is technically merge-able although it might need a quick patch in a few days as well because some things about how web3 is being installed are under review at Truffle atm ([see here](https://github.com/trufflesuite/truffle/pull/2558#pullrequestreview-312767609)).

So merge or wait idk - am watching developments over there and will fix as necessary. 

Also looks like [http confirmations handler is ok](https://github.com/trufflesuite/truffle/pull/2558/checks?check_run_id=291529310#step:6:1008) at Truffle now with 1.2.2.

PR #3177, opened 2019-11-05
Internal improvement: Add & adjust web3-core-helper typings

CruzMolina:


packages/web3-core-helpers/types/index.d.ts
```diff
@@ -84,8 +84,8 @@ export class WebsocketProviderBase {
 
     send(
         payload: JsonRpcPayload,
-        callback: (error: Error, result: any) => void
-    ): void;
+        callback: Callback<JsonRpcResponse>
+    ): any;
```
nivida:
```suggestion
    ): void;
```

packages/web3-core-helpers/types/index.d.ts
```diff
@@ -114,8 +114,8 @@ export class IpcProviderBase {
...
```
nivida:
```suggestion
    ): void;
```

packages/web3-core-helpers/types/index.d.ts
```diff
@@ -140,8 +140,8 @@ export class HttpProviderBase {
...
```
nivida:
```suggestion
    ): void;
```

nivida (changes requested):
Thanks for proposing this PR! I've added some suggestions with required changes.

nivida (approved):

cgewecke (commented):
@CruzMolina Thanks! Just have a couple questions...

Does this resolve the `tsc` errors you've been seeing in [truffle 2488][4]?

The provider bases look like they are defined in the core-helpers but extended in 
+ [web3-provider-http][1]
+ [web3-provider-ws][2]
+ [web3-provider-ipc][3]

The types there retain the format you've changed here. Will that be an issue - a more generic overloading in an inheriting type?

Those packages also contain types tests for the providers. Is it possible to write a `Callback` type example there so that future changes get caught for a closer check?

To be clear - would gladly defer to anyone with greater TS experience than I have - these are **genuinely questions** :)

[1]: https://github.com/ethereum/web3.js/blob/1.x/packages/web3-providers-http/types/index.d.ts#L29-L43
[2]: https://github.com/ethereum/web3.js/blob/1.x/packages/web3-providers-ws/types/index.d.ts#L43-L46
[3]: https://github.com/ethereum/web3.js/blob/1.x/packages/web3-providers-ipc/types/index.d.ts#L38-L41
[4]: https://github.com/trufflesuite/truffle/pull/2488

CruzMolina:
Ah, thanks for the heads up regarding the extensions @cgewecke.

Not sure if the typings added here will cause issues, but for consistency sake it makes sense to import these & implement the typings for each `Provider` class appropriately (along with adding tests).

Btw, this isn't a blocker on our end. We've been using the `@types/web3` typings and are in the process of migrating over so we can upgrade to latest.

cgewecke (approved):
@nivida Thanks for those tests, awesome. Are you fine with the overloading back to the old form in providers types files? 

Otherwise LGTM!

nivida:
@cgewecke Thanks for the hint! I've updated the tests and type definitions in those packages. 👌

cgewecke (approved):
Yeah! Make sense to me! @CruzMolina Are you happy with the added changes? Want to take second look?

CruzMolina:
I think typing the callback as I originally had it makes more sense (`Callback<JsonRpcResponse>`), but I can just type locally in one of our pkgs where we use the callback.

nivida:
@CruzMolina The reason for those changes was that ``dtslint`` tests were failing with the ``Callback<JsonRpcResponse>`` type definition and just worked because the ``send`` method got overwritten in the provider packages :)

PR #3184, opened 2019-11-06
Internal improvement: Adjust typings to allow an empty Web3 constructor

CruzMolina:


gnidan (approved):

packages/web3/types/index.d.ts
```diff
@@ -27,7 +27,7 @@ import { Shh } from 'web3-shh';
 import { Utils } from 'web3-utils';
 
 export default class Web3 {
-    constructor(provider: provider, net?: net.Socket);
+    constructor(provider?: provider, net?: net.Socket);
```
joshstevens19: Thanks for this can you please overload them, as this will make the intellisense for TS a lot nicer for people and overall make it simple to scan and read. 

It will state that there is 2 overloads if we do it the way below:

image

```
constructor(provider: provider);
constructor(provider: provider, net: net.Socket);
constructor();
```

and then write some tests for them please in `types/tests/web-test.ts`, testing each overload.

joshstevens19 (changes requested):

packages/web3/types/tests/web3-test.ts
```diff
@@ -31,7 +32,13 @@ Web3.modules;
 // $ExpectType Providers
 Web3.providers;
 
-const web3 = new Web3('https://localhost:5000/');
+let web3 = new Web3('https://localhost:5000/');
+
+const netSocket = new net.Socket();
```
joshstevens19: Can you move these new web3 stuff to the bottom so they don’t effect the main tests and then use the the expect type syntax to test it.

Like:

$ExpectType Web3
web3 = new Web3()

$ExpectType Web3
web3 = new Web3(‘https://localhost:5000/‘, netSocket)

Then looks good to merge.

joshstevens19 (changes requested):

joshstevens19 (approved):
LGTM

nivida (approved):

Issue #1646, opened 2018-05-21
issue with websocket connection to infura


cmaliwal:
I use websocket (ws) coneection to Infura node 

`web3 = new Web3(new Web3.providers.WebsocketProvider("wss://ropsten.infura.io/ws"));`

sometimes it's works fine or sometimes it give error on console : 

`Error: CONNECTION ERROR: Couldn't connect to node on WS.`

`Uncaught (in promise) Error: CONNECTION ERROR: Couldn't connect to node on WS.`

I am not getting why its happening.

nivida:
Can you determine when exactly this is happening? It's a bit hard to reproduce it this way.

minhtruong92:
I got the same issue, my code works with provider: "wss://ropsten.infura.io/ws" (testnet), but it doesn't work with "wss://mainnet.infura.io/ws". 
I use the code in introduction page https://infura.io/docs/wss/introduction

ryokomy:
I also had the same issue with "wss://rinkeby.infura.io/ws".

sanaterrae:
added some other details for same issue found here at https://github.com/INFURA/infura/issues/140

pasiphae00:
I'm having the same issue (using Kovan WS endpoint). It's extremely sporatic... About 25% of the time it works fine, and the other 75% it throws the `CONNECTION ERROR`.

I'll run my program once and it will fail, and then again with no changes and it will fail. 

Hard to see if this is an issue with Infura, or somewhere lower in the stack like the WS driver.

jamesmorgan:
I also get this almost exclusive failing with the latest version of web3 `1.0.37` - originally I thought it was an issue with Infura but then found this issue, any ideas on the solution - this is the ticket I raised with Infura https://community.infura.io/t/ropsten-connection-error-couldnt-connect-to-node-on-ws/272/5

jamesmorgan:
I have also found that adding this config fixed it - taken direct from https://github.com/ethereum/web3.js/issues/1217
```
{
  clientConfig: {
    maxReceivedFrameSize: 100000000,
    maxReceivedMessageSize: 100000000,
  }
}
```

nivida:
@jamesmorgan I've refactored the web3 providers and especially the WebsocketProvider (PR #2000). I will release this PR within the next days and probably this issue got fixed with it. I can't see anything that Web3.js is doing which could have an impact on establishing a connection.

holgerd77:
[...]

PR #3190, opened 2019-11-08
Provider related improvements

### LINKED ISSUE #1648: MaxListenersExceededWarning when sending contract (v1.0.0-beta.33)

nivida:
## Description

Fixes #1648, #3042, #3092, #1085, #1391, #1558, #1852, #1646, #1510 and #1933 

Because PR #3187, #3156, #3171, and #3167 do all depend strongly on each other and it's not possible to finish them separately. Did we move all of those PRs together and do improve and extend the provider handling of web3.js here in one bigger PR.  

### RequestManager 

The provider will now only get resolved once and the ``data`` listener is no longer added on each initiation of a ``Module``, ``Contract``, or on executing of the ``setProvider`` method on each package.

#### Old Behaviour

1. call ``setProvider``.
2. create RequestManager and register ``data`` listener.
3. call ``setProvider`` on all children's.
4. create RequestManager and register ``data`` listener for each child package.

#### New Behaviour

1. call ``setProvider``.
2. create RequestManager and register ``data`` listener.
3. call ``setRequestManager`` on all children's.

By side of improving the inheritance logic of the ``RequestManager`` did we add the ``error`` and ``connect`` event listener to pass errors to ``subscriptions`` and to call ``resubscribe`` on each ``subscription`` if the currently used provider has re-established his connection.

#### connection not open error:
Instead of just waiting 10ms with a setTimeout and to try it again does it now add the item to the ``requestQueue`` and will get executed as soon as the ``connect`` event does get emitted. If the connection does emit an ``error`` or ``close`` event does it terminate the request correctly as well.

#### Reconnecting:

_Options:_
```
{
  reconnect: {
    auto: false,
    delay: 5000,
    maxAttempts: false,
    onTimeout: false
  }
}
```

 Websocket reconnecting is a long-standing issue in the 1.x branch of this library and is the source of many issues when developing applications that for any reason use a not-so-reliable connection to the nodes.

Implementing of the reconnect handling means basically resolving several major problems:
- Reconnecting the transport layer, this is the websocket connection itself.
- Reattaching client's event listeners that handle the different websocket connection states.
- Resubscribing to higher-level events (subscriptions)

Those fixes do change the following:
- The WebsocketProvider inherits now from the EventEmitter object.
- `RequestManager`'s provider listeners updated:
  + Adds 'error' listener (does forward the error to all subs)
  + Adds 'connect' listener (resubscribes if subscriptions do exists)
  + Adds 'close' listener (does forward the close event as an error to all subs)
- Request queue added to WebsocketProvider
- WebsocketProvider error handling improved. 
- WebsocketProvider socket listener handling improved 
- WebsocketProvider ``reconnect`` method implemented

## Type of change

## Checklist:


nivida:
Open discussion from PR #3171:
Bildschirmfoto 2019-11-08 um 11 04 53

nivida:
@cgewecke 
> Open discussion from PR #3171:

I was wrong it is possible to listen to the ``ECONNREFUSED`` error there is just no ``code`` property for actually detecting this error in the error listener of the ``WebSocket`` object. But luckily does the ``WebSocket`` shim we do use for node emit the ``close`` event with the correct ``code`` property and message [here](https://github.com/theturtle32/WebSocket-Node/blob/master/lib/W3CWebSocket.js#L219).

This means we can detect the ``ECONNREFUSED`` in the ``close`` event listener and reject the already added requests correctly. But sadly, we can't suppress the error message in the console of the browser.

This also means that this line [here](https://github.com/autonomoussoftware/websocket-reconnector/blob/master/src/index.js#L122) in the ``websocket-reconnector`` package of @gabmontes doesn't make sense.

nivida:
I've added the ``on hold`` label because there is currently no way to close the ganache server programmatically in a clean way. (close code 1000; I'm in contact with @davidmurdoch from ganache)

wbt:
Web3 closing the Ganache server, cleanly or otherwise, would be unexpected behavior.

nivida:
@wbt This is handled correctly after this PR got merged. The issue is that I would like to call ``server.close`` in the test-case and ganache should close without emitting a close event with code ``1006``. Otherwise is it not possible to actually cleanly close my test case.

frozeman:
Just to add this here again for clarification. 
The provider should be set able globally when instating a global web3 object. But each module should be able to set it own provider. 

There are certain cases, like whisper and swarm, that is completely different providers than all other modules, those need to be handled differently on a global level.

packages/web3-eth-ens/src/contracts/Registry.js
```diff
@@ -39,10 +38,7 @@ function Registry(ens) {
     var self = this;
     this.ens = ens;
     this.contract = ens.checkNetwork().then(function (address) {
-        var contract = new Contract(REGISTRY_ABI, address);
-        contract.setProvider(self.ens.eth.currentProvider);
-
-        return contract;
+        return new self.ens.eth.Contract(REGISTRY_ABI, address);
```
nivida: @cgewecke Edit: I was able to remove the additional ``setProvider`` call because we do use now the ``web3-eth`` ``Contract`` constructor function to have the provider inheritance correctly working as described by Fabian. This is the simplest way to have the provider updated in the ``web3-eth-ens`` module on a ``web3.eth.setProvider(...)`` call. I would argue that this is within the scope of this PR because the inheritance of the provider is clearly an important thing here.

packages/web3-eth/src/index.js
```diff
@@ -68,14 +68,26 @@ var Eth = function Eth() {
...
     // overwrite setProvider
     var setProvider = this.setProvider;
     this.setProvider = function () {
         setProvider.apply(_this, arguments);
-        _this.net.setProvider.apply(_this, arguments);
-        _this.personal.setProvider.apply(_this, arguments);
-        _this.accounts.setProvider.apply(_this, arguments);
-        _this.Contract.setProvider(_this.currentProvider, _this.accounts);
+
+        _this.setRequestManager(_this._requestManager);
```
nivida, line 90: Does initiate a ``RequestManager`` object once with the given provider and does define the initiated manager downwards the module tree.

nivida:
@cgewecke I've kept the test file from you but it is probably worth to think about separating the reconnect test cases from the bare subscription tests. Do you agree to create a ``e2e.reconnect`` test file which will run against geth as well?

cgewecke (commented):
E2E tests look good and on the whole it almost seems like this PR reduces complexity a bit as well. It's nice. 

Also, PR description is fantastic.

 I think this is a good chance to follow the PR Guidelines proposed in #3224 as closely as possible, sorry. :) 

By category:

**Everything needs a test**
+ Web3 does not have existing E2E tests for Shh or ENS so we're missing a strong description of how those interact with a client. It's difficult to make changes like this safely without them. I will work on this this week in separate PRs.
+ All E2E gaps in the existing subscriptions tests should be filled before changing the subscription logic here, in a separate PR targeting 1.x

> Do you agree to create a e2e.reconnect test file which will run against geth as well?

Yes, definitely, that's a good idea esp if it's not possible to do this with ganache. Even if that test is weird, needs to be separated from everything else and takes a whole minute to execute it's fine. 

+ Test for the `close` event?
+ There for what happens when there are multiple subscriptions? 

**Keep the changeset to the bare minimum required**

+ Are any of the Types changes required? I see `ResponseQueue` but if that set of changes could be minimized it would be great because it helps to understand how the interface might be changing.
+ test/contract.js diff could be clean?
+ ~~packages/web3-providers-ws/src/helpers.js does not look like it's used.~~ Wrong, sorry just found it.
+ packages/web3-core-requestmanager/src/index.js has many "nice-to-haves" like var renamings ~~and`map` for `{}`~~ - think the changeset there could be reduced a lot.

Will add other notes in the diff rn... 

A big question I have is about interface changes to the sub-packages and whether these are breaking. @nivida Do you have a list of things you think might affect people?

test/setProvider.js
```diff
@@ -22,7 +22,6 @@ var Eth = require('../packages/web3-eth');
 var Shh = require('../packages/web3-shh');
 var Personal = require('../packages/web3-eth-personal');
 var Accounts = require('../packages/web3-eth-accounts');
-var Contract = require('../packages/web3-eth-contract');
```
cgewecke: This file can be removed from the changeset?

test/helpers/test.utils.js
```diff
@@ -57,6 +57,6 @@ module.exports = {
     mine: mine,
     extractReceipt: extractReceipt,
     getWeb3: getWeb3,
-    getWebsocketPort: getWebsocketPort,
+    getWebsocketPort: getWebsocketPort
```
cgewecke: Remove file from changeset

test/contract.js
```diff
@@ -9,306 +9,292 @@ var StandAloneContract = require('../packages/web3-eth-contract');
 
 
 var abi = [{
-    "type": "constructor",
```
cgewecke: This file needs a reset...

packages/web3-core-method/src/index.js
```diff
@@ -329,7 +329,6 @@ Method.prototype._confirmTransaction = function(defer, result, payload) {
                 })
                 // CHECK for CONTRACT DEPLOYMENT
                 .then(function(receipt) {
-
```
cgewecke: Remove file from changeset

nivida:
> Web3 does not have existing E2E tests for Shh or ENS so we're missing a strong description of how those interact with a client. It's difficult to make changes like this safely without them. I will work on this this week in separate PRs.

I think all other modules do use the ``core`` module the same way as for example the ``Shh`` and the ``ENS`` module does use the ``Contract`` object. Because of this do we have already enough proof for this PR. 
But definitely do we have to improve our test suite and to extend it with ``e2e`` tests for the ``ENS`` and ``Shh`` module. 

> All E2E gaps in the existing subscriptions tests should be filled before changing the subscription logic here, in a separate PR targeting 1.x

👍 

> Test for the close event?

I think we can add those tests now in this PR for the WS provider and we should improve the QA for all providers after. The IPC or HTTP provider does definitely also require some additional test cases.

> There for what happens when there are multiple subscriptions?

There shouldn't be a difference but this is definitely a missing test case thanks!

> Are any of the Types changes required?

Yes, moving of those types was required because the ws package does no longer have any dependency on any web3 package. This means I was "forced" to move those provider types to the related provider package otherwise would the ws package not have any type definitions. By side of moving them did I also update the WebsocketProvider class type definitions.

> packages/web3-core-requestmanager/src/index.js has many "nice-to-haves" like var renamings andmap for {} - think the changeset there could be reduced a lot.

I think there aren't many added or changed "nice-to-haves". I have just used the ``Map`` object for what it is thought to be used. :)

> Do you have a list of things you think might affect people?

I was only changing the ``RequestManager`` and ``WebsocketProvider`` object and pushed those changes upstream or downstream in the project structure. The only point a consumer is in contact with is the ``WebsocketProvider`` and the interface for it did only slightly change. 
The internal ``addDefaultEvents ``  method does no longer exist and is now called ``_addSocketListeners``.

test/contract.js
cgewecke: Did the fix not stick? Is it being done automatically by your text editor?

cgewecke (commented):
> But definitely do we have to improve our test suite and to extend it with e2e tests for the ENS and Shh module.

Ok cool - definitely separate PR 👍 



> I think we can add those tests now in this PR for the WS provider and we should improve the QA for all providers after. The IPC or HTTP provider does definitely also require some additional test cases.

👍 



> Yes, moving of those types was required because the ws package does no longer have any dependency on any web3 package. 

👍 

> The only point a consumer is in contact with is the WebsocketProvider and the interface for it did only slightly change. The internal addDefaultEvents method does no longer exist and is now called _addSocketListeners.

And also `responseCallbacks` and `notificationCallbacks` are now `responseQueue` and `requestQueue`... searched ganache-core on GH for all those and it doesn't seem like they refer to any of them. Hopefully all ok...

Anyway this is LGTM - I guess just filling in the tests a bit as discussed and the diff on test/contract.js? Almost there!

nivida:
> And also responseCallbacks and notificationCallbacks are now responseQueue and requestQueue...  

I will check some other big projects as well. (Embark, Aragon, Gnosis, 0x, etc.) 👍 

> Anyway this is LGTM - I guess just filling in the tests a bit as discussed and the diff on test/contract.js? Almost there!

Yep 👍

cgewecke:
I found one thing that might be a problem with the types changes. People are already importing the types directly from the web3-core-helpers package... 

https://github.com/trufflesuite/truffle/commit/7b02eef98305237504d12e8d1b1931fb25cb3021

Wdyt about reverting the types file re-org and doing it in a separate PR if necessary after this is complete? It would make the change-set here smaller and more narrowly focussed on the provider / subscript changes.

cgewecke:
Another note about the types is that people might wrap the providers (although I don't know of a WS wrapper for Web3, so everything's probably fine in this PR). 

A current example is at [truffle 2657][1] where their HDWalletProvider is missing some of the required Web3 interface and can't use 1.2.4.

[1]: https://github.com/trufflesuite/truffle/pull/2657

nivida:
> Wdyt about reverting the types file re-org and doing it in a separate PR if necessary after this is complete? It would make the change-set here smaller and more narrowly focussed on the provider / subscript changes.

This isn't possible without duplication of type definitions because the ``web3-providers-ws`` package does no longer require the ``web3-core-helpers`` package where those types were located. We could keep the types in ``web3-core-helpers`` to provide backward compatibility but we would have to duplicate the provider types at least for the ``WebsocketProvider``. It’s questionable why those devs do use types which are subfixed with ``Base`` and used in the provider packages. I thought it’s clearly visible that those are for internal usage. It is probably better to subfix them with ``Internal`` the next time.

cgewecke:
> This isn't possible without duplication of type definitions because the web3-providers-ws package does no longer require the web3-core-helpers package where those types were located.

Could you keep web3-core-helpers as a dep to get the types? I think that simplifies but it's up to you.

packages/web3-eth-contract/src/index.js
```diff
@@ -56,22 +56,24 @@ var Contract = function Contract(jsonInterface, address, options) {
...
     }
 
-    // sets _requestmanager
-    core.packageInit(this, [this.constructor.currentProvider]);
+    this.setProvider = function () {
+        core.packageInit(_this, arguments);
 
-    this.clearSubscriptions = this._requestManager.clearSubscriptions;
+        _this.clearSubscriptions = _this._requestManager.clearSubscriptions;
+    };
```
cgewecke, line 66: @nivida Could you eli5 me what this setProvider definition means and how it's different than Contract.setProvider? 

In Contract.setProvider it's possible to set accounts for example...

https://github.com/ethereum/web3.js/blob/a46dbcd7bfa6a0bb4c69e0ededcdb0646321c741/packages/web3-eth-contract/src/index.js#L241-L246

nivida: This is the static ``setProvider`` method of the ``Contract`` object which will be [used from the ``web3-eth``](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-eth/src/index.js#L304) module to pre-set the ``provider`` and the ``web3-eth-accounts`` for the initiation of the class later. This ``setProvider`` method is also documented as usage example in the [readme of the ``web3-eth-contract``](https://github.com/ethereum/web3.js/tree/1.x/packages/web3-eth-contract#usage) package.

nivida:
@cgewecke Btw.: During working on the 2.0 MVP and forwarding of the ``WebsocketProvider`` improvements did I found a bug I've fixed here with this [commit](https://github.com/ethereum/web3.js/pull/3190/commits/a46dbcd7bfa6a0bb4c69e0ededcdb0646321c741). I think we should also extend our test-cases to check if those queues are actually cleaned up and for example the [``reset``](https://github.com/ethereum/web3.js/blob/provider-related-improvements/packages/web3-providers-ws/src/index.js#L381) method should flush those queues or we have potential smaller memory leaks.

nivida:
Btw.:
> I will check some other big projects as well. (Embark, Aragon, Gnosis, 0x, etc.) 👍

I've checked the Embark, Aragon, Gnosis, and 0x projects and they do not use the ``responseCallbacks`` and ``notificationCallbacks`` properties of the ``WebsocketProvider``.

nivida:
@cgewecke 
The request and response queue does now get cleaned up if a consumer of the WS provider is calling the reset method of it. In other cases (error, close) does it already clear the Map if required and does throw the expected error. Because reset isn't called by us which means it would only get executed if a consumer does call it is it in my view of point not required to populate an error for it. 
I've additionally moved the provider types back to the ``web3-core-helpers`` package and added missing types in the tests of the provider type definitions.

Edit:
> not required to populate an error for it. 

This implies that the consumer does know that there aren't any open requests which I think he does when he is calling reset. I'm anyways looking for the exact use-case of it. 🤔

nivida:
@cgewecke I've extended the test suite related to the ``WebsocketProvider`` and his reconnect handling. The newly added ``websocket.reconnect.ganache`` test does as the name is saying test it with the dev node ``ganache``. By side of extending the test suite with additional cases did I removed the ``onError`` listener in the ``WebsocketProvider``. I was able to remove this listener because there is always a ``CloseEvent`` after an ``Error`` got emitted (I was reading the [WebSocket specification](https://html.spec.whatwg.org/multipage/web-sockets.html#network) and some explanatory blog posts about).

---

I've tested successfully the following cases manually with Geth (nodejs & browser):
1. A subscription is running and the node is restarting 
2. A subscription does get started but the provider has lost the connection.
3. An error will get thrown and correctly populated if the ``maxAttempts`` is reached during reconnecting (tested with subscriptions and normal requests).
4. If auto-reconnect is activated does the provider connect automatically as soon as the node is running (for example if you start your script but forget to start the node and you doing it after the script got executed/is running)
5. If a request got send but the response not received until now and the provider is losing his connection does it populate the correct error message. 
6. If the provider is connecting or re-connecting does it store the requests from ``provider.send`` in the ``requestQueue`` and will get executed as soon as the connection is established. (This test case does only slightly defer to case 4. The difference is that provider was already connected once successfully)


Edit: 
Because all ``WebSocket`` errors are followed by an ``CloseEvent`` do we have to discuss in which cases we would like to have the ``error`` event emitted and how subscriptions or normal requests should thread those errors. I will check now in which cases the old ``WebsocketProvider`` is emitting the ``error`` event.

cgewecke (commented):
@nivida Awesome reconnect tests, really nice.

Could you discuss the reasons for removing the error handler more? Do you just feel it's duplicate or is it causing a problem? One thing that worries me is that it emits the error again and getting rid of it might be more risky than necessary. 

Have added a few more tests in #3209 . 

Question about how auto-reconnect works. If my connection is dropped and I continue to make requests - is it expected that they will queue and be dispatched when the connection is re-established?

At the moment requests made while entering a reconnection cycle trigger the `PendingRequestsOnReconnecting` error - is it expected that the user will catch and re-try?

nivida:
> @nivida Awesome reconnect tests, really nice.

Thanks :)

> Question about how auto-reconnect works. If my connection is dropped and I continue to make requests - is it expected that they will queue and be dispatched when the connection is re-established?

Yes, they will get pushed into the ``requestQueue`` until the connection got established again and the request do get executed.

>  Could you discuss the reasons for removing the error handler more? Do you just feel it's duplicate or is it causing a problem?

The problem we have that on every ``ErrorEvent`` a ``CloseEvent`` is following which means the ``error`` listener of for example a subscription does get triggered twice. Once with the ``ErrorEvent`` and secondly with the ``CloseEvent`` which does have the exact ``reason`` with his ``code`` as property defined. 

> At the moment requests made while entering a reconnection cycle trigger the PendingRequestsOnReconnecting error - is it expected that the user will catch and re-try?

The naming of this error is probably a bit misleading (``PendingResponseOnReconnecting`` would be more explicit). This error does get thrown if a ``request`` is executed but the ``response`` not received and the provider starts to reconnect ([see line 391](https://github.com/ethereum/web3.js/blob/provider-related-improvements/packages/web3-providers-ws/src/index.js#L391)). If the developer is starting new ``requests`` and the provider is in the reconnecting state will it add those to the ``requestQueue`` on line [312](https://github.com/ethereum/web3.js/blob/provider-related-improvements/packages/web3-providers-ws/src/index.js#L312).

nivida:
Addition:

> The problem we have that on every ErrorEvent a CloseEvent is following which means the error listener of for example a subscription does get triggered twice.

The old WS provider did have an ``error`` listener which triggered the same error as the ``close`` listener and because it was cleaning up the response callback array in the provider on ``error``wasn't it triggering the error a second time when the ``close`` listener got triggered. Another difference is that the ``RequestManager`` wasn't listening to provider errors before. One of these provider-specific and not request specific errors is the ``MaxAttemptsReachedOnReconnectingError`` on line [412](https://github.com/ethereum/web3.js/blob/provider-related-improvements/packages/web3-providers-ws/src/index.js#L412).

cgewecke:
> The naming of this error is probably a bit misleading (PendingResponseOnReconnecting would be more explicit). This error does get thrown if a request is executed but the response not received and the provider starts to reconnect (see line 391). If the developer is starting new requests and the provider is in the reconnecting state will it add those to the requestQueue on line 312.

Ah ok! Thanks for clarifying. I added a test in #3302 for the basic case of a request queued while the server down and it works - great. 

> the error listener of for example a subscription does get triggered twice

Ok, I see. The problem is that it looks like the existing implementation just mirrors underlying event websocket interface [here](https://github.com/ethereum/web3.js/pull/3190/files#diff-215dff4ef1684c2278bb08b065ba5037L297-L317), with the exception that 'end' is translated as close. 

If it's the current public API, removing a piece of it is a big deal... this is a tricky issue though - will need to think about it a bit more.

packages/web3-core-requestmanager/src/index.js
```diff
@@ -20,7 +20,7 @@
  * @date 2017
  */
 
-"use strict";
+'use strict';
```
cgewecke:
```suggestion
"use strict";
```

Unlinting...

packages/web3-core-requestmanager/src/index.js
```diff
@@ -56,81 +58,102 @@ RequestManager.providers = {
...
-            throw new Error('Can\'t autodetect provider for "'+ p +'"');
+        } else if (provider) {
+            throw new Error('Can\'t autodetect provider for "' + provider + '"');
         }
     }
 
+
     // reset the old one before changing, if still connected
-    if(this.provider && this.provider.connected)
+    if (this.provider && this.provider.connected)
```
cgewecke:
```suggestion
    if(this.provider && this.provider.connected)
```

Unlinting...

packages/web3-core-requestmanager/src/index.js
```diff
@@ -56,81 +58,102 @@ RequestManager.providers = {
...
+    };
 
     if (!this.provider) {
         return callback(errors.InvalidProvider());
     }
 
     var payload = Jsonrpc.toPayload(data.method, data.params);
     this.provider[this.provider.sendAsync ? 'sendAsync' : 'send'](payload, function (err, result) {
-        if(result && result.id && payload.id !== result.id) return callback(new Error('Wrong response id "'+ result.id +'" (expected: "'+ payload.id +'") in '+ JSON.stringify(payload)));
+        if (result && result.id && payload.id !== result.id) return callback(new Error('Wrong response id "' + result.id + '" (expected: "' + payload.id + '") in ' + JSON.stringify(payload)));
```
cgewecke:
```suggestion
        if(result && result.id && payload.id !== result.id) return callback(new Error('Wrong response id "'+ result.id +'" (expected: "'+ payload.id +'") in '+ JSON.stringify(payload)));
```

Unlinting...

packages/web3-core-requestmanager/src/index.js
```diff
@@ -179,21 +204,21 @@ RequestManager.prototype.sendBatch = function (data, callback) {
...
+        this.subscriptions.set(
+            subscription.id,
+            {
+                callback: callback,
+                subscription: subscription
+            }
+        );
     } else {
-        throw new Error('The provider doesn\'t support subscriptions: '+ this.provider.constructor.name);
+        throw new Error('The provider doesn\'t support subscriptions: ' + this.provider.constructor.name);
```
cgewecke:
```suggestion
        throw new Error('The provider doesn\'t support subscriptions: '+ this.provider.constructor.name);
```

Unlinting...

packages/web3-core-requestmanager/src/index.js
```diff
@@ -227,18 +259,16 @@ RequestManager.prototype.removeSubscription = function (id, callback) {
...
+        this.subscriptions.forEach(function (value, id) {
+            if (!keepIsSyncing || value.name !== 'syncing')
                 _this.removeSubscription(id);
         });
     }
 
-
     //  reset notification callbacks etc.
-    if(this.provider.reset)
+    if (this.provider.reset)
```
cgewecke:
```suggestion
    if(this.provider.reset)
```

Unlinting....

packages/web3-core/src/index.js
```diff
@@ -20,65 +20,66 @@
  * @date 2017
  */
 
-"use strict";
+'use strict';
```
cgewecke:
```suggestion
"use strict";
```

Unlinting...

packages/web3-core/src/index.js
```diff
@@ -20,65 +20,66 @@
...
 
 
 var requestManager = require('web3-core-requestmanager');
 var extend = require('./extend.js');
 
 module.exports = {
-    packageInit: function (pkg, args) {
+    packageInit: function(pkg, args) {
```
cgewecke:
```suggestion
    packageInit: function (pkg, args) {
```

Unlinting...

packages/web3-core/src/index.js
```diff
@@ -20,65 +20,66 @@
...
 
         if (!pkg) {
             throw new Error('You need to instantiate using the "new" keyword.');
         }
 
-
         // make property of pkg._provider, which can properly set providers
         Object.defineProperty(pkg, 'currentProvider', {
-            get: function () {
+            get: function() {
```
cgewecke:
```suggestion
            get: function () {
```

Unlinting...

packages/web3-core/src/index.js
```diff
@@ -20,65 +20,66 @@
...
                 return pkg._provider;
             },
-            set: function (value) {
+            set: function(value) {
```
cgewecke:
```suggestion
            set: function (value) {
```

Unlinting...

packages/web3-core/src/index.js
```diff
@@ -20,65 +20,66 @@
...
         pkg.givenProvider = requestManager.Manager.givenProvider;
         pkg.providers = requestManager.Manager.providers;
 
-         pkg._provider =  pkg._requestManager.provider;
+        pkg._provider = pkg._requestManager.provider;
 
         // add SETPROVIDER function (don't overwrite if already existing)
         if (!pkg.setProvider) {
-            pkg.setProvider = function (provider, net) {
+            pkg.setProvider = function(provider, net) {
```
cgewecke:
```suggestion
            pkg.setProvider = function (provider, net) {
```

Unlinting...

packages/web3-core/src/index.js
```diff
@@ -20,65 +20,66 @@
...
+        };
+
         // attach batch request creation
         pkg.BatchRequest = requestManager.BatchManager.bind(null, pkg._requestManager);
 
         // attach extend function
         pkg.extend = extend(pkg);
     },
-    addProviders: function (pkg) {
+    addProviders: function(pkg) {
```
cgewecke:
```suggestion
    addProviders: function (pkg) {
```

Unlinting...

packages/web3-eth-ens/src/contracts/Registry.js
cgewecke: Ok! Fair enough.

cgewecke:
@nivida. Basically at the end here. All that's left is 

+ #3329 
+ #3347 
+ Left a few de-linting suggestions above. 

The optional npm publish tests are failing because all the packages here are one patch version behind latest now. That's ok - they can be double-checked after merge / before the next patch is published.

nivida:
@cgewecke 
> Left a few de-linting suggestions above.

I've only applied those which do increase the code style consistency and kept the other improvements as they are.

> The optional npm publish tests are failing because all the packages here are one patch version behind latest now. That's ok - they can be double-checked after merge / before the next patch is published.

I will merge 1.x into this branch and we can double-check it here. The other way around just feels wrong.

test/eth.subscribe.ganache.js
```diff
@@ -0,0 +1,301 @@
...
+    it('errors when the provider does not support subscriptions (callback)', function (done) {
+        web3 = new Web3('http://localhost:' + port);
+
+        web3.eth.subscribe('newBlockHeaders', function (err, result) {
+            assert(err.message.includes("provider doesn't support subscriptions: HttpProvider"));
+            done();
+        });
+    });
+
+    it('errors when the provider is not set (.on("error"))', function (done) {
```
cgewecke:
```suggestion
    it('errors when the provider does not support subscriptions (.on("error"))', function (done) {
```

ryanio:
Wow a lot of really awesome work here! I've personally experienced issues with websocket connections so it's really nice to see some queue/auto-reconnect logic :)

I'm seeing a handful of merge conflicts, should this PR be rebased from master for a final review?

holgerd77:
@nivida Can you still do a rebase here, or should @ryanio take this over for a final rebase?

nivida:
@holgerd77 
> Can you still do a rebase here, or should @ryanio take this over for a final rebase?

I will resolve the conflicts and will double-check the "npm publish e2e" tests. 👌 (I will do it during EthCC)

svanurin:
@nivida Thanks for the great work!
@cgewecke @holgerd77  Hello! Do you have any idea how far is it from merging this PR? 
As I have issues with websockets missing events from Infura and considering rewrite my app or wait for this fixes...

holgerd77:
@svanurin Won't make any promises here. If we get this merged next week during EthCC this will get its way into the following `v1.2.7` release, I would assume to be released sometime during the next 2-4 weeks.

svanurin:
@holgerd77  I see. Thank you and good luck with your work!

cgewecke:
@holgerd77 Rebased against latest 1.x now. 

One of the "allowed failure" jobs (ganache-core's unit tests) is failing on a single test. This has already been reported/discussed in [ganache-core 545][1] and they agree it's a bug in their test / websocket disconnection logic.

[1]: https://github.com/trufflesuite/ganache-core/pull/545

holgerd77 (approved):
Skimmed through the code a last time for a rough consistency check, LGTM, will merge.

wbt:
Thanks for all the work that went into this!  This has been a blocker to upgrading for me, blocking being able to report and/or fix issues in the latest.
Is there any general time-frame for when things might be tested and formally released?

cgewecke:
Hi @wbt. We're working on releasing `1.2.7-rc.0`  today (which will include this PR). 

And `1.2.7` should release next week if we don't discover any bugs while testing the `rc`.

dr3adx:
WHEN IS 1.2.7 GONNA BE RELEASED JESUS CHRIST THIS IS A MAJOR BUG

cgewecke:
@dr3adx This work is published in 1.2.7-rc.0 and we should cut a formal release for that this week. 

If you'd like to try it out right now you can install with
```
npm install --save web3@rc
```

frozeman:
@dr3adx what issues do you see?

frozeman:
@cgewecke do I understand this PR right and it removes the ability to give each module a different provider connection?
If so than this is a bug, as I had built it on purpose that way to show connection to different nodes / chains at the same time with the same web3 instance.

cgewecke:
@frozeman, thanks for checking this over...

> it removes the ability to give each module a different provider connection?

This topic was raised back in November when the change was first proposed in [this thread][1]. There it seemed like the main concern was that the `swarm` module preserved its own provider.

If you have a chance could you clarify what the expected behavior is with an example? 

[1]: https://github.com/ethereum/web3.js/pull/3156#issuecomment-550005464

hfa0:
@nivida seems like the providers `error` and `end` events are not firing anymore. Instead, the contracts `error` event does. Is this intended behaivoir ?
```javascript
provider = new Web3.providers.WebsocketProvider(url);
provider.on('end', () => {
  // not fired
});
provider.on('error', () => {
  // not fired
});
contract.events[type]({ fromBlock }).on('error', (err) => {
  // event fired
})
provider.disconnect(1012, 'restart');
```

cgewecke:
@hfa0 Apologies, this is a bug. Have opened this as a separate issue in #3485 and will open a PR patching rn.

PR #3192, opened 2019-11-09
New module structure (CJS -> ESM)

nivida:
## Description

Based on #3160 and does fix the line shift of the coverage report and decreases the minified UMD bundle size by ~40Kb.

## Type of change

- [x] Enhancement

## Checklist:


cgewecke (approved):
Wow! Coverage report looks good / seems fixed.

Have checked for mis-namings etc, everything LGTM. Think when merged to #3160 it should be double-checked to see where changes occurred that are not hit by tests because there's always a  possibility of an error creeping in when everything is renamed.

~~Not relevant for this PR but still needs to be addressed #3160 - the 'test & coverage' build should be isolated from any install or release build. On post-install, coverage mangle is being introduced to the .cjs by default.~~  [EDIT - apologies I see the ISTANBUL flag now, I didn't follow that correctly. I guess release is the really important part of this anyway so just ignore this].




Looks like this:
```js
var cov_1io38x4fw5 = function () {
  var path = "/Users/cgewecke/code/ef/origin/new-module-structure/web3.js/packages/web3/src/index.js";
```

cgewecke:
Another question:

A normal flow when working is to go back and forth between the unit tests and the code being changed, making fixes, using `.only` on a test suite etc. So it's quick.

`npm test` looks like it's missing a build step which is necessary to have changes persist when working this way. Adding `npm run build:all:cjs` increases the time to test start by ~20s. 

Do you have any experience using rollup's "watch" mode or know if this is less of an issue when using it?

nivida:
> Have checked for mis-namings etc, everything LGTM. Think when merged to #3160 it should be double-checked to see where changes occurred that are not hit by tests because there's always a possibility of an error creeping in when everything is renamed.

 👍
 
> npm test looks like it's missing a build step which is necessary to have changes persist when working this way. Adding npm run build:all:cjs increases the time to test start by ~20s.

We can speed up the mentioned development process with adding of the build commands to all web3 packages to be able to ``watch`` or manually ``build`` a single package. 

> Do you have any experience using rollup's "watch" mode or know if this is less of an issue when using it?

I've used the watch command in the 2.x branch and didn't have any problems with it:)

nivida:
> We can speed up the mentioned development process with adding of the build commands to all web3 packages to be able to watch or manually build a single package.

I will do the mentioned improvements in PR #3160.

PR #3195, opened 2019-11-12
Documentation and types improved for web3-eth and web3-eth-contract

### LINKED ISSUE #3194: Lots of typings are incorrect or documented incorrectly in 1.2.2
### Issue

Lots of typings are incorrect in 1.2.2.  In some cases this is because the typing matches the documentation rather than the implementation, but in these cases both the typing and the documentation should be updated (especially because in at least one case the documented typing is completely unworkable).

A number of examples follow.  This list is in no way intended to be exhaustive; it's simply the problems that I personally have run into.  But given the existence of these problems, I think you may want to give all the typings a thorough review.

#### getStorageAt

Let's start with the worst case.  The second argument to `getStorageAt` specifies the address of the storage slot from which to retrieve storage.  Currently it is typed as a `number`.  This is unworkable, because storage slot addresses are 256 bits long -- and yes, large ones, outside the range of a Javascript safe integer, are in fact commonly used.

Fortunately, the function does not *actually* only take numbers for that argument; numeric strings also work, as do `BN`s.  (Possibly `BigNumber` as well; I didn't check.)  But, currently, in order to use this function, one must perform a nonsensical coercion by means of `unknown`.

So, the type of that argument should be `number | string | BN` (or possibly something even more permissive depending on what the function actually takes), and the documentation should be updated to match.

#### Log.topics

The type of the `topics` field on the `Log` type is `(string | string[])[]`.  This is incorrect, in that it is overly broad, which on a return type is a problem.  A log returned by `getPastLogs` has, as its topic field, a simple array of strings (these strings being the topics).  That is, it should have type `string[]`.  Instead it has been typed as `(string | string[])[]`.

It looks to me like there was a mixup here between the `topics` field on the `Log` type, and the `topics` field on the `PastLogsOptions` type.  The latter should indeed be `(string | string[])[]`, as you can pass in either an individual address or a list of addresses.  However, there is no reason that these two `topics` fields on these two different types should have the same type.  They work differently, and their individual types should reflect that.

#### More on PastLogsOptions

PastLogsOptions's `topics` field actually shouldn't be `(string | string[])[]` (or rather, `Array<string | string[]>`... I'm not sure what the difference here is between `T[]` and `Array<T>`, so I'm going to ignore the distinction if you don't mind), but rather `(string | string[] | null)[]`, same as with LogsOptions.

The typings on `fromBlock` and `toBlock` are both `number | string`, which *works*, but is a bit permissive.  What's actually accepted is `number | "latest" | "pending"`, so you might want to restrict to that?

#### More on block typing

Let's talk some more about ways of specifying blocks.  Currently, functions that allow one to specify a block (not counting `getPastLogs`) take that block as `number | string`.  But actually, BN's and BigNumbers also work.  The typing, and documentation, should be updated to reflect this.

In fact, for many of these functions in `web3.eth`, the documentation is even more wrong than that, because the docs don't make clear that you can use a block hash instead of a block number, even though you can, for all of them (except `getPastLogs`, and possibly `subscribe` which I haven't tried).  So, this should also be fixed.

#### currentProvider

The `web3.currentProvider` object has type `string`.  Frequently, it is not a string, but rather a complex object with methods such as `send`.  A `Provider` type should be created to match how providers actually work, and should be documented.

nivida:
## Description 

Fixes #3194

## Type of change

- [x] Types
- [x] Documentation

## Checklist:

### DIFF
--- a/docs/web3-eth.rst
+++ b/docs/web3-eth.rst
@@ -677,7 +677,7 @@ Parameters
 ----------
 
 1. ``String`` - The address to get the balance of.
-2. ``Number|String`` - (optional) If you pass this parameter it will not use the default block set with :ref:`web3.eth.defaultBlock <eth-defaultblock>`.
+2. ``Number|String|BN`` - (optional) If you pass this parameter it will not use the default block set with :ref:`web3.eth.defaultBlock <eth-defaultblock>`.
 3. ``Function`` - (optional) Optional callback, returns an error object as first parameter and the result as second.
 
 -------
@@ -718,7 +718,7 @@ Parameters
 
 1. ``String`` - The address to get the storage from.
 2. ``Number`` - The index position of the storage.
-3. ``Number|String`` - (optional) If you pass this parameter it will not use the default block set with :ref:`web3.eth.defaultBlock <eth-defaultblock>`.
+3. ``Number|String|BN`` - (optional) If you pass this parameter it will not use the default block set with :ref:`web3.eth.defaultBlock <eth-defaultblock>`.
 4. ``Function`` - (optional) Optional callback, returns an error object as first parameter and the result as second.
 
 
@@ -756,7 +756,7 @@ Parameters
 ----------
 
 1. ``String`` - The address to get the code from.
-2. ``Number|String`` - (optional) If you pass this parameter it will not use the default block set with :ref:`web3.eth.defaultBlock <eth-defaultblock>`.
+2. ``Number|String|BN`` - (optional) If you pass this parameter it will not use the default block set with :ref:`web3.eth.defaultBlock <eth-defaultblock>`.
 3. ``Function`` - (optional) Optional callback, returns an error object as first parameter and the result as second.
 
 -------
@@ -795,7 +795,7 @@ Returns a block matching the block number or block hash.
 Parameters
 ----------
 
-1. ``String|Number`` - The block number or block hash. Or the string ``"genesis"``, ``"latest"`` or ``"pending"`` as in the :ref:`default block parameter <eth-defaultblock>`.
+1. ``String|Number|BN`` - The block number or block hash. Or the string ``"genesis"``, ``"latest"`` or ``"pending"`` as in the :ref:`default block parameter <eth-defaultblock>`.
 2. ``Boolean`` - (optional, default ``false``) If specified ``true``, the returned block will contain all transactions as objects. By default it is ``false`` so, there is no need to explictly specify false. And, if ``false`` it will only contains the transaction hashes.
 3. ``Function`` - (optional) Optional callback, returns an error object as first parameter and the result as second.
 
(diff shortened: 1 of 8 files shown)


nivida:
I didn't change the type for the ``web3.currentProvider`` because it's not typed as ``string`` as mentioned in the issue. 

``` typescript
export type provider =
    | HttpProvider
    | IpcProvider
    | WebsocketProvider
    | string
    | null;
```

docs/web3-eth.rst
```diff
@@ -795,7 +795,7 @@ Returns a block matching the block number or block hash.
 Parameters
 ----------
 
-1. ``String|Number`` - The block number or block hash. Or the string ``"genesis"``, ``"latest"`` or ``"pending"`` as in the :ref:`default block parameter <eth-defaultblock>`.
+1. ``String|Number|BN`` - The block number or block hash. Or the string ``"genesis"``, ``"latest"`` or ``"pending"`` as in the :ref:`default block parameter <eth-defaultblock>`.
```
cgewecke:
```suggestion
1. ``String|Number|BN`` - The block number or block hash. Or the string ``"earliest"``, ``"latest"`` or ``"pending"`` as in the :ref:`default block parameter <eth-defaultblock>`.
```

cgewecke: @nivida There's a few of these - does that change seem correct? 

Checked the codebase and I don't see "genesis" anywhere except the docs.

[JSON-RPC defaultBlock spec](https://github.com/ethereum/wiki/wiki/JSON-RPC#the-default-block-parameter)

nivida: > @nivida There's a few of these - does that change seem correct?

Yes, this is correct. It was always possible but not documented in the developer documentation or the function docs of those ``formatters``.

> Checked the codebase and I don't see "genesis" anywhere except the docs.

``genesis`` is handled in the ``inputDefaultBlockNumberFormatter`` of the ``web3-core-helpers`` package. 

Edit:
Existing predefined block numbers are: 
``genesis``, ``earliest``, ``latest``, and ``pending``

nivida: I've updated the related documentation and added if required a sentence about the pre-defined block numbers. Additionally got the missing and compatible type ``BigNumber`` added to the ``BlockNumber`` type of the web3-core module.

packages/web3-core/types/index.d.ts
```diff
@@ -396,6 +397,18 @@ export class WebsocketProvider extends WebsocketProviderBase {
...
+    toBlock?: BlockNumber;
+}
+
+export interface LogsOptions {
+    fromBlock?: BlockNumber;
+    address?: string | string[];
+    topics?: Array<string | string[] | null>;
+}
+
+export type BlockNumber = string | number | BN | BigNumber | 'latest' | 'pending' | 'earliest';
```
joshstevens19, line 410: :+1:

joshstevens19 (approved):
awesomeeee

docs/web3-eth.rst
cgewecke: Oh interesting! Thanks for double-checking that.

PR #3199, opened 2019-11-12
Module type definitions updated

### LINKED ISSUE #3196: Web3.givenProvider documentation/typing issue
Hello!

Thank you for this library!

However, it seems like you have a typing or documentation issue.

According to the docs:

```
For web3.js, check Web3.givenProvider. If this property is null you should connect to a remote/local node.

var web3 = new Web3(Web3.givenProvider || "ws://localhost:8545");
```

However, `Web3.givenProvider` doesn't exists according to the types:

```
export default class Web3 {
  ...
  readonly givenProvider: any;
}
```

Right now, `givenProvider` is typed as an instance property, not the class property. If the documentation is correct it should be updated to use the `static` keyword, e.g.:

```
export default class Web3 {
  ...
  static readonly givenProvider: any;
}
```

Thanks!

nivida:
## Description

Fixes #3196

Fixes the types for the static ``givenProvider`` property, improves the types tests, and does correct the types of the inherited core functions for each module (``currentProvider``, ``givenProvider``, ``extend``, etc.).

## Type of change

- [x] Types

## Checklist:

### DIFF
--- a/packages/web3-core/types/index.d.ts
+++ b/packages/web3-core/types/index.d.ts
@@ -219,14 +219,14 @@ export interface Log {
 // had to move `web3-net` due to other modules in `1.x` not referencing
 
 export class NetworkBase {
+    constructor();
     constructor(provider: provider);
     constructor(provider: provider, net: net.Socket);
-    constructor();
 
     readonly givenProvider: any;
+    readonly currentProvider: provider;
     static readonly givenProvider: any;
     static readonly providers: Providers;
-    readonly currentProvider: provider;
     BatchRequest: new () => BatchRequest;
 
     setProvider(provider: provider): boolean;
--- a/packages/web3-core/types/index.d.ts
+++ b/packages/web3-core/types/index.d.ts
@@ -219,7 +219,9 @@ export interface Log {
 // had to move `web3-net` due to other modules in `1.x` not referencing
 
 export class NetworkBase {
-    constructor(provider?: provider, net?: net.Socket);
+    constructor();
+    constructor(provider: provider);
+    constructor(provider: provider, net: net.Socket);
 
     readonly givenProvider: any;
     readonly currentProvider: provider;
@@ -249,7 +251,9 @@ export class NetworkBase {
 // had to move accounts from web3-eth-accounts due to other modules in 1.x not referencing
 
 export class AccountsBase {
-    constructor(provider?: provider, net?: net.Socket);
+    constructor();
+    constructor(provider: provider);
+    constructor(provider: provider, net: net.Socket);
 
     readonly givenProvider: any;
     readonly currentProvider: provider;
--- a/packages/web3-core/types/index.d.ts
+++ b/packages/web3-core/types/index.d.ts
@@ -219,12 +219,19 @@ export interface Log {
 // had to move `web3-net` due to other modules in `1.x` not referencing
 
 export class NetworkBase {
-    constructor(provider: provider, net?: net.Socket | null);
+    constructor(provider: provider);
+    constructor(provider: provider, net: net.Socket);
+    constructor();
 
     readonly givenProvider: any;
+    static readonly givenProvider: any;
+    static readonly providers: Providers;
     readonly currentProvider: provider;
     BatchRequest: new () => BatchRequest;
-    static readonly providers: Providers;
+
+    setProvider(provider: provider): boolean;
+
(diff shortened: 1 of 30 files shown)


cgewecke (approved):
LGTM! 

In #3194 they say `web3.currentProvider object has type string` but it looks correct here already. Are they just mistaken?

nivida:
> In #3194 they say web3.currentProvider object has type string but it looks correct here already. Are they just mistaken?

I do have the same impression. I think they are just confused because a provider in the case of ``setProvider`` can be of type ``string``. I will wait on the "thumbs up" from Josh and merge it after. 👌

packages/web3-core/types/index.d.ts
```diff
@@ -219,12 +219,13 @@ export interface Log {
 // had to move `web3-net` due to other modules in `1.x` not referencing
 
 export class NetworkBase {
-    constructor(provider: provider, net?: net.Socket | null);
+    constructor(provider?: provider, net?: net.Socket);
```
joshstevens19: it would be much nicer if you did

```
constructor();
constructor(provider: provider);
constructor(provider: provider, net: net.Socket);
```

packages/web3-core/types/index.d.ts
```diff
@@ -248,17 +249,13 @@ export class NetworkBase {
 // had to move accounts from web3-eth-accounts due to other modules in 1.x not referencing
 
 export class AccountsBase {
-    constructor(provider: provider, net?: net.Socket | null);
+    constructor(provider?: provider, net?: net.Socket);
```
joshstevens19: same as the above. Much nicer for ts users to look at the overloads.

packages/web3-eth-personal/types/index.d.ts
```diff
@@ -18,10 +18,23 @@
...
+import { provider, RLPEncodedTransaction, TransactionConfig, BatchRequest, Providers, Extension } from 'web3-core';
 
 export class Personal {
-    constructor(provider: provider);
+    constructor(provider?: provider, net?: net.Socket);
+
+    readonly givenProvider: any;
+    static readonly givenProvider: any;
+    static readonly providers: Providers;
+    readonly currentProvider: provider;
```
joshstevens19, line 32: shouldnt this be static as well?

packages/web3-eth/types/index.d.ts
```diff
@@ -45,7 +46,7 @@ import { Network } from 'web3-net';
 import { AbiItem } from 'web3-utils';
 
 export class Eth {
-    constructor(currentProvider: provider);
+    constructor(provider?: provider, net?: net.Socket);
```
joshstevens19: same as what i said above.

packages/web3-shh/types/index.d.ts
```diff
@@ -18,21 +18,24 @@
...
-    constructor(provider: provider, net?: net.Socket | null);
-
-    net: Network;
+    constructor(provider?: provider, net?: net.Socket);
 
     readonly givenProvider: any;
-    readonly currentProvider: provider;
+    static readonly givenProvider: any;
     static readonly providers: Providers;
+    readonly currentProvider: provider;
```
joshstevens19, line 32: again i think we need a static one as well?

packages/web3/types/index.d.ts
```diff
@@ -27,12 +27,11 @@ import { Shh } from 'web3-shh';
 import { Utils } from 'web3-utils';
 
 export default class Web3 {
-    constructor(provider: provider);
-    constructor(provider: provider, net: net.Socket);
-    constructor();
+    constructor(provider?: provider, net?: net.Socket);
```
joshstevens19: same as the above.

packages/web3-eth-personal/types/index.d.ts
nivida: The property ``currentProvider`` is an instance property. Only the defined properties [here](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-core/src/index.js#L81) are defined as static class properties.

nivida: ... and the referenced ``addProviders`` function is not called on every package.

joshstevens19 (approved):

nivida:
Merged because the ``dtslint`` runs locally with success overall packages.

PR #3200, opened 2019-11-13
web3-eth exporting of types from web3-core

### LINKED ISSUE #3197: web3-eth should export TransactionConfig
Hello!

I'm using `sendTransaction` method from the `web3-eth` package. The first argument has a type `TransactionConfig`, however, this type is not exported from `web3-eth`. I have to additionally and explicitly install the `web3-core` package just to import `TransactionConfig` from it. This is very cumbersome and creates tight coupling between modules.

You should re-export `TransactionConfig` from `web3-eth` package or even to define it in the package itself to decouple packages more (if it makes sense).

Thank you!

nivida:
## Description

This PR will improve the DX for TS developers. 

Fixes #3197 

## Type of change

- [x] Types

## Checklist:

### DIFF
--- a/packages/web3-eth/types/tests/eth.tests.ts
+++ b/packages/web3-eth/types/tests/eth.tests.ts
@@ -18,13 +18,22 @@
  * @date 2018
  */
 
+import {Log} from 'web3-core';
 import {
-    Log,
+    BlockTransactionObject,
+    BlockTransactionString,
+    BlockHeader,
+    Eth,
+    GetProof,
+    Syncing,
     RLPEncodedTransaction,
     Transaction,
-    TransactionReceipt
-} from 'web3-core';
-import { BlockTransactionObject, BlockTransactionString, BlockHeader, Eth, GetProof, Syncing } from 'web3-eth';
+    TransactionReceipt,
+    TransactionConfig,
+    hardfork,
+    Common,
+    chain
+} from 'web3-eth';
 
 const eth = new Eth('http://localhost:8545');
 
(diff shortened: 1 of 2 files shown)


cgewecke (commented):
I'm seeing the truffle e2e CI fail on `tsc` because it can't find `Block` exported from web3-eth. 
```
$ tsc
lib/interface-adapter/types/index.d.ts:1:10 - error TS2305: Module '"../../../node_modules/web3-eth/types"' has no exported member 'Block'.
1 import { Block as EvmBlock } from "web3-eth";
           ~~~~~
Found 1 error.
```

It looks like it was removed in [this commit](https://github.com/ethereum/web3.js/commit/73f338279655a108af9952926ca28009083a66f5#diff-db8d4b5a4b9c1b6661a28c635c5b062bL361). 

Is there any way of restoring that as an alias in this PR? 

See also #3150

nivida:
> Is there any way of restoring that as an alias in this PR?

Added the ``Block`` interface with a ``TODO`` comment to remove this type on a minor release.

packages/web3-eth/types/tests/eth.tests.ts
```diff
@@ -18,13 +18,22 @@
  * @date 2018
  */
 
+import {Log} from 'web3-core';
 import {
```
joshstevens19, line 22: why are you importing them when its not being tested?

joshstevens19 (approved):

packages/web3-eth/types/tests/eth.tests.ts
nivida: I have imported them as a test. This should give us the proof those forwarded types do exist in the ``web3-eth`` package. :)

PR #3215, opened 2019-11-15
fix: Fix typings export for web3-providers-ws

daffl:
The WebsocketProvider class is set as `module.exports` in https://github.com/ethereum/web3.js/blob/1.x/packages/web3-providers-ws/src/index.js#L417 which makes it a `default` export.

Currently, if you use the typings published in v1.2.4 you will get an error like `TypeError: WebsocketProvider is not a constructor` when trying to use it.

### DIFF
--- a/packages/web3-providers-ws/types/index.d.ts
+++ b/packages/web3-providers-ws/types/index.d.ts
@@ -22,4 +22,4 @@
 
 import { WebsocketProviderBase } from 'web3-core-helpers';
 
-export class WebsocketProvider extends WebsocketProviderBase { }
+export default class WebsocketProvider extends WebsocketProviderBase { }
--- a/packages/web3-providers-ws/types/tests/web3-provider-ws-tests.ts
+++ b/packages/web3-providers-ws/types/tests/web3-provider-ws-tests.ts
@@ -21,7 +21,7 @@
  */
 
 import { WebsocketProviderOptions, JsonRpcResponse } from 'web3-core-helpers';
-import { WebsocketProvider } from 'web3-providers';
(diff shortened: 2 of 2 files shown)


alcuadrado:
Hey @fmiras, does this fix what you mentioned me some time ago?

fmiras:
> Hey @fmiras, does this fix what you mentioned me some time ago?

Partially yes, but this same thing is happening with some other imports of Web3. In fact, with this we will have some modules without default export and other bunch of them within it.

alcuadrado:
Thanks for your reply @fmiras !

@nivida @cgewecke, FYI, what is happning right now is that type declarations use named exports, and the actual js files just reassign `module.exports`.  They should use this pattern instead: https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html

Note that fixing this is a breaking change for TS users, but their current situation is kinda shitty. They are forced to do things like this unless they use `esModuleInterop`:

```ts
import * as Web3 from 'web3';
import { Eth } from 'web3-eth';

const client: Eth = new (Web3 as any)(clientHost).eth;
```

nivida:
@alcuadrado Yes, but we have the required configurations documented [here](https://github.com/ethereum/web3.js#usage-with-typescript). It will run without doing anything special if those configuration properties are defined as documented. 

But yes you are right it’s probably worth to update our exports from our type definitions if such a change doesn't break existing DApps with those documented configuration properties set.

@joshstevens19 What do you think about as creator of those type definitions?

fmiras:
IMHO web3 should work out of the box with `ts-node` (no `tsconfig.json`)

PR #3219, opened 2019-11-19
eth_requestAccounts added to web3-eth package

nivida:
## Description

This PR does add the JSON-RPC method ``eth_requestAccounts`` to the web3-eth package.

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-eth/src/index.js
+++ b/packages/web3-eth/src/index.js
@@ -483,6 +483,11 @@ var Eth = function Eth() {
             params: 0,
             outputFormatter: utils.hexToNumber
         }),
+        new Method({
+            name: 'requestAccounts',
+            call: 'eth_requestAccounts',
+            params: 0
+        }),
 
         // subscriptions
         new Subscriptions({
--- a/packages/web3-eth/src/index.js
+++ b/packages/web3-eth/src/index.js
@@ -486,7 +486,8 @@ var Eth = function Eth() {
         new Method({
             name: 'requestAccounts',
             call: 'eth_requestAccounts',
-            params: 0
+            params: 0,
+            outputFormatter: utils.toChecksumAddress
         }),
 
         // subscriptions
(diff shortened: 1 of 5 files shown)


packages/web3-eth/src/index.js
```diff
@@ -483,6 +483,11 @@ var Eth = function Eth() {
             params: 0,
             outputFormatter: utils.hexToNumber
         }),
+        new Method({
+            name: 'requestAccounts',
+            call: 'eth_requestAccounts',
+            params: 0
```
cgewecke: `eth_accounts` [output-formats to checksum address][1].  Maybe the same here? 


[1]: https://github.com/ethereum/web3.js/blob/5a08984ca2c80e30bc59e429692b0b5960274493/packages/web3-eth/src/index.js#L338

packages/web3-eth/types/index.d.ts
```diff
@@ -364,6 +364,9 @@ export class Eth {
         blockNumber: BlockNumber,
         callback?: (error: Error, result: GetProof) => void
     ): Promise<GetProof>;
+
+    requestAccounts(): Promise<string[]>
+    requestAccounts(callback: (error: Error, result: string[]) => void): Promise<string[]>
```
cgewecke, line 369: In the types above here, a different style is used where callback is optional. Do you prefer having each format listed separately?

cgewecke (commented):
Apart from my question about address formatting this LGTM! It's cool. 

Did you try it with metamask?

packages/web3-eth/src/index.js
nivida: True that 👌

packages/web3-eth/types/index.d.ts
nivida: As I'm informed is the overloaded way the preferred way because of the better hints TS devs do get. Should I update the whole file you commented on to have the consistency we would like to have?

nivida:
> Did you try it with metamask? 

Yep, if I'm right does Status, imToken, and WalletConnect support this JSON-RPC method as well.

packages/web3-eth/types/index.d.ts
cgewecke: Oh I was just curious about this - that's good to know. I don't have any problem with mixed styles here - would probably leave as you have it just because changing all the other ones is a separate topic.

cgewecke (approved):
LGTM

PR #3226, opened 2019-11-20
Implements sha3Raw and soliditySha3Raw

### LINKED ISSUE #1961: web3.utils.sha3 and web3.utils.soliditySha3 return null when an empty byte sequence is hashed
Based on the code, this appears to be done intentionally for some reason, but it seems like a very bad idea to return the wrong answer. There's also no mention of this non-standard behavior in the documentation.

From https://github.com/ethereum/web3.js/blob/1.0/packages/web3-utils/src/utils.js#L434,L435:

```
var SHA3_NULL_S = '0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470';
...
    if(returnValue === SHA3_NULL_S) {
        return null;
```

Removing those lines would fix the problem.

nivida:
## Description

Fixes #1961

This PR does fix the referenced "bug" above where you get returned ``null`` instead of the hash.

@cgewecke I'm open for better function names ;)

## Type of change

## Checklist:


packages/web3-utils/src/soliditySha3.js
```diff
@@ -241,5 +241,18 @@ var soliditySha3 = function () {
     return utils.sha3('0x'+ hexArgs.join(''));
 };
 
+/**
+ * Hashes solidity values to a sha3 hash using keccak 256 but does return the hash of value `null` instead of `null`
+ *
+ * @method soliditySha3
```
cgewecke: Just a tiny thing here...
```suggestion
 * @method soliditySha3Raw
```

cgewecke (approved):
This LGTM! Really nice. 

Only thing I can think of that might be useful to add to the tests is maybe one non-empty-string test case borrowed from the other sha tests, just so the basic functionality is covered in case someone accidentally changes logic later.

Issue #1905, opened 2018-08-28
block timestamp - "Error: Number can only safely store up to 53 bits"


fritzmatias:
Hi, this issue is related to [1](https://github.com/ethereum/web3.js/issues/1215),[2](https://github.com/ethereum/web3.js/issues/1351). And because it is a problem related to 3rd parties implementation of the blockchain, must be the responsibility of them to manage the particular implementation of the precision related to the timestamp.
As the [documentation](https://github.com/ethereum/web3.js/issues/1215) defines, the timestamp must be interpreted as [unix time](https://en.wikipedia.org/wiki/Unix_time), but currently, an error breaks the execution of the application (if the timestamp value is bigger than expected) throwing an error. 
It is expected that the block timestamp data could change over time --given the [limitations](https://en.wikipedia.org/wiki/Unix_time#Representing_the_number) of unix time spec--  but the behavior of the web3 must not break because of it. 

Finally, the problem is related to a timestamp value bigger than a 32bit representation value. So, the right solution must be managing the timestamp as  BN (in little endian) and then convert it to the first 32bit representable values by the developer.
This means that a decimal number with more than 10 digits must be truncated to be interpreted as 32bit length.

My solution at release @beta.35 file web3-core-helpers/src/formatters.js 235
> `< block.timestamp = utils.hexToNumber(block.timestamp);`
> `> block.timestamp = outputBigNumberFormatter(block.timestamp);`

at relase @beta.36 file web3-core-helpers/src/formatters.js:185 
> `<     block.timestamp = utils.toDecimal(block.timestamp);`
> `>   block.timestamp = utils.toBigNumber(block.timestamp);`

a substring(0,10) operation is not recomended here to avoid problems with big-endian implementations 

Example:
[...]

PR #3234, opened 2019-11-22
Integer overflow fixed in formatters

### LINKED ISSUE #3227: Formatter has a large number of overflow bugs
outputBlockFormatter currently has a bug where multiple parameters overflow.
For example, the theoretical condition value of block.gasLimit is 0x7fffffffffffffff. Since this exceeds 53 bits that can be handled by javascript Number, it is necessary to use BigNumber.
https://github.com/ethereum/web3.js/blob/ca9869fba7a87087079c24195ebb522b587e148f/packages/web3-core-helpers/src/formatters.js#L263
Incorrect
```
block.gasLimit = utils.hexToNumber(block.gasLimit);
```

Correct
```
block.gasLimit = outputBigNumberFormatter(block.gasLimit);
```

There are other reports of similar problems.

#1905 block.timestamp
#1960 block.number

Probably the same problem will occur with block.gasUsed.
In addition, similar potential bugs are expected for outputTransactionFormatter, outputTransactionReceiptFormatter, and outputLogFormatter.

nivida:
## Description

Fixes #3227 & #1905

### Affected Methods:

- ``getBlockNumber``: ``web3-eth``
- ``getPastLogs``: ``web3-eth``
- ``isSyncing``: ``web3-eth``
- ``sendTransaction``: ``web3-eth``
- ``getBlock``: ``web3-eth``
- ``getUncle``: ``web3-eth``
- ``getTransaction``: ``web3-eth``
- ``getTransactionFromBlock``: ``web3-eth``
- ``getTransactionReceipt``: ``web3-eth``
- ``estimateGas``: ``web3-eth``
- ``getPastLogs``: ``web3-eth``
- ``getTransactionCount``: ``web3-eth``
- ``subscribe('newBlockHeaders')``: ``web3-eth``
- ``subscribe('logs')``: ``web3-eth``
- ``subscribe('syncing')``: ``web3-eth``
- [``contract-events``](https://web3js.readthedocs.io/en/v1.2.4/web3-eth-contract.html#contract-events): ``web3-eth-contract``
- [``contract-methods``](https://web3js.readthedocs.io/en/v1.2.4/web3-eth-contract.html#methods): ``web3-eth-contract``
- ``getPastEvents``: ``web3-eth-contract``

### Affected Interfaces

- [Transaction](https://github.com/ethereum/web3.js/blob/ae9c90e3ce0bac6d7e9f7ebd6f2cfd906e3e4ae3/packages/web3-core/types/index.d.ts)
- [TransactionReceipt](https://github.com/ethereum/web3.js/blob/ae9c90e3ce0bac6d7e9f7ebd6f2cfd906e3e4ae3/packages/web3-core/types/index.d.ts)
- [EventLog](https://github.com/ethereum/web3.js/blob/ae9c90e3ce0bac6d7e9f7ebd6f2cfd906e3e4ae3/packages/web3-core/types/index.d.ts)
- [Log](https://github.com/ethereum/web3.js/blob/ae9c90e3ce0bac6d7e9f7ebd6f2cfd906e3e4ae3/packages/web3-core/types/index.d.ts)
- [Syncing](https://github.com/ethereum/web3.js/blob/ae9c90e3ce0bac6d7e9f7ebd6f2cfd906e3e4ae3/packages/web3-eth/types/index.d.ts)
- [BlockHeader](https://github.com/ethereum/web3.js/blob/ae9c90e3ce0bac6d7e9f7ebd6f2cfd906e3e4ae3/packages/web3-eth/types/index.d.ts)
- [BlockTransactionBase](https://github.com/ethereum/web3.js/blob/ae9c90e3ce0bac6d7e9f7ebd6f2cfd906e3e4ae3/packages/web3-eth/types/index.d.ts)
- [Block](https://github.com/ethereum/web3.js/blob/ae9c90e3ce0bac6d7e9f7ebd6f2cfd906e3e4ae3/packages/web3-eth/types/index.d.ts)
- [BlockTransactionObject](https://github.com/ethereum/web3.js/blob/ae9c90e3ce0bac6d7e9f7ebd6f2cfd906e3e4ae3/packages/web3-eth/types/index.d.ts)
- [BlockTransactionString](https://github.com/ethereum/web3.js/blob/ae9c90e3ce0bac6d7e9f7ebd6f2cfd906e3e4ae3/packages/web3-eth/types/index.d.ts)

## Type of change

## Checklist:


nivida:
@cgewecke As mentioned in issue #3227 are the max. values defined in [EIP-1985](https://eips.ethereum.org/EIPS/eip-1985) bigger than the ``Number.MAX_SAFE_INTEGER`` of JS. 

That’s why I think adding a ``numbers`` options property is something we definitely should do. Do you agree?

Example: 
``` javascript
{
  numbers: 'default' | 'string' | 'BN' | Function
}
```

I've added ``Function`` as a possible type because it could be interesting to handle this more dynamically (another BigInt lib as BN, BigInt feature detection, etc.).

cgewecke:
@nivida I kind of like the interface proposed in this [comment at 3227](https://github.com/ethereum/web3.js/issues/3227#issuecomment-558977344) and think `extend`, is a good pattern for this case as well. 

To me it seems like there are many different clients who are either returning large values for *a specific field* (like `timestamp`) or not. And the example of [jpmorganchase/quorum.js][1] suggests that fork developers have tended to map outputs to big number on a field-by-field basis.  

On the client side, EIP-1985 is less important than it seems because 

+ Ethereum clients are *not* returning these values (no change is necessary, or desire-able)
+ Non-ethereum clients are not required to respect the EIPs (any field could be different & and users could have any set of expectations)

My instinct is to look at this question (on 1.x anyway) more from a practical standpoint and worry less about conforming to the clients' formal specification.  

@nivida What is you opinion of providing the ability to specify the format of number outputs as a separate extension lib published to the web3-js org? 

[1]: https://github.com/jpmorganchase/quorum.js/blob/master/lib/index.js

nivida:
> I kind of like the interface proposed in this comment at 3227 and think extend, is a good pattern for this case as well.

Yes, the ``extend`` method is great it does give our consumers the opportunity to customize such stuff in their own way. I'm not a big fan of the interface proposed in the mentioned issue. This because all option properties we have currently are defined as simple properties and only the ``setProvider`` function does exist which also does have a bit more logic behind as only to set a value from ``false`` to ``true``. I would provide such a configuration possibility the same way as all other configuration properties we have and we see the ``provider`` as "golden exception". :)

> @nivida What is you opinion of providing the ability to specify the format of number outputs as a separate extension lib published to the web3-js org?

I think this will confuse our customers more than necessary. I would add such a configuration property with four options or we close this PR and do provide this possibility in a clean way for 2.x 

> have tended to map outputs to big number on a field-by-field basis.

Yes, this is something we wouldn’t provide. The idea I had was to return all numbers on all places as string, BN, the current mixed types, or whatever the coerce function is returning. This would be a global setting applied to all numbers to improve the current situation for a lot of custom Ethereum chains.

After all: 
Do you agree to set this ``on-hold`` for now and to put our focus on bugs and features with a higher priority?

cgewecke:
> I think this will confuse our customers more than necessary. I would add such a configuration property with four options or we close this PR and do provide this possibility in a clean way for 2.x

Yeah, I don't think an extension lib is ideal either. And it seems like people will want it to be configurable?

> Do you agree to set this on-hold for now and to put our focus on bugs and features with a higher priority?

Yes, absolutely! We can always come back to this and there are several good solutions...

nivida:
I've closed this PR now and we can later re-open it again if required.

PR #3247, opened 2019-11-28
fixes defaultBlock property handling

### LINKED ISSUE #3244: eth.defaultBlock must be set to hex of number not regular number
#### Expected behavior

According to the documentation here: https://web3js.readthedocs.io/en/v1.2.0/web3-eth.html#defaultblock

defaultBlock can be set to string like 'latest' or Number, it gives the example below:

// set the default block
web3.eth.defaultBlock = 231;

#### Actual behavior

When you next make call you will get error: 
```
invalid argument 1: json: cannot unmarshal number into Go value of type string
```
See code at bottom.

Problem is you must supply Hex String of Number, so the following will work:

```
  const latestBlock = await web3.eth.getBlockNumber()
  web3.eth.defaultBlock = web3.utils.numberToHex( latestBlock - 1 )
```

So either, change documentation, or change something else I think.

#### js code to reproduce the behavior
```
const w3provider = 'wss://goerli.infura.io/ws/v3/<SECRETKEY>'
const Web3 = require('web3')
const web3 = new Web3(
  new Web3.providers.WebsocketProvider(w3provider)
)
;
( async () => {
  const latestBlock = await web3.eth.getBlockNumber()
  web3.eth.defaultBlock = latestBlock - 1
  const accountBal = await web3.eth.getBalance( '0x1addbb6462cb45c91f0dd116c6480aa586de0a29' )
  console.log(`(${accountBal})`)
})()
```

nivida:
## Description

Fixes #3244 

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,3 +109,5 @@ Released with 1.0.0-beta.37 code base.
 - ``getPendingTransactions`` added to web3-eth package (#3239)
 
 ### Fixed
+
+- ``defaultBlock`` property handling fixed (#3247)
--- a/docs/web3-eth-contract.rst
+++ b/docs/web3-eth-contract.rst
@@ -140,10 +140,11 @@ Property
 
 The default block parameters can be one of the following:
 
-- ``Number``: A block number
+- ``Number|BN|BigNumber``: A block number
 - ``"genesis"`` - ``String``: The genesis block
 - ``"latest"`` - ``String``: The latest block (current head of the blockchain)
 - ``"pending"`` - ``String``: The currently mined block (including pending transactions)
+- ``"earliest"`` - ``String``: The genesis block
 
 Default is ``"latest"``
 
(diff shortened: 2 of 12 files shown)


cgewecke (approved):
LGTM. Only thing that might be helpful is to add a regression test for this that uses the reproduction case in #3244. 
```javascript
// Regression test for #3244, should not throw.
it('formats the default block correctly', async function(){
    const accounts = await web3.eth.getAccounts();
    const latestBlock = await web3.eth.getBlockNumber();
    web3.eth.defaultBlock = latestBlock - 1
    await web3.eth.getBalance(accounts[0])
})
```

nivida:
> Only thing that might be helpful is to add a regression test for this that uses the reproduction case in #3244. 

I've added an additional test case to the ``eth.getBalance`` unit test which does cover the mentioned ``defaultBlock`` case. (number -> hex)

Issue #2197, opened 2019-01-17
Unstructured errors returned from reverted transactions


uluhonolulu:
When we have a revert in a transaction, webę throws an error, and the error message contains the revert reason, which is great. However, the message looks like `Returned error: VM Exception while processing transaction: revert Name of storage is required -- Reason given: Name of storage is required.`. Which means that we have to parse the message in order to get the revert reason, and the parser code is going to break every time the message formatting is changed (e.g. "Reason given" is changed to "Reason provided").

Would be great to have it structured, e.g. have a separate field that contains the revert reason. Or even something like this:
```
{
  message: "Returned error: VM Exception while processing transaction: revert Name of storage
is required -- Reason given: Name of storage is required.",
  inner: {
    message: "VM Exception while processing transaction: revert",
    status: "REVERT",  //something that won't be changed or localized
    inner: {
      message: "Name of storage is required."
    }
  }
}
```

The reason (pun intended) is that we need to know 1) if the error was due to a revert and 2) if yes, what was the revert reason. This would be particularly helpful in automated tests, but also in UI.

nivida:
Thanks for submitting this issue. The implementation of the solidity revert method is on our roadmap for the 1.0 stable release.

nivida:
We have implemented this feature for 1.x and will release it on the next minor release. We will implement this feature also for 2.x and I'm closing this issue for now because we will anyways synchronize the existing features between 1.x and 2.x later.

Issue #1707, opened 2018-06-15
Support reason string 


4000D:
According to [Solidity 0.4.22 release](https://github.com/ethereum/solidity/releases/tag/v0.4.22), we can specify the reason in `revert` or `require` function. 

After web3 1.0 alpha (or stable) version released, should plan to support reason string.

kikoncuo:
Any updates?
That'd be nice, right now, the current message is not helpful at all.

kikoncuo:
Quick update, it should work, we are missing the implementation on Geth and Parity, but ganache-cli client (>= 6.1.3) has it, you can test it with truffle v5.0.0

lastmjs:
Is this the reason there is no error message in the transaction data when using require? I have many requires with error messages, and when they fail I can't see anything in the transaction info on Etherscan about the reason. Remix shows the error when I use the JavaScript VM, but not on the test network. Is it true that the transaction information just doesn't have the error message yet when using require?

kikoncuo:
Seems like it, Ganache and Remix VM have it, Geth or Partity, don't.

abhayks1:
This is a very nice feature. Will save time lot of time while debugging.

nivida:
@abhayks1 Yep, it is on my todo list and should get released in v1.0.0-beta.39 or -beta.40.

a-thomas:
Any updates? It seems that we are already at beta.41

CryptoKiddies:
A friendly ping, is error message handling still in the works?

nivida:
@GeeeCoin We do currently discuss the steps until the stable release here in this issue https://github.com/ethereum/web3.js/issues/2684 feel free to join the discussion!

pcowgill:
This issue should be removed from [milestone `1.0`](https://github.com/ethereum/web3.js/milestone/3) or that milestone should be renamed, now that this has the `2.x` label.

Also, it should be moved from the [`1.0` project](https://github.com/ethereum/web3.js/projects/1) to the `2.0` project.

pcowgill:
But also, it's probably worth creating a parallel issue to add this into the stable `1.x` branch, since this is one of the most requested features for `web3.js`.

gorgos:
How does this not have higher priority? This is so important.

[...]

PR #3224, opened 2019-11-19
Add PR and release guidelines

cgewecke:
## Description

As discussed in chat with @nivida - these are ideas for a protocol around 1.x PRs and releases. A basic goal of 1.x has been to make improvements reliably without introducing new problems. This hasn't been achieved yet. Things we've tried include:

+ More tests for browsers, clients and publishing. They've caught some bugs early but also let through serious errors. They aren't adequate in themselves.
+ More thorough review - this is also missing avoidable problems.

In #3070, @alcuadrado  suggested both formalizing the release process and using an `rc` release to sanity-check changes. Following this advice would have yielded better results on 1.2.3 / 1.2.4 especially.

The idea of this doc is  to define a conservative approach to 
+ the kinds of changes we make to 1.x
+ the way releases are executed.

It's a draft - likely vague, too restrictive, and/or too permissive in some places. Any constructive advice on this topic from anyone in the community is more than welcome. 

Sources for the doc include
+ [Parity contributions docs][1]
+ [Ethers contributions docs][2]

[1]: https://github.com/paritytech/polkadot/blob/master/CONTRIBUTING.adoc#rules
[2]: https://docs.ethers.io/ethers.js/html/notes.html#contributing

PR #3248, opened 2019-11-28
Adds revert instruction handling

### LINKED ISSUE #176: Implement new error codes and error reason passing/throwing, when implemented in nodes
For GO see: https://github.com/ethereum/go-ethereum/issues/768

### LINKED ISSUE #1903: Incorrect return message for require() in public/external pure/view-type functions
when directly calling a public/external pure/view-type function with a failing require(), the return message is incorrect as shown here

ethereum/solidity#4840

for an example contract deployed on the mainnet
https://etherscan.io/address/0xd0b4e64c323186340ed2d8168ddae8f18677560e#code

```
pragma solidity ^0.4.24;

contract test {
    
    function sub1(uint256 _a, uint256 _b) public pure returns (uint256 result) {
        require(_a >= _b);
        return _a - _b;
    }
    
    function sub2(uint256 _a, uint256 _b) public pure returns (uint256 result) {
        require(_a >= _b, "_a cannot be less than _b");
        return _a - _b;
    }
    
}
```

And calling sub1(3, 5) and sub2(3, 5) with web3.js 

```
Web3 = require("web3")
web3 = new Web3("...")  // ... is some mainnet provider url, for example Infura.io
abi = ... // ... is the abi json object
contract = new web3.eth.Contract(abi);
ontract.options.address = "0xd0B4e64c323186340ed2D8168DdaE8f18677560E"
contract.methods.sub1(3, 5).call().then(console.log)
contract.methods.sub2(3, 5).call().then(console.log)
```

And for sub1(3, 5) it throws an "Error: Couldn't decode uint256 from ABI: 0x" which I think is quite vague and not informative at all.

Then for sub2(3, 5) it will not throw any error but return an absurd value of 

> 3963877391197344453575983046348115674221700746820753546331534351508065746944

While the error message returned by sub1(3, 5) could use some improvement, the return message by sub2(3, 5) is definitely wrong and should be fixed, as it is basically telling the Dapp front-end that 3 minus 5 equals to 3963877391197344453575983046348115674221700746820753546331534351508065746944 and that could even become a security vulnerability.

nivida:
## Description

Fixes #176 #1903 #1707 #2197

This PR does implement revert instruction handling in the web3.js lib.
 
If the ``handleRevert`` options property is set to ``true`` you will get revert reasons on calling of ``Contract`` object methods, ``sendTransaction``, and ``call`` of the ``eth`` module. Those revert reasons will get returned to the user over the error listener (``call`` doesn't have a error event) or the catch handler if you are using the native Promise object.

### Error Interface
``` typescript
interface RevertInstructionError extends Error {
    reason: string;
    signature: string;
}
```

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -43,5 +43,12 @@ module.exports = {
     },
     ConnectionTimeout: function (ms){
         return new Error('CONNECTION TIMEOUT: timeout of ' + ms + ' ms achived');
+    },
+    RevertInstructionError: function(reason, signature) {
+        var error = new Error('Your request got reverted with the following reason string: ' + reason);
+        error.reason = reason;
+        error.signature = signature;
+
+        return error;
     }
 };
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -50,5 +50,12 @@ module.exports = {
         error.signature = signature;
 
         return error;
+    },
+    TransactionRevertInstructionError: function(reason, signature, receipt) {
+        var error = new Error('Transaction has been reverted by the EVM:\n' + JSON.stringify(receipt, null, 2));
+        error.reason = reason;
+        error.signature = signature;
+
+        return error;
     }
 };
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -410,9 +410,12 @@ Method.prototype._confirmTransaction = function(defer, result, payload) {
 
                             if (receipt.status === false || receipt.status === '0x0') {
                                 try {
-                                    var revertMessage = await method.getRevertReason(payload.params[0], receipt.blockNumber);
+                                    var revertMessage = await method.getRevertReason(
+                                        payload.params[0],
+                                        receipt.blockNumber
+                                    );
 
-                                    if (revertMessage.reason) {
+                                    if (revertMessage) {
                                         utils._fireError(
                                             errors.RevertInstructionError(revertMessage.reason, revertMessage.signature),
                                             defer.eventEmitter,
@@ -723,11 +726,14 @@ Method.prototype.buildCall = function() {
 };
 
 /**
+ * Returns the revert reason string if existing or otherwise false.
  *
- * @param txOptions
- * @param revertedReceipt
- * @param abiCoder
- * @returns {Promise<*>}
+ * @method getRevertReason
+ *
+ * @param {Object} txOptions
+ * @param {Number} blockNumber
+ *
+ * @returns {Promise<Boolean|String>}
  */
 Method.prototype.getRevertReason = function(txOptions, blockNumber) {
     var self = this;
@@ -741,7 +747,7 @@ Method.prototype.getRevertReason = function(txOptions, blockNumber) {
         }))
             .createFunction(self.requestManager)(txOptions, utils.numberToHex(blockNumber))
             .then(function() {
-                resolve({reason: false});
+                resolve(false);
             })
             .catch(function(error) {
                 if (error.reason) {
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -772,7 +772,7 @@ Method.prototype.getRevertReason = function(txOptions, blockNumber) {
  * @returns {Boolean}
  */
 Method.prototype.isRevertReasonString = function (data) {
-    return ((data.length - 2) / 2) % 32 === 4 && data.substring(0, 10) === '0x08c379a0';
+    return _.isString(data) && ((data.length - 2) / 2) % 32 === 4 && data.substring(0, 10) === '0x08c379a0';
 };
 
 /**
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -413,7 +413,7 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
                                 try {
                                     var revertMessage = null;
 
-                                    if (method.handleRevert) {
+                                    if (method.handleRevert && method.call === 'eth_sendTransaction') {
                                         // Get revert reason string with eth_call
                                         revertMessage = await method.getRevertReason(
                                             payload.params[0],
@@ -425,7 +425,7 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
                                                 errors.RevertInstructionError(revertMessage.reason, revertMessage.signature),
                                                 defer.eventEmitter,
                                                 defer.reject,
-                                                payload.callback,
+                                                null,
                                                 receipt
                                             );
                                         } else {
@@ -445,7 +445,7 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
                                     );
                                 }
                             } else {
-                                // Throw OOG if status is true and provided gas and used gas are equal
+                                // Throw OOG if status is not existing and provided gas and used gas are equal
                                 utils._fireError(
                                     new Error('Transaction ran out of gas. Please provide more gas:\n' + receiptJSON),
                                     defer.eventEmitter,
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -61,7 +61,7 @@ var Method = function Method(options) {
     this.handleRevert = options.handleRevert;
 };
 
-Method.prototype.setRequestManager = function(requestManager, accounts) {
+Method.prototype.setRequestManager = function (requestManager, accounts) {
     this.requestManager = requestManager;
 
     // reference to eth.accounts
@@ -71,7 +71,7 @@ Method.prototype.setRequestManager = function(requestManager, accounts) {
 
 };
 
-Method.prototype.createFunction = function(requestManager, accounts) {
+Method.prototype.createFunction = function (requestManager, accounts) {
     var func = this.buildCall();
     func.call = this.call;
 
@@ -80,7 +80,7 @@ Method.prototype.createFunction = function(requestManager, accounts) {
     return func;
 };
 
-Method.prototype.attachToObject = function(obj) {
+Method.prototype.attachToObject = function (obj) {
     var func = this.buildCall();
     func.call = this.call;
     var name = this.name.split('.');
@@ -99,7 +99,7 @@ Method.prototype.attachToObject = function(obj) {
  * @param {Array} arguments
  * @return {String} name of jsonrpc method
  */
-Method.prototype.getCall = function(args) {
+Method.prototype.getCall = function (args) {
     return _.isFunction(this.call) ? this.call(args) : this.call;
 };
 
@@ -110,7 +110,7 @@ Method.prototype.getCall = function(args) {
  * @param {Array} arguments
  * @return {Function|Null} callback, if exists
  */
-Method.prototype.extractCallback = function(args) {
+Method.prototype.extractCallback = function (args) {
     if (_.isFunction(args[args.length - 1])) {
         return args.pop(); // modify the args array!
     }
@@ -123,7 +123,7 @@ Method.prototype.extractCallback = function(args) {
  * @param {Array} arguments
  * @throws {Error} if it is not
  */
-Method.prototype.validateArgs = function(args) {
+Method.prototype.validateArgs = function (args) {
     if (args.length !== this.params) {
         throw errors.InvalidNumberOfParams(args.length, this.params, this.name);
     }
@@ -136,14 +136,14 @@ Method.prototype.validateArgs = function(args) {
  * @param {Array}
  * @return {Array}
  */
-Method.prototype.formatInput = function(args) {
+Method.prototype.formatInput = function (args) {
     var _this = this;
 
     if (!this.inputFormatter) {
         return args;
     }
 
-    return this.inputFormatter.map(function(formatter, index) {
+    return this.inputFormatter.map(function (formatter, index) {
         // bind this for defaultBlock, and defaultAccount
         return formatter ? formatter.call(_this, args[index]) : args[index];
     });
@@ -156,11 +156,11 @@ Method.prototype.formatInput = function(args) {
  * @param {Object}
  * @return {Object}
  */
-Method.prototype.formatOutput = function(result) {
+Method.prototype.formatOutput = function (result) {
     var _this = this;
 
     if (_.isArray(result)) {
-        return result.map(function(res) {
+        return result.map(function (res) {
             return _this.outputFormatter && res ? _this.outputFormatter(res) : res;
         });
     } else {
@@ -175,7 +175,7 @@ Method.prototype.formatOutput = function(result) {
  * @param {Array} args
  * @return {Object}
  */
-Method.prototype.toPayload = function(args) {
+Method.prototype.toPayload = function (args) {
     var call = this.getCall(args);
     var callback = this.extractCallback(args);
     var params = this.formatInput(args);
@@ -195,7 +195,7 @@ Method.prototype.toPayload = function(args) {
 };
 
 
-Method.prototype._confirmTransaction = function(defer, result, payload) {
+Method.prototype._confirmTransaction = function (defer, result, payload) {
     var method = this,
         promiseResolved = false,
         canUnsubscribe = true,
@@ -216,7 +216,7 @@ Method.prototype._confirmTransaction = function(defer, result, payload) {
             name: 'getBlockByNumber',
             call: 'eth_getBlockByNumber',
             params: 2,
-            inputFormatter: [formatters.inputBlockNumberFormatter, function(val) {
+            inputFormatter: [formatters.inputBlockNumberFormatter, function (val) {
                 return !!val;
             }],
             outputFormatter: formatters.outputBlockFormatter
@@ -248,19 +248,19 @@ Method.prototype._confirmTransaction = function(defer, result, payload) {
     ];
     // attach methods to this._ethereumCall
     var _ethereumCall = {};
-    _.each(_ethereumCalls, function(mthd) {
+    _.each(_ethereumCalls, function (mthd) {
         mthd.attachToObject(_ethereumCall);
         mthd.requestManager = method.requestManager; // assign rather than call setRequestManager()
     });
 
 
     // fire "receipt" and confirmation events and resolve after
-    var checkConfirmation = function(existingReceipt, isPolling, err, blockHeader, sub) {
+    var checkConfirmation = function (existingReceipt, isPolling, err, blockHeader, sub) {
         if (!err) {
             // create fake unsubscribe
             if (!sub) {
                 sub = {
-                    unsubscribe: function() {
+                    unsubscribe: function () {
                         clearInterval(intervalId);
                     }
                 };
@@ -268,7 +268,7 @@ Method.prototype._confirmTransaction = function(defer, result, payload) {
             // if we have a valid receipt we don't need to send a request
             return (existingReceipt ? promiEvent.resolve(existingReceipt) : _ethereumCall.getTransactionReceipt(result))
             // catch error from requesting receipt
-                .catch(function(err) {
+                .catch(function (err) {
                     sub.unsubscribe();
                     promiseResolved = true;
                     utils._fireError(
@@ -281,7 +281,7 @@ Method.prototype._confirmTransaction = function(defer, result, payload) {
                     );
                 })
                 // if CONFIRMATION listener exists check for confirmations, by setting canUnsubscribe = false
-                .then(async function(receipt) {
+                .then(async function (receipt) {
                     if (!receipt || !receipt.blockHash) {
                         throw new Error('Receipt missing or blockHash null');
                     }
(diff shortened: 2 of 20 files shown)


packages/web3-core-method/src/index.js
```diff
@@ -526,16 +553,34 @@ var getWallet = function(from, accounts) {
...
     // actual send function
     var send = function() {
         var defer = promiEvent(!isSendTx),
             payload = method.toPayload(Array.prototype.slice.call(arguments));
 
-
         // CALLBACK function
         var sendTxCallback = function(err, result) {
+            if (isCall && (method.isRevertReasonString(result) && method.abiCoder)) {
+                var reason = method.abiCoder.decodeParameter('string', '0x' + result.substring(10));
```
nivida, line 570: It could be done with the utility functions web3.js does have. But there will be the possibility in the future to have custom error signatures and the AbiCoder would then anyways be used here.

packages/web3-core-method/src/index.js
```diff
@@ -683,6 +725,56 @@ Method.prototype.buildCall = function() {
...
+ * Checks if the given hex string is a revert message from the EVM
+ *
+ * @method isRevertReasonString
+ *
+ * @param {String} data - Hex string prefixed with 0x
+ *
+ * @returns {Boolean}
+ */
+Method.prototype.isRevertReasonString = function (data) {
+    return _.isString(data) && ((data.length - 2) / 2) % 32 === 4 && data.substring(0, 10) === '0x08c379a0';
```
nivida, line 780:
```
length = (data.length - 2); // exclude prefix
length = length / 2; // Transforming length to bytes length
argsCount = ((length % 32) === 4 ) // We have 4 arguments: signature, offset, length, data
isErrorSignature = data.substring(0, 10) === '0x08c379a0'; // The first 4 bytes are the Error signature

isRevertReasonString = (argsCount && isErrorSignature)
```

cgewecke (commented):
Hi, just leaving a quick note. This looks really cool! 

The are couple things that seem like breaking changes
+ Throwing on eth_call
+ Throwing with a different error message on eth_send

Additionally, I'm a little worried about additional server calls to fetch the reason since many wrappers or libs around Web3 are already doing that. In a contract testing context with lots error cases being run it could slow things down to duplicate this logic. 

What do you think about making reason fetching available as an option that is toggled off by default and on via a new contract option? 

Super interested in how `eth_call` works as well...do you think it might be possible to spoof a 'revert' message for a call? Like if data can be set by a contract's users, they could set it so that on 'read', web3 would decode it as an error message and throw?

nivida:
@cgewecke 
> This looks really cool!

Thanks!:)

> What do you think about making reason fetching available as an option that is toggled off by default and on via a new contract option?

Sounds good to me but we have to be careful to not have too many configuration options because it could have a bad impact on the DX in the long run.  I would also not have this only as a ``Contract``option instead would I prefer to have this as a "global" ``web3-eth`` module option with the possibility to disable and enable the revert instruction handling for single ``Contract`` objects. This because of ``eth.sendTransaction``, ``eth.sendSignedTransaction``, and ``eth.call`` could also be used to call ``Contract`` functions with the help of the ``eth.abi`` module.

>  Super interested in how eth_call works as well...do you think it might be possible to spoof a 'revert' message for a call? Like if data can be set by a contract's users, they could set it so that on 'read', web3 would decode it as an error message and throw?

I'm not sure if I understood you correctly. But you would like to have the revert instruction handling also for the bare ``eth`` JSON-RPC methods as ``eth.call(...)``? 
This would already be implemented for  ``eth.sendTransaction``, ``eth.sendSignedTransaction``, and ``eth.call``.

nivida:
@cgewecke I have added all the missing test cases and improved the solution I had with the ``handleRevert`` module option which does default to ``false``. I'm not sure if this module option name is the most explicit one we can have, so feel free to propose a better name! ;) 

During testing of the revert instruction handling did I notice that only module options which are defined before the ``Contract`` instance got created do get applied and all others not. This behavior is existing since the beginning of 1.0 and is actually not that easy to fix as I thought first. 
That’s why I will check that closer later and will open a related issue with a possible solution after.

Example: 
``` javascript
web3.eth.handleRevert = true;
contract = new web3.eth.Contract(..);
contract.methods.method(...).send(...);  // works as expected

-----

contract = new web3.eth.Contract(..); 
contract.handleRevert = true;
contract.methods.method(...).send(...);  // doesn't work as expected
```

Edit:

I've added all new properties to the contract options with commit https://github.com/ethereum/web3.js/pull/3248/commits/9db572c4560f9780abdd4f2f0e84065412794cfd.

Example:

``` javascript
contract = new web3.eth.Contract([], '0x...', {handleRevert: true, ...}); 
contract.methods.method(...).send(...);  // does work as expected
```

ToDo: 

- [x] Extending of ``Contract`` options tests

packages/web3-core-helpers/src/errors.js
```diff
@@ -43,5 +43,12 @@ module.exports = {
     },
     ConnectionTimeout: function (ms){
         return new Error('CONNECTION TIMEOUT: timeout of ' + ms + ' ms achived');
+    },
+    RevertInstructionError: function(reason, signature) {
+        var error = new Error('Your request got reverted with the following reason string: ' + reason);
```
cgewecke, line 48: I wonder if it's possible to have this message resemble the existing transaction message more... it looks like this:
```javascript
new Error('Transaction has been reverted by the EVM:\n' + receiptJSON),
```

Lots of people grep that and parse the receipt out of the string as well. The message I propose below is  awkwardly written, but it might provide the same information at a lower cost to people who already have error handling logic in place.

```javascript
'Request errored with revert reason: ' + reason + '. Transaction has been reverted by the EVM:\n' + receiptJSON'
```

And maybe to handle `.call` you would have the function flow be
```javascript
RevertInstructionError: function(reason, signature, receipt){
  if (receipt !== undefined){
    // it's a transaction
 } else {
    // Message is only: 'Request errored with revert reason: ' + reason.'
 }
}
```

packages/web3-core-method/src/index.js
```diff
@@ -408,14 +410,42 @@ Method.prototype._confirmTransaction = function(defer, result, payload) {
...
+                                    utils._fireError(
+                                        new Error('Transaction has been reverted by the EVM:\n' + receiptJSON),
+                                        defer.eventEmitter,
+                                        defer.reject,
+                                        null,
+                                        receipt
+                                    );
+                                }
                             } else {
+                                // Throw OOG if status is true and provided gas and used gas are equal
```
cgewecke: Wait! This only a comment but that change was reverted in 1.2.4. If status is true the tx succeeded and shouldn't throw. This gas logic should only run if there's no `status`.

packages/web3-eth-contract/src/index.js
```diff
@@ -180,12 +180,13 @@ var Contract = function Contract(jsonInterface, address, options) {
...
-    this.defaultChain = this.constructor.defaultChain;
-    this.defaultHardfork = this.constructor.defaultHardfork;
-    this.defaultCommon = this.constructor.defaultCommon;
+    this.transactionBlockTimeout = this.options.transactionBlockTimeout || this.constructor.transactionBlockTimeout;
+    this.transactionConfirmationBlocks = this.options.transactionConfirmationBlocks || this.constructor.transactionConfirmationBlocks;
+    this.transactionPollingTimeout = this.options.transactionPollingTimeout || this.constructor.transactionPollingTimeout;
+    this.defaultChain = this.options.defaultChain || this.constructor.defaultChain;
+    this.defaultHardfork = this.options.defaultHardfork || this.constructor.defaultHardfork;
+    this.defaultCommon = this.options.defaultCommon || this.constructor.defaultCommon;
+    this.handleRevert = this.options.handleRevert || this.constructor.handleRevert;
```
cgewecke: Is this one a little tricky because it's boolean? For example if I set 
```javascript
web3.eth.handleRevert = true;
```
and later I change my mind for a specific contract
```javascript
web3.eth.Contract.handleRevert = false
```
I think it will still be `true`.

cgewecke (commented):
Yes this is looking great @nivida! Just left a couple comments about the err message format and the option behavior. 

Earlier I asked a question about spoofing the revert message with `.call`. With a transaction, the `status` field is checked which means that the tx definitely errored. But with a call, you can only make a guess about what the returned data means. 

I'm wondering if it's possible for a user to set some string value in the contract so that when another person tries to fetch the value, Web3 interprets it as an error and throws?  If so, then it's possible to try to trick someone. 

For example, suppose there is method like:
```solidity
function setNameOfCat(string oldName, string newName) public onlyCatOwner {
  require(isCatOwner(msg.sender, oldName));
  ...etc
}

// Anyone is allowed to read this...
function getCatNameByAddress(address _address) public view returns (string){
  return cats[_address];
}
```

Suppose I am a cat owner and instead of a cat name I set a message with the error prefix that said:
```
In order to read data from this contract you must set an approval for address '0xabc....'  etc...
```

It might be dangerous if someone read my cat's name. Does that make sense? 

**EDIT: Notes on whether this is possible.**

~~Fwiw, I've tried generate a case for this and it's not obvious how it could be done. The `bytes` and `strings` formats returned by the client are very different than the error message format.~~

~~Actually it's possible with `bytes32[]`....~~

Nope, the length check in the reason string evaluator look like it prevents this hack from working.

packages/web3-core-helpers/src/errors.js
nivida: > Lots of people grep that and parse the receipt out of the string as well.

Because the new error message only appears if the ``handleRevert`` property is set to ``true`` do I think can we keep the error message as it is.  

> And maybe to handle .call you would have the function flow be

I thought to keep the logic to only return the receipt for the error listener and to not duplicate the return values with adding it to the ``Error`` object. If the mentioned behavior above is the wished behavior do I propose because of consistency reasons to add the ``receipt`` to all ``Error`` objects.

packages/web3-core-method/src/index.js
nivida: Oh, sorry this is the logic of 2.x. I will update this inline comment.

packages/web3-eth-contract/src/index.js
nivida: The example from you will work and is implemented in the same way [as all other properties](https://github.com/ethereum/web3.js/blob/feature/revert-reason-string/packages/web3-eth/src/index.js#L303). The difference to the ``defaultAccount`` property, for example, is that we are able to define those options over the passed contract constructor options parameter. 

Edit:
Also the ``defaultAccount`` can get defined with the passed options object (``from``).

nivida:
> It might be dangerous if someone read my cat's name. Does that make sense?

This is true. But I think it should be handled correctly in the contract or from our consumer with validating the input values. They are able then to show their users the correct error message.

packages/web3-core-method/src/index.js
```diff
@@ -408,14 +410,42 @@ Method.prototype._confirmTransaction = function(defer, result, payload) {
...
+                                            payload.params[0],
+                                            receipt.blockNumber
+                                        );
+
+                                        if (revertMessage) { // Only throw a revert error if a revert reason is existing
+                                            utils._fireError(
+                                                errors.RevertInstructionError(revertMessage.reason, revertMessage.signature),
+                                                defer.eventEmitter,
+                                                defer.reject,
+                                                payload.callback,
```
cgewecke: Ah yes, what's in this callback? Why is that param different than the existing handler which uses `null`?

nivida: Oh, thanks for seeing this! Yes, the callback should not get passed here because only the ``transactionHash`` does get returned if a callback is existing.

packages/web3-core-helpers/src/errors.js
cgewecke: > Because the new error message only appears if the handleRevert property is set to true do I think can we keep the error message as it is.

What is the advantage of your design?  

>  add the receipt to all Error objects

Yes that seems reasonable to me. 

About serializing the receipt in the message - that decision was made a long time ago and people have adapted to it. They've written all kinds of logic to manage it. Suddenly deciding not to support that interface seems weird to me, idk.

packages/web3-eth-contract/src/index.js
cgewecke: Oh! Thanks for linking to that part of the code - I misunderstood this.

cgewecke:
@nivida Could this also be rebased against 1.x at some point so it's possible to see it run against the Truffle E2E?

packages/web3-core-helpers/src/errors.js
nivida: > What is the advantage of your design?

It doesn't break existing code and will only break it if you use the newly provided property which was, in the beginning, the reason to add it. 

> Yes that seems reasonable to me.

Great I will open a separate PR for this!

> About serializing the receipt in the message - that decision was made a long time ago and people have adapted to it. They've written all kinds of logic to manage it. Suddenly deciding not to support that interface seems weird to me, idk.

I've used another error message to only have one new ``Error`` which is used on ``eth_call`` and ``eth_sendTransaction``. If the serialized receipt is a must-have for those error messages can we also have two different ``Error``'s. This means I would use the old one (``Transaction has been reverted by the EVM:\n' + receiptJSON'``) for ``eth_sendTransaction`` calls and the new one for direct ``eth_call`` calls.

nivida: This would probably allow us to remove the ``handleRevert`` options property at all because the ``Error`` is still the same and does only have additional properties if a revert message is existing.

cgewecke: > I've used another error message to only have one new Error which is used on eth_call and eth_sendTransaction. If the serialized receipt is a must-have for those error messages can we also have two different Error's. 

Wait I don't understand that part. Keeping the error formats similar doesn't seem like it causes lots of problems. 

> This would probably allow us to remove the handleRevert options property at all because the Error is still the same and does only have additional properties if a revert message is existing.

I don't think this true because making additional calls that are already duplicated upstream for this lib potentially slows things down, so it should be optional. There's also completely different behavior for `.call` and potentially some security considerations for Solidity writers.

nivida: I have kept the ``handleRevert`` property as it is and I've added the ``TransactionRevertInstructionError``. This error does have the same error message as we currently do have. With the difference of two additional added object properties called ``reason`` and ``signature``.

cgewecke: I'm unresolving this to look at it again tomorrow, sorry.

nivida: > some security considerations for Solidity writers.

I can't really see what security considerations are here because someone could also call the Ethereum blockchain with his own node and scripts and would get those details returned. There is nothing I think web3.js has to do in this direction. 

> Keeping the error formats similar doesn't seem like it causes lots of problems.

I've chosen a generic error message because the same error was used for ``eth_call`` and ``eth_sendTransaction``. I have created now, in addition, a second error especially for the transaction case which does have the same properties as the generic one but with the small difference of having the same error message as we have currently in 1.2.4. (``'Transaction has been reverted by the EVM:\n' + receiptJSON``)

cgewecke: I think it makes sense to be **as consistent as possible with the existing implementation**. The library has many users who want to see the reason. They'd also like to get it while making the smallest number changes to their own code. That means:

+ New error messages should include the reason. (If this isn't done, there will immediately be issues opened saying feature doesn't work).
+ New error messages should resemble the old message. (So existing handlers are not broken.)
+ New error message for `call` and `send` should resemble each other as much a possible. (So, similar error handlers can be written/used. DRY.) 

I think the above conditions are difficult to get perfect but atm the logic in this PR does not satisfy these requirements.

> I can't really see what security considerations are here because someone could also call the Ethereum....There is nothing I think web3.js has to do in this direction.

Ok, I'm not sure how easy it is to do the fake error I suggested below anyway. But perhaps we could get more opinions about the security implications. I think Web3 *should* worry a little about what it's like to use the library without understanding all the small details:

+ Most users assume that error messages are 'official' and believe them. 
+ Many engineers would be quite surprised if their users could modify the contract such that ordinary data was interpreted by the JS API as an error with a custom message the contract writers had not approved.

nivida: All points above are implemented as described with the last commit. The error message is still the same I've only extended the object property list with ``signature`` and ``reason`` for the ``eth_sendTransaction`` call. The only difference and also only if ``handleRevert`` is set to true is that ``eth_call`` does throw an error if a revert instruction got detected.

 > Many engineers would be quite surprised if their users could modify the contract such that ordinary data was interpreted by the JS API as an error with a custom message the contract writers had not approved.

This is something we never can forbid in JS. The only way to stop a user to modify a contract object would be if web3.js would only run in a JavaScript runtime which does have SES and the related Realms object implemented. The other case you mentioned with spoofing of the requests and returning of wrong errors (malicious node) is also something we only could partially taking action against with having of simultaneously running providers which are connected to different nodes and we compare the returned result of them. A normal middle man attack can get worked against completely by our users using SSL encrypted communication between the client and the connected node.

cgewecke: > All points above are implemented as described with the last commit. The error message is still the same I've only extended the object property list with signature and reason for the eth_sendTransaction call.

Here are the points again:

+  New error messages should include the reason. (If this isn't done, there will immediately be issues opened saying feature doesn't work).
+ New error messages should resemble the old message. (So existing handlers are not broken.)
+ New error message for `call` and `send` should resemble each other as much a possible. (So, similar error handlers can be written/used. DRY.) 

He are the ways it's currently different from the points:
1. The TransactionRevertInstructionError does not include the reason in the error message string. 
2. (Only making this point so that when 1 is done, 2 is also satisfied.)
3. `call` message contains the reason in the error message string, `send` does not. 

Also the message for `call` is this:
```
Your request got reverted with the following reason string:
```

It might make sense to reserve the word 'revert' for state-changing transactions. Most `calls` will just be to read from a view fn. Writing the message so it captures that distinction could be helpful. I would change my original wording for the reason phrase to:
```
Request errored with Solidity reason: ' + reason
```

RE: MITM etc, ok fair enough although I think those issues affect all applications - this question is quite specific to expectations about how Solidity behaves that might not be clear to everyone. I'm not suggesting that the call handling be removed - only that this issue is discussed with others to get their opinion about it's severity and to think about how to disclose it to users (if necessary).

cgewecke (commented):
Apart from the fight about the error message, this LGTM.

WDYT about including test cases for? 
```
web3.eth.call()
web3.eth.sendTransaction()
```

packages/web3-core-helpers/src/errors.js
nivida: > The TransactionRevertInstructionError does not include the reason in the error message string.

Yes, to not break the existing code. I would keep this as it is and we can inform the user about the interface in the release and our documentation. The proposed error message from you would be a really ugly solution which I will not implement.

> call message contains the reason in the error message string, send does not.

Yes, because it is a completely new error and does anyways throw where it wasn't throwing an error before. I think this is totally okay to have it like this. 

> Request errored with Solidity reason: ' + reason

This would exclude Vyper and other contract languages. The word ``reverted`` is also correct because the revert instruction got executed in the EVM.

nivida:
> WDYT about including test cases for? 

The test cases would be the same and would also test the same code. But I definitely can create a generic test helper for this and extend the test cases for ``eth.call`` and ``eth.sendTransaction``. 👍

packages/web3-core-helpers/src/errors.js
cgewecke: Good point about Vyper.  I also see what you mean about the opcode as well - there's a case there. 

> The proposed error message from you would be a really ugly solution which I will not implement.

Ok. This seems like the root of the disagreement. This discussion has been a waste of everyone's time.

cgewecke (approved):
In summary, this review ends in approval from me, per the PR review guidelines proposed in #3224. AFAIK it doesn't violate any of the requirements. 

(There's are a few dozen unnecessary lints that go against the idea of making the smallest number of changes required, but the core changes seem well-restricted in scope).

packages/web3-core-helpers/src/errors.js
nivida: > This discussion has been a waste of everyone's time.

Yes, the learnings I take away from this is that also changing of an error message string does count as breaking change and that such changes in the "heart" of web3.js should be discussed in detail before the implementation does start. This will probably save time on both sides the proposer and reviewer of a PR. Thanks for your inputs and the review at all!

cgewecke:
@nivida Final note! This is really cool,  Web3 might be the first lib to collect the revert string from `view` functions and will be very useful. 

Looked pretty safe to me as well - the length of the reason message is hard to replicate using conventional datatypes.

knaderi:
Hi,
I upgraded from web3 1.2.4 to 1.2.7.
I have this call in my code:
const transactionReceipt:TransactionReceipt = await web3.eth.sendSignedTransaction(answer.rawTransaction)

error TS2322: Type 'TransactionRevertInstructionError | TransactionReceipt' is not assignable to type 'TransactionReceipt'.
  Type 'TransactionRevertInstructionError' is missing the following properties from type 'TransactionReceipt': status, transactionHash, transactionIndex, blockHash, and 7 more.

57       const transactionReceipt:TransactionReceipt = await web3.eth.sendSignedTransaction(answer.rawTransaction)

cgewecke:
@knaderi 

It looks like many of the promise types in web3-eth have mis-defined optional error types. Am opening an issue to correct this right now.

If you *don't* specify the type of your variable (e.g. just let TS infer it) does everything work ok?
```ts
 const transactionReceipt = await web3.eth.sendSignedTransaction(answer.rawTransaction)
```

PR #3260, opened 2019-12-07
Add bundlesize to CI

cgewecke:
## Description

Adds a [bundlesize][1] checker to CI. First stage of plan for the build pipeline per [comment 3160][2].

Service and tool lgtm, they have [grant funding from Google Chrome][3]. Added their token in CI to the Travis settings.

There are various config options including the ability to inspect multiple bundles, with and without compression.  Chose a single report for the `packages/web3/dist` (uncompressed) because that seemed to give the clearest report. And set the maxSize high for now - can be tightened as size is optimized.

Looks like this:

Screen Shot 2019-12-07 at 1 12 08 PM

@nivida if you look in the CI build job you'll see a server error from buildsize - it's possible this is because you'll need to authorize the app for Web3? Currently just using a token from my auth. 

[1]: https://github.com/siddharthkp/bundlesize
[2]: https://github.com/ethereum/web3.js/pull/3160#issuecomment-558340099
[3]: https://github.com/siddharthkp/bundlesize/issues/306#issuecomment-505201605

## Type of change
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] Any dependent changes have been merged and published in downstream modules.
- [x] I ran ```npm run dtslint``` with success and extended the tests and types if necessary.
- [x] I ran ```npm run test:unit``` with success and extended the tests if necessary.
- [x] I ran ```npm run build-all``` and tested the resulting file/'s from  ```dist``` folder in a browser.
- [ ] I have updated the ``CHANGELOG.md`` file in the root folder.
- [ ] I have tested my code on the live network.

PR #3206, opened 2019-11-14
Release - 1.2.3

nivida:
## Description

This release improves the TS type definitions and does include two smaller patches related to the OOG detection and the provided minified web3 bundle.

### Fixed

- Fix perfect gas usage causes tx to error (#3175)
- Fix regenerator runtime error in web3.min.js (#3155)
- Fix TS types for eth.subscribe syncing, newBlockHeaders, pendingTransactions (#3159)
- Improve web3-eth-abi decodeParameters error message (#3134)

### Compare View

[v1.2.2 -> v1.2.3](https://github.com/ethereum/web3.js/compare/v1.2.2...80ee9a5)

## Type of change

- [x] Release

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] Any dependent changes have been merged and published in downstream modules.
- [x] I ran ```npm run test``` with success and extended the tests if necessary.
- [x] I ran ```npm run build``` and tested the resulting file from ```dist``` folder in a browser.
- [x] I have tested my code on the live network.

Issue #3155, opened 2019-10-24
web3.min.js v1.2.2: Uncaught ReferenceError: regeneratorRuntime is not defined


gpersoon:
### Description
Error "Uncaught ReferenceError: regeneratorRuntime is not defined" occurred when calling a smart contract function. The error shows in the developer console.

#### Expected behavior
No error
Note: with a browserified version of web3.js 1.2.1 this error doesn't occur

#### Actual behavior
An error is shown in the developer console

#### Workaround
A workaround is to include polyfill.min.js:
```JS
<script src="https://cdn.jsdelivr.net/npm/babel-polyfill/dist/polyfill.min.js"></script> 
```

#### Steps to reproduce the behavior
```JS
npm install web3
```
Use chrome and metamask on rinkeby network to load following html page:
```JS
<html>
    <head>
        <script src=".........../node_modules/web3/dist/web3.min.js"></script>               
    </head>
<body>
<script>   
    async function asyncloaded() { 
        await ethereum.enable();
        web3 = new Web3(Web3.givenProvider);
        var ERC777ABI = [{"constant": false,"inputs": [{"internalType": "address","name": "recipient","type": "address"},{"internalType": "uint256","name": "amount","type": "uint256"},{"internalType": "bytes","name": "data","type": "bytes"}],        "name": "send","outputs": [],"payable": false,"stateMutability": "nonpayable","type": "function"}];
        const ERC777 = new web3.eth.Contract(ERC777ABI, "0x51a8e7fEC5639Eb6861a674E4F46148561d5c5c6");  // rinkeby
        accounts=await web3.eth.getAccounts();        
        var result = await ERC777.methods.send(accounts[0],1,"0x77").send({from: accounts[0]});
    }
    window.addEventListener('load', asyncloaded);  
    </script>     
</body>
</html>
```
The deployed smart contact is an ERC777 contract.
#### Error Logs
8inpage.js:1 Uncaught ReferenceError: regeneratorRuntime is not defined
    at web3.min.js:1
    at n (web3.min.js:1)
    at web3.min.js:1
    at Array.forEach (<anonymous>)
    at Object.callback (web3.min.js:1)
    at MetamaskInpageProvider.<anonymous> (web3.min.js:1)
    at u (inpage.js:1)
    at MetamaskInpageProvider.a.emit (inpage.js:1)
    at a.<anonymous> (inpage.js:1)
    at u (inpage.js:1)
### Versions
* web3.js: 1.2.2
[...]

PR #3261, opened 2019-12-07
Fix minified bundle

### LINKED ISSUE #3256: Loading web3.min.js in browser fails with "require is not defined"
I've installed `web3@1.2.4` with `npm install`.
I copied the file `node_modules/web3/dist/web3.min.js` into my web app.
I'm loading the file with:

    <script type="text/javascript" src="js/web3.min.js"></script>

In the browser console (Chrome 70+), I see this error:

    Uncaught ReferenceError: require is not defined
        at web3.js:1

cgewecke:
## Description

**NB: This PR targets the buildsize checker PR #3260, prerequisite for merging here**

The bundle entry point is corrupted due to fix proposed in #3155 & implemented in #3206 
There are two settings for the babel presets and in web3's case, picking either breaks the build in a different way.

|                      | regeneratorRuntime error | "require" is not defined error |
|----------------------|--------------------------|--------------------------------|
| useBuiltIns: "usage" |           false          | true                           |
| useBuiltIns: "entry" |           true           | false                          |
 
The fix is to use "entry" and add `@babel/polyfill` to gulp's browserify step. Increases the minified bundle's size about 150 kb (~10%). This wouldn't affect anyone whose packing things themselves though.
 
PR adds
+  tests for the "require not defined" case (git repo and published bundles). The "runtime" error is already caught by the browser tests.
+ an E2E validation which models the CDN use-case. It [publishes][4] a simple website which uses the bundle to a random url at surge.sh. This site uses Web3 to query the latest block from mainnet over Infura.  An example from this PR's CI run is at: 

  [sudden-playground.surge.sh][3]

(Sometimes surge.sh fails with an auth error because it's generated a name which is already taken or something. The tests will pass though - the surge check is just meant to be a convenience.)

The commits are sequenced with initially failing tests so their correctness can be checked: 

+ First commit 245b4ec9c232da2f8639ff381394e74361780aa9 has no build changes and reproduces the "require not defined" error [here][1]
+ Second commit 72bf5f2273071094efefb23bee7c8fd69cf09128 fixes "require not defined" but fails with the "runtime" error [here][2]
+ Third commit 953c2e1b102707d930ec3e5553831b605b6eaa7c fixes both

[1]: https://travis-ci.org/ethereum/web3.js/jobs/622114843#L734-L741
[2]: https://travis-ci.org/ethereum/web3.js/jobs/622119454#L567-L571
[3]: http://sudden-playground.surge.sh/
[4]: https://travis-ci.org/ethereum/web3.js/jobs/622125877#L757-L768

Fixes #3256

## Type of change

## Checklist:


gulpfile.js
```diff
@@ -188,7 +189,7 @@ packages.forEach(function(pckg, i) {
         stream = stream.pipe(source(pckg.fileName + '.js'))
             .pipe(streamify(babel({
                 compact: false,
-                presets: [[ '@babel/preset-env', { "useBuiltIns": "usage", "corejs": 2 } ]]
+                presets: [[ '@babel/preset-env', { "useBuiltIns": "entry", "corejs": 2 } ]]
```
cgewecke: Was just looking at the [babel docs][1] and honestly I don't know if useBuiltIns is the right option if the polyfill is being injected.... 

[1]: https://babeljs.io/docs/en/babel-preset-env#usebuiltins-entry

gulpfile.js
```diff
@@ -167,7 +167,8 @@ packages.forEach(function(pckg, i) {
         var stream = browserify(browserifyOptions)
             .require(pckg.src, {expose: pckg.expose})
             .require('bn.js', {expose: 'BN'}) // expose it to dapp developers
-            .add(pckg.src);
+            .add(pckg.src)
+            .add(require.resolve("@babel/polyfill"));
```
nivida: Babel would handle this for us with the ``preset-env`` and the ``corejs`` polyfills.

karma.conf.js
```diff
@@ -4,18 +4,30 @@ if (!process.env.TRAVIS){
     process.env.CHROME_BIN = require('puppeteer').executablePath();
 }
 
+function getTestFiles(){
```
nivida, line 9: Is it possible to move out those helper functions? and are you sure the minified file still does get ignored from browserify?

karma.conf.js
```diff
@@ -4,18 +4,30 @@ if (!process.env.TRAVIS){
...
+    switch (process.env.BROWSER_BUNDLE_TEST){
+        case 'publishedDist': return ["packages/web3/dist/web3.min.js", "test/e2e.minified.js"]
+        case 'gitRepoDist':   return ["dist/web3.min.js", "test/e2e.minified.js"]
+        default:              return ["test/**/e2e*.js"]
+    }
+}
+
+// Only loads preprocessor for the logic unit tests
+function getPreprocessors(){
+    if (!process.env.BROWSER_BUNDLE_TEST){
```
nivida, line 19: We use more and more such environment vars and I think it would be the time now or at least any time soon to document those environment variables and to define the naming rule behind it. It is already hard to follow the different execution paths depending on the defined environment variables. :)

karma.conf.js
```diff
@@ -4,18 +4,30 @@ if (!process.env.TRAVIS){
...
         ],
-        files: [
-            'test/**/e2e*.js'
-        ],
-        preprocessors: {
-            'test/**/e2e*.js': [ 'browserify' ]
-        },
+        files: getTestFiles(),
+        preprocessors: getPreprocessors(),
+        failOnEmptyTestSuite: false,
```
nivida: Why is ``failOnEmptyTestSuite`` defined as ``false`` required? 🤔

package.json
```diff
@@ -40,6 +40,8 @@
         "test:e2e:clients": "npm run test:e2e:ganache; npm run test:e2e:geth:insta; npm run test:e2e:geth:auto",
         "test:e2e:chrome": "./scripts/e2e.chrome.sh",
         "test:e2e:firefox": "./scripts/e2e.firefox.sh",
+        "test:e2e:min": "./scripts/e2e.min.sh",
+        "test:e2e:surge": "./scripts/e2e.surge.sh",
```
nivida: Detail but ``test:e2e:surge`` should probably be called ``test:e2e:cdn`` which is more descriptive.

scripts/e2e.min.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+# -------------------------------------------------------------------------------------
+# Runs basic min bundle loading check on chrome headless
+# -------------------------------------------------------------------------------------
+
+# Exit immediately on error
```
nivida, line 7: Which version of chrome will be used? It could be important to not use the latest version which probably does support some features older supported browsers don't have.

scripts/e2e.surge.sh
```diff
@@ -0,0 +1,29 @@
...
+echo "======================================================================"
+
+cp packages/web3/dist/web3.min.js scripts/html
+cd scripts
+(echo "" && exit && cat ) | npx surge html
+
+# Might be running locally in development
+rm html/web3.min.js
+
+# Script doesn't always work...
```
nivida: Is it possible to fix the described error from the PR desc? could you add some more details about it? 🤔

test/e2e.minified.js
```diff
@@ -0,0 +1,9 @@
+// Simple check to make sure window.web3 is defined for the karma min loading tests
+describe('window.web3 is defined', function () {
+
+  // Referencing `process` throws w/out browserify.
+  // We want to be skipped during the browserified units tests here.
+  try { if (process) return } catch (e) {}
+
+  if (!window.Web3) throw new Error('"window.Web3" was not defined');
```
nivida: I think it would be great if we extend this test case with expectations for the values we do request from infura. The ``innerHtml`` property of those should at least not be empty.

test/e2e.minified.js
```diff
@@ -0,0 +1,9 @@
+// Simple check to make sure window.web3 is defined for the karma min loading tests
+describe('window.web3 is defined', function () {
```
nivida:
```suggestion
describe('window.Web3 is defined', function () {
```

nivida (changes requested):
> The fix is to use "entry" and add @babel/polyfill to gulp's browserify step. Increases the minified bundle's size about 150 kb (~10%). This wouldn't affect anyone whose packing things themselves though.

``entry`` is not only a fix it does also improve the way babel does add those imports and this should save some lines. An increase of 10% is in my view of point to much and should already show us that we do add to much stuff.
I've used the babel runtime and transform runtime plugin by side of the ``preset-env`` to be sure the required regenerator runtime is existing for each bundle and also to have just one absolute ``@babel/runtime`` package.

Could you elaborate on why exactly the ``usage`` or ``entry`` ``useBuiltIns`` options do make the difference as mentioned in the PR desc? As I know and only if babel does transpile our packages correctly should this not affect us in this way. The only difference is where exactly those imports are located (globally in the bundle or for each file).

> Was just looking at the babel docs and honestly I don't know if useBuiltIns is the right option if the polyfill is being injected....

Yes, it isn't the correct thing. The goal would be to only use the ``useBuiltIns`` options and the corejs polyfills. The ``@babel/polyfill`` package you have added is already marked as deprecated by babel and I think we shouldn't rely on deprecated packages for this fix. 

["Since @babel/polyfill was deprecated in 7.4.0, we recommend directly adding core-js and setting the version via the corejs option."](https://babeljs.io/docs/en/babel-preset-env#usebuiltins) - babel docs

cgewecke:
@nivida 
> Could you elaborate on why exactly the usage or entry useBuiltIns options do make the difference as mentioned in the PR desc?

The clearest description I could find for this was [in babel polyfill docs][1]. But before I also tried many combinations of runtime plugins in the preset env block and could not get both tests passing. Maybe there's a way with `usage` and some kind of module option setting? Newer babel? 

> I've used the babel runtime and transform runtime plugin by side of the preset-env to be sure the required regenerator runtime is existing for each bundle and also to have just one absolute @babel/runtime package.

Please feel free to experiment with this! If you can get rid of the polyfill that would be incredible. The only thing I'd ask is that the tests remain as they are - they trigger the two errors we've seen. 

> An increase of 10% is in my view of point to much and should already show us that we do add to much stuff.

Agree, this is not a fix. It's a patch to stop something crashing. Over desktop the difference in loading the min over CDN with another 10% [should be +25ms][2] and it will not affect anyone bundling by themselves either.  

Lastly - we agree the way forward is with rollup. If gulp will be ripped out within 1 month and the tests here are correct, then getting the babel setting perfect might be less important.

[2]: https://github.com/ethereum/web3.js/issues/1178#issuecomment-552072346
[1]: https://babeljs.io/docs/en/babel-polyfill#usage-in-node-browserify-webpack

test/e2e.minified.js
cgewecke: This test doesn't have an html associated with it. It's only the injection of the test into the karma runner with the min. There's no external browserification so if I put a `require('path')` at the top of the file, I would see
```
Uncaught ReferenceError: require is not defined
  at test/e2e.minified.js:1:1
```
Likewise, if I set the preset env to `usage`, I will also see `require is not defined`,  but coming from the bundle.

That's all the test does - it only replicates the crash from #3256.

The E2E with the html requires visual inspection because it happens out on a published website.

scripts/e2e.min.sh
cgewecke: Yes this is possible, I'm not sure. We use whatever version is installed in Travis. I think you asked me to do this with #3148 because you didn't like the chromium download wait.

karma.conf.js
cgewecke: No! Good catch.

karma.conf.js
cgewecke: Yes, I wonder where should we define them and how improve their clarity though. I will have to think about this a bit - probably a separate PR because we have 5 or 6 of these now.

karma.conf.js
cgewecke: These are only used here. I think moving them out would reduce a clarity a bit / make the logic harder to follow. What is your reasoning for putting outside? And where?  

> are you sure the minified file still does get ignored from browserify?

Yes! You should check this yourself, its the whole point of the test.

gulpfile.js
cgewecke: Ok! I believe you - I'm completely confused by this and tried different combos for a few hours. Smh. If you can configure the presets to work that would be awesome.

karma.conf.js
nivida: > What is your reasoning for putting outside? And where?

To have all helpers in one place which do use the environment vars this would increase the clarity in this case. But yes it does otherwise decrease the readability in the karma file because you have to follow the execution path over multiple files. 
The conclusion from my side is: Let us keep it as it is now and if the complexity with the environment variables would increase over time can we look around for another more "centralized" structure.

nivida: > Yes! You should check this yourself, its the whole point of the test.

I will run it on the finish. :)

scripts/e2e.surge.sh
cgewecke: Added in https://github.com/ethereum/web3.js/pull/3261/commits/331b01176fc0c4f01bd066f172d0b303f84b8ab9

package.json
cgewecke: Changed to cdn in 2d43bb1489e4bf26bc8979d6db5cc37d5acce0f4

karma.conf.js
nivida: I only know [convict](https://github.com/mozilla/node-convict) which could improve the situation here (I've never used it until now).

karma.conf.js
cgewecke: > if the complexity with the environment variables would increase over time can we look around for another more "centralized" structure.

Yes, definitely.

scripts/e2e.min.sh
nivida: Yes, I thought our ``e2e`` tests could use the locally installed chrome bin if existing on Travis or from our PC's to speed up the installation process of web3.js. I think this is something we could also address in another separate PR because it would affect other parts of our QA as well. 

Let us discuss this closer on the next call or in an issue here on GitHub.

karma.conf.js
cgewecke: Removed in 331b01176fc0c4f01bd066f172d0b303f84b8ab9

test/e2e.minified.js
nivida: Oh! Sorry than I was wrong :) 

> The E2E with the html requires visual inspection because it happens out on a published website.

I know from other ``e2e`` test suites that they do compare screenshots for such cases would this be possible or is this too hard to achieve?

scripts/e2e.surge.sh
nivida: Just curious about, does this mean the ``surge`` error which appears sometimes isn't fixable? :)

test/e2e.minified.js
cgewecke: I think it is possible and it's very common in websites. 

The thing I like about the surge test is that it's simple and real. You make a bundle and publish it to a site. You look at it and it "works". There's no testing framework.

scripts/e2e.surge.sh
cgewecke: I think it's something about surge - below is from a  [surge guide](https://sunlightmedia.org/using-surge-for-deploying-static-sites/) and the error described is what I've seen, using random names. But if you want to make an account for yourself since you're the Web3 maintainer, it would probably work to publish to same site each time. 

It's [surge.sh](https://surge.sh/help/) and free.

> Surge will automatically provide a domain name using random words. You can also create any custom domain if it is not already taken. Enter your chosen domain name, then hit enter.

> (Note: If the custom domain name you input is already taken, you’ll see an error message.

> Aborted - you do not have permission to publish to [<custom-domain.surge.sh>]

gulpfile.js
nivida: The issue we have here is that ``browserify`` does handle the ``require`` statement accordingly but the babel transformation after does add new require statements that aren't "browserified". I will try to change the order and will remove the babel polyfill dependency.

nivida: I've fixed it with using of ``babelify`` and ``browserify.transform``. The bundle size is now at  ``976.96KB``.

test/e2e.minified.js
nivida: I wrote that down in my notes and we can discuss further advanced in-browser tests on the next team call. A small demo ERC-20 DApp as I was creating for my presentation once is probably something we could add as an in-browser e2e test.

gulpfile.js
alcuadrado: Sam is right, babel will add the `require` needed for the polyfills if `entry` is used. And also about the order in which babel and browserify need to be run :)

alcuadrado: Have you tried removing it and comparing the bundle sizes? I suspect the difference comes from this line.

I believe the most promising config is using `usage`, setting the `corejs` option explicitly, installing `core-js`, and removing `@babel/polyfill`.

nivida: Still not fixed the regeneratorRuntime error appeared again.. I will now add the transform runtime plugin from babel to enforce the transformation.

nivida: Finally, I was able to fix it. I've first tried to use the ``@babel/plugin-transform-runtime`` but sadly brought this up a bunch of other errors. This is why I've decided to add the ``regenerator-runtime`` package directly in the gulp pipeline which does work as expected.

nivida:
@cgewecke As already commented did I fix the ``require`` and ``regeneratorRuntime`` error with adding of the ``regenerator-runtime`` package in our gulp build pipeline and using of ``babelify``. What I'm wondering is why the ``preset-env`` from babel is transforming the ``async`` call but doesn't add the required ``regenerator-runtime``. Is this a bug from babel or is this thought to be the case? 🤔 

Edit:
Using of the ``regenerator-runtime`` package directly instead of the ``@babel/polyfill`` package does increase the minified bundle size by 20Kb instead of 150Kb.

scripts/e2e.surge.sh
cgewecke: @nivida I've just pushed a change to make this always publish to sudden-playground.surge.sh since that name works. If you need to change this, it super easy - there are a couple tokens to set in CI.

cgewecke:
@nivida Ah great!! Thanks so much for figuring that out.

package.json
```diff
@@ -14,9 +14,9 @@
     },
     "bundlesize": [
         {
-          "path": "./packages/web3/dist/web3.min.js",
-          "maxSize": "1.5 mB",
-          "compression": "none"
+            "path": "./packages/web3/dist/web3.min.js",
+            "maxSize": "1.5 mB",
+            "compression": "none"
```
cgewecke: This formatting change is correct - believe it's how npm format the package.

cgewecke:
@nivida  Couple additional notes:

+ Rebased against 1.x
+ e2e.cdn.sh now publishes to `sudden-playground.surge.sh` each time. It shouldn't fail anymore and you can change that to a site you control if your prefer via tokens in CI

nivida (approved):

alcuadrado (approved):
I re-reviewed this, and it looks really good now :)

holgerd77:
Phew, what a detective work. Great!! 😄

Issue #3186, opened 2019-11-07
Unpin every dependency


alcuadrado:
This project has lots of pinned dependencies (i. e. exact version defined in the `dependencies` entries). This is not good in a library and should be removed. 

Here's why I'm convinced we should do that.

## Why is it bad?

1. It leads to dependencies duplication, which can heavily impact bundle sizes when using web3 on the web.
2. It prevents users to get new fixes.
3. It makes it super complicated for us to push fixes to multiple versions of web3. See #3181 for an example.

## What about all the people pinning their dependencies?

Pinning dependencies is recommended practices in many cases, and for a reason. It makes your builds (more) reproducible, and prevents malicious actors from pushing exploits into your application. 

The keyword here is "application". When building an app, and not a library, you are responsible for bundling it and need more control over your dependencies. You can do whatever you want, as nobody will use it as a dependency. Of course, this makes you are responsible for building it, deploying fixes when appropriate, and making it secure. This things are **hard**, but that's how the npm ecosystem is designed.

Pinning dependencies in a library is overstepping, removing the user from that needed flexibility, and leading to the problems I mentioned.

nivida:
@alcuadrado I totally agree with you. Only Ethers has to be kept pinned until we have updated our AbiCoder test cases etc. This because of breaking changes in the AbiCoder of ethers (introduced with version ``4.0.0-beta.4``).

alcuadrado:
A side effect of unpinning dependencies in a library is that using `package-lock.json`s in the CI gives you a different experience than your users'. `npm` and `yarn` only care about lock files when you install the project with an `npm/yarn install`, not when you install it as a dependency.

We don't use `package-lock.json`s in EthereumJS nor in Buidler.

cgewecke:
> A side effect of unpinning dependencies in a library is that using package-lock.jsons in the CI gives you a different experience than your users'. npm and yarn only care about lock files when you install the project with an npm/yarn install, not when you install it as a dependency.

@alcuadrado Out of curiosity, what is your view of [this issue][1] at Truffle? There, Micah Zoltu is pressing to pin deps in library-like packages for reproducibility reasons. 

Is a fair summary that by unpinning, Web3 would say: 'this lib assumes its primary users will bundle it and deploy as a dapp' vs. 'this lib assumes its primary users will include it as a dep in an npm module'? 

[1]: https://github.com/trufflesuite/truffle/issues/2555

alcuadrado:
[...]

PR #3262, opened 2019-12-08
Standardize bn.js dep across monorepo

cgewecke:
## Description

**NB: This PR targets the buildsize checker PR #3260, prerequisite for merging here**

(Tries to) reduce the bundle size to provide a baseline measurement for how small the gulp/browserify build could get.  

+ Internalizes the two silentcicero dependencies in web3-utils so that BN.js can be standardized to 4.11.8 across the entire project. The deps seem [abandoned][4].
  + commit cce191fa48fa1a120968f16b4fe9ffbaf028ab85 
  + build size: 1.14 MB :/
  + improvement: none (See [build][1]) 

[1]: https://travis-ci.org/ethereum/web3.js/jobs/622209876#L1148
[2]: https://github.com/browserify/common-shakeify
[3]: https://github.com/browserify/common-shakeify#onexportdeletefilename-exportname
[4]: https://github.com/ethjs/ethjs-unit

## Type of change

- [x] Dependencies

## Checklist:

### DIFF
--- a/packages/web3-utils/package.json
+++ b/packages/web3-utils/package.json
@@ -16,9 +16,8 @@
         "bn.js": "4.11.8",
         "eth-lib": "0.2.7",
         "ethereum-bloom-filters": "^1.0.6",
-        "ethjs-unit": "0.1.6",
-        "number-to-bn": "1.7.0",
         "randombytes": "^2.1.0",
+        "strip-hex-prefix": "1.0.0",
         "underscore": "1.9.1",
         "utf8": "3.0.0"
     },
--- a/packages/web3-utils/src/ethjsUnit.js
+++ b/packages/web3-utils/src/ethjsUnit.js
@@ -1,0 +1,190 @@
+// This file was internalized from silentcicero/ethjs-unit
+// in order to reduce the bundle size. Package shared a
+// bn.js dependency which wasn't being updated resulting
+// in duplicate versions of that dep included in webpacks
+
+/*
+The MIT License (MIT)
+
+Copyright (c) 2016 Nick Dodson
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.*/
+
+/*
+Primary Attribution
+Richard Moore <ricmoo@me.com>
+https://github.com/ethers-io
+Note, Richard is a god of ether gods. Follow and respect him, and use Ethers.io!
+*/
+
+const BN = require('bn.js');
+const numberToBN = require('./numberToBN');
+
+const zero = new BN(0);
+const negative1 = new BN(-1);
+
+// complete ethereum unit map
+const unitMap = {
+  'noether':      '0', // eslint-disable-line
+  'wei':          '1', // eslint-disable-line
+  'kwei':         '1000', // eslint-disable-line
+  'Kwei':         '1000', // eslint-disable-line
+  'babbage':      '1000', // eslint-disable-line
+  'femtoether':   '1000', // eslint-disable-line
+  'mwei':         '1000000', // eslint-disable-line
+  'Mwei':         '1000000', // eslint-disable-line
+  'lovelace':     '1000000', // eslint-disable-line
+  'picoether':    '1000000', // eslint-disable-line
+  'gwei':         '1000000000', // eslint-disable-line
+  'Gwei':         '1000000000', // eslint-disable-line
+  'shannon':      '1000000000', // eslint-disable-line
+  'nanoether':    '1000000000', // eslint-disable-line
+  'nano':         '1000000000', // eslint-disable-line
+  'szabo':        '1000000000000', // eslint-disable-line
+  'microether':   '1000000000000', // eslint-disable-line
+  'micro':        '1000000000000', // eslint-disable-line
+  'finney':       '1000000000000000', // eslint-disable-line
+  'milliether':   '1000000000000000', // eslint-disable-line
+  'milli':        '1000000000000000', // eslint-disable-line
+  'ether':        '1000000000000000000', // eslint-disable-line
+  'kether':       '1000000000000000000000', // eslint-disable-line
+  'grand':        '1000000000000000000000', // eslint-disable-line
+  'mether':       '1000000000000000000000000', // eslint-disable-line
+  'gether':       '1000000000000000000000000000', // eslint-disable-line
+  'tether':       '1000000000000000000000000000000', // eslint-disable-line
+};
+
+/**
+ * Returns value of unit in Wei
+ *
+ * @method getValueOfUnit
+ * @param {String} unit the unit to convert to, default ether
+ * @returns {BigNumber} value of the unit (in Wei)
+ * @throws error if the unit is not correct:w
+ */
+function getValueOfUnit(unitInput) {
+  const unit = unitInput ? unitInput.toLowerCase() : 'ether';
+  var unitValue = unitMap[unit]; // eslint-disable-line
+
+  if (typeof unitValue !== 'string') {
+    throw new Error(`[ethjs-unit] the unit provided ${unitInput} doesn't exists, please use the one of the following units ${JSON.stringify(unitMap, null, 2)}`);
+  }
+
+  return new BN(unitValue, 10);
+}
+
+function numberToString(arg) {
+  if (typeof arg === 'string') {
+    if (!arg.match(/^-?[0-9.]+$/)) {
+      throw new Error(`while converting number to string, invalid number value '${arg}', should be a number matching (^-?[0-9.]+).`);
+    }
+    return arg;
+  } else if (typeof arg === 'number') {
+    return String(arg);
+  } else if (typeof arg === 'object' && arg.toString && (arg.toTwos || arg.dividedToIntegerBy)) {
+    if (arg.toPrecision) {
+      return String(arg.toPrecision());
+    } else { // eslint-disable-line
+      return arg.toString(10);
+    }
+  }
+  throw new Error(`while converting number to string, invalid number value '${arg}' type ${typeof arg}.`);
+}
+
+function fromWei(weiInput, unit, optionsInput) {
+  var wei = numberToBN(weiInput); // eslint-disable-line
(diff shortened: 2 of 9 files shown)


cgewecke:
@nivida My question for the initial review is: 

Do you think this is worth doing? e.g Could remove the `shakefiy` and get CI passing. The bn.js would be deduped. But there'd be no measurable build size improvement - is it still a good idea?

It might help people using webpack.

nivida:
> Do you think this is worth doing? e.g Could remove the shakefiy and get CI passing. The bn.js would be deduped. But there'd be no measurable build size improvement - is it still a good idea?

I've achieved the same with adding of custom resolver rules (the ``dedupe`` parameter of the ``rollup`` config function). As the build size reduction does show are there still some duplications of ``bn.js`` around (``bn.js`` ~80kb; ``webpack`` bundle analyzer should show them). Complete deduplication of ``js-sha3``, ``bn.js``, and ``elliptic`` was bringing up the bundle size improvement. I think if we use ``shakeify`` and update the ``cjs`` imports to a common shakable way in all packages will it probably improve the situation as well.

I think we better keep the code as it is and try to add custom resolver rules to the ``browserify`` settings to dedupe the ``bn.js``, ``js-sha3``, and ``elliptic`` dependency. By side of the bundle size improvement is it probably anyways better to manage the included version of those listed dependencies on our own (security risky dependencies).

cgewecke:
@nivida 

> I think we better keep the code as it is and try to add custom resolver rules to the browserify settings to dedupe the bn.js, js-sha3, and elliptic dependency. By side of the bundle size improvement is it probably anyways better to manage the included version of those listed dependencies on our own (security risky dependencies).

For clarification, you think these changes should stay in and maybe I should remove shakeify from this PR and the optimizations should go in separately?

nivida:
@cgewecke 
> For clarification, you think these changes should stay in and maybe I should remove shakeify from this PR and the optimizations should go in separately?

I think it will be possible to achieve the same or even more with custom resolver rules instead of using the shakeify package and copying of dependencies over. ~~Such resolver rules are also way more future proof than going through the deps and copying them over if required or checking for duplications on the addition of a dependency.~~

alcuadrado:
Can you elaborate on why custom resolution rules are needed? I've never seen that case before, so I can't follow it. Is there a shared limitation by browserify and rollup that makes them needed with both tools?

nivida:
> Can you elaborate on why custom resolution rules are needed? I've never seen that case before, so I can't follow it. 

It is to deduplicate those dependencies over the whole lib which does, in the end, lower the minified bundle size. 

> Is there a shared limitation by browserify and rollup that makes them needed with both tools?

Rollup and also Browserify does not remove all duplicated dependencies as it should during tree shaking and that’s why those resolver rules are required. The used ``bn.js`` versions are all on the same minor version which means Rollup and Browserify should both be able to de-duplicate it ~~but sadly they can't handle it correctly.~~

alcuadrado:
> Rollup and also Browserify does not remove all duplicated dependencies as it should during tree shaking and that’s why those resolver rules are required. The used `bn.js` versions are all on the same minor version which means Rollup and Browserify should both be able to de-duplicate it but sadly they can't handle it correctly.

I'm inclined to think that the fault is on our side and that it's just a misconfiguration or a poor dependencies choice. I'd need a link to their docs explaining this limitation, or to a bug report describing this behavior, to be convinced. The reason is that otherwise, custom resolution rules would be super common, and they aren't.

I'll open an issue to discuss this, as properly solving this will probably need lots of planning and discussing.

nivida:
@alcuadrado 
> or a poor dependencies choice

Yes, this is the case and the reason why @cgewecke did copy the ``ethjs`` and ``number-to-bn`` dependency over to web3.js. We can keep those copied dependencies, we use the utility methods from ethers, or we define global resolver rules for our minified bundle. But I think the most future proof and also the most stable way would be to use the ethers utilities because those are small, do use bn.js, and are managed by Richard from the EF-JS-Team. With further dependency clean up PRs is it probably also possible to achieve the bundle size from PR #3160.

> I'd need a link to their docs explaining this limitation

If I do remember correctly was rollup able to de-duplicate all ``bn.js`` dependencies which are in our ``package.json`` files but it wasn't able to de-duplicate it if a sub-package dependency has a ``bn.js`` dependency listed in their ``package.json``. ``webpack`` does have the same problem as ``rollup`` and this was the reason why my [starter test env](https://github.com/nivida/web3-webpack-starter/blob/master/webpack.config.js) has some resolver rules defined.

Edit: 
I've noticed the de-duplication issue we have is because some dependencies have ``bn.js`` and other dependencies defined as fixed version and it can't de-duplicate them also if only the patch level differs.

alcuadrado:
Update on the bn.js deduplication: 

I further analyzed the dependencies tree while preparing #3267. Rollup and browserify behavior (i.e. not deduplicating bn.js) is correct. Forcing it to deduplicate is a huge risk, as nothing guarantees that it will work. 

There are dependencies that require incompatible versions of bn.js, and that should be respected by any bundler, and this is done by bundling two or more versions.

For example, `web3-utils` require the exact version `4.11.8`, and also `number-to-bn` which in turn requires `4.11.6`.

I think the path followed by @cgewecke in this PR is a movement in the right direction.

Note that not being able to deduplicate dependencies is what I've been warning about for some time, for example in #3186. Because of pinning dependencies, the situation is already complicated for us and is at least as complicated, if not worse, for our users.

packages/web3-utils/src/ethjsUnit.js
```diff
@@ -0,0 +1,190 @@
...
+THE SOFTWARE.*/
+
+/*
+Primary Attribution
+Richard Moore <ricmoo@me.com>
+https://github.com/ethers-io
+Note, Richard is a god of ether gods. Follow and respect him, and use Ethers.io!
+*/
+
+const BN = require('bn.js');
```
alcuadrado, line 36: Isn't ethers a dependency? Why is this duplicated?

packages/web3-utils/package.json
```diff
@@ -16,9 +16,8 @@
         "bn.js": "4.11.8",
         "eth-lib": "0.2.7",
         "ethereum-bloom-filters": "^1.0.6",
-        "ethjs-unit": "0.1.6",
-        "number-to-bn": "1.7.0",
         "randombytes": "^2.1.0",
+        "strip-hex-prefix": "1.0.0",
```
alcuadrado, line 20: This is probably causing a dependency duplication in another packages, as IIRC `strip-hex-prefix` is a dependency of `ethereumjs-util`, which is inderectly used by web3.

packages/web3-utils/src/numberToBN.js
```diff
@@ -0,0 +1,67 @@
...
+var BN = require('bn.js');
+var stripHexPrefix = require('strip-hex-prefix');
+
+/**
+ * Returns a BN object, converts a number value to a BN
+ * @param {String|Number|Object} `arg` input a string number, hex string number, number, BigNumber or BN object
+ * @return {Object} `output` BN object of the number
+ * @throws if the argument is not an array, object that isn't a bignumber, not a string number or number
+ */
+module.exports = function numberToBN(arg) {
```
alcuadrado, line 39: Is this function exposed in web3's API? If not, I'd remove it. It's a a huge mess.

nivida:
> Forcing it to deduplicate is a huge risk, as nothing guarantees that it will work.

Because only the patch level does differ did I thought is there no risk to apply such custom resolver rules but thanks for the closer check!

> I think the path followed by @cgewecke in this PR is a movement in the right direction.

Yes, it is definitely the right direction but I would recommend replacing those dependencies with equivalent utility functions from ethers. This would be aligned with the overall EF-JS-Team goal and would also improve the situation for us here :)

packages/web3-utils/src/numberToBN.js
nivida: This will be used internally for some utility functions but could probably get replaced with the ethers BigNumber wrapper object.

packages/web3-utils/package.json
cgewecke: Looked into further - the weight of this package (which uses it's own tiny dependency, [is-hex-prefix][3]) is < [20 lines][1], un-minified. It has a single published [version][2] from 3 years ago and is maintained by silentcicero.

It looks pinned to 1.0.0 all the way through the tree:
+ ethjs-util pins both deps [here][4].
+ ethereumjs-utils pins ethjs-util [here][5]

This stuff is small and could easily be internalized somewhere although the benefit might be hard to measure. 

SilentCicero continues to actively develop eth stuff  but is working on [optimistic rollup sidechains][7]...

That's everything I found out :) and am persuadable on changing this, what's your assessment?

[1]: https://github.com/SilentCicero/strip-hex-prefix/blob/94058ad57053b4be0c7946b1abcbef0ae6c34cc8/src/index.js#L8-L14
[2]: https://www.npmjs.com/package/strip-hex-prefix
[3]: https://github.com/SilentCicero/is-hex-prefixed/blob/master/src/index.js#L7-L12
[4]: https://github.com/ethjs/ethjs-util/blob/master/package.json#L106-L109
[5]: https://github.com/ethereumjs/ethereumjs-util/blob/93544654c1b38775c2e87c8e8d6e66bf4ef7b8b3/package.json#L94
[7]: https://medium.com/@fuellabs/fuel-scaling-stablecoin-payments-10x-today-50x-in-the-near-future-961e3113b39b

packages/web3-utils/src/numberToBN.js
cgewecke: It internal to [toBN][1] and [fromWei][2] both of which are heavily used.  

Personally I'd be more comfortable removing this in a different PR that rationalized use of Ethers utils and where we have tests in place to make sure we don't accidentally introduce unexpected behavior. Also where we double-check the Ethers issues about differences people see in how Ethers utils behave vs. Web3 since the code here is 3 years old. 

PR bumps BN across some patch versions - might be worth isolating.

[1]: https://github.com/ethereum/web3.js/blob/6ca8bba4e5cb9f76672b326541ba6c620af40e70/packages/web3-utils/src/utils.js#L61-L67
[2]: https://github.com/ethereum/web3.js/blob/aeb1f467923671f933d5ab8a4bbd688d6a792bf9/packages/web3-utils/src/ethjsUnit.js#L111

packages/web3-utils/package.json
cgewecke: [EDITED] On a fresh install of ethereumjs-util in an empty package the `npm list bn.js` is:
```
└─┬ ethereumjs-util@6.2.0
  ├── bn.js@4.11.8 
  ├─┬ rlp@2.2.4
  │ └── bn.js@4.11.8  deduped
  └─┬ secp256k1@3.7.1
    ├── bn.js@4.11.8  deduped
    └─┬ elliptic@6.5.2
      └── bn.js@4.11.8  deduped
```
It looks like we should get deduped...

packages/web3-utils/src/ethjsUnit.js
cgewecke: That's a good question...discussed a bit below. This code is a few years old now and swapping it for ethers will take a little additional research.

cgewecke:
> I think the path followed by @cgewecke in this PR is a movement in the right direction.

Is this still right? The scope of the PR is to reduce the bundle size a bit but there's no measurable benefit here. 

In turn, looking over this stuff makes it clear that we should depend more on the Ethers implementations but doing that might require more care (see [comment][1]).

@nividia are you confident this change will pay dividends if the bundling is fixed?
 
[1]: https://github.com/ethereum/web3.js/pull/3262#discussion_r357302938

nivida:
> Is this still right? The scope of the PR is to reduce the bundle size a bit but there's no measurable benefit here.

The scope of this PR didn't change in my point of view it is still flatting of the dependency tree and improving the bundle size. This because flatting/deduping of the dependency tree should improve the bundle size of bundles created with ``parcel`` or ``webpack`` or we do not actually dedupe/flatten it.

> @nividia are you confident this change will pay dividends if the bundling is fixed?

If no measurable benefit is existing will this also not improve anything with the fixed build pipeline.

Edit:
> or we do not actually dedupe/flatten it.

It could also be the case that the bundler is deduplicating it (parcel is probably doing this) and we do not notice any change.

nivida:
@cgewecke I think we should held a call related to the bundle size improvements for splitting this up in smaller tasks with clear goals etc. WDYT?

cgewecke:
@nivida Yes definitely.

nivida:
Closed because we will plan/discuss the bundle size improvements first together.

PR #3263, opened 2019-12-08
Build is smaller on second install?

cgewecke:
## Description
**NB: PR targets #3262 which standardizes the bn.js version across the project**

PR reduces the bundlesize ~20%, but in a way that is super weird. Suggests something is wrong with the way browserify de-duplicates things in a lerna monorepo. 

TLDR; if you install the project and *then* go to packages/web3 (the bundle entry point) and install the most widely duplicated deps there, there's a big bundle reduction. 

+ [with common-shakeify][1]:  **885kb**  (Broken build)
+ without common-shakeify:  **906kb**

Also tried browserify's `preserveSymlinks` option to address this and was quickly OOM within 4GB - @nivida was this the same thing you were seeing when you looked at this in October?

**[EDITED notes...]**
+ ~~treat the published min as a convenience and not worry about it's size too much (although it would be great to get something on the CDN that was small)~~
+ Actually, this PR and #3262 show rollup is the best approach. Browserify is not bundling correctly.
+ We could also try to measure size by :
  + running virtual npm publish, 
  + create a tiny fake project with installed web3, 
  + webpack with minnification
  + measure size of webpack build.

[1]: https://travis-ci.org/ethereum/web3.js/jobs/622370838#L1171

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -17,7 +17,7 @@ var exorcist = require('exorcist');
 var bower = require('bower');
 var streamify = require('gulp-streamify');
 var replace = require('gulp-replace');
-var shakeify = require('common-shakeify');
+//var shakeify = require('common-shakeify');
 var exec = require('child_process').exec;
 
 var DEST = path.join(__dirname, 'dist/');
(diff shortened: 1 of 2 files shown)


cgewecke:
My question for the initial view is: what is going on here?!

nivida:
> @nivida was this the same thing you were seeing when you looked at this in October?

I had the same experience when I started to improve the build pipeline. The OOM issue does especially occur if you try to create bundles for each package web3.js does have. 
But I was able to run the build pipeline (it still sometimes failed) if I executed ``npm dedupe`` in the root of the project. This does remove all duplications and will install all dependencies in the root ``node_modules``. 

> My question for the initial view is: what is going on here?!

I was asking me the same and tried several ways until I've decided to back-port and extend the rollup setup from 2.x.

cgewecke:
Ok thanks for answering my question 😄 Closing...

PR #3265, opened 2019-12-10
Provider improvements/geth disconnect test

cgewecke:
## Description

Mix of good and bad news, I guess. 

**Good**
+ 3 (real) tests pass.

**Less good**
+ The simple test case for `eth.subscribe` when the connection is already closed fails. There *is* a "connection not open on send()" error caught in Subscription.subscribe but it doesn't bubble up. This emitter seems disconnected from anything:
https://github.com/ethereum/web3.js/blob/8804708330c25de72c13dcb6182729e52693ba06/packages/web3-core-subscriptions/src/subscription.js#L300
  In the ganache tests, it looks like this error case is handled by the WS module [close handler][1], the client behavior is different somehow.

[1]: https://github.com/ethereum/web3.js/blob/provider-related-improvements/packages/web3-providers-ws/src/index.js#L199-L226

+ The passing tests error when run in sequence - will have to debug that further.

+ The last, more complex autoreconnect case fails.

## Checklist:

### DIFF
--- a/test/e2e.subscribe.geth.js
+++ b/test/e2e.subscribe.geth.js
@@ -1,0 +1,179 @@
+const assert = require('assert');
+const utils = require('./helpers/test.utils');
+const geth = require('./helpers/test.geth.api');
+const Web3 = utils.getWeb3();
+const util = require('util')
+
+// NB: there are parallel tests for ganache at `eth.subscribe.ganache` in the main unit tests
+describe('subscription connect/reconnect (geth) [ @E2E ]', function() {
+    this.timeout(15000);
+
+    // Geth instamine over WS is unreliable
+    if (!process.env.GETH_AUTOMINE) return;
+
+    let web3;
+    let accounts;
+
+    // Setup a seperate RPC connection for the admin startWS calls
+    // Once we close the websocket with the 'test Web3 instance',
+    // it cannot be used to make more admin calls.
+    before(function(){
+        geth.connectHTTP(Web3, 8545);
+    });
+
+    beforeEach(async function() {
+        web3 = new Web3('ws://localhost:8546');
+        accounts = await web3.eth.getAccounts();
+    });
+
+    afterEach(async function() {
+        await geth.admin.wsStart();
+    });
+
+    // Usually passes, unstable when run with other suites.
+    // Test closes the connection and is notified of the drop.
+    it.skip('(1) admin.close closes the connection (baseline)', async function(){
+        try {
+            await geth.admin.wsStop(web3);
+            assert(fail);
+        } catch (err) {
+            assert(err.message.includes('CONNECTION ERROR'));
+            assert(err.message.includes('1006'));
+        }
+    });
+
+    // Usually passes, unstable when run with other suites.
+    it.skip('(2) eth calls fail after the connection has been dropped (baseline)', async function(){
+        await geth.admin.wsStopQuietly(web3)
+
+        try {
+            const val = await web3.eth.getBlockNumber();
+            assert.fail();
+        } catch (err){
+            assert(err.message.includes('connection not open on send()'))
+        }
+    });
+
+
+    it('(3) subscribes (baseline)', function() {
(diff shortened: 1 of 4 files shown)


nivida:
@cgewecke Thanks for testing it! 

> The simple test case for eth.subscribe when the connection is already closed fails. There is a "connection not open on send()" error caught in Subscription.subscribe but it doesn't bubble up. This emitter seems disconnected from anything:

I've checked this closer and the issue we have here is a timing issue. The ``error`` event does get dispatched before the ``Subscription`` object does get returned in the ``Subscription.subscribe`` method. We would have to add a delay in the ``WebsocketProvider`` or we find another cleaner solution.

nivida:
> In the ganache tests, it looks like this error case is handled by the WS module close handler, the client behavior is different somehow.

The ganache test does cover another case. If the request was sent to the node but the connection got closed before the expected response of the node got returned.

Edit:
Btw.: Another finding I've made is the different versions of ``EventEmitter`` we currently do use in the subscriptions and ws package of web3. We can update this dependency in the subscriptions package without any problem (I've checked the release announcements of the ``eventemitter3`` package)

test/e2e.subscribe.geth.js
```diff
@@ -0,0 +1,158 @@
...
+                    assert(err.message.includes('CONNECTION ERROR'));
+                    resolve();
+                });
+        });
+    });
+
+    // This test passes on it's own, but not when run in sequence with the others.
+    // Might be easier to debug when test #3 passes
+    it.skip('errors when subscr is already running & connection gets closed', function() {
+        let stage = 0; // Required to not trigger server.close a second time
```
nivida: I know it is a draft PR but this could get removed with using of the ``once`` method instead of the ``on`` method of the ``EventEmitter`` class the ``Subscription`` class is extending from.

cgewecke: Actually I'm seeing things that have no listeners attached fail sometimes, especially in longer test runs. Idk if the problems here are geth, the tests, or the code, or what.... 

I will keep this PR in draft and try it against 1.x for tests 1-5 (which should already "work" on web3, maybe). It might be good to have as many tests on this topic for 1.x before the provider improvements go in so that there's a baseline... we'd know if the tests are bad or should work, whatever. 

Ganache works well but geth seems super race-conditiony. It may be the tests.

cgewecke:
> The ganache test does cover another case. If the request was sent to the node but the connection got closed before the expected response of the node got returned.

@nivida Could you explain this further to me - which request is being sent? The ganache test I'm thinking of looks like this 

https://github.com/ethereum/web3.js/blob/c9db70fddde3937452ee6c33dba64cab820074ec/test/eth.subscribe.ganache.js#L38-L49

And I've modeled that with geth like this:

https://github.com/ethereum/web3.js/blob/c9db70fddde3937452ee6c33dba64cab820074ec/test/e2e.subscribe.geth.js#L70-L84

Do you feel like these are testing something different?

test/e2e.subscribe.geth.js
nivida: Great I will do some tests as well! It’s probably worth to think about adding parity to our test suite as well because of the different behaviour we have noticed during writing tests for geth. 👌

nivida:
> Could you explain this further to me - which request is being sent? The ganache test I'm thinking of looks like this

Those tests do actually look similar but there is a small difference between those. In the ganache case is the connection closed after ``eth_subscribe`` got sent and before the response from the node got returned and in the Geth case is the connection already closed before the ``eth_subscribe`` JSON-RPC request is sent.

nivida:
After many tries to implement well-working test cases for Geth did @cgewecke and I agreed to extend the existing ganache tests and to do additional manual tests with Geth and parity.
The problem we have is that Geth doesn't close the WS connection as expected on calling of ``admin_stopWS`` and Parity doesn't provide such an API.

The mentioned manual test with Parity and Geth will get documented in PR #3190 from me.

PR #3276, opened 2019-12-17
Add tests for core-subscriptions edge cases

cgewecke:
## Description

**This PR targets #3190 (Provider Improvements)**
 
Adds a few tests for edge cases in web3-core-subscriptions. Found two tiny bugs - have changed the code to fix them. 

+ `web3.eth.subscribe('newBlockHeaders', 5)`, which fails in _invalidParams, gave a message like: "invalid params for "5". Expected 1, Got 1."

+ `web3.eth.subscribe('does-not-exist')` resulted in a syntax error. It should just log a warning to std.out.

## Type of change

- [x] Unit tests

## Checklist:

### DIFF
--- a/packages/web3-core-subscriptions/src/index.js
+++ b/packages/web3-core-subscriptions/src/index.js
@@ -59,7 +59,7 @@ Subscriptions.prototype.buildCall = function() {
         }
 
         var subscription = new Subscription({
-            subscription: _this.subscriptions[arguments[0]],
+            subscription: _this.subscriptions[arguments[0]] || {}, // Subscript might not exist
             requestManager: _this.requestManager,
             type: _this.type
         });
--- a/test/eth.subscribe.ganache.js
+++ b/test/eth.subscribe.ganache.js
@@ -7,6 +7,7 @@ describe('subscription connect/reconnect', function() {
     let server;
     let web3;
     let accounts;
+    let subscription;
     const port = 8545;
 
     beforeEach(async function() {
@@ -24,44 +25,34 @@ describe('subscription connect/reconnect', function() {
         }
     });
 
-    it('subscribes (baseline)', function() {
-        return new Promise(async function (resolve) {
-            web3.eth
-                .subscribe('newBlockHeaders')
-                .on('data', function(result) {
-                    assert(result.parentHash);
-                    resolve();
-                });
-        });
-    });
-
-    it('subscribes with a callback', function(){
-        let subscription;
-
-        return new Promise(async function (resolve) {
-            subscription = web3.eth
-                .subscribe('newBlockHeaders', function(err, result){
-                    assert(result.parentHash);
-                    subscription.unsubscribe(); // Stop listening..
-                    resolve();
+    it('subscribes (baseline)', function(done) {
+        web3.eth
+            .subscribe('newBlockHeaders')
+            .on('data', function(result) {
+                assert(result.parentHash);
+                done();
             });
-         })
     });
 
-    it('resubscribes', function(){
-        let subscription;
-
-        return new Promise(async function (resolve) {
-            subscription = web3.eth.subscribe('newBlockHeaders');
-            subscription.unsubscribe();
-            subscription.resubscribe()
-
-            subscription.on('data', function(result){
+    it('subscribes with a callback', function(done){
+        subscription = web3.eth
+            .subscribe('newBlockHeaders', function(err, result){
                 assert(result.parentHash);
                 subscription.unsubscribe(); // Stop listening..
-                resolve();
-            })
-         })
+                done();
+            });
+    });
+
+    it('resubscribes', function(done){
+        subscription = web3.eth.subscribe('newBlockHeaders')
+        subscription.unsubscribe();
+        subscription.resubscribe()
+
+        subscription.on('data', function(result){
+            assert(result.parentHash);
+            subscription.unsubscribe(); // Stop listening..
+            done();
+        })
     });
 
     it('allows a subscription which does not exist', function(){
@@ -70,7 +61,7 @@ describe('subscription connect/reconnect', function() {
 
     it('errors when zero params subscrip. is called with the wrong arguments', function(){
         try {
-            web3.eth.subscribe('newBlockHeaders', 5);
+            web3.eth.subscribe('newBlockHeaders', 5)
             assert.fail();
         } catch (err) {
             assert(err.message.includes('Invalid number of parameters for "newHeads"'))
@@ -79,15 +70,13 @@ describe('subscription connect/reconnect', function() {
     });
 
     // Could not get the .on('error') version of this to work - maybe a race condition setting it up.
-    it('errors when the provider is not set', function(){
+    it('errors when the provider is not set', function(done){
         web3 = new Web3();
 
-        return new Promise(async function (resolve) {
-            web3.eth.subscribe('newBlockHeaders', function(err, result){
-                assert(err.message.includes('No provider set'));
-                resolve()
-            })
-        });
+        web3.eth.subscribe('newBlockHeaders', function(err, result){
+            assert(err.message.includes('No provider set'));
+            done();
+        })
     });
 
     it('errors when the `eth_subscribe` request got send, the reponse isnt returned from the node, and the connection does get closed in the mean time', async function() {
--- a/test/eth.subscribe.ganache.js
+++ b/test/eth.subscribe.ganache.js
@@ -35,6 +35,62 @@ describe('subscription connect/reconnect', function() {
         });
     });
 
+    it('subscribes with a callback', function(){
+        let subscription;
+
+        return new Promise(async function (resolve) {
+            subscription = web3.eth
+                .subscribe('newBlockHeaders', function(err, result){
+                    assert(result.parentHash);
+                    subscription.unsubscribe(); // Stop listening..
+                    resolve();
+            });
+         })
+    });
+
+    it('resubscribes', function(){
+        let subscription;
+
+        return new Promise(async function (resolve) {
+            subscription = web3.eth.subscribe('newBlockHeaders');
+            subscription.unsubscribe();
+            subscription.resubscribe()
+
+            subscription.on('data', function(result){
+                console.log(result.parentHash)
+                assert(result.parentHash);
+                subscription.unsubscribe(); // Stop listening..
+                resolve();
+            })
+         })
+    });
+
+    it('allows a subscription which does not exist', function(){
+        web3.eth.subscribe('subscription-does-not-exists');
+    });
+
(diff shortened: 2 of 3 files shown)


packages/web3-core-subscriptions/src/index.js
```diff
@@ -59,7 +59,7 @@ Subscriptions.prototype.buildCall = function() {
         }
 
         var subscription = new Subscription({
-            subscription: _this.subscriptions[arguments[0]],
+            subscription: _this.subscriptions[arguments[0]] || {}, // Subscript might not exist
```
nivida, line 62: @cgewecke Great finding! This is an undocumented and questionable feature and we should probably document it in our documentation if it actually does make sense. This because we have already implemented all subscriptions for the ``shh`` and ``eth`` module and it isn't possible to start for example a ``parity`` subscription with the ``web3.eth.subscribe`` method because the property ``type`` of the ``Subscriptions`` class is hardcoded to ``eth`` which should be ``parity`` in the case someone wants to use not implemented ``parity`` specific subscriptions. Because no one was using this un-documented feature until now (because of the syntax error would we definitely have an open issue) can we think about to extend the capabilities of the ``web3.eth.subscribe`` method to support fully custom-defined subscriptions. The possible existing options object we already have in the function signature of the ``web3.eth.subscribe`` method would probably give us the opportunity to do it in a non-breaking way.

Edit:
But this is probably more something we can address with a separate issue and PR.

test/eth.subscribe.ganache.js
```diff
@@ -35,6 +35,61 @@ describe('subscription connect/reconnect', function() {
         });
     });
 
+    it('subscribes with a callback', function(){
+        let subscription;
+
+        return new Promise(async function (resolve) {
```
nivida: ``async`` isn't required

test/eth.subscribe.ganache.js
```diff
@@ -35,6 +35,61 @@ describe('subscription connect/reconnect', function() {
...
+            subscription = web3.eth
+                .subscribe('newBlockHeaders', function(err, result){
```
nivida: Here we have no spaces between the brackets and on line 41 do we have them.

test/eth.subscribe.ganache.js
```diff
@@ -35,6 +35,61 @@ describe('subscription connect/reconnect', function() {
...
+
+    it('resubscribes', function(){
+        let subscription;
+
+        return new Promise(async function (resolve) {
+            subscription = web3.eth.subscribe('newBlockHeaders');
+            subscription.unsubscribe();
+            subscription.resubscribe()
+
+            subscription.on('data', function(result){
```
nivida: Shouldn't we register here the listener directly after the call of ``web3.eth.subscribe`` to be sure the ``data`` event listener is kept?

test/eth.subscribe.ganache.js
```diff
@@ -35,6 +35,61 @@ describe('subscription connect/reconnect', function() {
...
+                    subscription.unsubscribe(); // Stop listening..
+                    resolve();
+            });
+         })
+    });
+
+    it('resubscribes', function(){
+        let subscription;
+
+        return new Promise(async function (resolve) {
```
nivida: ``async`` isn't required here and the function code style isn't consistent as well. I think this applies to all newly created test cases.

test/eth.subscribe.ganache.js
```diff
@@ -35,6 +35,61 @@ describe('subscription connect/reconnect', function() {
         });
     });
 
+    it('subscribes with a callback', function(){
+        let subscription;
```
nivida: This variable could be registered once as a global existing ``subscription`` let in the current ``describe`` closure.

test/eth.subscribe.ganache.js
```diff
@@ -35,6 +35,61 @@ describe('subscription connect/reconnect', function() {
...
+                resolve();
+            })
+         })
+    });
+
+    it('allows a subscription which does not exist', function(){
+        web3.eth.subscribe('subscription-does-not-exists');
+    });
+
+    it('errors when zero params subscrip. is called with the wrong arguments', function(){
```
nivida, line 107: Is it possible to test here the ``error`` event listener and the callback ``error`` parameter?

test/eth.subscribe.ganache.js
```diff
@@ -35,6 +35,61 @@ describe('subscription connect/reconnect', function() {
...
+            assert(err.message.includes('Invalid number of parameters for "newHeads"'))
+            assert(err.message.includes('Got 1 expected 0'));
+        }
+    });
+
+    // Could not get the .on('error') version of this to work - maybe a race condition setting it up.
+    it('errors when the provider is not set', function(){
+        web3 = new Web3();
+
+        return new Promise(async function (resolve) {
```
nivida: If I'm correct is ``Promise`` and ``async`` not required in those test cases and we could use the ``done`` callback which would look a bit cleaner.

nivida (changes requested):

test/eth.subscribe.ganache.js
cgewecke: 👍

test/eth.subscribe.ganache.js
cgewecke: Apparently not - neither hears anything. An error is thrown.

test/eth.subscribe.ganache.js
cgewecke: Fixed all of these, sorry. I copy pasted the format of the baseline test (and I fixed that one too)

test/eth.subscribe.ganache.js
cgewecke: Rewrote this in 21d854c362475a694d79f32dc38dbfa4272e41f9. Is that closer to what you think it should be?

packages/web3-core-subscriptions/src/index.js
cgewecke: Ok great - that's a good idea. The parity subscriptions you're talking about are [these](https://wiki.parity.io/JSONRPC-parity_pubsub-module)?

It's like 'subscribing'  over websockets to the standard RPC methods (ex:`eth_getBalance`) to get a single notification?

test/eth.subscribe.ganache.js
nivida: Not exactly what I thought about. We should be able to register one ``data`` listener and this listener should still get triggered after the call of ``resubscribe``. It is probably easier to test this with the ``connected`` event. I've also noticed that we only have to call ``unsubscribe`` once because the ``resubscribe`` is already unsubscribing the subscription.

Pseudo-code: 
``` javascript
it('...', (done) => {
  stage = 0;
  subscription = web3.eth.subscribe('newBlockHeaders');
  subscription.on('data',() => {
    if (stage === 0) {
      subscription.resubscribe();
      stage = 1;
      return;
    }

    assert(result.parentHash);
    subscription.unsubscribe(); // Stop listening..
    done();
  });
});

```

packages/web3-core-subscriptions/src/index.js
nivida: Yes, they do offer some "normal" JSON-RPC method calls as subscriptions for WebSocket connections. It's definitely better to use such a subscription if existing instead of polling :)


Btw. and off-topic: I've created a new subscription type in the 2.0 MVP to have polling subscriptions ([PollingSubscription](https://github.com/ethereum/web3.js/blob/architecture/expose-only-one-package/internal/core/src/json-rpc/subscriptions/polling/PollingSubscription.ts))

test/eth.subscribe.ganache.js
cgewecke: @nivida Have added as a separate case in ac880f656b564bf275ff2d417c3f2d62830a9ea2. 

In your view, are these actually different scenarios?

If not, what is the intended use-case of the `resubscribe` method?

nivida: > In your view, are these actually different scenarios?

Yes, because the ``resubscribe`` is doing an ``unsubscribe`` as well and doesn't remove the event listeners of the current subscription. Because ``unsubscribe`` and ``resubscribe`` is called in your case is it doing a double ``unsubscribe`` call once with removing of the listeners and once without.

test/eth.subscribe.ganache.js
nivida: The passed callback should get called successfully with the error. In the case of the error event listener do we have a timing issue we already noticed when the ``provider.send`` call is failing before the ``EventEmitter`` got returned. We would have to add a synthetic delay or another solution to fix this. 🤔

nivida: The same problem should exist in this case as well: https://github.com/ethereum/web3.js/blob/provider-related-improvements/packages/web3-core-subscriptions/src/subscription.js#L198

cgewecke: @nivida What do you think about instead of firing the callbacks / emitters just throwing in those cases. 

It seems more reliable and it's before anything is sent - maybe it makes sense to "fail hard and fast" everywhere within [this code block](https://github.com/ethereum/web3.js/blob/ac880f656b564bf275ff2d417c3f2d62830a9ea2/packages/web3-core-subscriptions/src/subscription.js#L191-L218) at `.subscribe` .

nivida: Yes, this would probably be the cleanest solution but would introduce a breaking-change because all DApp devs would have to add a try/catch around the ``subscribe`` call and they would get errors where they were silently ignored before. 🤔

nivida: Additionally, would this mean that some errors are thrown, some are only returned in the callback, and some are emitted through with the error event and also are returned to the callback.

cgewecke: Ah yes! 

In d21d746d2567f0cf7adc6bf699404d9bae3bf11b have wrapped the 'no provider' and 'no provider.on' handling in a zero-length timeout (to schedule the emitter in the next execution loop) and added tests for all those cases. 

Preserved the current handling for bad params because that's the existing behavior. Not sure there's much value in adding two other ways that error can express itself - we'd still need to throw if people currently expect that. Also seems consistent with much of the other param validation.  

What is your view of it?

nivida: The currently applied fix does look good to me 👍 

Is it possible to fix the error handling on line 221 and 247 in the ``subscription.js`` file the same way as it's done for the parameter validation in this PR as well? 

Because when the call of ``requestManager.send`` does fail immediately do we have the same problem as we have for the parameter validation. This is probably an edge-case but it will definitely happen in some dev/test setups.

cgewecke: Added in 9038a7a9d4a6872234302d933bb6517a26052c2b.

nivida (approved):

PR #3285, opened 2019-12-20
Make user supplied callback fire in accounts.signTransaction

### LINKED ISSUE #3283: Bad return in web3.eth.accounts.signTransaction
See https://github.com/ethereum/web3.js/blob/1.x/packages/web3-eth-accounts/src/index.js#L238.

The inner `signed` function returns from that `try` block, but instead it seems like it should assign the object to `result`. In the case that `return` is reached then L252 is not executed. That's what I was running into in #3281 — I finally realized when stepping through `signTransaction` in vscode's debugger.

---

There is another concern I have about `web3.eth.accounts.signTransaction`, and I'm not sure how many other function throughout the web3 codebase may be in the same situation. While I'll explain it here, my guess is there should be another issue to discuss the matter more specifically.

So, `signTransaction` can take a callback, and if none is provided it sets up a dummy callback (cf. L134). However, it always both calls the callback and returns a promise. That's really not good 💣 💥 , but there may be something I've misunderstood so I'm hoping the web3 team can provide feedback on this point.

The implication is that callers that want to use a callback also need to guard against unhandled promise rejections at their call sites, and they should not need to do that. In other words, if I'm passing `(err, val) => {...}` as the callback then handling of the error should be total with respect to the callback and it should not be necessary to do:

```js
web3.eth.accounts.signTransaction(..., callback).catch(...)
```

When implementing functions that optionally take a Node.js style callback as the last argument and otherwise return a promise, it might be good to implement with Node's [`callbackify`](https://nodejs.org/docs/latest-v10.x/api/util.html#util_util_callbackify_original) ([src](https://github.com/nodejs/node/blob/94365f0c1980a48b156899705a18bbd390e933f0/lib/util.js#L340-L365)), or the [callbackify](https://www.npmjs.com/package/callbackify) package may be even better since it handles *"callback-mode vs. promise-mode"* internally while with Node's built-in `callbackify` you have to do that yourself.

The main idea is to author all functions/methods that do something asynchronous with a "promise first" mentality.  And then consistently provide for callback usage by passing them through callbackify. When used in callback-mode the function/method always returns `undefined` and err/val handling is total with respect to the callback (which is never called immediately, also very important); while in promise-mode a promise is always returned and the caller needs to use it with `await` or `.then`/`.catch`.

#### Versions
web3 1.2.4

cgewecke:
## Description

Fixes a bug which disabled the ability to call `eth.accounts.signTransaction` using the callback form. 

@michaelsbradleyjr This only addresses the narrowest part of your issue. Have not used `Object.assign` as you suggest there, mostly to keep with the pattern of Web3's existing code. Please advise if you believe that's critical though. 

Fixes #3283 

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-eth-accounts/src/index.js
+++ b/packages/web3-eth-accounts/src/index.js
@@ -128,7 +128,6 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
     var _this = this,
         error = false,
         transactionOptions = {},
-        result,
         hasTxSigningOptions = !!(tx && ((tx.chain && tx.hardfork) || tx.common));
 
     callback = callback || function() {
@@ -235,7 +234,7 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
             var rawTransaction = '0x' + rlpEncoded;
             var transactionHash = utils.keccak256(rawTransaction);
 
-            return {
+            var result = {
                 messageHash: '0x' + Buffer.from(ethTx.hash(false)).toString('hex'),
                 v: '0x' + Buffer.from(ethTx.v).toString('hex'),
                 r: '0x' + Buffer.from(ethTx.r).toString('hex'),
@@ -244,13 +243,13 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
                 transactionHash: transactionHash
             };
 
+            callback(null, result);
+            return result;
+
         } catch (e) {
             callback(e);
             return Promise.reject(e);
         }
-
-        callback(null, result);
-        return result;
     }
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,7 +105,7 @@ Released with 1.0.0-beta.37 code base.
 - ``eth_requestAccounts`` as ``requestAccounts`` added to web3-eth package (#3219)
 - ``sha3Raw`` and ``soliditySha3Raw`` added to web3-utils package (#3226)
 - ``eth_getProof`` as ``getProof`` added to web3-eth package (#3220)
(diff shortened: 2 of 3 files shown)


packages/web3-eth-accounts/src/index.js
```diff
@@ -235,7 +234,7 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
             var rawTransaction = '0x' + rlpEncoded;
             var transactionHash = utils.keccak256(rawTransaction);
 
-            return {
+            var result = {
```
michaelsbradleyjr, line 237: Not a big deal because of how `var` works (hoisting) but since `result` is referenced outside the try block that starts on L173 it might be more clear if `result` is declared outside of it with `var result;` and then the assignment is made with `result =` instead of `var result =`.

michaelsbradleyjr (commented):
Looks good, and sorry if I didn't communicate clearly when I mentioned "assign":

>  ...it seems like it should assign the object to result

By that I meant exactly what you did (though see my comment below re: declaration) and not `Object.assign`.

michaelsbradleyjr (approved):

packages/web3-eth-accounts/src/index.js
cgewecke: Ah! It's no longer referenced anywhere outside the try block... believe the old references deleted just below this point were the only ones.  

Thanks for looking this over @michaelsbradleyjr!

michaelsbradleyjr: Ah, I see, I missed that aspect of the changes. Thanks for clarifying!

nivida (approved):
LGTM!

PR #3299, opened 2020-01-10
Add websocket / misc tests (#3190)

cgewecke:
## Description

**PR targets #3190 (Provider Improvements)**

Part of another round of review for provider improvements :) 

Mostly just closing some gaps in the coverage report and testing the behavior of auto-reconnect.

There's one logic change, will discuss in the diff below.


packages/web3-providers-ws/src/index.js
```diff
@@ -178,7 +178,7 @@ WebsocketProvider.prototype._onClose = function (event) {
 
     if (this.requestQueue.size > 0) {
         this.requestQueue.forEach(function (request, key) {
-            request.callback(errors.ConnectionClosedError(event));
+            request.callback(errors.ConnectionNotOpenError(event));
```
cgewecke: On 1.x atm when someone tries to `send` over websockets and the provider doesn't really exist - a `connection not open on send` is fired. It's a common error that happens if you haven't started your client.

The provider-improvements branch adds request clearance on close and throws a different error for this case. 

Reverting to the original message here. WDYT?

nivida: I've checked the old WS provider and it does throw in the ``error`` or ``close`` case the ``InvalidConnection`` error with the msg ``on WS`` which is actually not helpful. This means I would prefer to use the ``ConnectionClosedError`` and to provide our consumers more details about it.

test/websocket.misc.js
```diff
@@ -0,0 +1,37 @@
+var assert = require('assert');
+var utils = require('./helpers/test.utils');
+var Web3 = utils.getWeb3();
+
+describe('Websocket etc & errors', function () {
+    let web3;
+
+    beforeEach(() => web3 = new Web3('ws://localhost:' + 8777));
```
nivida: Can you replace this arrow function with a normal function?

test/websocket.misc.js
```diff
@@ -0,0 +1,37 @@
+var assert = require('assert');
+var utils = require('./helpers/test.utils');
+var Web3 = utils.getWeb3();
+
+describe('Websocket etc & errors', function () {
```
nivida: The description of this test file is to open and the file name itself as well. Is it possible to have a more describing file name and test description? If this isn't possible do we have to move those cases to somewhere else or we have to create additional smaller test files with more explicit file names and test descriptions.

test/websocket.misc.js
```diff
@@ -0,0 +1,37 @@
...
+    let web3;
+
+    beforeEach(() => web3 = new Web3('ws://localhost:' + 8777));
+
+    // This test's error is fired by the request queue checker in the onClose handler
+    it('errors when there is no connection', async function(){
```
nivida: The ``connection not open`` error doesn't get triggered from the ``onClose`` listener this error does get thrown in the ``send`` method of the provider.

test/websocket.misc.js
```diff
@@ -0,0 +1,37 @@
...
+            await web3.eth.getBlockNumber();
+            assert.fail();
+        } catch (err) {
+            assert(err.message.includes('connection not open'));
+        }
+    });
+
+    // This test's error is fired by the readyState check in .send
+    it('errors when requests continue after socket closed', async function(){
+        try { await web3.eth.getBlockNumber() } catch (err) {
```
nivida: Is this try/catch required?

test/websocket.misc.js
```diff
@@ -0,0 +1,37 @@
...
+                await web3.eth.getBlockNumber();
+                assert.fail();
+            } catch (err){
+                assert(err.message.includes('connection not open'));
+            }
+        }
+    });
+
+    it('supports subscriptions', async function(){
+        assert(web3.eth.currentProvider.supportsSubscriptions());
```
nivida: Oh! great! We forgot to add test cases for this simple function to the providers.. good finding 💪

nivida (commented):
I've only added some comments about the ``websocket.misc`` test file because it looks like something we do not really want to have in this format. :)

test/websocket.misc.js
cgewecke: Yes, see above. There are two different locations the error is thrown from. The first failure is thrown from `_onClose`, the second from `send.`

test/websocket.misc.js
cgewecke: I've reverted to the old logic in 0b962dae0c2cf3ae689e351b497c35a64ac7f5e0, while keeping the test expectation so you can see this is not the case [in travis here](https://travis-ci.org/ethereum/web3.js/jobs/635338577#L3658-L3682).

In the current 1.x branch, close doesn't throw the error itself but changes the readyState which causes the request to hit the logic in send. That's no longer the case on the provider-improvements branch.

packages/web3-providers-ws/src/index.js
cgewecke: > it does throw in the error or close case the InvalidConnection

Have opened draft pr #3302 to show these tests pass - e.g produce ConnectionNotOpen on 1.x. 

How did you generate InvalidConnection? That would be a good case to add here.

test/websocket.misc.js
cgewecke: Sorry, yes the name is bad. I've made it more generic now (not sure if that's better) - if you have a good name llmk.

packages/web3-providers-ws/src/index.js
nivida: the ``InavlidConnection`` error does get triggered as soon as the private [_timeout](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-providers-ws/src/index.js#L252) method does get called.

test/websocket.misc.js
nivida: I would move those test cases to the ``websocket.reconnect.ganache`` test file and rename it to ``e2e.websocket.js`` and let us create a real unit test withing the ``websocket.js`` test file.

test/websocket.misc.js
nivida: The described behaviour from you only happens if the underlying ``WebSocket`` connection does have the ``readyState`` set to ``0`` when calling the ``send`` method of the ``WebsocketProvider``. If the provider does have the ``readyState`` sett to ``2`` or ``3`` will it throw the ``connection not open`` error but also only if the provider didn't start to reconnect. Which means it is the same logic as before only with the ``reconnecting`` case added. (see: https://github.com/ethereum/web3.js/blob/1.x/packages/web3-providers-ws/src/index.js#L252)

test/websocket.misc.js
nivida: Thanks for your explanation! Could you add such a note as an in-line comment to the related test case? (A future developer will definitely like such a hint) :)

test/websocket.misc.js
cgewecke: Do you remain concerned that the wrong error will be fired with the proposed change to the request clearance? Or do you think the change makes sense?

test/websocket.misc.js
cgewecke: Added more comments in 4044687b2992782eb2dffea26014a7f9ca56ef75

test/websocket.misc.js
cgewecke: In 4044687b2992782eb2dffea26014a7f9ca56ef75 I combined the files and renamed it `websocket.ganache.js`. The files with the `e2e` prefix are grepped by mocha so that all the clients and browsers run them. 

But many of these tests rely on turning the server on and off, so they fit better with the units, idk. Similar to `eth.subscribe.ganache.js`

cgewecke (commented):
@nivida Made the recommended changes (+/-) and also added a couple tests for mid-chunk parsing errors in 7ad8d34c004e9ef79a0dc776ba37bee6656ee19f

test/websocket.misc.js
nivida: Sounds good to me! We can probably improve "the e2e situation" later! :)

test/websocket.misc.js
nivida: hmm.. we could keep the error in the close listener as it is in the base branch of this PR and extend the conditions in the ``WebsocketProvider.send`` to throw the ``connection not open`` error when the ``readyState`` is set to ``CLOSING``. If I'm right is it with this addition the same "error logic" as in the currently released version. Let me check that quickly.

nivida: I was sadly wrong the current conditions do already handle the ``CLOSING`` state correctly.

nivida: The current logic would be to throw the ``ConnectionNotOpenError`` for all items in the ``requestQueue`` as you already have done. We could there extend the error object with additional properties as for example the ``code`` and ``reason`` from the close event. The ``responseQueue`` should throw the ``InvalidConnection`` error and also there could we add the ``code`` and ``reason`` property from the close event to the error object. 

WDYT?

cgewecke: @nivida Yes this sounds reasonable to me - are you open to merging this as a correct description of 1.x's behavior for the most common error states and making those changes in a separate set of commits (maybe even a PR directed at #3190)?

nivida: @cgewecke Yes, sounds good to me! 👌

nivida (approved):
I've improved the code style consistency (missing semicolons and not consistent formatting of functions)

PR #3301, opened 2020-01-10
Adds the possibility to define a custom registry for the ENS module

### LINKED ISSUE #3010: ENS: option to use custom registry addresses
### Description 

The current web3.eth.ens module only supports the official ENS registries on various networks, but there are some custom ENS deployments (especially on testnets) that users may want to resolve from.

#### Expected behavior

Upon creating the ENS module, the user could pass an option to specify the registry address ([e.g. see `ethereum-ens`'s options](https://github.com/ensdomains/ensjs#ens)).

#### Actual behavior

The ENS module self-selects a registry address based on the current network, without giving the user a chance to override it.

### Versions 
* web3.js: v1.2.1 and v2

It would be ideal for this to be included in both v1 and v2, but if we'd like to avoid new features to v1, it would make sense to add it only to v2.

nivida:
## Description

This PR does extend the ``ENS`` module with the ``registryAddress`` property to define a custom registry and to store an already detected registry address. By side of the improvement of a stored registry address did I also improve the anyways changed ``checkNetwork`` method to only check the sync state of the node if at least an hour is gone. This because otherwise, it doesn't make sense to check it.

Fixes #3010

## Type of change

## Checklist:

- [x] I have tested my code on a local test network.


docs/web3-eth-ens.rst
```diff
@@ -8,6 +8,35 @@ The ``web3.eth.ens`` functions let you interacting with ENS.
...
 ------------------------------------------------------------------------------
 
+registryAddress
+=====================
+
+.. code-block:: javascript
+
+    web3.eth.ens.registryAddress;
+
+The ``registryAddress`` property can be used to define a custom registry address in the case you are connected to an unknown chain.
```
cgewecke:
```suggestion
The ``registryAddress`` property can be used to define a custom registry address when you are connected to an unknown chain.
```

docs/web3-eth-ens.rst
```diff
@@ -8,6 +8,35 @@ The ``web3.eth.ens`` functions let you interacting with ENS.
...
+
+.. note::
+   If no address is defined will it try to detect the registry on the chain you are currently connected with and on the call of ``setProvider`` in the Eth module will it keep the defined address and use it for the ENS module.
```
cgewecke, line 21:
```suggestion
If no address is defined it will try to detect the registry on the chain you are currently connected with.  When  ``setProvider`` in the Eth module is called it will keep the defined address and use it for the ENS module.
```

packages/web3-eth-ens/src/ENS.js
```diff
@@ -33,21 +34,40 @@ var ResolverMethodHandler = require('./lib/ResolverMethodHandler');
...
-    },
-    enumerable: true
-});
+    Object.defineProperty(this, 'registryAddress', {
+        get: function () {
+            return registryAddress;
+        },
+        set: function (value) {
+            if (value === null) {
+                registryAddress = value;
```
cgewecke, line 61: How does this setter work? Just curious ... I see the registryAddress being set in the tests and everything looks right there.

cgewecke (approved):
This looks perfect to me, awesome. 💯 

I left one question and a couple small rewording suggestions for the docs.

packages/web3-eth-ens/src/ENS.js
nivida: I've used the ``inputAddressFromatter`` to format and validate the ``address`` as we are doing on other points as well. With the exception of the additional valid value ``null``. The ``null`` value can be used to "clear" the custom-defined registry address and to "re-activate" the automatic registry search the ``ENS`` module does have.

docs/web3-eth-ens.rst
nivida: [As it looks like](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html) can we keep those spaces.

nivida:
@cgewecke Thanks for your review! I've fixed the commented typos with [69001c7](https://github.com/ethereum/web3.js/pull/3301/commits/69001c7dc4c65bac6d6f2637908f075a50450781)

packages/web3-eth-ens/src/ENS.js
cgewecke: Ah thanks for the explanation.

PR #3302, opened 2020-01-10
Add tests for websocket connection errors

cgewecke:
## Description

These are the same tests proposed for #3190 but aimed at 1.x. They show that the current 1.x behavior when connecting to a non-existent url is to fire an error with this message 
```
connection not open on send
```

### DIFF
--- a/test/websocket.js
+++ b/test/websocket.js
@@ -1,0 +1,39 @@
+var assert = require('assert');
+var utils = require('./helpers/test.utils');
+var Web3 = utils.getWeb3();
+
+describe('Websockets', function () {
+    let web3;
+
+    beforeEach(function(){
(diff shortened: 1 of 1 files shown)


nivida:
@cgewecke Thanks for creating this! I'm wondering why the requirement now is to have this error message in PR #3190 and why you were arguing to not have it in PR #3171.

cgewecke:
@nivida RE: #3171 could you explain further? I've just checked and the only thing I see is [a discussion][1] about whether it's safe to remove the `console.error('connection not open on send()')`. 

[1]: https://github.com/ethereum/web3.js/pull/3171#discussion_r341287741

cgewecke:
This was discussed at the web3 meeting and is close-able.

PR #3303, opened 2020-01-11
Minimize changes to test/contract.js (#3190)

cgewecke:
## Description

**PR targets #3190**

(Part of ongoing #3190 review. Opening for further discussion.) 

PR reverts all the changes to test/contract.js except the three necessary to get the tests passing.

Diff vs. 1.x looks like this:

Screen Shot 2020-01-10 at 4 15 46 PM

Many tests fail unless the call to eth.setProvider is commented out in the test fixture. Why?
https://github.com/ethereum/web3.js/blob/5e1aec2f07c82d412123229bef8a865142a8668a/test/contract.js#L242-L244

### DIFF
--- a/docs/include_package-core.rst
+++ b/docs/include_package-core.rst
@@ -15,7 +15,6 @@ Will change the provider for its module.
 
 .. note::
     When called on the umbrella package ``web3`` it will also set the provider for all sub modules ``web3.eth``, ``web3.shh``, etc EXCEPT ``web3.bzz`` which needs a separate provider at all times.
-    Be aware to clean up the listeners of your provider if you pass the exact same provider instance a second time to the ``setProvider`` method of any module. This because the ``setProvider`` method will register the listeners required from web3.js again.
 
 ----------
 Parameters
--- a/docs/include_package-core.rst
+++ b/docs/include_package-core.rst
@@ -13,7 +13,9 @@ setProvider
 
 Will change the provider for its module.
 
-.. note:: When called on the umbrella package ``web3`` it will also set the provider for all sub modules ``web3.eth``, ``web3.shh``, etc EXCEPT ``web3.bzz`` which needs a separate provider at all times.
+.. note::
+    When called on the umbrella package ``web3`` it will also set the provider for all sub modules ``web3.eth``, ``web3.shh``, etc EXCEPT ``web3.bzz`` which needs a separate provider at all times.
+    Be aware to clean up the listeners of your provider if you pass the exact same provider instance a second time to the ``setProvider`` method of any module. This because the ``setProvider`` method will register the listeners required from web3.js again.
 
 ----------
 Parameters
--- a/test/contract.js
+++ b/test/contract.js
@@ -240,7 +240,7 @@ var getEthContractInstance = function(abi, address, options, provider) {
     }
 
     var eth = new Eth(provider);
-    eth.setProvider(provider);
+    //eth.setProvider(provider);
     return new eth.Contract(abi, address, options);
 }
 
--- a/test/contract.js
+++ b/test/contract.js
@@ -9,231 +9,247 @@ var StandAloneContract = require('../packages/web3-eth-contract');
 
 
 var abi = [{
-    'type': 'constructor',
-    'inputs': [{
-        'name': 'who',
-        'type': 'address'
-    }, {
-        'name': 'myValue',
-        'type': 'uint256'
+    "type": "constructor",
+    "inputs": [{
+        "name": "who",
+        "type": "address"
+    },{
+        "name": "myValue",
+        "type": "uint256"
     }]
-}, {
-    'constant': false,
-    'inputs': [
+},{
+    "constant": false,
+    "inputs": [
         {
-            'components': [
-                {'name': 'status', 'type': 'bool'}
+            "components": [
+                {"name": "status", "type": "bool"}
             ],
-            'name': 'nestedStruct',
-            'type': 'tuple'
+            "name": "nestedStruct",
+            "type": "tuple"
         }
     ],
-    'name': 'addStruct',
-    'outputs': [],
-    'payable': false,
-    'stateMutability': 'nonpayable',
-    'type': 'function'
-}, {
-    'constant': true,
-    'inputs': [
+    "name": "addStruct",
+    "outputs": [],
+    "payable": false,
+    "stateMutability": "nonpayable",
+    "type": "function"
+},{
+    "constant": true,
+    "inputs": [
         {
-            'name': '',
-            'type': 'address'
+            "name": "",
+            "type": "address"
         }
     ],
-    'name': 'listOfNestedStructs',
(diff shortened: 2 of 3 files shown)


test/contract.js
```diff
@@ -9,231 +9,247 @@ var StandAloneContract = require('../packages/web3-eth-contract');
...
- *
- * @param {Array} abi
- * @param {String} address
- * @param {providers} provider
- * @param {Object} options
- *
- * @returns {Contract}
- */
-var getStandAloneContractInstance = function (abi, address, provider, options) {
+var getStandAloneContractInstance = function(abi, address, options, provider) {
```
nivida, line 193: I've improved those two factory functions because line 195-213 is not required if you change the function parameters order. This means I would vote to keep those changes in  ``getStandAloneContractInstance`` & ``getEthContractInstance``.

test/contract.js
```diff
@@ -9,231 +9,247 @@ var StandAloneContract = require('../packages/web3-eth-contract');
...
- */
-var getEthContractInstance = function (abi, address, provider, options) {
     var eth = new Eth(provider);
-    const contract = new eth.Contract(abi, address, options);
+    //eth.setProvider(provider);
+    return new eth.Contract(abi, address, options);
+}
 
-    return contract;
+var account = {
```
nivida, line 247: Could you move this ``var`` up to the other global variables of this test file?

test/contract.js
```diff
@@ -9,231 +9,247 @@ var StandAloneContract = require('../packages/web3-eth-contract');
...
- * @param {String} address
- * @param {providers} provider
- * @param {Object} options
- *
- * @returns {Contract}
- */
-var getEthContractInstance = function (abi, address, provider, options) {
     var eth = new Eth(provider);
-    const contract = new eth.Contract(abi, address, options);
+    //eth.setProvider(provider);
```
nivida, line 243: The call of ``setProvider`` is not required because the ``constructor`` of the ``Eth`` module does set the provider already as expected. The tests fail because the added validations to the ``FakeIpcProvider`` aren't in the correct order (I will check that closer). But we can say that the logic behind ``setProvider`` is still correct or the test case on line 3112 would fail.

nivida (changes requested):

test/contract.js
cgewecke: I think we disagree about this - the simplification you propose in #3190 is good, but it makes the file have a large number of changes that aren't strictly necessary to get the tests to pass. 

I would prefer to see the existing tests go in with an absolute minimum of difference so they are easily legible if someone looks at the PR later to figure what went wrong or what might have changed. 

With this diff, we will see that only the data type of subscriptions object and the question about eth.setProvider presented difficulties in getting the existing tests to pass. 

In line with  "making the smallest change possible" discussed in [part 4][1] of  the review guidelines.

[1]: https://github.com/ethereum/web3.js/blob/1.x/CONTRIBUTIONS.md#guidelines-for-pull-requests-and-releases-web3-1x

test/contract.js
cgewecke: Yes, it seems possible everything is ok but it would be good to certain as well :)

test/contract.js
nivida: This is totally true. It will be way easier to follow the changes of PR #3190 this way 👌 (we can improve such stuff later)

test/contract.js
cgewecke: Resolving this under the 'minimum change' principal.

test/contract.js
nivida: @cgewecke I've checked the issue we have with the additional call of ``eth.setProvider`` closer. Because we pass the exact same provider again with ``setProvider`` does it register the ``data`` listener again and the ``eth_getTransactionReceipt`` call will get executed more than required. We could implement the removal of all listeners on ``setProvider`` to prevent this behaviour but this would probably break existing DApps which do register event listeners before ``setProvider`` got called. 

After all, if we add an explanatory note to the ``setProvider`` method to inform the user about this behaviour can we ignore it. WDYT?

cgewecke: Ok - that sounds reasonable...I'm going to walk the code again as you describe just to satisfy my paranoia. The way the test set's itself up with with two provider settings doesn't seem very normal anyway.

nivida: @cgewecke Great! 👌

nivida:
@cgewecke I've extended the note of the ``setProvider`` documentation to inform our consumers to care about the provider listeners if you pass the exact same provider instance again. I will wait on a short review of you and will merge this PR after. 👌

docs/include_package-core.rst
```diff
@@ -13,7 +13,9 @@ setProvider
 
 Will change the provider for its module.
 
-.. note:: When called on the umbrella package ``web3`` it will also set the provider for all sub modules ``web3.eth``, ``web3.shh``, etc EXCEPT ``web3.bzz`` which needs a separate provider at all times.
+.. note::
+    When called on the umbrella package ``web3`` it will also set the provider for all sub modules ``web3.eth``, ``web3.shh``, etc EXCEPT ``web3.bzz`` which needs a separate provider at all times.
+    Be aware to clean up the listeners of your provider if you pass the exact same provider instance a second time to the ``setProvider`` method of any module. This because the ``setProvider`` method will register the listeners required from web3.js again.
```
cgewecke: @nivida 

I'm not sure this is correct behavior - tbh it reads like a bug description.

Can we remove this docs addition and merge the PR into #3190 as is, leaving the question of what should be done here open? As discussed in chat, #3190 requires additional testing and behavioral descriptions for the new provider/request-manager arch - am very aware of the issues raised by this test change (now).

In [this review comment][1] you note:

>  We could implement the removal of all listeners on setProvider to prevent this behaviour but this would probably break existing DApps which do register event listeners before setProvider got called.

[1]:  https://github.com/ethereum/web3.js/pull/3303#discussion_r367886182

However, it's possible that `setProvider` *should* clear existing subscriptions since toggling networks in a Dapp is pretty common. Intuitively, going `a --> b --> a` would "just work".

nivida: > However, it's possible that setProvider should clear existing subscriptions since toggling networks in a Dapp is pretty common. Intuitively, going a --> b --> a would "just work".

We were always talking about EventEmitter listeners and not subscriptions. Subscriptions do get cleaned up as it was before. (please read the whole code and try to understand what changes I have done otherwise will this take years)

> Can we remove this docs addition and merge the PR into #3190 as is, leaving the question of what should be done here open? 

We can do that but could you next time please answer a comment more precisely? ([comment from me](https://github.com/ethereum/web3.js/pull/3303#discussion_r367886182))

>  am very aware of the issues raised by this test change (now).

There aren't really any bigger problems. We do solve a lot of bugs with it and we have extended the test suite as well. I think you are panicking without any reason. We also do have now an RC version we will release first which does help us to cover unseen problems which I can't see right now.

PR #3314, opened 2020-01-15
WebsocketProvider - Error Handling

nivida:
## Description

I've compared the current errors with the errors we do throw in the refactored ``WebsocketProvider`` and adjusted those who would introduce a breaking change. 

During the comparison of those did I notice that the old ``WebsocketProvider`` does have two different timeouts. The first place the old provider does check for a timeout is in the ``_parseResponse`` method and is per chunk and the other configurable timeout in ``_addResponseCallback`` is on a per ``request`` base. I will check that closer to be sure it won't break anything.

## Type of change

- [x] Error Handling

## Checklist:


cgewecke (approved):
Yes, makes sense and looks good! 👍

PR #3316, opened 2020-01-20
Ensure '0x' prefix exists (web3-eth-accounts)

### LINKED ISSUE #3041: web3.eth.accounts.privateKeyToAccount accepts private key without "0x" prefix but returns incorrect result
### Description

According to the example in the documentation, function `web3.eth.accounts.privateKeyToAccount` expects private key to be passed to hexadecimal string prefixed with "0x".  If private key is passed without "0x" prefix, the function silently returns incorrect result.

#### Expected behavior

Function `web3.eth.accounts.privateKeyToAccount` throws on incorrect private key format, and always returns correct result on correctly formatted private keys.  Whether hexadecimal private key without "0x" prefix is correct or not is discussible.

#### Actual behavior

On hexadecimal private key without "0x" prefix function `web3.eth.accounts.privateKeyToAccount` silently returns incorrect result.

Call function `web3.eth.accounts.privateKeyToAccount` with parameters `0x348ce564d427a3311b6536bbcff9390d69395b06ed6c486954e971d960fe8709` and `348ce564d427a3311b6536bbcff9390d69395b06ed6c486954e971d960fe8709` and compare the results.

### Versions 
* web3.js: 1.0.0-beta.55

nivida:
## Description

This PR ensures the '0x' prefix is always existing on a method call of the Accounts module.

Fixes #3041

## Type of change

- [x] Enhancement

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,3 +133,7 @@ Released with 1.0.0-beta.37 code base.
 ### Added
 
 - ENS module extended with the possibility to add a custom registry (#3301)
+
+### Changed
+
+- Ensure '0x' prefix is existing for Accounts.sign and Accounts.privateKeyToAccount (#3041)
--- a/packages/web3-eth-accounts/src/index.js
+++ b/packages/web3-eth-accounts/src/index.js
@@ -121,6 +121,10 @@ Accounts.prototype.create = function create(entropy) {
 };
 
 Accounts.prototype.privateKeyToAccount = function privateKeyToAccount(privateKey) {
+    if (!privateKey.startsWith('0x')) {
+        throw new Error('Required prefix "0x" is missing.');
+    }
+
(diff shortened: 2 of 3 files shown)


cgewecke (approved):
LGTM! 💯

DalderupMaurice:
Out of curiousity, is there a reason why the implementation differs from the v2 branch?
https://github.com/ethereum/web3.js/pull/2929/files#diff-0150e6b63e1674be31f24602a91ea0f5R126

This would result in an extra breaking change going from v1 to v2

Either way, thanks for adding it in!

nivida:
@DalderupMaurice Oh thanks for commenting here! This is true, the implemented solution in 2.0 is the more advanced solution. This shouldn't introduce an extra breaking change but I will open a PR against 1.x asap with the solution implemented in 2.x thanks!

PR #3317, opened 2020-01-20
Extend `_txInputFormatter` with hex prefix check

### LINKED ISSUE #1082: Request: ensure '0x' prefixing for all data fields
Ok, so this started as an [issue](https://github.com/paritytech/parity/issues/6531) I filled to Parity, but it turns out the fix should be made in `web3.js` instead.

There's a very subtle and hard to trace bug caused (for example) when you submit the contract bytecode data without the `0x` prefix. Sample code:

```javascript
let web3 = new Web3(new Web3.providers.HttpProvider("http://localhost:8545"));

let compiledCode = solc.compile(code);
let abiDefinition = JSON.parse(compiledCode.contracts[':Voting'].interface);
let VotingContract = web3.eth.contract(abiDefinition);
let byteCode = compiledCode.contracts[':Voting'].bytecode;

let deployedContract = VotingContract.new(['Rama','Nick','Jose'],{data: byteCode, from: web3.eth.accounts[0], gas: 4700000});
```

The last line fails because on the line before `solc` yields hex-encoded data without the `0x` prefix. Both Geth and TestingRPC implementations are flexible enough to accept that payload as is, even though it doesn't strictly conform to the standard, but when I tested this code against a Parity backend I got a very unhelpful `invalid params` exception.

An argument could be made to change the `solc` implementation to return an `0x`-prefixed string, but I think it would be best to add some code to the `web3.js` implementation to automatically do that prefixing when required. That is already the case for `parity.js`.

**Note that nowhere in the JS API docs there's mention of how such data should be encoded**, the only guide is the JSONRPC documentation, which does ask for an `0x` prefix for all hex data, but the user cannot be expected to be well informed about the lower level protocol to use the higher level JS API. Since the JS API already performs type conversion whenever convenient, it should fall into its responsibilities to handle this detail as well. It would also be helpful for `web3.js` to give a meaningful warning in case the prefix is omitted (as it's more sensible to include it anyway).

nivida:
## Description

``_txInputFormatter`` extended with hex prefix check. I've kept the ``isHex`` utility function as it is because it is a part of the public API and would probably introduce a breaking change for some.

Edit: 
I've changed the implementation to prepend the ``0x`` prefix if not existing instead of throwing an ``Error``. This will improve the consistency of the sent ``data`` property for ``eth_call`` and ``eth_sendTransaction`` as well just without breaking something. @cgewecke 

Fixes #1082

## Type of change

## Checklist:


nivida:
@cgewecke As our test does bring up would this be a breaking change for Ganache. The more strict validation which does require any hex string to be prefixed with the known hex prefix '0x' does throw an error in many cases. This means we are probably forced to wait with a more strict validation of hex strings until 2.0. WDYT?

cgewecke:
@nivida Yes, agree it looks that way to me as well. It's too bad.

Ganache has 7 failing tests.  The original comment at the reporting issue says:

> Both Geth and TestingRPC implementations are flexible enough to accept that payload as is, even though it doesn't strictly conform to the standard, but when I tested this code against a Parity backend I got a very unhelpful invalid params exception. 

which means some people are probably passing the raw solc bytecode output directly to Geth too.

cgewecke (commented):
@nivida It looks like were forced to put the Hold label on this for the time being.

nivida:
@ryanio This could get merged because it is backward compatible. web3.js will send anything with a ``0x`` prefix but doesn't have a strict validation of the inputs.

ryanio:
@nivida ah just coming across your comment, sorry I missed it when you first posted it. ok thanks we will include this in the next release.

ryanio (approved):
LGTM!

Issue #3198, opened 2019-11-12
Eth is not a constructor


slavafomin:
Hello!

According to the typings, the `web3-eth` package exports the `Eth` class.

Here's the usage code which is 100% correct according to the provided typings:

```typescript
import { Eth } from 'web3-eth';

const eth = new Eth(provider);
```

However, at runtime it causes the `Eth is not a constructor` error.

I have to use this typing augmentation:

```typescript
import { Eth } from 'web3-eth';

declare module 'web3-eth' {
  export = Eth;
}
```

And to import it like this: `import Eth from 'web3-eth';` in order to make it work.

You need to replace `export class Eth { ... }` with `export = class Eth { ... }` in your declarations.

Thank you!

nivida:
Thanks for opening this issue! I've checked your problem closer and you are missing the required ``allowSyntheticDefaultImports`` and ``esModuleInterop`` configuration options as documented [here](https://github.com/ethereum/web3.js#usage-with-typescript).

Feel free to ping me here again if this wasn’t the case and I will re-open this issue.

slavafomin:
I'm using `esModuleInterop` and it's not the case here. When your are exporting a class like this: `export class Eth { ... }` you are using named exports, which is effectively translates to `modules.exports = { Eth: Eth };` and should be imported with `import { Eth } from '...'`.

I believe my issue still stands true.

nivida:
@slavafomin Thanks for commenting here on this issue! I will reopen it and will try again to reproduce your issue. 👌

benjlevesque:
I have a similar issue with `AbiCoder` in `web3-eth-abi`:

```typescript
import { AbiCoder } from "web3-eth-abi";
new AbiCoder().encodeEventSignature("");
```
=> `web3_eth_abi_1.AbiCoder is not a constructor`

___

My suggestion is to append these 2 lines to `web3-eth-abi/types/index.d.ts`
```typescript
declare const coder: AbiCoder;
export default coder;
```
It enables to use this code 

```typescript
import  AbiCoder  from "web3-eth-abi";
AbiCoder.encodeEventSignature("");
```

Indeed, the JS implementation exports an instance of AbiCoder, which should be reflected by the declaration file.

AuHau:
Hey there,
I believe this is still an issue even with the required compiler flags. I created reproducible bug repo, just to be sure I am not missing something: https://github.com/AuHau/web3.js_types

AuHau:
Still is an issue

[...]

Issue #2363, opened 2019-02-10
Import Web3 not working in Typescript


rhlsthrm:
#### Expected behavior
Documentation should accurately describe the way to import Web3 into a Typescript project. There are conflicting documentations about the correct import, neither of which work.

https://github.com/ethereum/web3.js#usage-with-typescript
```
import Web3 from 'web3';
const web3 = new Web3("ws://localhost:8546");
```

https://web3js.readthedocs.io/en/1.0/web3.html
```
import {Web3} from 'web3';

// "Web3.providers.givenProvider" will be set if in an Ethereum supported browser.
const web3 = new Web3(Web3.givenProvider || 'ws://some.local-or-remote.node:8546');
```

#### Actual behavior
Errors from both approaches, see below for logs.

#### Steps to reproduce the behavior

1. Create Typescript project.
2. Import using either of the documented processes.

#### Error Logs
```
[1] const web3 = new web3_1.Web3("ws://localhost:8546");
[1]              ^
[1]
[1] TypeError: web3_1.Web3 is not a constructor
[1]     at Object.<anonymous> (/Users/rahul/Desktop/connext/connext/moloch-monorepo/packages/moloch-backend/dist/src/index.js:6:14)
[1]     at Module._compile (internal/modules/cjs/loader.js:734:30)
[1]     at Object.Module._extensions..js (internal/modules/cjs/loader.js:745:10)
[1]     at Module.load (internal/modules/cjs/loader.js:626:32)
[1]     at tryModuleLoad (internal/modules/cjs/loader.js:566:12)
[1]     at Function.Module._load (internal/modules/cjs/loader.js:558:3)
[1]     at Module.require (internal/modules/cjs/loader.js:663:17)
[1]     at require (internal/modules/cjs/helpers.js:20:18)
[1]     at Object.<anonymous> (/Users/rahul/Desktop/connext/connext/moloch-monorepo/packages/moloch-backend/dist/index.js:6:10)
[1]     at Module._compile (internal/modules/cjs/loader.js:734:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:745:10)
[1]     at Module.load (internal/modules/cjs/loader.js:626:32)
[1]     at tryModuleLoad (internal/modules/cjs/loader.js:566:12)
[1]     at Function.Module._load (internal/modules/cjs/loader.js:558:3)
[1]     at Module.require (internal/modules/cjs/loader.js:663:17)
[...]

PR #3321, opened 2020-01-21
Add ts config info to docs

cgewecke:
## Description

PR adds the current TS config guidelines to the `Getting Started` section of the docs. At the moment these are in the GitHub README where fewer people might see them.
 
#2363 
#3198 
#3215 

There are several issues where TS users report difficulty importing Web3 modules because the underlying JS exports like this: (example)
```
module.exports = Accounts;
```
but the types don't define a default export. This does not affect people who configure their tsconfig with `esModuleInterop: true`. This setting is the default in a tsconfig generated with `tsc --init`. 

The issue affects 16 of 20 Web3 modules. It does not affect the main module or utils.  

In #3125 @alcuadrado comments:
> what is happening right now is that type declarations use named exports, and the actual js files just reassign module.exports. They should use this pattern instead: https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html

> ...Note that fixing this is a breaking change for TS users

**Questions**:

+ @joshstevens19, how do you see this issue? 
+ @nivida has suggested it might be possible to export in both styles and satisfy everyone. Is this possible or does any change necessarily break one party or another? 
+ how problematic is `esModuleInterop`? Does it cause significant levels of inconvenience or break other common packages for users?

## Type of change

- [x] Documentation

## Checklist:

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -37,3 +37,28 @@ If this property is ``null`` you should connect to a remote/local node.
     var web3 = new Web3(Web3.givenProvider || "ws://localhost:8545");
 
 That's it! now you can use the ``web3`` object.
+
+
+Usage with TypeScript
+=====================
+
+We support types within the repo itself. Please open an issue here if you find any wrong types.
+
+You can use ``web3.js`` as follows:
+
+.. code-block:: typescript
+
+    import Web3 from 'web3';
+    const web3 = new Web3('ws://localhost:8546');
+
+
+If you are using the types in a ``commonjs`` module like a Node app you need to enable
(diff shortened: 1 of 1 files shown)


nivida:
> @nivida has suggested it might be possible to export in both styles and satisfy everyone. 

I didn't suggest it to just satisfy everyone. The idea behind having both export types (default & named) is to not introduce a breaking change for those who have changed their configuration to the documented one. I didn't test it until now how ``tsc`` would handle it with both export types. I've also didn't test what happens if we just change it to the correct ones in a project with the documented config. I will definitely do that asap.

@alcuadrado WDYT?

PR #3322, opened 2020-01-21
Add network-timeout to fix Mosaic E2E

cgewecke:
## Description

The Mosaic run is constantly erroring with some kind of yarn network connectivity problem on install. According to [yarn 4890 comment][1] adding the --network-timeout <seconds> flag should help.

Looks like [it did][2] for this PR anyway...

[1]: https://github.com/yarnpkg/yarn/issues/4890#issuecomment-348479446
[2]: https://travis-ci.org/ethereum/web3.js/jobs/639803905?utm_medium=notification&utm_source=github_status


nivida (approved):
Thanks for investigating and fixing it!

PR #3323, opened 2020-01-21
back-ports privateKey validation of the 2.x Accounts module to 1.x

nivida:
## Description

Thanks to @DalderupMaurice to give me a hint about the more advanced solution in the 2.x branch.

## Type of change

- [x] Enhancement

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] Any dependent changes have been merged and published in downstream modules.
- [x] I ran ```npm run dtslint``` with success and extended the tests and types if necessary.
- [x] I ran ```npm run test:unit``` with success and extended the tests if necessary.
- [x] I ran ```npm run build-all``` and tested the resulting file/'s from  ```dist``` folder in a browser.
- [ ] I have updated the ``CHANGELOG.md`` file in the root folder.
- [ ] I have tested my code on the live network.

PR #3328, opened 2020-01-22
Make key length check optional, do not enforce on encrypt/decrypt (#3323)

cgewecke:
## Description

**PR targets #3323** 

Restores existing non-standard key length encrypt/decrypt tests and has accounts.encrypt / decrypt skip the key length check. 

## Type of change

- [x] PR Review suggestion


nivida:
@cgewecke Thanks for your input! Didn't thought about those non-ethereum standards-compliant secretes someone could store. I will close this one and update the original one.

PR #3329, opened 2020-01-22
Check for error codes on close in RequestManager (#3190)

cgewecke:
## Description
**PR targets #3190**

Addresses an issue exposed by the ganache E2E tests: clean closes are throwing errors to all attached error listeners. 

Adds logic to the `.on("close")` listener in the RequestManager to check the event code and only run errored callbacks if something definitely went wrong. 

Commit 044a4d5c3a07ea880a3a6ea87b7f025824410e2c contains failing tests that show the [problem in Travis][1].

Commit b1a71b27b7cdc89e5d346f49dd9c20167a124708 passes.

NB: the ganache test continues to fail because their close is `1006` even though they run `connection.close` before shutting the client down. Tried to duplicate their failure within Web3's suite and was unable to reproduce. Am forking them to get to the root of the problem. 

[1]: https://travis-ci.org/ethereum/web3.js/jobs/640668821#L602-L609
## Type of change

- [x] Review suggestion

### DIFF
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -48,17 +48,16 @@ module.exports = {
         return this.ConnectionError('connection not open on send()', event);
     },
     ConnectionCloseError: function (event){
-        let msg = 'CONNECTION ERROR: The connection closed unexpectedly';
-
         if (typeof event === 'object' && event.code && event.reason) {
-            msg = 'CONNECTION ERROR: The connection got closed with ' +
+            return this.ConnectionError(
+                'CONNECTION ERROR: The connection got closed with ' +
                 'the close code `' + event.code + '` and the following ' +
-                'reason string `' + event.reason + '`';
-
-            return this.ConnectionError(msg, event);
+                'reason string `' + event.reason + '`',
+                event
+            );
         }
 
-        return new Error(msg);
+        return new Error('CONNECTION ERROR: The connection closed unexpectedly');
     },
     MaxAttemptsReachedOnReconnectingError: function (){
         return new Error('Maximum number of reconnect attempts reached!');
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -47,6 +47,17 @@ module.exports = {
     ConnectionNotOpenError: function (event){
         return this.ConnectionError('connection not open on send()', event);
     },
+    ConnectionCloseError: function (event){
+        const msg = (typeof event === 'object' && event.code && event.reason)
+
+            ? 'CONNECTION ERROR: The connection got closed with ' +
+              'the close code `' + event.code + '` and the following ' +
+              'reason string `' + event.reason + '`'
+
+            : 'CONNECTION ERROR: The connection closed unexpectedly';
+
+        return new Error(msg);
+    },
     MaxAttemptsReachedOnReconnectingError: function (){
         return new Error('Maximum number of reconnect attempts reached!');
     },
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -48,13 +48,15 @@ module.exports = {
         return this.ConnectionError('connection not open on send()', event);
     },
     ConnectionCloseError: function (event){
-        const msg = (typeof event === 'object' && event.code && event.reason)
+        let msg = 'CONNECTION ERROR: The connection closed unexpectedly';
 
-            ? 'CONNECTION ERROR: The connection got closed with ' +
-              'the close code `' + event.code + '` and the following ' +
-              'reason string `' + event.reason + '`'
+        if (typeof event === 'object' && event.code && event.reason) {
+            msg = 'CONNECTION ERROR: The connection got closed with ' +
+                'the close code `' + event.code + '` and the following ' +
+                'reason string `' + event.reason + '`';
 
-            : 'CONNECTION ERROR: The connection closed unexpectedly';
+            return this.ConnectionError(msg, event);
+        }
 
         return new Error(msg);
     },
--- a/packages/web3-core-helpers/types/index.d.ts
+++ b/packages/web3-core-helpers/types/index.d.ts
@@ -69,7 +69,7 @@ export class errors {
     static InvalidResponse(result: Error): Error;
     static ConnectionTimeout(ms: string): Error;
     static ConnectionNotOpenError(): Error;
-    static ConnectionCloseError(event: WebSocketEvent | boolean): Error;
+    static ConnectionCloseError(event: WebSocketEvent | boolean): Error | ConnectionError;
     static MaxAttemptsReachedOnReconnectingError(): Error;
     static PendingRequestsOnReconnectingError(): Error;
     static ConnectionError(msg: string, event?: WebSocketEvent): ConnectionError;
--- a/packages/web3-core-helpers/types/index.d.ts
+++ b/packages/web3-core-helpers/types/index.d.ts
@@ -69,6 +69,7 @@ export class errors {
     static InvalidResponse(result: Error): Error;
     static ConnectionTimeout(ms: string): Error;
     static ConnectionNotOpenError(): Error;
+    static ConnectionCloseError(event: WebSocketEvent | boolean): Error;
     static MaxAttemptsReachedOnReconnectingError(): Error;
     static PendingRequestsOnReconnectingError(): Error;
     static ConnectionError(msg: string, event?: WebSocketEvent): ConnectionError;
--- a/packages/web3-core-requestmanager/src/index.js
+++ b/packages/web3-core-requestmanager/src/index.js
@@ -279,9 +279,9 @@ RequestManager.prototype.clearSubscriptions = function (keepIsSyncing) {
  * Evaluates WS close event
  *
  * @method _isCleanClose
- * @param event close event
+ * @param {object | boolean} event WS close event or exception flag
  *
- * @returns boolean
+ * @returns {boolean}
  */
 RequestManager.prototype._isCleanCloseEvent = function (event) {
     return typeof event === 'object' && ([1000].includes(event.code) || event.wasClean === true);
@@ -291,9 +291,9 @@ RequestManager.prototype._isCleanCloseEvent = function (event) {
  * Detects Ipc close error. The node.net module emits ('close', isException)
  *
  * @method _isIpcCloseError
- * @param event close event
+ * @param {object | boolean} event WS close event or exception flag
  *
- * @returns boolean
+ * @returns {boolean}
  */
 RequestManager.prototype._isIpcCloseError = function (event) {
     return typeof event === 'boolean' && event;
--- a/packages/web3-core-requestmanager/src/index.js
+++ b/packages/web3-core-requestmanager/src/index.js
@@ -279,20 +279,21 @@ RequestManager.prototype.clearSubscriptions = function (keepIsSyncing) {
  * Evaluates WS close event
  *
  * @method _isCleanClose
+ * @param event close event
  *
- * @returns {boolean}
+ * @returns boolean
  */
 RequestManager.prototype._isCleanCloseEvent = function (event) {
-    return typeof event === 'object' &&
-           ([1000].includes(event.code) || event.wasClean === true);
+    return typeof event === 'object' && ([1000].includes(event.code) || event.wasClean === true);
 };
 
 /**
  * Detects Ipc close error. The node.net module emits ('close', isException)
  *
  * @method _isIpcCloseError
+ * @param event close event
  *
- * @returns {boolean}
+ * @returns boolean
  */
 RequestManager.prototype._isIpcCloseError = function (event) {
     return typeof event === 'boolean' && event;
--- a/packages/web3-core-requestmanager/src/index.js
+++ b/packages/web3-core-requestmanager/src/index.js
@@ -279,7 +279,8 @@ RequestManager.prototype.clearSubscriptions = function (keepIsSyncing) {
  * Evaluates WS close event
  *
  * @method _isCleanClose
- * @param {object | boolean} event WS close event or exception flag
+ *
+ * @param {CloseEvent | boolean} event WS close event or exception flag
  *
  * @returns {boolean}
  */
@@ -291,7 +292,8 @@ RequestManager.prototype._isCleanCloseEvent = function (event) {
  * Detects Ipc close error. The node.net module emits ('close', isException)
  *
  * @method _isIpcCloseError
- * @param {object | boolean} event WS close event or exception flag
+ *
+ * @param {CloseEvent | boolean} event WS close event or exception flag
  *
  * @returns {boolean}
  */
--- a/packages/web3-core-requestmanager/src/index.js
+++ b/packages/web3-core-requestmanager/src/index.js
@@ -123,7 +123,7 @@ RequestManager.prototype.setProvider = function (provider, net) {
             });
         });
 
-        // notify all subscriptions about bad close conditions and unsubscribe
+        // notify all subscriptions about bad close conditions
         this.provider.on('close', function close(event) {
             if (![1000, 1001].includes(event.code) || event.wasClean === false) {
                 _this.subscriptions.forEach(function (subscription) {
--- a/packages/web3-core-requestmanager/src/index.js
+++ b/packages/web3-core-requestmanager/src/index.js
@@ -123,11 +123,13 @@ RequestManager.prototype.setProvider = function (provider, net) {
             });
         });
 
-        // notify all subscriptions about the close condition
+        // notify all subscriptions about bad close conditions and unsubscribe
         this.provider.on('close', function close(event) {
-            _this.subscriptions.forEach(function (subscription) {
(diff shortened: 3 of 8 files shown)


packages/web3-core-requestmanager/src/index.js
```diff
@@ -123,11 +123,13 @@ RequestManager.prototype.setProvider = function (provider, net) {
             });
         });
 
-        // notify all subscriptions about the close condition
+        // notify all subscriptions about bad close conditions and unsubscribe
         this.provider.on('close', function close(event) {
-            _this.subscriptions.forEach(function (subscription) {
-                subscription.callback(new Error('CONNECTION ERROR: The connection got closed with the close code `' + event.code + '` and the following reason string `' + event.reason + '`'));
-            });
+            if (![1000, 1001].includes(event.code) || event.wasClean === false) {
```
nivida: - Do those rules also work with an IPC provider? 
- Shouldn't such additional error rules get handled in the provider classes? 
- Aren't we hiding errors with those changes?

cgewecke: > Do those rules also work with an IPC provider?

Good question!

> Aren't we hiding errors with those changes?

At the moment we are throwing errors when there aren't any. (See tests). Can you think of a test which would show an error being swallowed here? Agree that's super important...and will think about it more.

> Shouldn't such additional error rules get handled in the provider classes? 

I think that would be ideal - but this is where the listener callbacks are processed and this block of code is where the error is already being managed (see existing error). 

What would you propose as an alternative?

packages/web3-core-requestmanager/src/index.js
```diff
@@ -123,11 +123,13 @@ RequestManager.prototype.setProvider = function (provider, net) {
             });
         });
 
-        // notify all subscriptions about the close condition
+        // notify all subscriptions about bad close conditions and unsubscribe
```
cgewecke:
```suggestion
        // notify all subscriptions about bad close conditions
```

packages/web3-core-requestmanager/src/index.js
nivida: > Good question!

I've checked the ``net`` module [code on GitHub](https://github.com/nodejs/node/blob/master/lib/net.js) and the documentation of nodejs. The close listener doesn't have all those information passed as we know from the ``WebSocket`` object. The only information you get is a boolean called ``hadError``.

> At the moment we are throwing errors when there aren't any. (See tests). Can you think of a test which would show an error being swallowed here? Agree that's super important...and will think about it more.

I see your point. From a networking perspective, they aren't errors but from a "running-subscription perspective" they are.
We have removed once the error listener because after an error does a ``WebSocket`` ``CloseEvent`` follow. 
~~The correct implementation would probably be to have your addition from the ``RequestManager`` moved to the ``_onClose`` listener and to emit the ``WebsocketProvider`` error event if those conditions are fulfilled.~~

> What would you propose as an alternative? 

~~I would move this logic to the ``_onClose`` listener of the ``WebsocketProvider`` with adding of the ``error`` event listener in the ``RequestManager``. The ``_onClose`` listener should then detect if the ``WebSocket`` connection got closed because of an ``Error`` (un-expected/un-clean close) and emit the correct ``WebsocketProvider`` events.~~

nivida (changes requested):
Could you apply the changes I've talked about in [this comment](https://github.com/ethereum/web3.js/pull/3329/files#r370550760)? Thanks!:)

packages/web3-core-requestmanager/src/index.js
nivida: I've done some tests with the idea I had here and it sadly doesn't work as I thought it would. The solution you proposed is the only one which is working as expected. One addition we could (have to) do is to clean up the ``subscriptions`` map on ``close`` and ``error`` to not have a memory leak.

nivida: ... and it is required to extend the condition to handle the subscriptions with the ``IpcProvider`` nicely as well.

``` javascript 
if ((typeof event === 'boolean' && event) || (![1000, 1001].includes(event.code) || event.wasClean === false)) {
```

nivida: > they aren't errors but from a "running-subscription perspective" they are.

This is from my point of view probably something we can't handle in a good way in any case. For example, if the connected node does close the connection in a clean way but it's unexpected behavior for the client. The developer would have to listen to the close event of his provider to handle this. 
One possible solution we could implement is to add a flag to the ``CloseEvent`` to later detect if it was a clean close from the client-side or remote side.  🤔 

Edit: 
Oh! I totally forget about the meaning of error code [1001](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent) this error code would exactly flag this.

cgewecke: Ok interesting. 

191ae06eb0d595ee07a96f6d58ff3f76d24b6087 adds handling for the IPC case, plus made the test assertions a little more detailed.

35663ec65fe78c18a3914b9956810fc5d5ab9b1d cleans up the event handlers in the tests - am seeing them fire (usually trying to close the server) in a few places after the test is completed.

cgewecke:
Also investigated the failing ganache test, experimenting with different approaches on a fork. 

In simplified form, after attaching an error listener to an event...
```
web3.currentProvider.disconnect()
await pify(server.close)()
```
...throws a 1006 close error (for them). Their tests are set up in a weird way that allows the server shutdown to trigger the _onClose handler before the disconnect logic runs to completion. The events  in question are heard in a subsequent test - they're all delayed relative to the main program thread.  

When sever close is put in a 100ms timeout everything works as expected. 

**Question**: What is the intention behind removing the socketListeners before close in this segment? To prevent `_onClose` from being triggered by the `connection.close`?  

https://github.com/ethereum/web3.js/blob/5e3ad84484741e8b8a48ce2ca95e4ae499d7a157/packages/web3-providers-ws/src/index.js#L364-L367

packages/web3-core-requestmanager/src/index.js
```diff
@@ -123,11 +123,27 @@ RequestManager.prototype.setProvider = function (provider, net) {
...
-            });
+            // The node.net module emits ('close', isException)
+            const ipcCloseError = typeof event === 'boolean' && event;
+
+            // The WS package has a close event w/ codes & reasons
+            const hasErrorCode = ![1000, 1001].includes(event.code);
+
+            if (ipcCloseError || hasErrorCode || event.wasClean === false){
+                _this.subscriptions.forEach(function (subscription) {
+                    const message = (hasErrorCode)
```
nivida: We could move the error message logic out to the ``web3-core-helpers`` package with his error functions. This would clean up those lines a bit.

packages/web3-core-requestmanager/src/index.js
```diff
@@ -123,11 +123,27 @@ RequestManager.prototype.setProvider = function (provider, net) {
...
+
+                        ? 'CONNECTION ERROR: The connection got closed with ' +
+                          'the close code `' + event.code + '` and the following ' +
+                          'reason string `' + event.reason + '`'
+
+                        : 'CONNECTION ERROR: The connection closed unexpectedly';
+
+                    subscription.callback(new Error(message));
```
nivida: Isn't here a missing [subscriptions.delete()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete) call?

packages/web3-core-requestmanager/src/index.js
```diff
@@ -123,11 +123,27 @@ RequestManager.prototype.setProvider = function (provider, net) {
...
+        // notify all subscriptions about bad close conditions
         this.provider.on('close', function close(event) {
-            _this.subscriptions.forEach(function (subscription) {
-                subscription.callback(new Error('CONNECTION ERROR: The connection got closed with the close code `' + event.code + '` and the following reason string `' + event.reason + '`'));
-            });
+            // The node.net module emits ('close', isException)
+            const ipcCloseError = typeof event === 'boolean' && event;
+
+            // The WS package has a close event w/ codes & reasons
+            const hasErrorCode = ![1000, 1001].includes(event.code);
```
nivida: ``isCleanCloseEvent`` and inverting it in the condition would be more explicit and I would prefer to move this out to a small private helper function on the ``RequestManager`` object. ``event.wasClean === false`` should be included in this helper function as well.

packages/web3-core-requestmanager/src/index.js
```diff
@@ -123,11 +123,27 @@ RequestManager.prototype.setProvider = function (provider, net) {
...
         });
 
-        // notify all subscriptions about the close condition
+        // notify all subscriptions about bad close conditions
         this.provider.on('close', function close(event) {
-            _this.subscriptions.forEach(function (subscription) {
-                subscription.callback(new Error('CONNECTION ERROR: The connection got closed with the close code `' + event.code + '` and the following reason string `' + event.reason + '`'));
-            });
+            // The node.net module emits ('close', isException)
+            const ipcCloseError = typeof event === 'boolean' && event;
```
nivida: This could be moved out to a private helper function on the ``RequestManager`` object. Such a function would make this un-clean inline comment obsolete. :)

packages/web3-core-requestmanager/src/index.js
nivida: ``1000, 1001`` I've talked about to handle the error code ``1001`` differently because the end-point went away which is clearly an error from a client perspective. Only the error code ``1000`` does signal a clean close from the client. Do you think we can improve this?

test/eth.subscribe.ganache.js
```diff
@@ -130,6 +130,7 @@ describe('subscription connect/reconnect', function () {
             .subscribe('newBlockHeaders')
             .on("error", function (err) {
                 assert(err.message.includes('No provider set'));
+                this.removeAllListeners();
```
nivida: Why exactly is here in this test always ``this.removeAllListeners()`` required?

nivida (changes requested):

nivida:
> What is the intention behind removing the socketListeners before close in this segment? To prevent _onClose from being triggered by the connection.close?

There isn't any clear intention I still can remember but I think it would be better to remove those listeners after it was able to successfully disconnect. This because the developer wouldn't have to handle this special case separately to the probably anyways already implemented "on connection close business logic" in the ``close`` event listener of the provider. WDYT?

packages/web3-core-requestmanager/src/index.js
nivida: > https://github.com/ethereum/web3.js/pull/3329#discussion_r371194063

test/eth.subscribe.ganache.js
cgewecke: I'm seeing the listeners continue to fire in the background sometimes. Esp. noticeable when  `pify(server.close)()` is called on a closed server - an unhandled rejection gets thrown. 

Have done this to all of them as "best practice" to clean-up on finish.

nivida: You could also use ``once`` instead of ``on`` which will do the same :)

cgewecke: Ok I'll go through and see which one's can be once - some of them probably can't because they are multi-call or multi-listener.

cgewecke:
@nivida The handler refactor is in d8159e4bfb4e79caa30578a6902f34dccdcdc0f9. The once changes are in 3dd1d7b2e83e351dcf05c13ee7954db1b2a9430a

packages/web3-core-helpers/types/index.d.ts
```diff
@@ -69,6 +69,7 @@ export class errors {
     static InvalidResponse(result: Error): Error;
     static ConnectionTimeout(ms: string): Error;
     static ConnectionNotOpenError(): Error;
+    static ConnectionCloseError(event: WebSocketEvent | boolean): Error;
```
nivida: Could you change the error handling to have the same signature as we have it with the other connection errors? 
``static ConnectionCloseError(event: WebSocketEvent | boolean): Error | ConnectionError;``

packages/web3-core-requestmanager/src/index.js
```diff
@@ -272,6 +275,29 @@ RequestManager.prototype.clearSubscriptions = function (keepIsSyncing) {
...
 
+/**
+ * Evaluates WS close event
+ *
+ * @method _isCleanClose
+ *
+ * @returns {boolean}
+ */
+RequestManager.prototype._isCleanCloseEvent = function (event) {
+    return typeof event === 'object' &&
```
nivida: Could you change it to have this on one line? If I'm right does it not exceed the maximum characters per line.

nivida: ... and the parameters are missing in the funcDoc as well

packages/web3-core-requestmanager/src/index.js
```diff
@@ -272,6 +275,29 @@ RequestManager.prototype.clearSubscriptions = function (keepIsSyncing) {
...
+};
+
+/**
+ * Detects Ipc close error. The node.net module emits ('close', isException)
+ *
+ * @method _isIpcCloseError
+ *
+ * @returns {boolean}
+ */
+RequestManager.prototype._isIpcCloseError = function (event) {
```
nivida, line 300: missing params in the funcDoc

packages/web3-core-requestmanager/src/index.js
nivida: I've just added some comments to improve the code style but anything else looks good! Thanks! :)

packages/web3-core-helpers/types/index.d.ts
cgewecke: I was wondering about that - this error is just a normal error with a message though - it doesn't have additional properties like `ConnectionError`.

packages/web3-core-requestmanager/src/index.js
nivida: Is it possible to follow the JSDoc standard? (https://jsdoc.app/tags-param.html)

packages/web3-core-requestmanager/src/index.js
```diff
@@ -272,6 +275,30 @@ RequestManager.prototype.clearSubscriptions = function (keepIsSyncing) {
...
+ */
+RequestManager.prototype._isCleanCloseEvent = function (event) {
+    return typeof event === 'object' && ([1000].includes(event.code) || event.wasClean === true);
+};
+
+/**
+ * Detects Ipc close error. The node.net module emits ('close', isException)
+ *
+ * @method _isIpcCloseError
+ * @param event close event
```
cgewecke:
```suggestion
 * @param {object | boolean} event WS close event or exception flag
```

packages/web3-core-requestmanager/src/index.js
```diff
@@ -272,6 +275,30 @@ RequestManager.prototype.clearSubscriptions = function (keepIsSyncing) {
         this.provider.reset();
 };
 
+/**
+ * Evaluates WS close event
+ *
+ * @method _isCleanClose
+ * @param event close event
+ *
+ * @returns boolean
```
cgewecke:
```suggestion
 * @returns {boolean}
```

packages/web3-core-requestmanager/src/index.js
```diff
@@ -272,6 +275,30 @@ RequestManager.prototype.clearSubscriptions = function (keepIsSyncing) {
         this.provider.reset();
 };
 
+/**
+ * Evaluates WS close event
+ *
+ * @method _isCleanClose
+ * @param event close event
```
cgewecke:
```suggestion
 * @param {object | boolean} event WS close event or exception flag
```

packages/web3-core-requestmanager/src/index.js
```diff
@@ -272,6 +275,30 @@ RequestManager.prototype.clearSubscriptions = function (keepIsSyncing) {
...
+    return typeof event === 'object' && ([1000].includes(event.code) || event.wasClean === true);
+};
+
+/**
+ * Detects Ipc close error. The node.net module emits ('close', isException)
+ *
+ * @method _isIpcCloseError
+ * @param event close event
+ *
+ * @returns boolean
```
cgewecke:
```suggestion
 * @returns {boolean}
```

packages/web3-core-helpers/types/index.d.ts
nivida: I've updated the ``errors.ConnectionCloseError`` type and the related function in the ``web3-core-helpers`` module.

packages/web3-core-requestmanager/src/index.js
nivida: I've improved the ``param`` notation of the ``_isCleanCloseEvent`` funcDoc.

nivida:
@cgewecke I've applied the missing requested changes and will merge it after a quick look at it from you.

nivida (approved):

packages/web3-core-helpers/src/errors.js
```diff
@@ -47,6 +47,18 @@ module.exports = {
     ConnectionNotOpenError: function (event){
         return this.ConnectionError('connection not open on send()', event);
     },
+    ConnectionCloseError: function (event){
+        if (typeof event === 'object' && event.code && event.reason) {
+            return this.ConnectionError(
```
cgewecke, line 52: 👍

cgewecke:
@nivida LGTM thanks!

PR #3330, opened 2020-01-23
Revert non-essential changes (#3190)

cgewecke:
## Description

**PR targets #3190**

@nivida, I doubt you will be fond of this 🙂 Apologies in advance.

Continuing in the spirit of #3303 and getting #3190 into compliance with the PR Guidelines, this PR drops ~40 files from #3190 

Removed are: 
+ package-lock changes for modules without dep changes
+ non-essential TS re-org. (Have preserved the key changes in core-helpers & ws-provider)
+ all build artifacts
+ ENS module changes. (as out of scope)

All tests pass and the change-set is reduced from 63 to 24 files. The new scope can be seen most clearly by re-targeting this PR at 1.x.

Am not questioning the value of the things I'm proposing to remove here - just trying to reduce #3190's complexity and make its diff more legible. 

#3190 rewrites some of the most sensitive and least-tested parts of the code, introducing excellent fixes. This PR just tries to tighten its focus to what's essential. 
   
## Type of change

- [x] Review suggestion


packages/web3-core-method/types/index.d.ts
```diff
@@ -17,6 +17,8 @@
  * @date 2018
  */
 
+import {JsonRpcPayload} from 'web3-core-helpers';
```
nivida, line 20: This import isn't in usage and can get removed because it is totally obvious

packages/web3-core-method/package-lock.json
```diff
@@ -134,7 +134,7 @@
 				"fs-extra": "^6.0.1",
 				"strip-json-comments": "^2.0.1",
 				"tslint": "^5.12.0",
-				"typescript": "^3.8.0-dev.20191128"
+				"typescript": "^3.7.0-dev.20191015"
```
nivida, line 137: Means this package-lock doesn't get updated anyways and I was misleading updating it? 🤔

packages/web3-core-helpers/types/index.d.ts
```diff
@@ -81,16 +81,16 @@ export class errors {
     static TransactionOutOfGasError(receipt: object): TransactionError
 }
 
-export class IpcProviderBase {
-    constructor(path: string, net: net.Server);
+export class WebsocketProviderBase {
```
nivida, line 84: I've just changed the order during copy them back to the old place but if the order is important to you will we definitely revert it. :)

packages/web3-eth-ens/src/contracts/Registry.js
```diff
@@ -38,7 +39,10 @@ function Registry(ens) {
     var self = this;
     this.ens = ens;
     this.contract = ens.checkNetwork().then(function (address) {
-        return new self.ens.eth.Contract(REGISTRY_ABI, address);
+        var contract = new Contract(REGISTRY_ABI, address);
+        contract.setProvider(self.ens.eth.currentProvider);
```
nivida: Those lines can get removed as explained 1-2 times because it was only required because the provider inheritance didn't work as expected in this case. This will also bring up the issue we had with the listener in the last review PR from you.

packages/web3-eth-ens/src/contracts/Registry.js
```diff
@@ -87,7 +91,9 @@ Registry.prototype.resolver = function (name) {
     return this.contract.then(function (contract) {
         return contract.methods.resolver(namehash.hash(name)).call();
     }).then(function (address) {
-        return new self.ens.eth.Contract(RESOLVER_ABI, address);
+        var contract = new Contract(RESOLVER_ABI, address);
+        contract.setProvider(self.ens.eth.currentProvider);
```
nivida: Same here...

packages/web3-providers-http/types/index.d.ts
```diff
@@ -20,8 +20,6 @@
  * @date 2018
  */
 
-import {HttpProviderBase} from 'web3-core-helpers';
+import { HttpProviderBase } from 'web3-core-helpers';
 
-export {HttpProviderOptions, HttpHeader, JsonRpcPayload, JsonRpcResponse} from 'web3-core-helpers';
-
-export class HttpProvider extends HttpProviderBase { }
+export class HttpProvider extends HttpProviderBase {}
```
nivida, line 25: Those additional exports are required to not import http related types from two places it is just a small improvement without any logical change.

packages/web3-providers-http/types/tests/web3-provider-http-tests.ts
```diff
@@ -40,15 +39,13 @@ const options: HttpProviderOptions = {
...
         https: new https.Agent({})
     }
-};
-
-const httpProvider = new HttpProvider('http://localhost:8545', options);
+});
 
 // $ExpectType void
-httpProvider.send(payload, (error: Error | null) => {});
+httpProvider.send({} as any, (error: Error | null) => {});
```
nivida, line 45: This test is weaker than before and I think those type tests should be as strict as possible.

packages/web3-providers-http/types/tsconfig.json
```diff
@@ -11,7 +11,7 @@
         "allowSyntheticDefaultImports": false,
         "baseUrl": ".",
         "paths": {
-            "web3-providers-http": ["."]
+            "web3-providers": ["."]
```
nivida, line 14: This was completely misleading and if possible should we keep the``web3-providers-http`` path. We probably can have both paths to not break anything.

packages/web3-providers-http/types/tslint.json
```diff
@@ -5,7 +5,6 @@
         "no-import-default-of-export-equals": false,
         "file-name-casing": [true, "kebab-case"],
         "whitespace": false,
-        "no-unnecessary-class": false,
-        "no-redundant-jsdoc": false
+        "no-unnecessary-class": false
```
nivida, line 8: Did you never have this strange error message? dtslint was not always happy with having the same copyright/author comment. (@joshstevens19 knows this error too)

packages/web3-providers-ipc/types/index.d.ts
```diff
@@ -20,9 +20,6 @@
  * @date 2018
  */
 
-import * as net from "net";
-import {IpcProviderBase} from 'web3-core-helpers';
-
-export {JsonRpcPayload, JsonRpcResponse} from 'web3-core-helpers';
+import { IpcProviderBase } from 'web3-core-helpers';
```
nivida, line 23: The additional exports can still stay because it isn't any logical change and does improve the TS DX.

packages/web3-providers-ipc/types/tests/web3-provider-ipc-tests.ts
```diff
@@ -49,10 +48,10 @@ ipcProvider.addDefaultEvents();
 ipcProvider.supportsSubscriptions();
 
 // $ExpectType void
-ipcProvider.send(payload, (error: Error | null) => {});
+ipcProvider.send({} as any, (error: Error | null) => {});
```
nivida, line 51: More explicit test cases are to prefer

packages/web3-providers-ipc/types/tsconfig.json
```diff
@@ -11,7 +11,7 @@
         "allowSyntheticDefaultImports": false,
         "baseUrl": ".",
         "paths": {
-            "web3-providers-ipc": ["."]
+            "web3-providers": ["."]
```
nivida, line 14: Same as in the http package.

packages/web3-providers-ws/types/index.d.ts
```diff
@@ -20,14 +20,6 @@
...
-import {WebsocketProviderBase} from 'web3-core-helpers';
-
-export {
-    WebsocketProviderOptions,
-    ReconnectOptions,
-    JsonRpcResponse,
-    JsonRpcPayload,
-    RequestItem
-} from 'web3-core-helpers';
+import { WebsocketProviderBase } from 'web3-core-helpers';
```
nivida, line 23: The additional exports can still stay because it isn't any logical change and does improve the TS DX.

packages/web3-providers-ws/types/tests/web3-provider-ws-tests.ts
```diff
@@ -55,10 +57,10 @@ wsProvider.connected;
 wsProvider.supportsSubscriptions();
 
 // $ExpectType void
-wsProvider.send(payload, (error: Error | null) => {});
+wsProvider.send({} as any, (error: Error | null) => {});
```
nivida, line 60: More explicit test cases are to prefer

packages/web3-providers-ws/types/tsconfig.json
```diff
@@ -11,7 +11,7 @@
         "allowSyntheticDefaultImports": false,
         "baseUrl": ".",
         "paths": {
-            "web3-providers-ws": ["."]
+            "web3-providers": ["."]
```
nivida, line 14: Same as in the other provider packages.

nivida (changes requested):
Thanks for going through. We have to discuss the following points,

- Would you say it is required to have a separate PR for updating package-lock files?
- When do we update the minified file? 
- How are we doing reviews? The duration of the provider improvements reviews and implementation are taking way too long from my point of view. (soon 4 months)

nivida:
@cgewecke btw.: 
Could you please start to use the labels we have and to label those issues and your PRs correctly :)

Edit:
> Continuing in the spirit of #3303 

Is it possible to first finish PR #3303 before we are creating more such PRs? And is it not also possible to comment in the initial PR and I will do those changes as in a normal review process? It will just get harder to follow anything if we create more PRs. We have moved already PRs together to have a better overview of those depending changes.

cgewecke:
> Would you say it is required to have a separate PR for updating package-lock files?

I think if a dependency is changed in a module, the package-lock should also be changed. But otherwise it should not be included. In #3190 every module had its package-lock changed.

> When do we update the minified file? 

In a separate PR - maybe as part of the publication process? It's common to see a build step included as a [prepublish][1] directive in the npm package. 

[1]: https://docs.npmjs.com/misc/scripts#description

> The duration of the provider improvements reviews and implementation are taking way too long from my point of view. (soon 4 months)

Ah! There's a document describing how to do [PRs][4] and [Reviews][3] [here][2]! 

Some parts relevant to #3190 contributing to the long delay and many (unwelcome?) review recommendations are:

+ any ... logic change should include tests 
+ have a narrow, well-defined focus.
+ make the smallest set of changes possible to achieve their goal.
+ preserve the conventions and stylistic consistency of any files they modify.
+ Given the choice between a conservative change that mostly works and an adventurous change which seems better but introduces uncertainty - prefer the conservative change.

[2]: https://github.com/ethereum/web3.js/blob/1.x/CONTRIBUTIONS.md#pull-requests-for-substantive-changes-eg-everything-except-comments-and-docs
[3]: https://github.com/ethereum/web3.js/blob/1.x/CONTRIBUTIONS.md#reviews
[4]: https://github.com/ethereum/web3.js/blob/1.x/CONTRIBUTIONS.md#pull-requests-for-substantive-changes-eg-everything-except-comments-and-docs

packages/web3-core-method/types/index.d.ts
cgewecke: @nivida - Tbh I've just scanned your review - each comment provides a rationale for the change. As I note in the opening comment, I think many of the changes *are* legitimate. But they're also out of scope, add lots of complexity, reduce PR clarity and can be introduced separately. 40 files can be removed from #3190 (non of them test files) with all tests passing. It's not a minimized set.

packages/web3-core-helpers/types/index.d.ts
cgewecke: The diff in this PR is clearer / not mangled.

cgewecke:
> Could you please start to use the labels we have and to label those issues and your PRs correctly :)

Apologies, I will do this.

> Is it possible to first finish PR #3303 before we are creating more such PRs

Can you merge #3303? It just clarifies the test file and now we know that there are test failures when resetting the provider....that's under investigation.

nivida:
> In #3190 every module had its package-lock changed.

This was probably the case because I've changed the dependency graph for the TS improved I've applied which sadly were out of scope. 

> In a separate PR - maybe as part of the publication process? It's common to see a build step included as a prepublish directive in the npm package.

Sounds good to me! 

> Ah! There's a document describing how to do PRs and Reviews here!

True that. I think we should just not loos us in over correct PRs. Bigger out of scope changes as you have for example detected here is definitely something I have to pretend but sometimes do we save 15min - 30min if we just keep smaller improvements instead of doing them twice just because of the correctness. Good examples are improving a funcDoc, adding an explanatory in-line comment while going through a code that isn't in the scope, or increasing of the code style quality in a file you anyways have changed.

nivida:
> Can you merge #3303? It just clarifies the test file and now we know that there are test failures when resetting the provider....that's under investigation.

Yep! I will check this PR again and merge it if anything is fine 👌

packages/web3-core-method/types/index.d.ts
nivida: I see those points as well but instead of just reverting those smaller improvements should we probably also create issues out of them to track those un-clean parts of the ``web3.js`` code base. This is probably also a good chance to create issues labeled with ``good first issue``. WDYT? :)

packages/web3-providers-http/types/index.d.ts
nivida: I've resolved this and created an issue https://github.com/ethereum/web3.js/issues/3334 for doing it later.

packages/web3-providers-http/types/tests/web3-provider-http-tests.ts
nivida: I've resolved this and created a related issue for it (https://github.com/ethereum/web3.js/issues/3335)

packages/web3-providers-http/types/tsconfig.json
nivida: I've resolved this and created a related issue (https://github.com/ethereum/web3.js/issues/3336)

packages/web3-providers-http/types/tslint.json
nivida: I've resolved this because it does only happen sometimes and I will investigate further the next time it happens.

packages/web3-core-method/types/index.d.ts
nivida: I've created issue #3334, #3335, and #3336 to not forget about those required types improvements and to show what I thought with the comment above.

nivida (approved):

PR #3333, opened 2020-01-23
Check build in ci

cgewecke:
## Description

Example for [release 1.2.5 comment][1]

After commenting out the browser job build step, the tests error with
```
Error: Cannot find module '../../packages/web3/dist/web3.min' from '/home/travis/build/ethereum/web3.js/test/helpers'
```
[1]: https://github.com/ethereum/web3.js/pull/3315#issuecomment-577800474

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -28,7 +28,7 @@ elif [ "$TEST" = "unit_and_e2e_clients" ]; then
 
 elif [ "$TEST" = "e2e_browsers" ]; then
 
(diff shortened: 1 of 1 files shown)


nivida:
Yes, this behaviour is because the ``packages/web3/dist/`` folder is excluded with the [.gitignore](https://github.com/ethereum/web3.js/blob/1.x/.gitignore) file and without executing the bundler build pipeline does it not exist.

PR #3344, opened 2020-01-27
Implement FakeIpcProvider.reset (#3190)

cgewecke:
## Description

**PR targets #3190**

This tries to make sure the injection-based tests in `contract.js` run the subscriptions clearing correctly when RequestManager.setProvider is called. 

In #3303 there was quite a bit of confusion about whether or not additional listeners are being attached on a double-call to setProvider...but it looks like the test fixture might just not reflect the code path correctly. 

## Type of change

- [x] Review recommendation

## Checklist:

### DIFF
--- a/test/helpers/FakeIpcProvider.js
+++ b/test/helpers/FakeIpcProvider.js
@@ -68,7 +68,7 @@ FakeIpcProvider.prototype.on = function (type, callback) {
     }
 };
 
-FakeIpcProvider.prototype.reset = function (type, callback) {
+FakeIpcProvider.prototype.reset = function () {
     this.notificationCallbacks = [];
 };
 
--- a/test/helpers/FakeIpcProvider.js
+++ b/test/helpers/FakeIpcProvider.js
@@ -31,6 +31,7 @@ var FakeIpcProvider = function IpcProvider() {
     this.error = [];
     this.validation = [];
     this.notificationCallbacks = [];
+    this.connected = true;
 };
 
 
@@ -65,6 +66,10 @@ FakeIpcProvider.prototype.on = function (type, callback) {
     if(type === 'data') {
         this.notificationCallbacks.push(callback);
     }
+};
+
+FakeIpcProvider.prototype.reset = function (type, callback) {
+    this.notificationCallbacks = [];
 };
 
 FakeIpcProvider.prototype.getResponseOrError = function (type, payload) {
--- a/test/contract.js
+++ b/test/contract.js
@@ -240,7 +240,7 @@ var getEthContractInstance = function(abi, address, options, provider) {
     }
 
     var eth = new Eth(provider);
-    //eth.setProvider(provider);
+    eth.setProvider(provider);
     return new eth.Contract(abi, address, options);
 }
 


test/helpers/FakeIpcProvider.js
```diff
@@ -31,6 +31,7 @@ var FakeIpcProvider = function IpcProvider() {
     this.error = [];
     this.validation = [];
     this.notificationCallbacks = [];
+    this.connected = true;
```
cgewecke, line 34: This field is checked in the RequestManager to clear subscriptions on setProvider here:

https://github.com/ethereum/web3.js/blob/5e3ad84484741e8b8a48ce2ca95e4ae499d7a157/packages/web3-core-requestmanager/src/index.js#L95-L97


It's set on the http provider on send but I can't find where/if it's set in the WS Provider (in the provider-improvements branch anyway). It's part of the TS interface. 

In the web3-js/websocket [implementation][1], it looks like `connection.connected` is set to true in the constructor, and unset on close. 

@nivida Do you think this property should be getting set on connect and unset on close in ws-provider?

[1]: https://github.com/web3-js/WebSocket-Node/blob/polyfill/globalThis/lib/WebSocketConnection.js

nivida: The ``connected`` property does exist as an object accessor [here](https://github.com/ethereum/web3.js/blob/5e3ad84484741e8b8a48ce2ca95e4ae499d7a157/packages/web3-providers-ws/src/index.js#L82) in the ``WebsocketProvider`` class.

test/helpers/FakeIpcProvider.js
```diff
@@ -67,6 +68,10 @@ FakeIpcProvider.prototype.on = function (type, callback) {
     }
 };
 
+FakeIpcProvider.prototype.reset = function (type, callback) {
+    this.notificationCallbacks = [];
```
nivida, line 72: Shouldn't it also reset the ``validation``, ``error``, and ``response`` array as well? and have you tested if this also works with the existing ``FakeIpcProvider2`` class? :)

test/helpers/FakeIpcProvider.js
cgewecke: Ah great!  Everything looks good then.

test/helpers/FakeIpcProvider.js
cgewecke: What would be gained by doing that?

nivida: It would really reset the provider with the anyways no longer used injected items.

cgewecke: I guess, but it's a fake provider - it's seems like its only purpose is to exercise what you want to test and there aren't failing tests for those properties.

nivida: > but it's a fake provider 

Yes, it is. It is just the correct way to clean anything up. Who knows if this has once an impact on our test cases.

cgewecke: Do you mean the lack of realism could have an impact in the future?  Idk - I think this is fine as it is tbh and answers the question in #3303.

nivida: Yes, it could. Alone the wordings we are using do show that we think/thought and do not know. That's why we should implement the ``reset`` method as I was recommending above.

cgewecke: > Alone the wordings we are using do show that we think/thought and do not know. 

Could you clarify this? Do you mean you're uncertain that the reason the tests pass is because the reset is mocked in now?

nivida: No, Chris, all I mean is if we implement this reset method than in the correct way and not just what we think is enough for now. Could you please just do it.

cgewecke: I don't think it's useful or necessary at the moment. If you think it is, add it though.

nivida: Next time will I directly do it and just merge it instead of discussing it with you in such a childish way :)

nivida: > Shouldn't it also reset the validation, error, and response array as well?

The answer would have been: "no this isn't possible because many test cases will fail". :)

nivida (approved):

Issue #3309, opened 2020-01-13
Survey and add tests for all error messages


cgewecke:
As part of 1.x stabilization, Web3 needs a good behavioral description of when and how its error messages are thrown. This is a pre-condition for any more extensive refactoring because it's common for users to grep the errors and base further logic on them. In many cases the message strings themselves have become a de-facto interface that needs to be supported. 

Below is a tracking table - line numbers are based on the repo state at 513116fb05428327df35c8c5c9528a678d6f498c. As things change those numbers will likely float a bit.

| Cov | Package    | File    | Line    |  PR |
| ---------- | ------- | ------- | ------- | ------ |
|  ❌   | bzz | index.js  | 69 |  |
|  ❌   | core| index.js  | 34 |  |
|   ✅  | helpers | formatters.js | 141 |  |
|   ✅  | helpers | formatters.js | 150 |  |
|  ❌   | helpers | formatters.js | 204 |  |
|   ❌  | helpers | formatters.js | 263 |  |
|  ✅   | helpers | formatters.js | 471 |  |
|   ✅  | method | index.js | 36 |  |
|   ✅  | method | index.js | 128 |  |
|   ❌  | method | index.js | 275 |  |
|   ✅  | method | index.js | 287 |  |
|   ✅  | method | index.js | 345 |  |
|   ✅  | method | index.js | 386 |  |
|   ✅  | method | index.js | 432 |  |
|   ✅  | method | index.js | 447 |  |
|   ❌  | method | index.js | 457 |  |
|   ❌ | method | index.js | 484 |  |
|   ✅  | method | index.js | 494 |  |
|   ❌  | method | index.js | 507 |  |
|   ✅  | method | index.js | 581 |  |
|   ✅  | requestmngr | batch.js | 57 |  |
|   ❌  | requestmngr | batch.js | 61 |  |
|   ❌ | requestmngr | givenprovider.js | 56 |  |
|   ❌  | requestmngr | givenprovider.js | 68 |  |
|   ✅  | requestmngr | index.js | 85 |  |
|   ❌   | requestmngr | index.js | 119 |  |
|   ✅   | requestmngr | index.js | 127 |  |
|   ❌  | requestmngr | index.js | 158 |  |
|   ✅  | requestmngr | index.js | 170 |  |
|   ✅  | requestmngr | index.js | 174 |  |
|   ❌  | requestmngr | index.js | 196 |  |
|   ❌ | requestmngr | jsonrpc.js | 42 |  |
[...]

PR #3365, opened 2020-02-05
Add tests for eth-contract errors

cgewecke:
## Description

Part of #3309 . 

Adds tests for errors thrown in the web3-eth-contract module. Per discussion in #3309, where an error is newly covered, its text is also migrated to web3-core-helpers so that these messages start to get consolidated in one place. 

There are some uncovered cases here...

In `_encodeMethodABI` and `_processExecuteArguments`, there are errors for missing deployment data and method param correctness which don't look reachable - those problems are checked in the `deploy` and `_createTxObject` calls which sit earlier in the call chain. 

Haven't removed this logic because it seems more risky than necessary but they could probably be addressed in a later code cleanup.

## Type of change

- [x] Tests

## Checklist:

### DIFF
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -119,19 +119,16 @@ module.exports = {
     ContractEventDoesNotExistError: function(eventName) {
         return new Error('Event "' + eventName + '" doesn\'t exist in this contract.');
     },
-    ContractUnsetContractAddressError: function() {
-        return new Error('This contract object doesn\'t have address set yet, please set an address first.');
-    },
     ContractReservedEventError: function(type) {
         return new Error('The event "'+ type +'" is a reserved event name, you can\'t use it.');
     },
     ContractMissingDeployDataError: function() {
         return new Error('No "data" specified in neither the given options, nor the default options.');
     },
-    ContractMissingToAddressError: function() {
+    ContractNoAddressDefinedError: function() {
         return new Error('This contract object doesn\'t have address set yet, please set an address first.');
     },
-    ContractMissingFromAddressError: function() {
+    ContractNoFromAddressDefinedError: function() {
         return new Error('No "from" address specified in neither the given options, nor the default options.');
     }
 };
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -109,5 +109,29 @@ module.exports = {
     },
     ResolverMethodMissingError: function(address, name) {
         return new Error('The resolver at ' + address + 'does not implement requested method: "' + name + '".');
+    },
+    ContractMissingABIError: function() {
+        return new Error('You must provide the json interface of the contract when instantiating a contract object.');
+    },
+    ContractOnceRequiresCallbackError: function() {
+        return new Error('Once requires a callback as the second parameter.');
+    },
+    ContractEventDoesNotExistError: function(eventName) {
+        return new Error('Event "' + eventName + '" doesn\'t exist in this contract.');
+    },
+    ContractUnsetContractAddressError: function() {
+        return new Error('This contract object doesn\'t have address set yet, please set an address first.');
+    },
+    ContractReservedEventError: function(type) {
+        return new Error('The event "'+ type +'" is a reserved event name, you can\'t use it.');
+    },
+    ContractMissingDeployDataError: function() {
+        return new Error('No "data" specified in neither the given options, nor the default options.');
+    },
+    ContractMissingToAddressError: function() {
+        return new Error('This contract object doesn\'t have address set yet, please set an address first.');
+    },
+    ContractMissingFromAddressError: function() {
+        return new Error('No "from" address specified in neither the given options, nor the default options.');
     }
 };
--- a/packages/web3-eth-contract/src/index.js
+++ b/packages/web3-eth-contract/src/index.js
@@ -71,7 +71,7 @@ var Contract = function Contract(jsonInterface, address, options) {
     this.clearSubscriptions = this._requestManager.clearSubscriptions;
 
     if(!jsonInterface || !(Array.isArray(jsonInterface))) {
-        throw new Error('You must provide the json interface of the contract when instantiating a contract object.');
+        throw errors.ContractMissingABIError();
     }
 
     // create the options object
@@ -342,7 +342,7 @@ Contract.prototype._getCallback = function getCallback(args) {
  */
 Contract.prototype._checkListener = function(type, event){
     if(event === type) {
-        throw new Error('The event "'+ type +'" is a reserved event name, you can\'t use it.');
+        throw errors.ContractReservedEventError(type);
     }
 };
 
@@ -598,9 +598,13 @@ Contract.prototype.deploy = function(options, callback){
     options = this._getOrSetDefaultOptions(options);
 
 
-    // return error, if no "data" is specified
+    // throw error, if no "data" is specified
     if(!options.data) {
-        return utils._fireError(new Error('No "data" specified in neither the given options, nor the default options.'), null, null, callback);
+        if (typeof callback === 'function'){
+            callback(errors.ContractMissingDeployDataError());
+        } else {
+            throw errors.ContractMissingDeployDataError();
+        }
     }
 
     var constructor = _.find(this.options.jsonInterface, function (method) {
@@ -644,11 +648,11 @@ Contract.prototype._generateEventOptions = function() {
         });
 
     if (!event) {
-        throw new Error('Event "' + eventName + '" doesn\'t exist in this contract.');
+        throw errors.ContractEventDoesNotExistError(eventName);
     }
 
     if (!utils.isAddress(this.options.address)) {
-        throw new Error('This contract object doesn\'t have address set yet, please set an address first.');
+        throw errors.ContractUnsetContractAddressError();
     }
 
     return {
@@ -685,7 +689,7 @@ Contract.prototype.once = function(event, options, callback) {
     callback = this._getCallback(args);
 
     if (!callback) {
-        throw new Error('Once requires a callback as the second parameter.');
+        throw errors.ContractOnceRequiresCallbackError();
     }
 
     // don't allow fromBlock
@@ -856,7 +860,7 @@ Contract.prototype._processExecuteArguments = function _processExecuteArguments(
 
     // add contract address
     if(!this._deployData && !utils.isAddress(this._parent.options.address))
-        throw new Error('This contract object doesn\'t have address set yet, please set an address first.');
+        throw errors.ContractMissingToAddressError();
 
     if(!this._deployData)
         processedArgs.options.to = this._parent.options.address;
@@ -945,7 +949,7 @@ Contract.prototype._executeMethod = function _executeMethod(){
 
             // return error, if no "from" is specified
             if(!utils.isAddress(args.options.from)) {
-                return utils._fireError(new Error('No "from" address specified in neither the given options, nor the default options.'), defer.eventEmitter, defer.reject, args.callback);
+                return utils._fireError(errors.ContractMissingFromAddressError(), defer.eventEmitter, defer.reject, args.callback);
             }
 
             if (_.isBoolean(this._method.payable) && !this._method.payable && args.options.value && args.options.value > 0) {
--- a/packages/web3-eth-contract/src/index.js
+++ b/packages/web3-eth-contract/src/index.js
@@ -601,10 +601,9 @@ Contract.prototype.deploy = function(options, callback){
     // throw error, if no "data" is specified
     if(!options.data) {
         if (typeof callback === 'function'){
-            callback(errors.ContractMissingDeployDataError());
-        } else {
-            throw errors.ContractMissingDeployDataError();
+            return callback(errors.ContractMissingDeployDataError());
         }
+        throw errors.ContractMissingDeployDataError();
     }
 
     var constructor = _.find(this.options.jsonInterface, function (method) {
--- a/packages/web3-eth-contract/src/index.js
+++ b/packages/web3-eth-contract/src/index.js
@@ -651,7 +651,7 @@ Contract.prototype._generateEventOptions = function() {
     }
 
     if (!utils.isAddress(this.options.address)) {
-        throw errors.ContractUnsetContractAddressError();
+        throw errors.ContractNoAddressDefinedError();
     }
 
     return {
@@ -859,7 +859,7 @@ Contract.prototype._processExecuteArguments = function _processExecuteArguments(
 
     // add contract address
     if(!this._deployData && !utils.isAddress(this._parent.options.address))
-        throw errors.ContractMissingToAddressError();
+        throw errors.ContractNoAddressDefinedError();
 
     if(!this._deployData)
         processedArgs.options.to = this._parent.options.address;
@@ -948,7 +948,7 @@ Contract.prototype._executeMethod = function _executeMethod(){
 
             // return error, if no "from" is specified
             if(!utils.isAddress(args.options.from)) {
-                return utils._fireError(errors.ContractMissingFromAddressError(), defer.eventEmitter, defer.reject, args.callback);
+                return utils._fireError(errors.ContractNoFromAddressDefinedError(), defer.eventEmitter, defer.reject, args.callback);
             }
 
             if (_.isBoolean(this._method.payable) && !this._method.payable && args.options.value && args.options.value > 0) {
--- a/test/contract.errors.js
+++ b/test/contract.errors.js
@@ -98,8 +98,8 @@ describe('contract: errors', function () {
             "type": "event"
         }];
 
-        const address = '0xEE221E529cF6DB20179E7bAeb4442e9CbdCa83d7';
-        const contract = new eth.Contract(newListenerEventAbi, address);
+        const contractAddress = '0xEE221E529cF6DB20179E7bAeb4442e9CbdCa83d7';
+        const contract = new eth.Contract(newListenerEventAbi, contractAddress);
 
         try {
             contract.once('newListener', () => {});
@@ -137,7 +137,7 @@ describe('contract: errors', function () {
         });
     })
 
-    it('errors when send is called without a contract address set', function(){
+    it('errors when send is called without a *contract address* set', function(){
         const expected = 'This contract object doesn\'t have address set yet, ' +
                          'please set an address first.';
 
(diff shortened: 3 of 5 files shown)


nivida:
@cgewecke Is this ready for review? :)

nivida:
@cgewecke I've scrolled through shortly and I think we have a problem of the scope here. It is a code refactoring which actually does make sense but the scope is to extend the test cases and not to refactor any other code. (Violates [point 4.2](https://github.com/ethereum/web3.js/blob/1.x/CONTRIBUTIONS.md#pull-requests-for-substantive-changes-eg-everything-except-comments-and-docs) of the PR contribution rules).

cgewecke:
> Is this ready for review

There are couple of errors which need a little more investigation before this is done. 

> It is a code refactoring

Do you mean the transfer of error messages to helpers/errors? I've tried to follow what you advise in this [3309 comment][1] but I can break it up into separate PRs if you prefer - it's no problem. 

[1]: https://github.com/ethereum/web3.js/issues/3309#issuecomment-575039387

nivida:
> There are couple of errors which need a little more investigation before this is done.

👌

> Do you mean the transfer of error messages to helpers/errors? I've tried to follow what you advise in this 3309 comment but I can break it up into separate PRs if you prefer - it's no problem.

Oh, I see. Sorry, I was only reading the issue description itself and didn't know you will do this as well. I think the answer was missing on my proposal. :) 
~~Is it possible to add the scope of each PR to the issue description as clear tasks? It would help us to see the clear scope of any PR related to this issue/[epic](https://www.atlassian.com/agile/project-management/epics).~~ 

Edit: 

Late-night comments aren't a good idea. I see you have a PR column. Just add this PR to the correct items. This will help me to see the on-going and missing PRs of this epic.

nivida:
@cgewecke Btw.:
Are such force pushes required?

Bildschirmfoto 2020-02-05 um 23 01 03

cgewecke:
> Are such force pushes required?

Ah! I used `rebase -i` vs. 1.x to resolve the conflicts from yesterday. How would you prefer to do this?

nivida:
> Ah! I used rebase -i vs. 1.x to resolve the conflicts from yesterday. How would you prefer to do this?

Oh, I see! I'm currently just doing normal git merges. This is probably something we should discuss and add to our contribution guidelines. I will open a small issue for this tomorrow. Thanks for explaining the reason of this force push! :)

packages/web3-eth-contract/src/index.js
```diff
@@ -581,9 +581,13 @@ Contract.prototype.deploy = function(options, callback){
...
 
-    // return error, if no "data" is specified
+    // throw error, if no "data" is specified
     if(!options.data) {
-        return utils._fireError(new Error('No "data" specified in neither the given options, nor the default options.'), null, null, callback);
+        if (typeof callback === 'function'){
+            callback(errors.ContractMissingDeployDataError());
+        } else {
+            throw errors.ContractMissingDeployDataError();
+        }
```
cgewecke: This block is the only real logic change. Fixes bug where `utils._fireError` is usually swallowing this error.

`contract.deploy` doesn't have an event emitter or a promise associated with it (and AFAIK, isn't really invoked with a callback) 

Have rewritten it into a thrown error but kept backwards compat for the callback in case anyone actually doing that.

packages/web3-eth-contract/src/index.js
```diff
@@ -581,9 +581,12 @@ Contract.prototype.deploy = function(options, callback){
...
 
 
-    // return error, if no "data" is specified
+    // throw error, if no "data" is specified
     if(!options.data) {
-        return utils._fireError(new Error('No "data" specified in neither the given options, nor the default options.'), null, null, callback);
+        if (typeof callback === 'function'){
+            return callback(errors.ContractMissingDeployDataError());
+        }
+        throw errors.ContractMissingDeployDataError();
```
cgewecke, line 606: This block is the only real logic change. Fixes bug where `utils._fireError` is usually swallowing this error.

`contract.deploy` doesn't have an event emitter or a promise associated with it (and AFAIK, isn't really invoked with a callback) 

Have rewritten it into a thrown error but kept backwards compat for the callback in case anyone actually doing that.

packages/web3-core-helpers/src/errors.js
```diff
@@ -76,5 +76,29 @@ module.exports = {
     },
     TransactionOutOfGasError: function(receipt) {
         return this.TransactionError('Transaction ran out of gas. Please provide more gas:\n' + JSON.stringify(receipt, null, 2), receipt);
+    },
+    ContractMissingABIError: function() {
```
nivida, line 113: Just an idea we probably could implemented on a later stage. The improvement I'm thinking about is to extend the returned ``Error`` object with additional properties but only if it actually does make sense. This will help our consumers to show their users better error messages or to improve their code (they do not have to parse a string).

packages/web3-core-helpers/src/errors.js
```diff
@@ -76,5 +76,29 @@ module.exports = {
...
+        return new Error('You must provide the json interface of the contract when instantiating a contract object.');
+    },
+    ContractOnceRequiresCallbackError: function() {
+        return new Error('Once requires a callback as the second parameter.');
+    },
+    ContractEventDoesNotExistError: function(eventName) {
+        return new Error('Event "' + eventName + '" doesn\'t exist in this contract.');
```
nivida, line 120: Sorry.. another improvement idea we should implement. Instead of string concatenating should we use the template literal.

packages/web3-eth-contract/src/index.js
```diff
@@ -627,11 +630,11 @@ Contract.prototype._generateEventOptions = function() {
         });
 
     if (!event) {
-        throw new Error('Event "' + eventName + '" doesn\'t exist in this contract.');
+        throw errors.ContractEventDoesNotExistError(eventName);
     }
 
     if (!utils.isAddress(this.options.address)) {
-        throw new Error('This contract object doesn\'t have address set yet, please set an address first.');
+        throw errors.ContractUnsetContractAddressError();
```
nivida: I would rename this error too: ``ContractNoAddressDefinedError``

packages/web3-eth-contract/src/index.js
```diff
@@ -832,7 +835,7 @@ Contract.prototype._processExecuteArguments = function _processExecuteArguments(
 
     // add contract address
     if(!this._deployData && !utils.isAddress(this._parent.options.address))
-        throw new Error('This contract object doesn\'t have address set yet, please set an address first.');
+        throw errors.ContractMissingToAddressError();
```
nivida: This error is a duplicated error. Please use the ``ContractUnsetContractAddressError`` and after renaming it ``ContractNoAddressDefinedError``.

packages/web3-eth-contract/src/index.js
```diff
@@ -921,7 +924,7 @@ Contract.prototype._executeMethod = function _executeMethod(){
 
             // return error, if no "from" is specified
             if(!utils.isAddress(args.options.from)) {
-                return utils._fireError(new Error('No "from" address specified in neither the given options, nor the default options.'), defer.eventEmitter, defer.reject, args.callback);
+                return utils._fireError(errors.ContractMissingFromAddressError(), defer.eventEmitter, defer.reject, args.callback);
```
nivida: I would rename this for consistency reasons from ``ContractMissingFromAddressError`` to ``ContractNoFromAddressDefinedError``

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
+const assert = require('assert');
+const Eth = require('../packages/web3-eth');
+const FakeIpcProvider = require('./helpers/FakeIpcProvider');
+
+const abi = [{
```
nivida, line 5: It would probably be smart to use an ABI item from the testing contracts we have in our test-suite.

Side note: 
We should create a QA improvement issue to generate those ABIs on the execution of a test. This because the output of the compiler could change.

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+    stateMutability: "view",
+    inputs: [{ name: "a", type: "bytes32" }],
+    outputs: [{ name: "b", type: "bytes32" }],
+}];
+
+describe('contract: errors', function () {
+    let provider;
+    let eth;
+
+    before(function(){
```
nivida: Missing spaces ``function () { ...``

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+        provider = new FakeIpcProvider();
+        eth = new Eth(provider);
+    })
+
+    it('errors when no ABI is provided / ABI is not array', function () {
```
nivida: Is it possible to have descriptions that do have the error function we test included? This would increase the readability of the test outputs.

packages/web3-core-helpers/src/errors.js
nivida: Is it possible to have those error messages covered with real unit tests? What I mean with this would be to have those functions tested as a unit and also tested in the unit/integration mix tests we have in the contract tests of the legacy codebase.

packages/web3-eth-contract/src/index.js
nivida: Isn't it possible to handle this the same way as on all other places e.g.:
```
utils._fireError(errors.ContractMissingFromAddressError(), defer.eventEmitter, defer.reject, args.callback);
```

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+        const expected = 'You must provide the json interface of the ' +
+                         'contract when instantiating a contract object.';
+
+        // Existence check
```
nivida: Can you split up these two test cases to two test cases?

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+        // Array check
+        try {
+            new eth.Contract({});
+            assert.fail();
+        } catch(err){
+            assert(err.message.includes(expected));
+        }
+    })
+
+    it('errors when event listener is not provided with callback', function(){
```
nivida:
```suggestion
    it('errors when event listener is not provided with callback', function () {
```

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+
+        try {
+            contract.once('data');
+            assert.fail();
+        } catch(err){
+            assert(err.message.includes(expected));
+        }
+    })
+
+    it('errors when a non-existant event is listened for', function(){
```
nivida:
```suggestion
    it('errors when a non-existant event is listened for', function () {
```

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+        const expected = 'Event "void" doesn\'t exist in this contract';
+        const contract = new eth.Contract(abi);
+
+        try {
+            contract.once('void', () => {});
+            assert.fail();
+        } catch(err){
```
nivida:
```suggestion
        } catch (err) {
```

nivida: Could you apply those code style improvements anywhere? I think you can already read out the pattern I would like to have.

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+            assert(err.message.includes(expected));
+        }
+    })
+
+    it('errors when listening for event without setting address', function(){
+        const expected = 'This contract object doesn\'t have address set yet, ' +
+                         'please set an address first';
+
+        const eventAbi = [{
```
nivida, line 77: Why not using the Basic contract ABI we have?

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+        } catch(err){
+            assert(err.message.includes(expected));
+        }
+    })
+
+    it('errors when an event name is reserved', function(){
+        const expected = 'The event "newListener" is a reserved event name, ' +
+                         'you can\'t use it.';
+
+        const newListenerEventAbi = [{
```
nivida, line 98: Please add missing ABI items to a shared ABI file or add it to the contract if those files do get compiled for our tests.

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+        const newAbi = [{
+          "inputs": [],
+          "payable": false,
+          "stateMutability": "nonpayable",
+          "type": "constructor"
+        }];
+
+        const contract = new eth.Contract(newAbi);
+
+        // Thrown error format
```
nivida: Can you split up those two test cases to two test cases?

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+        });
+    })
+
+    it('errors when send is called without a *contract address* set', function(){
+        const expected = 'This contract object doesn\'t have address set yet, ' +
+                         'please set an address first.';
+
+        const contract = new eth.Contract(abi);
+        try {
+            contract.methods.simpleMethod("0xaaa").send();
```
nivida, line 148: Can you add the callback and promise case as separate test cases?

nivida (dismissed):
Thanks for opening the first PR of the epic #3309. I've added the first requested changes and will after check the coverage report to be sure anything is covered as expected.

test/contract.errors.js
nivida: This isn't required if we create real unit tests for the newly added error functions.

packages/web3-eth-contract/src/index.js
nivida: This just because of consistency reasons. In the future, it can have some advantages for us.

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+    name: "simpleMethod",
+    payable: false,
+    constant: true,
+    type: "function",
+    stateMutability: "view",
+    inputs: [{ name: "a", type: "bytes32" }],
+    outputs: [{ name: "b", type: "bytes32" }],
+}];
+
+describe('contract: errors', function () {
```
nivida, line 15: I think we should define a clear pattern to name test files and test cases. This will increase the readability of our test outputs extremely. Can you work something out as QA officer? ;-P

test/contract.errors.js
```diff
@@ -0,0 +1,166 @@
...
+        const contract = new eth.Contract(abi);
+        try {
+            contract.methods.simpleMethod("0xaaa").send();
+            assert.fail();
+        } catch(err){
+            assert(err.message.includes(expected));
+        }
+    });
+
+    it('errors when send is called without a *from address* being set', async function(){
```
nivida: Can you add the callback and promise case as separate test cases?

nivida:
> In _encodeMethodABI and _processExecuteArguments, there are errors for missing deployment data and method param correctness which don't look reachable

Anything will be reachable or the error will never get thrown on usage. :)
We can check that closer together after the first request changes are implemented.

nivida:
> Per discussion in #3309, where an error is newly covered, its text is also migrated to web3-core-helpers so that these messages start to get consolidated in one place.

Thanks for updating the PR description. But could you update the list [here](https://github.com/ethereum/web3.js/issues/3309) with the related PR? This helps me to see what is in-progress what is finished and which tests are missing.

packages/web3-core-helpers/src/errors.js
cgewecke: > Is it possible to have those error messages covered with real unit tests?

If the error tests include the entire message I think we get good test coverage for this without needing to duplicate anything. 

Another thing is that if you look at the coverage report for helpers you can see the some errors are *not* being thrown in the tests - that information is useful but would be hidden by direct units.

test/contract.errors.js
cgewecke: I think this is clearer and more explicit.

test/contract.errors.js
cgewecke: Tbh I think it's helpful to have a small, simple ABIs included here. It resembles the pattern in `test/contract.js` which this test file is a sibling of and gives a later test writer a pattern to adapt if they need to quickly make a specific ABI case. 

These small ABIs let people see a minimal input to the test cases for clarity's sake. It's normal to have the inputs declared in or near the test if they aren't too big.

test/contract.errors.js
cgewecke: This one is directly thrown, rather than passed to utils._fireError

test/contract.errors.js
cgewecke: I don't understand the readability issue you're identifying here. These test titles are typical behavioral descriptions, similar to all the other tests. 

If one fails, the behavioral change should be looked into. The specific error function doesn't seem relevant - it just wraps a string.

test/contract.errors.js
cgewecke: Same, I'd prefer to keep these small ABIs here, where they are immediately available to the test reader. 

I've removed a couple that weren't actually required.

packages/web3-eth-contract/src/index.js
cgewecke: Nice catch!

packages/web3-eth-contract/src/index.js
cgewecke: I'm not sure, actually. `deploy` isn't async - it returns a "transaction object". The eventEmitter and promise args to _fireError are explicitly set to null in the original.

So this is a pre-flight error that's accidentally getting swallowed by utils._fireError when it's invoked synchronously like
```
myContract.deploy(options);
myContract.deploy(options).encodeABI(); 
```

packages/web3-core-helpers/src/errors.js
nivida: > If the error tests include the entire message I think we get good test coverage for this without needing to duplicate anything.

Green coverage !== good unit tests. I would test them as a unit as the definition of unit tests are.

test/contract.errors.js
nivida: I agree. 👌

test/contract.errors.js
nivida: One ``it`` one case is the rule. Could you split it up in two cases with a more specific test case description?

test/contract.errors.js
nivida: You would also create two test cases if we would test the ``utils._fireError`` function. One test case one ``it``.

test/contract.errors.js
nivida: Any idea here @cgewecke?

nivida: We should at least create an issue out of it.

nivida:
> Thanks for updating the PR description. But could you update the list here with the related PR? This helps me to see what is in-progress what is finished and which tests are missing.

Reminder :-)

nivida:
> > In _encodeMethodABI and _processExecuteArguments, there are errors for missing deployment data and method param correctness which don't look reachable

> Anything will be reachable or the error will never get thrown on usage. :)

Can you elaborate closer why this isn't possible? I have a strong feeling this is testable ;-P

packages/web3-core-helpers/src/errors.js
cgewecke: Have looked at this again... I think we just disagree. Hitting these errors via the methods which call them seems like the correct approach to me.

test/contract.errors.js
cgewecke: This problem is detected before anything gets routed to a callback or promise.

ryanio (approved):
This is great, really nice to standardize the errors like this.

ryanio:
If you have ideas for further work on this, perhaps you can document in an issue or add in a follow-up comment to #3309?

PR #3378, opened 2020-02-18
Assign window.ethereum to givenProvider when available

### LINKED ISSUE #3374: Make web3.js side-effect free
[Web3.js is modifying the `window.ethereum`](https://github.com/ethereum/web3.js/blob/a5e832fc20088cf4c4e916bc0ec3107ceb4428d0/packages/web3-core-requestmanager/src/givenProvider.js#L44-L45) object, which can cause different issues with other libraries trying to interact with `window.ethereum`.

I think `window.ethereum` should be considered immutable, and any modification to its API, if needed, should be done internally.

More generally, I think a library like web3.js should not have any side effect.

cgewecke:
## Description

#3374 

Metamask is rolling out their upgrade to an [EIP-1193][1] compliant provider while preserving  `web3.currentProvider.sendAsync` method for backwards compatibility.

It looks like the original idea in `RequestManager/givenProvider` was to defer to the injected provider, but the standard's name for this changed at some point in the last 2 years and code here was not updated. 

**Resources**:
+ patch for overwrite at [aragon/use-wallet][2]
+ declaration of intent at [Metamask 7710][3] (Dec 2019)

**Questions:**
+ Is this a breaking change?  (Probably not, see [comment][4])
+ Can web3 consume the new Metamask provider? (Probably, see [comment][5] but this may need more investigation wrt EIP 1193)

[1]: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md
[2]: https://github.com/aragon/use-wallet/pull/21/files#diff-2b4ca49d4bb0a774c4d4c1672d7aa781R22
[3]: https://github.com/MetaMask/metamask-extension/issues/7710#issuecomment-568534489
[4]: https://github.com/ethereum/web3.js/pull/3378#issuecomment-588943517
[5]: https://github.com/ethereum/web3.js/pull/3378#issuecomment-589445107

Fixes #3374 (sort of...the issue makes a more general point about givenProvider side-effects)

## Type of change

- [x] Bug fix (possibly breaking...)

## Checklist:


nivida:
> Is this a breaking change?

This shouldn't break anything from my point of view. 

> Can web3 consume the new Metamask provider?

If they provide the mentioned backward compatibility yes. Otherwise, would we have to update the ``RequestManager`` and probably also the subscription object to support the EIP-1193. But I would vote to first finalize the EIP and to add the ``getCapabilitites`` method we talked about during devcon in Osaka. The ``getCapabilitites`` is required for libs to detect the EIP-1193 provider accordingly.

nivida:
Open discussion EIP-1193: https://github.com/ethereum/EIPs/issues/2319

cgewecke:
@nivida Thanks! 

> If they provide the mentioned backward compatibility yes

Dan Finlay says explicitly in Metamask 7710 that:

> The sendAsync() method is not going to change or break, so using it is one safe way to avoid these issues.

...so that base seems covered. 

Additional background: there's a Metamask [blog post][1] from Nov 6 advising developers to the follow this pattern wrt account access:
```javascript
window.addEventListener('load', async () => {
    // Modern dapp browsers...
    if (window.ethereum) {
        window.web3 = new Web3(ethereum);
        try {
            // Request account access if needed
            await ethereum.enable();
            // Acccounts now exposed
            web3.eth.sendTransaction({/* ... */});
        } catch (error) {
            // User denied account access...
        }
    }
    // Legacy dapp browsers...
    else if (window.web3) {
        window.web3 = new Web3(web3.currentProvider);
        // Acccounts always exposed
        web3.eth.sendTransaction({/* ... */});
    }
    // Non-dapp browsers...
    else {
        console.log('Non-Ethereum browser detected. You should consider trying MetaMask!');
    }
});
```

[1]: https://medium.com/metamask/https-medium-com-metamask-breaking-change-injecting-web3-7722797916a8

nivida:
FYI: ``await ethereum.enable();`` !== EIP-1193 :)

docs/getting-started.rst
```diff
@@ -27,7 +27,8 @@ First you need to get web3.js into your project. This can be done using the foll
 - pure js: link the ``dist/web3.min.js``
 
 After that you need to create a web3 instance and set a provider.
-Ethereum supported Browsers like Mist or MetaMask will have a ``ethereumProvider`` or ``web3.currentProvider`` available. For  web3.js, check ``Web3.givenProvider``.
+Ethereum supported Browsers like Mist or MetaMask will have a ``ethereumProvider`` or ``ethereum``
```
ryanio: We can probably remove the reference to `Mist` here since it is deprecated. Can add e.g. "Trust Wallet" which injects a provider.

cgewecke:
```suggestion
Ethereum supported Browsers like TrustWallet or MetaMask will have an ``ethereumProvider`` or ``ethereum``
```

cgewecke: Corrected in a1750bd

ryanio: thx!

packages/web3-core-requestmanager/src/givenProvider.js
```diff
@@ -34,9 +34,16 @@ try {
 }
 
 // EthereumProvider
+//
+// TODO: removable? This `if` clause looks like an early, subsequently abandoned
```
ryanio: Yes I believe `ethereumProvider` is a pre-EIP1193 artifact. It could make sense to keep it around but not sure if anyone is offering it anymore.

docs/getting-started.rst
```diff
@@ -27,7 +27,8 @@ First you need to get web3.js into your project. This can be done using the foll
 - pure js: link the ``dist/web3.min.js``
 
 After that you need to create a web3 instance and set a provider.
-Ethereum supported Browsers like Mist or MetaMask will have a ``ethereumProvider`` or ``web3.currentProvider`` available. For  web3.js, check ``Web3.givenProvider``.
+Ethereum supported Browsers like TrustWallet or MetaMask will have an ``ethereumProvider`` or ``ethereum``
```
ryanio: Hm, could be reworded like:
> Most Ethereum-supported browsers like MetaMask and TrustWallet have an [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) compliant provider available at `window.ethereum`. For web3.js, check `Web3.givenProvider`.

edit: added in https://github.com/ethereum/web3.js/pull/3378/commits/0223059d04e4a777cf864029c2c108e3f7ff56e9

cgewecke:
Note: this PR was originally put on hold because the mosaic repo instantiated Web3 for their test utilities as below:

```js
const web3 = new Web3(Web3.givenProvider || 'http://localhost:8545');
```

and called it like this:
```js
    web3.currentProvider.send({
      method: 'evm_mine',
      jsonrpc: '2.0',
      id: 1337,
    },
    (err) => {
      if (err) {
        return reject(err);
      }

      const newBlockHash = web3.eth.getBlock('latest').hash;

      return resolve(newBlockHash);
    });
  });
```

However in Buidler, `global.ethereum`'s `send` method is async.  These calls errored with:
```
Error: Method [object Object] not supported.
    at HttpProvider.send (/home/travis/build/ethereum/web3.js/mosaic-1/node_modules/@nomiclabs/buidler/src/internal/core/providers/http.ts:55:34)
    at wrapper_1.wrapSend (/home/travis/build/ethereum/web3.js/mosaic-1/node_modules/@nomiclabs/buidler/src/internal/core/providers/accounts.ts:219:21)
```

I've just removed the mosaic fork's Web3 instantiation logic for unrelated reasons and the test now passes, but I think it might still be a case to be aware of. 

Buidler's `ethereum` provider is being revised in for backwards compatibility with `sendAsync` in [buidler PR 608][1]. 

[1]: https://github.com/nomiclabs/buidler/pull/608

ryanio:
great thanks, good to know.

I'd also like to add and use the new [`request`](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#request) method from EIP-1193 with fallback to `sendAsync`, but can do in a separate PR if you think that would be better. I've been relayed that it should be in MetaMask prod sometime next week, available now as a rc build [here](https://github.com/MetaMask/metamask-extension/pull/8718#issuecomment-637247504).

cgewecke:
> I'd also like to add and use the new request method from EIP-1193 with fallback to sendAsync

Sounds good to me!

ryanio:
great thanks, on second thought I'll do it in a follow-up PR to keep this one targeted for the intended issue.

ryanio (approved):
LGTM!

Issue #2211, opened 2019-01-23
web3 -- web worker environment isn't supported


7-of-9:
Hi all,

Beta 34 (and/or its dependencies, e.g. any-promise) seems to have dependencies on "window" instead of "self" which prevents it working in a web worker context. I've not yet tested on later betas, but suspect they suffer from the same.

Related:  https://github.com/ethereum/web3.js/issues/1774
Related: https://ethereum.stackexchange.com/questions/56405/integrating-web3-from-metamask-in-react

nivida:
This issue got solved in the version 1.0.0-beta.38 which I will release within the next days.

7-of-9:
@nivida fyi -- re. beta41: I'm seeing "window is not defined" when creating a  new Web3.providers.WebsocketProvider(...) inside A web worker context.

(< beta38 threw much earlier: at "import Web3...", so it's closer to working!)

UPDATE: `self.window = self`
In web worker before new WebsockProvider() seems to fix this.

Fix at source would be line 1162 of web3-provider-umd.js (beta41) - just remove "Window."

naturalfreak:
same problem with version 1.2.4

timothywangdev:
still broken

keirongulrajani:
When I've tried `importScripts` in a Web Worker using `web3.js` instead of `web3.min.js` from the node module i can see the `window is not defined` error being thrown when `register-shim` is being loaded; this is a file inside `any-promise`, and the `any-promise` library is used by `web3-core-promievent` in the current version.

The `loadImplementation()` inside `register-shim.js` is attempting to grab `window.Promise` when there is no `window`:
```
/**
 * Browser specific loadImplementation.  Always uses `window.Promise`
 *
 * To register a custom implementation, must register with `Promise` option.
 */
function loadImplementation(){
  if(typeof window.Promise === 'undefined'){
    throw new Error("any-promise browser requires a polyfill or explicit registration"+
      " e.g: require('any-promise/register/bluebird')")
  }
  return {
    Promise: window.Promise,
    implementation: 'window.Promise'
  }
}
```

perhaps doing as it says to register a custom implementation with a `Promise` option is the solution?

EDIT: for those experiencing this issue with the window being undefined, I found a solution [here](https://stackoverflow.com/a/22160937/755370) about setting `window` as `self` first before the `importScripts` statements so that it has the correct environment (effectively tricking `any-promise` under the hood while loading to treat the worker as `window`:

```
var window = self;
[...]

Issue #1774, opened 2018-07-13
Using web3 inside a web worker


26000:
Is it possible? When i `importScripts("web3.min.js")` inside one, says `Uncaught ReferenceError: Web3 is not defined` when i try to call it. I tried browserify with webworkify, but it said `Uncaught ReferenceError: window is not defined`. So does web3 necessarily need `window`, which is inaccessible from a web worker context, or there's a workaround?

nivida:
Please ask this question on https://ethereum.stackexchange.com/

PR #3381, opened 2020-02-21
Remove any-promise dependency

### LINKED ISSUE #3377: web3 -- Web Worker environment still is not supported in 1.2.4
Hi all,

`1.2.4` (and/or its dependencies, e.g. any-promise) seems to have dependencies on "window" instead of "self" which prevents it working in a Web Worker context.

When I've tried `importScripts` in a Web Worker using `web3.js` instead of `web3.min.js` from the node module i can see the `window is not defined` error being thrown when `register-shim` is being loaded; this is a file inside `any-promise`, and the `any-promise` library is used by `web3-core-promievent` in the current version.

The `loadImplementation()` inside `register-shim.js` is attempting to grab `window.Promise` when there is no `window`:
```
/**
 * Browser specific loadImplementation.  Always uses `window.Promise`
 *
 * To register a custom implementation, must register with `Promise` option.
 */
function loadImplementation(){
  if(typeof window.Promise === 'undefined'){
    throw new Error("any-promise browser requires a polyfill or explicit registration"+
      " e.g: require('any-promise/register/bluebird')")
  }
  return {
    Promise: window.Promise,
    implementation: 'window.Promise'
  }
}
```

It is possible to work around this by setting `window` as `self` first before the `importScripts` statements so that it has the correct environment (effectively tricking `any-promise` under the hood while loading to treat the worker as `window`:

```
var window = self;
importScripts(/*path to web 3 goes here*/);
```

however, there is still an issue of passing/accessing the current provider details to the web worker environment, as there's no way to access this information from the browser instance of web3. It is also not ideal having to create an entirely new instance of web3 inside the web worker instead of using an existing one.

Related: #2211 #1774
Related: https://ethereum.stackexchange.com/questions/56405/integrating-web3-from-metamask-in-react

cgewecke:
## Description

The [any-promise][5] package has an [unsafe dereferencing][1] of the `window` object in its code that throws in the Web Worker environment. This PR removes it and falls back on native promises.

**What does any-promise do?** (from their README)
> NOTE: You probably want native promises now

> Let your library support any ES 2015 (ES6) compatible Promise and leave the choice to application authors. The application can optionally register its preferred Promise implementation and it will be exported when requiring any-promise from library code.

> If no preference is registered, defaults to the global Promise for newer Node.js versions. The browser version defaults to the window Promise, so polyfill or register as necessary.

Essentially, consumers of Web3 can decide which promise to use and Web3 will respect that, since it (usually) requires Promise from `any-promise`.

Hard to know if anyone is relying on this...

**Resources**
+ [Most browsers support native Promises][2] - exceptions are IE11 and Opera mini. 
+ IE 11 [lifecycle FAQ][3] advises users to switch to IE Edge
+ PR #1415 (March 2018) standardized use of `any-promise` at Web3
 
[1]: https://github.com/harttle/liquidjs/issues/73
[2]: https://caniuse.com/#feat=promises
[3]: https://support.microsoft.com/en-us/help/17454/lifecycle-faq-internet-explorer
[5]: https://github.com/kevinbeaty/any-promise

Fixes #3377 #2211 #1774

## Type of change

- [x] Bug fix (probably non-breaking?)

## Checklist:

- [x] I have executed ``npm run test:cov`` and my test cases do cover all lines and branches of the added code.

### DIFF
--- a/packages/web3-core-promievent/package-lock.json
+++ b/packages/web3-core-promievent/package-lock.json
@@ -4,11 +4,6 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
-		"any-promise": {
-			"version": "1.3.0",
-			"resolved": "https://registry.npmjs.org/any-promise/-/any-promise-1.3.0.tgz",
-			"integrity": "sha1-q8av7tzqUugJzcA3au0845Y10X8="
-		},
 		"eventemitter3": {
 			"version": "3.1.2",
 			"resolved": "https://registry.npmjs.org/eventemitter3/-/eventemitter3-3.1.2.tgz",
--- a/packages/web3-core-promievent/package.json
+++ b/packages/web3-core-promievent/package.json
@@ -9,7 +9,6 @@
     },
     "main": "src/index.js",
     "dependencies": {
-        "any-promise": "1.3.0",
         "eventemitter3": "3.1.2"
     }
 }
--- a/packages/web3-core-promievent/src/index.js
+++ b/packages/web3-core-promievent/src/index.js
@@ -23,7 +23,6 @@
 "use strict";
 
 var EventEmitter = require('eventemitter3');
-var Promise = require("any-promise");
 
 /**
  * This function generates a defer promise and adds eventEmitter functionality to it
--- a/packages/web3-eth-accounts/package-lock.json
+++ b/packages/web3-eth-accounts/package-lock.json
@@ -52,6 +52,18 @@
 				"semver": "^6.3.0"
 			}
 		},
+		"@web3-js/websocket": {
+			"version": "1.0.30",
+			"resolved": "https://registry.npmjs.org/@web3-js/websocket/-/websocket-1.0.30.tgz",
+			"integrity": "sha512-fDwrD47MiDrzcJdSeTLF75aCcxVVt8B1N74rA+vh2XCAvFy4tEWJjtnUtj2QG7/zlQ6g9cQ88bZFBxwd9/FmtA==",
+			"requires": {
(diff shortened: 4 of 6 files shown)


nivida:
> [EDIT - this coverage drop is due to web3-eth-accounts/src/index.js getting smaller after removing a require statement, increasing the proportion of uncovered lines]

Verified ✅

nivida (approved):

nivida:
@cgewecke Oh! Have you done some further tests with the ``PromiEvent`` object? It would probably be the perfect moment to add some additional test cases for it :) 

... and did you also thought about to fix this upstream instead of removing it? (I think the possibility is here that some depending projects do want to support IE11)

Edit:
> add some additional test cases for it

"add some explicit test cases for it" this is probably the better wording.

holgerd77 (approved):
Would also agree that keeping this dependency causes more problems than helping to address. Also adding a [Can-I-use-Web-Workers](https://caniuse.com/#feat=webworkers) table here for reference. Generally have to say that I have not very much of a picture of how widely web workers are used atm.

cgewecke:
@nivida 

> did you also thought about to fix this upstream instead of removing it?

Yes. There have been no changes to any-promise for ~4 years. The problem with the `window` object has been raised there before in [any-promise 29][1].

On the surface, the lib seems abandoned / a historical artifact but according to npm it continues see significant increases in usage. Its weekly downloads have almost doubled in the last year (base = 1.5 mil)

Screen Shot 2020-02-24 at 8 59 37 AM

RE: testing. I don't know what test to add here. If you can think of one, will happily add.  
 

[1]: https://github.com/kevinbeaty/any-promise/issues/29

Issue #3389, opened 2020-02-24
WebsocketProvider with `fromBlock` doesnt return events from the past


VladoDemcak:
Maybe I am facing the same issue as described here: https://github.com/ethereum/web3.js/issues/3379.

**Current behaviour**
When I am subscribed I get only events from last block. Eg if there are events in block 0,1,2,3 and I set `fromBlock`: 0 and the current block is 3 I get only events from block 3 and not from 0.

**Expected behaviour**
Once I am subscribed and I specify `fromBlock`: 0, I should get all events from block 0, 1 ... latest.

 I've created the [post on stackoverflow](https://ethereum.stackexchange.com/questions/80029/web3js-events-fromblock-0-doesnt-work) as well so more details about the problem I am facing is described [here](https://ethereum.stackexchange.com/questions/80029/web3js-events-fromblock-0-doesnt-work).

I am using `WebsocketProvider` and `quorum-v2.4.0` and `web3js 1.2.6`. So I am connected like this:
```javascript
const web3 = new Web3("ws://127.0.0.1:23000"); 
```
 
When I create subscription for my contract it looks like below. I specify `fromBlock` because I want to get all events from block 0 when I am creating subscription.

```javascript
 contract.events.Transfer({
    fromBlock: 0,
  })
    .on("connected", function (x) {
      // do somethinng
    })
    .on('data', function (x) {
      // do somethinng
    })
    .on('error', function (x) {
      // do somethinng
    });
```

However it doesn't return events from block 0 even though there are some events. It returns only the events from last block. So if current block is 10 and there was an event in block 9 it returns only this event but doesnt return event from block 8,7,6...0.

After some time of investigating the issue I have found there is a weird part of the implementation in [subscription.js](https://github.com/ethereum/web3.js/blob/c20bcf09b04f773406ce3532e88fd105bb04e244/packages/web3-core-subscriptions/src/subscription.js#L246) where `fromBlock` is removed with `delete` keyword. Check the snipped from `subscription.js` below.

```javascript
    // create subscription
    // TODO move to separate function? so that past logs can go first?

    if(typeof payload.params[1] === 'object')
        delete payload.params[1].fromBlock;
```

Since this code deletes the`fromBlock` from params it is missing in [WebsocketProvider.prototype.send](https://github.com/ethereum/web3.js/blob/c20bcf09b04f773406ce3532e88fd105bb04e244/packages/web3-providers-ws/src/index.js#L286).

[...]

PR #3391, opened 2020-02-24
Reproduction case for issue 3389

cgewecke:
## Description

**DO NOT MERGE / EXAMPLE ONLY**

For further discussion in #3389, where the `from` parameter for an event subscription doesn't seem to be honored. 

Test written here vs geth 1.9.x, auto-mining at 2 second intervals suggests its working as expected.


cgewecke:
Closing - problem looks infura related

PR #3401, opened 2020-03-03
Fix historical event subscriptions when made as first provider request 

### LINKED ISSUE #3389: WebsocketProvider with `fromBlock` doesnt return events from the past

cgewecke:
## Description

**PR targets #3190** (because it fixes a Websocket bug)

Resolves a race condition that arises when the first request made by a Websocket provider is a subscription to an event which includes a request for historical logs. 

Web3 was deleting the `fromBlock` parameter made for the historical logs request while the Websocket connection established itself, resulting in the "missing logs" bug reported in #3389.

Commit 64c1302671c90c3d6497ea473599c8006a62f5af is a test that can be seen failing in [Travis here][1]
Commit  7f3f42b121c366a65da2dbf26509da8237ec0ee6 fixes it by copying the params. 

The copied params object looks like....
```js
{
 "fromBlock": "0x0",
 "topics": [
  "0x929af6b98ce8455e51b1e90cceeeec03ce38d7c7396f1f0b0233d043c898a29c",
  null
 ],
 "address": "0xc30a9b6f84a3d660585437a00ddd0090773ea9dc"
}
```

[1]: https://travis-ci.org/ethereum/web3.js/jobs/657937790#L664

Fixes #3389

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-core-subscriptions/src/subscription.js
+++ b/packages/web3-core-subscriptions/src/subscription.js
@@ -232,9 +232,13 @@ Subscription.prototype.subscribe = function() {
     // get past logs, if fromBlock is available
     if(payload.params[0] === 'logs' && _.isObject(payload.params[1]) && payload.params[1].hasOwnProperty('fromBlock') && isFinite(payload.params[1].fromBlock)) {
         // send the subscription request
+
+        // copy the params to avoid race-condition with deletion below this block
+        var blockParams = Object.assign({}, payload.params[1]);
+
         this.options.requestManager.send({
             method: 'eth_getLogs',
-            params: [payload.params[1]]
+            params: [blockParams]
         }, function (err, logs) {
             if(!err) {
                 logs.forEach(function(log){
--- a/test/e2e.contract.events.js
+++ b/test/e2e.contract.events.js
@@ -13,6 +13,7 @@ describe('contract.events [ @E2E ]', function() {
     var accounts;
     var basic;
     var instance;
+    var port;
 
     var basicOptions = {
         data: Basic.bytecode,
@@ -21,7 +22,7 @@ describe('contract.events [ @E2E ]', function() {
     };
 
     beforeEach(async function(){
-        var port = utils.getWebsocketPort();
+        port = utils.getWebsocketPort();
 
         web3 = new Web3('ws://localhost:' + port);
         accounts = await web3.eth.getAccounts();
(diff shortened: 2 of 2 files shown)


nivida:
@cgewecke Is it not possible to have the base branch defined as 1.x? (I think we shouldn't blow up the PRI branch more)

cgewecke:
@nivida Yes, that's one possibility. The main reason to point it at #3190 is that it will need to be rebased into #3190 anyway since it touches the same files, and there's a growing list of merge conflicts there. 

But maybe it makes sense to go through #3190 and pull out *everything* that could go into 1.x on it's own. There are tests and fixes for misc bugs discovered in the course of the work there which could be addressed as separate items. 

That would reduce the size / complexity of #3190 further and perhaps it could also be re-opened as a smaller set of commits that a reviewer can easily follow. 

Will wait for feedback from @holgerd77 and @ryanio about how best to proceed...

cgewecke:
Closing in favor #3455 (which targets 1.x directly) for simplicity. 

This PR is 180+ commits  ahead of 1.x & it's simplest to retarget in a fresh request.

PR #3421, opened 2020-03-18
Fix windows CI & lengthen verdaccio timeouts

cgewecke:
## Description

Some CI tuning for the "allowed failure" jobs that stems from Ryan's work on Github Actions. 
+ kills a zombie Node process that was making the windows CI hang
+ increases verdaccio's http-agent timeouts (should make those jobs more stable) 

## Type of change

- [x] CI improvement

## Checklist:


ryanio (commented):
Looks good.

Did you want to add a few more zeroes to keepAliveMsecs?

Did you by any chance try if other values like max_timeout, max_fail, or server: keepAliveTimeout would additionally help at all? https://github.com/ryanio/web3.js/blob/c9e9c3051e88f0ad07a81d9f6c78fcb488a19b85/verdaccio.yml

cgewecke:
@ryanio Ah thanks. Looking at the logs you ran, my impression was that the other timeouts (e.g 10m) weren't getting hit. Is anything waiting 10 minutes in those jobs? 

And `e2e_ganache` (which usually passes on Travis) is timing out on GH actions, unless the http-agent timeouts are set wide. Maybe that setting is the decisive one?

Tbh I don't really have any idea what's going on 🙂 Making all the settings long as you suggest is probably safest.

> Did you want to add a few more zeroes to keepAliveMsecs?

Yes!

What is your opinion about leaving the allowed failures in Travis and using GH Actions for the real jobs? Only reason I suggest this is that Travis' UI seems slightly better for that use-case - you can easily see if the job failed but the run as a whole passes.

ryanio:
> Is anything waiting 10 minutes in those jobs?

In my runs it would randomly time out on me when downloading deps, so I bumped everything to 10mins to play it safe, but I never saw it wait that long. It would usually just immediately fail on retries up to the `max_fail` limit. Seemed to fail at a different dep every time so it was confusing indeed, although I think it did succeed a few times.

> What is your opinion about leaving the allowed failures in Travis and using GH Actions for the real jobs? Only reason I suggest this is that Travis' UI seems slightly better for that use-case - you can easily see if the job failed but the run as a whole passes.

Sure that seems nice, however unfortunately last I heard the `ethereum` org still doesn't have access to GH Actions as a feature so it might not be possible to actually switch :(

ryanio (approved):

cgewecke:
@ryanio 

> Sure that seems nice, however unfortunately last I heard the ethereum org still doesn't have access to GH Actions as a feature so it might not be possible to actually switch :(

Oh ok. I've ported over a couple improvements you made to `eth.accounts.wallet` to make those less flaky too then.

ryanio (approved):
Great, LGTM :)

scripts/e2e.windows.sh
```diff
@@ -14,12 +14,12 @@ cd windows_test
...
 
 # Shutdown verdaccio server
 cd ..
 source verdaccio_pid
 kill -9 $VERDACCIO_PID
 
-# Debugging...
-ps -ef
+# Terminate stray node process
+kill -9 $(grep node <(ps -ef) | sed 's/.*\travis *\([0-9]*\).*/\1/')
```
holgerd77: Hmm, these kind of things seem pretty dangerous to me if run in a local environment. E.g. I tested on my machine and the non-`kill` part returned a totally unrelated "Atom Helper" process having something with `node` in the passed CLI parameters.

Is there any way to make this more robust to avoid side effects like this?

cgewecke: @holgerd77 

Oh yes...do you think checking that the `CI` environment variable is set before running this would be enough? 

Here's [a list of env variables set in TravisCI][1]

[1]: https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

cgewecke: Added a check for CI in 2f1ebc5bd8a36bdfab651a09cd6be4086c4837ca

holgerd77 (approved):
LGMT now.

PR #3422, opened 2020-03-18
Clarify that 0x prefix is needed for imported private keys

miohtama:
## Description

Purely documentation update.

I run an issue that `web3.eth.accounts.privateKeyToAccount` private key format was behaving differently from other Ethereum packages

* truffle-hd-wallet

* ethereumjs-wallet

This may lead to a problem that users are freaking out and think they lost their account when dealing with raw private keys.

web3.js generated different accounts from the other packages because private key was handled differently. This was because of the disagreement if hexadecimal private keys should be prefixed with `0x` or not.

I clarified this in the documentation and also added a pointer about creating a HD wallet.

## Type of change

## Checklist:

- [X] I have selected the correct base branch. (Hopefully docs are still being updated for 1.x)

### DIFF
--- a/docs/web3-eth-accounts.rst
+++ b/docs/web3-eth-accounts.rst
@@ -98,12 +98,18 @@ privateKeyToAccount
 
 Creates an account object from a private key.
 
+For more advanced hierarchial address derivation, see 
+[truffle-hd-awallet-provider](https://github.com/trufflesuite/truffle/tree/develop/packages/hdwallet-provider)
+package.
+
 ----------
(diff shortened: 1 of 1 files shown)


cgewecke (approved):
Thanks @miohtama, this looks good and makes things clearer. 

For context, changes made in #3316 and #3323 will enforce the 0x prefix by default. They're scheduled to be published with `1.2.7`.

holgerd77 (approved):
LGTM

PR #3428, opened 2020-03-22
Add support for ENS contenthash methods

### LINKED ISSUE #3392: web3.eth.ens.getContent seems to do not work properly
#### Expected behavior

Given a ENS name that has an associated content, I should be able to receive the content hash when I call the web3.eth.ens.getContent('myName');

#### Actual behavior

I run this code (in typescript)

```typescript
const web3 = new Web3(new Web3.providers.HttpProvider(
   "https://mainnet.infura.io/v3/my-infura-key"
));
web3.eth.ens.getContent('luca.cervone.eth')
   .then((result: String) => {
      console.log(result);
   }).catch((error: Error) => {
      console.log(error);
   });
```

The above code returns the following error: 

```console
Uncaught (in promise) Error: Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using the correct ABI for the contract you are retrieving data from, requesting data from a block number that does not exist, or querying a node which is not fully synced.
    at ABICoder.decodeParameters (index.js?bbaf:239)
    at Contract._decodeMethodReturn (index.js?d100:557)
    at Method.outputFormatter (index.js?d100:910)
    at Method.formatOutput (index.js?6248:167)
    at sendTxCallback (index.js?6248:596)
    at eval (index.js?176c:147)
    at XMLHttpRequest.request.onreadystatechange (index.js?8148:110)
```

I attach a screenshot  proving that the name as an associated content 

Schermata 2020-02-25 alle 11 11 45

cgewecke:
## Description

Adds `getContenthash` and `setContenthash` methods to the ENS module.

**Example**
```js
await web3.eth.ens.getContenthash('ethereum.eth');

> {
  "protocolType": "ipfs",
  "decoded": "QmaEBknbGT4bTQiQoe2VNgBJbRfygQGktnaW5TbuKixjYL"
}

await web3.eth.ens.setContenthash(
  'ethereum.eth', 
  'ipfs://QmaEBknbGT4bTQiQoe2VNgBJbRfygQGktnaW5TbuKixjYL,
);
```
As discussed in #3392, there are now two widely used resolver contracts for ENS with different interfaces. 

~~PR adds support for the newer PublicResolver ABI, toggling between the two candidates by making an EIP 165 `supportsInterface` call for the `contenthash` method before selecting which ABI to use. If the Resolver address associated with the ENS name supports contenthash, the PublicResolver ABI is used, otherwise module defaults to the legacy Resolver ABI.~~

PR updates the ENS Resolver ABI to include contenthash interface as well as deprecated `content`. Makes EIP 165 supportsInterface calls for a subset of resolver methods to verify that the target contract implements them.

ENS has [a feature][1] that allows nodes to register ABIs on-chain for contract addresses but this does not appear to be implemented for the publicly deployed Resolver contracts themselves. 

Running a [reverse address check][2] on Etherscan for the mainnet PublicResolver at: [0x4976fb03C32e5B8cfe2b6cCB31c09Ba78EBaBa41][3] does not discover an ENS node which might supply an ABI. 

[1]: https://eips.ethereum.org/EIPS/eip-205
[2]: https://etherscan.io/enslookup
[3]: https://etherscan.io/address/0x4976fb03C32e5B8cfe2b6cCB31c09Ba78EBaBa41

Fixes #3392

## Type of change

## Checklist:


packages/web3-eth-ens/src/lib/contentHash.js
```diff
@@ -0,0 +1,85 @@
+/*
+  Adapted from ensdomains/ui
+  Permalink: https://github.com/ensdomains/ui/blob/3e62e440b53466eeec9dd1c63d73924eefbd88c1/src/utils/contents.js#L1-L85
+*/
```
cgewecke, line 30: Have opened an issue at [@ensdomains/ui 62][1] asking for clarification about how to license & attribute this file's contents appropriately. The repo currently has no license. 

[1]: https://github.com/ensdomains/ui/issues/62

ricmoo:
ABI is supported in the public resolver (resolver.eth) See line 72: https://etherscan.io/address/0x4976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41#code

cgewecke:
@ricmoo I might misunderstand this but isn't that ABI method similar to `contentHash`, e.g it lets you associate an ABI with a name? 

In this case we're looking for the ABI *of the resolver itself* and I wasn't able to see a way of retrieving it. 

If it is possible it would be the best solution by far though...

ricmoo:
Oh! I see what you mean.

We could talk to @Arachnid about adding the ABI to resolver.eth, but I think it would be fine to just hardcode the getters (and maybe the setters) into it; each has an associated EIP.

Also, maybe it should be `contenthash`, not `contentHash` (and `setContenthash`, not `setContentHash`) to keep it inline with the [EIP-1577 standard](https://eips.ethereum.org/EIPS/eip-1577)?

cgewecke:
> Also, maybe it should be contenthash, not contentHash (and setContenthash, not setContentHash) to keep it inline with the standard?

@ricmoo Ah yes, that's a good point.

cgewecke:
Have revised per initial review - think this is moving in a better direction...
 
+ Removed ABI toggling
+ Added logic to query supportsInterface for each resolver method call Web3 wraps and ENS [advertises an interface ID for in their docs][2]. This is tolerant - e.g it will error if the Resolver doesn't support `content` but ignore methods it doesn't know about.
+ Replaced ABI with the `Resolver` [generic][1] from @ensdomain/resolver@0.2.4. (This includes definitions for deprecated methods like `content`).
+ Added an E2E test for pubkey set/get - its interface is also getting validated now.
+ Removed camel-casing of `contenthash` 

[1]: https://github.com/ensdomains/resolvers/blob/master/contracts/Resolver.sol#L39-L43
[2]: https://docs.ens.domains/contract-developer-guide/writing-a-resolver

package.json
```diff
@@ -137,15 +139,13 @@
...
         "nyc": "^14.1.1",
         "puppeteer": "^1.20.0",
+        "regenerator-runtime": "^0.13.3",
         "sandboxed-module": "^2.0.3",
         "surge": "^0.21.3",
+        "truffle": "^5.1.8",
         "typescript": "next",
         "underscore": "^1.9.1",
         "vinyl-source-stream": "^2.0.0",
-        "wait-port": "^0.2.6",
```
nivida: there are unnecessary changes in this file as the wait-port and truffle dependency movement :-)

packages/web3-eth-ens/package.json
```diff
@@ -13,6 +13,7 @@
     },
     "main": "src/index.js",
     "dependencies": {
+        "content-hash": "^2.5.2",
```
nivida, line 16: I would recommend asking @ricmoo here again.. there are some cases this package wasn't doing his job nicely.. and also here.. is this external dependency required from a security perspective?

packages/web3-eth-ens/src/lib/ResolverMethodHandler.js
```diff
@@ -22,8 +22,22 @@
 
 var PromiEvent = require('web3-core-promievent');
 var namehash = require('eth-ens-namehash');
+var errors = require('web3-core-helpers').errors;
 var _ = require('underscore');
 
+// These ids obtained at ensdomains docs:
+// https://docs.ens.domains/contract-developer-guide/writing-a-resolver
+var interfaceIds = {
```
nivida: This should get out-sourced to the config file I think..

packages/web3-eth-ens/src/lib/ResolverMethodHandler.js
```diff
@@ -41,14 +55,15 @@ function ResolverMethodHandler(registry) {
  * @param {function} callback
  * @returns {Object}
  */
-ResolverMethodHandler.prototype.method = function (ensName, methodName, methodArguments, callback) {
+ResolverMethodHandler.prototype.method = function (ensName, methodName, methodArguments, outputFormatter, callback) {
```
nivida, line 46: The outputFormatter addition wouldn't be required.. you can just wrap the promise as on other places :-)

nivida: (This ugly handler has to get refactored/removed anyways.. just ping me.. I do have a solution somewhere on a handwritten paper)

packages/web3-eth-ens/src/lib/ResolverMethodHandler.js
```diff
@@ -69,9 +84,11 @@ ResolverMethodHandler.prototype.call = function (callback) {
     var self = this;
     var promiEvent = new PromiEvent();
     var preparedArguments = this.parent.prepareArguments(this.ensName, this.methodArguments);
+    var outputFormatter = this.outputFormatter || null;
 
-    this.parent.registry.getResolver(this.ensName).then(function (resolver) {
-        self.parent.handleCall(promiEvent, resolver.methods[self.methodName], preparedArguments, callback);
+    this.parent.registry.getResolver(this.ensName).then(async function (resolver) {
+        await self.parent.checkInterfaceSupport(resolver, self.methodName);
```
nivida, line 78: I remember a discussion I had once with @cgewecke about additional network requests and their impact. Shouldn't this ``checkInterface`` call not also be optional with a ``strictEns`` flag or so? Otherwise, the arguments in the discussion we had are false:)

packages/web3-eth-ens/src/lib/ResolverMethodHandler.js
```diff
@@ -208,4 +230,31 @@ ResolverMethodHandler.prototype.prepareArguments = function (name, methodArgumen
...
+/**
+ *
+ *
+ * @param {Contract} resolver
+ * @param {string} methodName
+ *
+ * @returns {Promise}
+ */
+ResolverMethodHandler.prototype.checkInterfaceSupport = async function (resolver, methodName) {
+    // Skip validation for undocumented interface ids (ex: multihash)
```
nivida, line 230: FYI: Multihash is here for backward compatibility

packages/web3-eth-ens/src/lib/ResolverMethodHandler.js
```diff
@@ -208,4 +230,31 @@ ResolverMethodHandler.prototype.prepareArguments = function (name, methodArgumen
...
+    if (!interfaceIds[methodName]) return;
+
+    var supported = false;
+    try {
+        supported = await resolver
+            .methods
+            .supportsInterface(interfaceIds[methodName])
+            .call();
+    } catch(err) {
+        console.warn('Could not verify interface of resolver contract at "' + resolver.options.address + '". ');
```
nivida, line 240: Couldn't it also throw other errors? ;-)

packages/web3-eth-ens/src/lib/contentHash.js
```diff
@@ -0,0 +1,111 @@
...
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
```
nivida, line 30: Instead of just copy it should we probably create an EF package for it with working together with ENS @Arachnid

packages/web3-eth-ens/src/ressources/ABI/Resolver.js
```diff
@@ -1,23 +1,158 @@
 "use strict";
 
+// This ABI taken from @ensdomains/resolver@0.2.4
+// @ensdomains/resolver/build/contracts/Resolver.json
```
nivida: I once thought about using those ABI's from the ENS packages itself.. we could/should do that and add the additional required ABI items for backward compatibility with a clear deprecation guide/timeline. :-)

packages/web3-eth-ens/src/ressources/ABI/Resolver.js
```diff
@@ -34,33 +169,39 @@ var RESOLVER = [
         "name": "ABI",
         "outputs": [
             {
-                "name": "contentType",
+                "name": "",
```
nivida: Are those return values also unnamed in legacy Dapps?:)

packages/web3-eth-ens/src/ressources/ABI/Resolver.js
```diff
@@ -242,114 +495,153 @@ var RESOLVER = [
...
             }
         ],
-        "name": "PubkeyChanged",
-        "type": "event"
+        "name": "setMultihash",
+        "outputs": [],
+        "payable": false,
+        "stateMutability": "nonpayable",
+        "type": "function"
     }
```
nivida, line 407: Overall do I think are there some unrelated changes in this ABI.. we should verify every change or undo the non-required changes.

packages/web3-eth-ens/types/index.d.ts
```diff
@@ -22,6 +22,12 @@ import { TransactionRevertInstructionError } from 'web3-core-helpers';
 import { Eth } from 'web3-eth';
 import { Contract } from 'web3-eth-contract';
 
+export interface ContentHash {
```
nivida, line 25: Noice! 🎊

test/e2e.ens.js
```diff
@@ -50,5 +57,126 @@ describe('ENS [ @E2E ]', function () {
             '0x0000000000000000000000000000000000000001'
         );
     });
+
+    it('should get/set a publickey', async function(){
```
nivida, line 61: Some great new test cases thanks Chris! But also here.. aren't there unnecessary changes that would be against the guidelines of this project? :-)

test/eth.ens.js
```diff
@@ -103,6 +103,8 @@ function isExpectedReceipt(receipt) {
 }
 
 describe('ens', function () {
+    this.timeout(5000);
```
nivida: Can you elaborate on why exactly this overall timeout is required? 5s are quite a time :/

cgewecke:
Rebased on bcf248f895fe14d466e6fe60e9096f0fc5e7a766

packages/web3-eth-ens/src/lib/ResolverMethodHandler.js
cgewecke: Yes, it could but is spliced before another call to the resolver. If there's a problem with the connection or something, the error will be thrown then. I don't think this will swallow anything unexpectedly.

packages/web3-eth-ens/src/lib/ResolverMethodHandler.js
cgewecke: The origin of this design is the discussion beginning in [#3392 comment][1]. In this case there are two widely deployed resolvers with different APIs, so there's a strong case for validating the request before-hand. 

I think you're alluding here to a [debate in #3428][1] about whether or not to gate revert-with-reason behind an option flag. The ENS module is not subject to the similar considerations AFAIK, personally don't think this will impact people in the same way 🤷  

[1]: https://github.com/ethereum/web3.js/pull/3248#pullrequestreview-324469285




[1]: https://github.com/ethereum/web3.js/issues/3392#issuecomment-591918724

packages/web3-eth-ens/package.json
cgewecke: This package is [used by @ensdomains/ui][1] for the same purpose. Intention is to make the decoding here as close to theirs as possible without pulling in lots of other deps. 

[1]: https://github.com/ensdomains/ui/blob/d2bfbe92fce7fc18e3caeb6ccec969de804b3f86/package.json#L60

test/e2e.ens.js
cgewecke: Idk. This PR introduces to interface validation as part of the solution to supporting contentHash per design discussion in #3392. We have existing methods that will be validated this way so we should also test them.

packages/web3-eth-ens/src/lib/ResolverMethodHandler.js
cgewecke: Not sure I follow, but yes this could be refactored at some point.

package.json
cgewecke: The package.json has been normalized alphabetically in some other PR now. This isn't in the diff anymore.

packages/web3-eth-ens/src/lib/contentHash.js
cgewecke: This code comes out of [@ensdomains/ui][1] which is relatively large. This seemed like the lightest approach.

[1]: https://github.com/ensdomains/ui

packages/web3-eth-ens/src/lib/ResolverMethodHandler.js
cgewecke: Fixed in 
a6e19cb

packages/web3-eth-ens/src/ressources/ABI/Resolver.js
cgewecke: Agree, that's better. Fixed in efd5cbf

packages/web3-eth-ens/src/ressources/ABI/Resolver.js
cgewecke: Fixed in efd5cbf

packages/web3-eth-ens/src/ressources/ABI/Resolver.js
cgewecke: Fixed in efd5cbf

test/eth.ens.js
cgewecke: Thanks for catching. Fixed in f037c31

cgewecke:
Update: Have made suggested improvements per @nivida's review:
+ interfaceIds moved to config
+ preserved the legacy ABI and only added contentHash methods as necessary
+ deleted a stray mocha timeout

andrewheadricke:
Just stumbled across this issue, looking forward to getting this incorporated into web3

ryanio (approved):
LGTM! Nice comprehensive work @cgewecke.

New tests are passing successfully in CI:
```
    ✓ should error when calling "getContent" if resolver does not support it
    ✓ should error when calling "setContent" if resolver does not support it
    ✓ getContenthash return object keys are null if no contentHash is set
    ✓ should get/set an IPFS contenthash (ipfs://)
    ✓ should get/set an IPFS contenthash (/ipfs/)
    ✓ should get/set a bzz contenthash
    ✓ should get/set an onion contenthash
    ✓ should get/set an onion3 contenthash
    ✓ setContenthash errors when encoding an invalid contenthash (promise)
    ✓ setContentHash errors when encoding an invalid contenthash (callback)
```

Issue #2697, opened 2019-04-16
Implementing of EIP-712


nivida:
### Description 

Implementing of the ``eth_signTypedData`` method as described [here](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md).

https://github.com/ethereum/web3.js/issues/1975

#### Expected behavior

``web3.eth.signTypedData()`` exists.

### Versions 
* web3.js: latest

princesinha19:
This method already exists in web3-eth. https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth/src/methods/EthSignMethod.js#L23

The method name is web3.eth.sign(). We can make an alias for web3.eth.signTypedData(). @nivida

the-jackalope:
The EIP states that `web3.eth.signTypedData` will be a separate method in Web3 1.0 and will have a different signature than `web3.eth.sign`.

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#specification-of-the-web3-api

What's the source of truth on this issue?

Ankarrr:
I'm wondering is someone implementing `web3.eth.signTypedData` ?

e00dan:
This would be super useful as previous signing method is basically deemed deprecated as of now - at least that's what MetaMask is going to tell you when trying to use existing web3.js methods for signing.

Amxx:
After reading @princesinha19 's message I tried having a look at the linked source file (which does not exist).

What I managed to find (https://github.com/ethereum/web3.js/blob/1.x/packages/web3-eth/src/index.js) does NOT contain any reference to eth_signTypedData, and the [inputSignFormatter](https://github.com/ethereum/web3.js/blob/67ea9d097737a1fb3c26cb2ae9a237245d600947/packages/web3-core-helpers/src/formatters.js#L220-L222) used to format eth_sign's inputs clearly does not support eth_signTypedData's format.

I'm almost mad that I was given false hopes that this issue was addressed ... it should have been years ago ... Being forced to do rpc call myself, I wonder if there is a point to using web3.js anymore in 2020

princesinha19:
@Amxx Actually, the structure of the repo got changed that's why the given link reference doesn't exist anymore. Sorry if my message caused any inconvenience to you.

azmain84:
I do not know ... why is this problem difficult to fix .. please help me.?

Amxx:
I got sick of it. If you have any interest in the issue being fixxed, please test my PR

[...]

Issue #1975, opened 2018-09-27
signature for EIP712 


Amxx:
[EIP712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) proposes a standard for hashing and signing of typed structured data as opposed to just bytestrings.

Similarly to ethereum message who are sign using
``
ecsign(keccak256("\x19Ethereum Signed Message:\n" | message.length | message))
``
EIP712 proposes to sign arbitrary structures using:
``
ecsign(keccak("\x19\x01" | domainSeparator | hashStruct(message)))
``
with `domainSeparator` et `hashStruct` described by the proposal.

I have implemented these functions, meaning I am able to compute `keccak("\x19\x01" | domainSeparator | hashStruct(message)))` ... I am only missing the ecsign part. I understand that exposing ecsign can be dangerous in production application (you don't control what people sign) but for the sake of testing and debuging my smartcontract in truffle/ganache I need a way to sign my custom message.

Using `web3.eth.sign` includes the `"\19Ethereum Signed Message:\n" + message.length` salting so it doesn't meet my needs.

`ethereumjs-util.ecsign` could be a solution but it requires the private key that I don't know how to access.

Any solution for me ? Is support for EIP712 signature planned ?

PR #3449, opened 2020-04-01
Add eth.signTypedData (ERC712)

Amxx:
## Description

Attempt to finally implement my original issue #1975 that was re raised in #2697
I've requested this >18 months ago :/

[@cgewecke edited here...]
**PR needs**
- [ ] Documentation updates
- [ ] Typescript definitions
- [ ] Manual test with Metamask
- [ ] Custom input validation for eth_signTypedData [per EIP 712][1]?

[1]: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#specification-of-the-eth_signtypeddata-json-rpc 

## Type of change

## Checklist:

- [x] I have executed ``npm run test:cov`` and my test cases do cover all lines and branches of the added code.


holgerd77:
@Amxx Thanks for taking this constructive approach. This is a highly frequented library and unfortunately not all feature requests can be approached as desired.

packages/web3-eth/src/index.js
```diff
@@ -485,6 +485,27 @@ var Eth = function Eth() {
...
+            inputFormatter: [null, formatter.inputAddressFormatter],
+            outputFormatter: function (result) {
+                var vrs = this.accounts._decodeSignature(result);
+                return {
+                    v: vrs[0],
+                    r: vrs[1],
+                    s: vrs[2],
+                    signature: result
+                }
+            },
```
cgewecke, line 504: @Amxx 

Have modified this so it resembles the `.accounts` output more. 

Could not get ganache to accept the object test input when using `JSON.stringify` as an input formatter - I've removed that. 

**Questions**
+ Could you discuss in more detail your TODO comment about the input formatting?
+ Is ganache the only client that supports this endpoint? (Geth doesn't).

cgewecke (commented):
@Amxx This looks good! I've pushed a couple things and am leaving comments

+ minimized the package-lock changes
+ ported a ganache-core test for `eth_signTypedData`

This PR still needs:
+ documentation
+ typescript definitions

Happy to add these, just want to get your ok on the inital changes first.

packages/web3-eth/src/index.js
Amxx: - Details about my TODO comment:

The first argument (equivalent of message) shuld be a javascript object that should contain 4 fields: 'types', 'primaryType', 'domain', 'message'. Other objects are not valid. Both 'types', 'domain' and 'message' should be objects. 'primaryType' should be a string that also happens to be a key of the 'types' field. Message should have the structure describe in types (by the 'primaryType') ... there are MANY checks that could be performed.

Of course, just JSON.stringify the object, doing the RPC call, and catching errors is possible. I'm not sure how extensive you want the pre-checks to be.

- Metamask has been the first provider to support this endpoint. Not sure about others.

I've done what I could. Feel free to ask me more or do whatever you want with my draft.

cgewecke: @amxx Ok thanks!

Will add a checklist of things still needed the original PR description to make sure we cover everything here...

Amxx: Not sure how to proceed with ganache. In case that helps, here is an old piece of code I used to use with metamask:

```
function signStruct(typename, message, wallet)
{
	return new Promise((resolve, reject) => {
		web3.currentProvider.sendAsync({
			method: "eth_signTypedData_v4",
			params: [ wallet, JSON.stringify({ types: TYPES, domain: DOMAIN, primaryType: typename, message: message }) ],
			from: wallet,
		}, (err, result) => {
			if (result.error == undefined)
			{
				message.sign = result.result;
				resolve(message);
			}
			else
			{
				reject(result);
			}
		});
	});
}

```

cgewecke: @amxx Thanks, that's helpful. We'll need to test vs. metamask...

Amxx: The thing with metamask is they have implemented various iterations of the EIP draft with different names. Most commonly used might be eth_signTypedData_v3, latest is eth_signTypedData_v4, not sure what eth_signTypedData fallback to (if first version that is bad)

nivida:
One good question here is.. what version of EIP-712 got implemented and is it backward compatible? (I think there are 4-5 versions existing)

... and additionally, we use the eth-sig-util package for it which is quite risky from a security perspective.. isn't there a solution to do it without an additional dependency? does have @ricmoo already implemented it in ethers which would be EF internal?

ricmoo:
I haven't created an EIP-712 formatter and parser yet... It's on my backlog. :)

cgewecke:
This languished for a bit, sorry.

After reviewing the various versions of the signTypedData EIP and the range of implementations out in the wild, the consensus among the maintainers here is that this might need to be implemented in experimental form. It's not clear that the broader community has coalesced around a single idea of how things should work. 

That said, there's also strong agreement that Web3 should support APIs like this and provide access to experimental features in some form. This topic needs further design discussion.

cc: @GregTheGreek

su45fg:
ได้ดูการ เเก้ไข ไฟล์ข้อมูล เเล้ว

packages/web3-eth/src/index.js
su45fg: การเพิ่ม เเล้วข้อมูล จะอัพเดต ต้องใช้เวลา นานไหม?

su45fg: ดีครับ รอดูว่าจะ เเก้ปัญหา ตรงeth error ได้ไหม
   ผมก็มี ปัญหานี้เอง เหมือนกัน

su45fg: > ```
>   การเพิ่ม เเล้วข้อมูล จะอัพเดต ต้องใช้เวลา นานไหม?
> ```

PR #3451, opened 2020-04-01
Fix 3380: Add AbstractProvider

### LINKED ISSUE #3380: Web3 constructor does not accept custom/non-standard providers in TypeScript
#### Overview

Related to https://github.com/ethereum/web3.js/issues/2266

There should be a way to overload the TypeScript definition of `Web3` constructor to accept custom providers. I've tried everything, even [declaration merging](https://github.com/Microsoft/TypeScript/issues/14080) won't work. The easiest solution I foresee, which would not be a breaking change, is to provide a basic `Provider` interface which all providers must implement. Using a [generic type parameter](https://www.typescriptlang.org/docs/handbook/generics.html) would be even more ideal, because it means properties like `currentProvider` could be strongly typed, even if they are custom/non-standard.

Right now the only workaround is to cast the provider to `any`, which is not ideal.

1. Set up TS project
2. Attempt to construct `Web3` instance with non-standard provider

odanado:
## Description
Fixes #3380 

The only method that `provider` should implement is the `send` function.
I'd like you to consider whether the interface of `EventEmitter` should be implemented or not.

## Type of change

## Checklist:


packages/web3-core/types/index.d.ts
```diff
@@ -410,9 +413,14 @@ export interface LogsOptions {
 
 export type BlockNumber = string | number | BN | BigNumber | 'latest' | 'pending' | 'earliest' | 'genesis';
 
+export interface AbstractProvider extends EventEmitter {
```
odanado: I would like to consider the following points
- Using the `extends EventEmitter`.
- Remove `extends EventEmitter` (Leaving only the `send` function in the interface).
- Remove `extends EventEmitter` and directly define the `on`, `once`, and `removeListener` functions as well as the `WebsocketProviderBase`.

smithki: My `$0.02` is to separate the core functionality from the subscription functionality. Not all custom providers support subscription features by default. Classes can implement multiple interfaces in TypeScript, so it shouldn't be problematic to export two interfaces.

I am not an active contributor to this code-base. This is my opinion only.

packages/web3-core/types/index.d.ts
```diff
@@ -410,9 +413,14 @@ export interface LogsOptions {
...
+    send(payload: JsonRpcPayload, callback: (error: Error | null, result?: JsonRpcResponse) => void): void;
```
odanado, line 415: The interface may also need to require sendAsync function for a library that only supports a legacy provider.

e.g. https://github.com/MetaMask/eth-json-rpc-middleware/blob/9464aa2085c63b43f1e5ed569bd08b0697c53d39/providerAsMiddleware.js#L6

packages/web3-core/types/index.d.ts
ryanio: Hey @odanado thanks for this PR :) It LGTM. For your 3 points, perhaps just on/once/removeListener is sufficient as the "bare minimum".

What are the types of `WebsocketProviderBase`?

Also, in light of [recent updates to EIP-1193](https://github.com/ethereum/EIPs/issues/2319#issuecomment-610610675), there is a new `request` method that would be great to add. I don't think having `sendAsync` would hurt either even though it's deprecated it has more long term support.

odanado: @ryanio  Thank you for the comment.
I didn't know the `request` method of EIP-1193.

I thought that the interface of the `AbstractProvider` should be an only of `send` and `sendAsync`.
The reasons are as follows.
- The status of EIP-1193 is a draft.
- EIP-1193 is not yet fully implemented in Metamask

I think it should define a provider that has methods such as on/once/removeListener/request separately from `AbstractProvider`.

packages/web3-core/types/index.d.ts
ryanio: Hey @odanado thanks again for this PR and sorry for the delay, finally getting back to it :)

I think adding `sendAsync` would be a great idea, do you mind adding it? [ethers.js/web3-provider](https://github.com/ethers-io/ethers.js/blob/56af4413b1dd1787db68985e0b612b63d86fdf7c/packages/providers/src.ts/web3-provider.ts#L17) defines it like so:

```typescript
sendAsync?: (request: { method: string, params?: Array<any> }, callback: (error: any, response: any) => void) => void
```

I'll add the latest EIP-1193 compatibility in a subsequent PR since we just moved the document into Last Call with a 6-week review period and MetaMask should be deploying their implementation soon.

ryanio (changes requested):

packages/web3-core/types/index.d.ts
odanado: @ryanio Thank you for the comment.
I updated.

- remove EventEmitter
- add sendAsync method

ryanio:
Thanks for adding sendAsync  🙌 looks like there is [a small dtslint issue](https://github.com/ethereum/web3.js/pull/3451/checks?check_run_id=642041392#step:5:151).

odanado:
@ryanio I forgot that fix a test.
I updated the test.

ryanio (approved):
Looks great, thanks!

PR #3459, opened 2020-04-08
add prefix to a private key on the web3-eth-accounts

nphsu:
Relates to https://github.com/ethereum/web3.js/issues/3458

Added `0x` prefix before executing functions if it not included.


holgerd77:
Hi @shunp, thanks for the PR! There is merged PR on this which will ensure the `0x` prefix, this will go into the upcoming `1.2.7` release.

Will close here.

PR #3467, opened 2020-04-13
Add docs note for Geth v1.9.12 `eth_call` change

### LINKED ISSUE #3456: Issue with method call result with msg.sender and Geth 1.9.12
#### Sample contract
```
contract GetMyBalance {

    // below function cant get right result
    // using Remix I get:
    // 0: uint256: 115792089237316195423570985008687907853269984665640564039457581007913129639935
    function _getMyBalance()
        // external
        public
        view
        returns(uint256)
    {
        return msg.sender.balance;
    }

    // Below can get right result
    // by calling getMyBalance()
    // and then get right 
    uint256 public logMyBalance_;
    function getMyBalance()
        external
    {
        logMyBalance_ = _getMyBalance();
    }

    // I can get right result
    uint256 public logMyBalance;
    function writeMyBalance() external {
        logMyBalance = msg.sender.balance;
    }
}
```

#### Expected behavior
Actual balance of caller

#### Actual behavior
115792089237316195423570985008687907853269984665640564039457581007913129639935

1. Deploy contract on any testnet except kovan or on Mainnet
2. call `_getMyBalance` method

#### Related Issues on different repos
https://github.com/ethereum/remix/issues/1430
https://github.com/ethereum/solidity/issues/8545

#### Versions
web3js: 1.2.6

ryanio:
## Description

It was identified in https://github.com/ethereum/web3.js/issues/3456 that geth released a breaking change regarding the default value of `from` in `eth_call`.

This means that if you're calling `view` methods that refer to a `msg.sender` without explicitly setting a `from` address in your request options, you may see unexpected behavior.

This PR adds some notes to the changelog and documentation about this change.

After some discourse we decided that we should not brush this change under the rug by modifying the behavior of `web3.js`, but instead stay consistent with the best practices and recommendations from the node clients. 

In `web3.js`, the `from` address can be specified on a per-call basis or by setting the `defaultAccount` property.

Closes https://github.com/ethereum/web3.js/issues/3456

## Type of change

## Checklist:


docs/web3-eth.rst
```diff
@@ -1600,7 +1600,7 @@ Executes a message call transaction, which is directly executed in the VM of the
 Parameters
 ----------
 
-1. ``Object`` - A transaction object see :ref:`web3.eth.sendTransaction <eth-sendtransaction-return>`, with the difference that for calls the ``from`` property is optional as well.
+1. ``Object`` - A transaction object, see :ref:`web3.eth.sendTransaction <eth-sendtransaction-return>`. For calls the ``from`` property is optional however it is highly recommended to explicitly set it or it may default to `address(0)` depending on your node or provider.
```
cgewecke, line 1603: 👍 Perfect.

WDYT about adding a similar note for `method.call` here?
https://web3js.readthedocs.io/en/v1.2.6/web3-eth-contract.html#methods-mymethod-call

ryanio: good idea, just added in latest commit 👍

cgewecke (dismissed):

cgewecke (approved):

PR #3479, opened 2020-04-17
Fix CI build error from `dtslint`

ryanio:
## Description

This PR fixes intermittent CI build failures due to `dtslint` erroring out on installing TypeScript `3.9`.

As a workaround we install the latest typescript into each package and use the `--localTs` option. It's not ideal since we only test the types against one version, but it was the existing behavior prior to https://github.com/ethereum/web3.js/pull/3464 where I attempted to expand it. This PR also removes the now unnecessary `// Minimum TypeScript Version` comments I put in to satisfy testing against prior versions in that PR.

This PR also removes the devDep `definitelytyped-header-parser` since`dtslint` automatically includes it as a child dep and outdated pinned versions may cause future dtslint issues.

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -183,3 +183,5 @@ Released with 1.0.0-beta.37 code base.
 - Change CI provider from Travis to Github Actions. (#3468)
 
 ### Fixed
+
+- Fix intermittent CI build issues with `dtslint`. (#3479)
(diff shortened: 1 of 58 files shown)


cgewecke (approved):
Sweet. LGTM.

PR #3481, opened 2020-04-19
Fix mosaic_e2e job error when CI run for fork PR

cgewecke:
## Description

@ryanio noted on gitter that in #3451 the mosaic_e2e job is erroring CI. 

The verdaccio jobs are marked continue-on-error because Lerna refuses to publish when the git head is in a detached state. We can checkout the branch if PR is made in the root repo but we'd need to add logic to pull in branches from an upstream remote to get this to work for PRs from forks.

Job's just error-ing because of a typo, should be fixed with d4fe796


ryanio (approved):
Thanks :)

PR #3486, opened 2020-04-26
Emit "error" and "end" events on WebsocketProvider close

### LINKED ISSUE #3485: Websocket Provider "error" and "end" events not fired when client disconnects
As reported by @hfa0  in #3190 [comment][1], 1.2.7 introduces a bug where "error" and "end" listeners directly attached to the Websocket provider aren't firing as expected on close. 

Subscriptions and events **do** have their error handlers fired, and pending requests should also error correctly.

```javascript
provider = new Web3.providers.WebsocketProvider(url);
provider.on('end', () => {
  // not fired
});
provider.on('error', () => {
  // not fired
});
provider.on('close', (event) => {
  // event fired
})
contract.events[type]({ fromBlock }).on('error', (err) => {
  // event fired
})
provider.disconnect(1012, 'restart');
```

The "close" event contains an object lots of information about what happened fwiw:
```js
"type": "close",
 "isTrusted": false,
 "_yaeti": true,
 "code": 1006,
 "reason": "Connection dropped by remote peer.",
 "wasClean": false,
  ... etc
```

[1]: https://github.com/ethereum/web3.js/pull/3190#issuecomment-619438025

Web3.js versions: 1.2.7, 1.2.7-rc.0

cgewecke:
## Description

Adds "error" and "end" event emissions to the Provider "close" logic, restoring behavior that existed in the WebsocketProvider prior to 1.2.7. 

At the moment if you've attached these listeners directly to the provider, they're not firing as expected when the client drops.

For background, this topic was raised in the #3190 review [here][1]. @nivida noted:

> The old WS provider did have an error listener which triggered the same error as the close listener and because it was cleaning up the response callback array in the provider on errorwasn't it triggering the error a second time when the close listener got triggered. Another difference is that the RequestManager wasn't listening to provider errors before. One of these provider-specific and not request specific errors is the MaxAttemptsReachedOnReconnectingError on line 412.

Believe the change in this PR addresses that concern. The subscription error callbacks are fired and deleted in the close handler before the final error event is emitted. Have added test logic in 4a3ef1e to verify that subscription error handlers are only run once when the connection drops.
 
Fixes #3485

[1]: https://github.com/ethereum/web3.js/pull/3190#issuecomment-572943326
## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -185,3 +185,4 @@ Released with 1.0.0-beta.37 code base.
 ### Fixed
 
 - Fix intermittent CI build issues with `dtslint`. (#3479)
+- Fix provider "error" / "end" events not fired when Websocket provider disconnects (#3485)
--- a/packages/web3-core-requestmanager/src/index.js
+++ b/packages/web3-core-requestmanager/src/index.js
@@ -130,6 +130,13 @@ RequestManager.prototype.setProvider = function (provider, net) {
                     subscription.callback(errors.ConnectionCloseError(event));
                     _this.subscriptions.delete(subscription.subscription.id);
                 });
+
+                if(_this.provider.emit){
+                    _this.provider.emit('error', errors.ConnectionCloseError(event));
+                }
+            }
+            if(_this.provider.emit){
+                _this.provider.emit('end', event);
             }
         });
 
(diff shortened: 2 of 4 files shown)


hfa0 (approved):
LGTM

ryanio (approved):
Looks great, and really nice tests.

PR #3472, opened 2020-04-15
Add --ignore-scripts to lerna publish command for E2E publishing tests

cgewecke:
## Description

**PR targets #3470**

In #3470, the publishing steps are managed with npm lifecycle scripts. However the E2E npm installation tests don't install the dev dependencies required for the build (in order to preserve as much dep. isolation as possible) so that step fails. 

We don't use the minified bundle in these tests so we can safely skip here... 

The target branch for this is an open question...it might be nice to see those tests pass for the release and they're intended as pre-publication sanity checks. On the other hand maybe this fix belongs outside the release process.

(The tests pass here ... it could be floated in next week after the release is complete)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ ] I have selected the correct base branch.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules.
- [ ] I ran ```npm run dtslint``` with success and extended the tests and types if necessary.
- [ ] I ran ```npm run test:unit``` with success.
- [ ] I have executed ``npm run test:cov`` and my test cases do cover all lines and branches of the added code.
- [ ] I ran ```npm run build-all``` and tested the resulting file/'s from  ```dist``` folder in a browser.
- [ ] I have updated the ``CHANGELOG.md`` file in the root folder.
- [ ] I have tested my code on the live network.

PR #3470, opened 2020-04-15
Release - 1.2.7

ryanio:
## Description

This PR introduces `web3.js` version `1.2.7` beginning with `1.2.7-rc.0`.

[Please see the release notes](https://github.com/ethereum/web3.js/releases/tag/1.2.7-rc.0) for more.

### Added

- Add revert reason support to sendSignedTransaction (#3345)
- ENS module extended with the possibility to add a custom registry (#3301)
- Missing ENS Registry methods and Resolver.supportsInterface method added (#3325)
- Add optional gas type to AbiItem typescript definitions (for ABIs generated by Vyper) (#3437)
- Add auto-reconnect option for Websockets (#3092, #1085, #1391, #1558, #1852, #1646)

### Changed

- Ensure '0x' prefix is existing for Accounts.sign and Accounts.privateKeyToAccount (#3041)
- Repository cleanup (#3443)
  - Removed old `docs/_build` folder
  - Removed old bower and meteor artifacts
  - Moved logo assets to own folder
  - Moved github assets to own folder
- Remove @types/node from (non-dev) dependency tree (#3965, #3227)

### Fixed

- Add missing subscription.on('connected') TS type definition (#3319)
- Add missing bignumber.js dependency for TS types (#3386)
- Upgrade swarm-js to 0.1.40 to remove npm vulnerability warning (#3399)
- Upgrade devDeps to resolve security warnings (#3464)
  - dtslint 0.4.2 => 3.4.1
  - definitelytyped-header-parser 1.0.1 => 3.9.0
- Race-condition when subscribing to historical logs as first client request (#3389)
- Fix crash when using Web-Workers by removing any-promise dependency (#3377 #2211 #1774)
- MaxListenersExceededWarning event emitter warning mitigated (#1648)

## Compare View

[v1.2.6 -> v1.2.7](https://github.com/ethereum/web3.js/compare/v1.2.6...v1.2.7-rc.0) 

## Type of change

- [x] Release

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules.
- [ ] I ran ```npm run dtslint``` with success and extended the tests and types if necessary.
- [ ] I ran ```npm run test:unit``` with success.
[...]

PR #3487, opened 2020-04-27
Add --ignore-scripts to lerna publish command for E2E publishing tests

cgewecke:
## Description

**Revival of #3472**

In #3470, the publishing steps are managed with npm lifecycle scripts. However the E2E npm installation tests don't install the dev dependencies required for the build (in order to preserve as much dep. isolation as possible) so that step fails. 

We don't use the minified bundle in these tests so we can safely skip here... 

The target branch for this is an open question...it might be nice to see those tests pass for the release and they're intended as pre-publication sanity checks. On the other hand maybe this fix belongs outside the release process.

(The tests pass here ... it could be floated in next week after the release is complete)

## Type of change

## Checklist:

### DIFF
--- a/scripts/e2e.npm.publish.sh
+++ b/scripts/e2e.npm.publish.sh
@@ -60,6 +60,7 @@ lerna version minor \
   --no-git-tag-version \
   --no-push \
   --allow-branch $BRANCH \
+  --ignore-scripts \
   --yes
 
 # Set identity prior to publishing (necessary for Windows)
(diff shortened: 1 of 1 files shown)


cgewecke:
Looks like the ganache job is hanging open due to additional test failures related to the changes in #3486. Will investigate.

cgewecke:
Auto-closed by accident..

ryanio (approved):

PR #3494, opened 2020-04-29
Throw when requests made after WS reconnect attempts are exhausted

### LINKED ISSUE #3492: WebSocket reconnect doesn't throw error on maxAttempts
#### Expected behavior
An error should be thrown when the max attempts is reached while reconnecting. Next time a method gets called that requires this connection then it should try to reconnect again.
#### Actual behavior
WebSocket continues to try to reconnect until it succeeds. This only happens after the initial connection is successful. When the initial connection fails then the reconnection works as expected

1. Create Web3 instance using WebSocket provider with the following configuration
```typescript
{
      reconnect: {
        auto: true,
        delay: 1000, // ms
        onTimeout: false,
        maxAttempts: 1
      }
}
```
2. Attempt to get a transaction receipt
3. Observe that the transaction receipt is received 
4. Disconnect your internet
5. Attempt to get same transaction receipt
6. Wait for a minute
7. Reconnect internet
8. Observe that the transaction receipt is received

I noticed that this also happens even when `auto` is set to false. I was expecting to get the following error instead: `Error: CONNECTION TIMEOUT: timeout of 5000 ms achived`

#### Versions
Node: v12.16.2
Web3.js: 1.2.7
OS: Ubuntu 18.04.4 LTS

cgewecke:
## Description

Addresses #3492. 

Requests sent when the connection is dead after max reconnection attempts have been exhausted, 
aren't throwing. 

The `reconnecting` flag isn't unset correctly so new requests begin queueing here: 

https://github.com/ethereum/web3.js/blob/bcf248f895fe14d466e6fe60e9096f0fc5e7a766/packages/web3-providers-ws/src/index.js#L311-L315

~~(Opening as a draft pending feedback from #3492 because their expected behavior is different from what's proposed here.)~~

Fixes #3492 

## Type of change

## Checklist:


ryanio (approved):
LGTM! Thanks for the additional test cases.

Issue #1178, opened 2017-11-15
Compiled web3 bundle is HUGE


josephg:
The browserify-compiled web3 bundle is crazy huge. I'm using the 1.0 beta and pre-minification web3 is 1.3 megs. Minification halves that, but its still way bigger than it needs to be.

```
$ cat foo.js 
require('web3')
$ browserify foo.js -o bundle.js
$ ls -l bundle.js 
-rw-r--r--  1 josephg  staff  1365873 15 Nov 16:00 bundle.js
```

Most of the bundle is transitive dependancies of dubious usefulness that the dapp I'm writing certainly won't need - like an ASN1 parser, an elliptic curve cyper, a 160k promise wrapper (wat) and the distributed filesystem libraries.

Have a browse of the [contents of the compiled bundle](http://josephg.com/web3disc.html).

Fixing this will not just lower download size, it will also make dapps load faster and improve the experience from nodejs. Node currently needs nearly 1 second to parse all that JS even on my fancy macbook:

```
$ node
> console.time('load'); require('web3'); console.timeEnd('load')
load: 750.081ms
```

josephg:
... It also includes 2 copies of bn.js for some reason

williamchong007:
![image](https://user-images.githubusercontent.com/6198816/39127722-6e5e158c-4738-11e8-8927-9e56b7450a5c.png)
as of beta34, I am having 6 bn.js of 2 different versions (4.11.6, 4.11.8) just by installing web3
please suggest

pubkey:
![screenshot from 2018-07-11 19-44-39](https://user-images.githubusercontent.com/8926560/42589986-e15299d0-8542-11e8-8805-d89dbae6e9b5.png)
It looks like in `beta.34`, all submodules use the same `bn.js@4.11.6`. Maybe we can figure out why npm will install bn.js into each module instead of once. Webpack will then also only use bn.js once.

pubkey:
The next step would be to move to es6-modules so we do not require the whole `bn.js` or `underscore` each time. Or is there already a plan for that?

nivida:
Yes we have an 1.0ES6 branch @pubkey please feel free to contribute.
I will have a look on the dependencies.

roshanrags:
@nivida Is there a status page or something summarizing the current status of the ES6 branch? I can contribute, but don't really know where to start or what has been done so far.

Maybe a Github Project tracking it would be helpful.

sweetpalma:
[...]

PR #3495, opened 2020-04-29
Expose bundlesize token in GH Actions CI

cgewecke:
## Description

In #3260 we added a bundle size checker to CI to track minified package size increases (cf: #1178). It relies on a CI env secret that needs to be ported to GH actions.

Looks like this on Travis:

Screen Shot 2019-12-07 at 1 12 08 PM

## Type of change

- [x] CI tuning 🎶 

## Checklist:


ryanio (approved):
👍

PR #3497, opened 2020-05-04
Add string event integration tests

cgewecke:
## Description

Adds integration tests for illegal utf-8 string and wide unicode (emoji) event decoding.

## Type of change

- [x] Tests

## Checklist:


test/sources/Basic.sol
```diff
@@ -32,4 +34,15 @@ contract Basic {
...
+    function firesStringEvent(string memory _str) public {
+        emit StringEvent(_str);
+    }
+
+    function firesIndexedStringEvent(string memory _str) public {
+        emit IndexedStringEvent(_str);
+    }
+
+    function firesIllegalUtf8StringEvent() public {
+        emit StringEvent('�������');
```
cag, line 46: Try `emit StringEvent(string(0x9f90e274657374))`.

cgewecke: @cag 

solc is not letting me do this (have tried a few variants). Keep getting something along these lines:
```
TypeError: Explicit type conversion not allowed from "int_const 44913823586808692" 
           to "string memory".
        emit StringEvent(string(0x9f90e274657374));
                         ^----------------------^
```

cag: My bad, I don't usually use bytes literals. Try `emit StringEvent(string(hex"9f90e274657374"))`

cgewecke: @cag Ah that works but it's detecting that it's invalid. Also tried w/ 0.4.26 and it's the same...
```
TypeError: Explicit type conversion not allowed from "literal_string (contains invalid UTF-8 sequence 
           at position 0)" to "string memory".
        emit StringEvent(string(hex"9f90e274657374"));
                         ^-------------------------^
```

cgewecke: We don't have to use this kind of test fwiw, we could also try to trigger it using a raw log, [as shown here][1]. 


[1]: https://github.com/ethereum/web3.js/blob/3e10668745ddbff8c785351f36754343ec680acf/test/event.decode.js#L49-L57

cgewecke: Also tried using a raw log style test without success (e.g failure)

cgewecke:
More utf-8 stuff: https://github.com/ethereum/web3.js/issues/3480

ryanio:
really nice to have these extra tests 👍

were you able to try decoding a malformed utf-8 string from a contract to see how web3 would react? @ricmoo mentioned flipping a top bit of any byte:

> Basically, take any UTF-8 string and flip any top bit of any byte. If you flip a continuation byte, the code unit will underrun, if you flip a start-byte, the next continuation byte is invalid as a sequence, and if you flip a non-ASCII bit, the next byte won't be the needed continuation byte...

If ether’s default behavior is to error in this case (but also provides a way to recover the string) I wonder if it would be of benefit to the community for web3.js to have a differing default behavior. 

Since we have relaxed strictness around some encoding formats, one mentality could be for web3 try to “fail as little as possible” in decoding so if a user is interacting with a smart contract he doesn’t own and someone has stored or emits bad string data, the code tries to continue to run instead of crashing (if not caught). We could introduce a param to enforce strict validation on decoding strings that is default false.

Easier said than done, but just an idea.

cgewecke:
> were you able to try decoding a malformed utf-8 string from a contract to see how web3 would react?

I have tried to do this [here][1], but have been unsure about how to malform the strings. The recipe you suggest is great,  will see if I can do that, thanks! 

[1]: https://github.com/ethereum/web3.js/pull/3497/files#diff-6a4ac45451154cf7496e7ff9ab356c3bR46

cgewecke:
@ryanio 

Took another look at this and remembered that I got the inputs for these from an [w3.org test case set][1] for malformed utf-8. (Apologies - I should have documented this.)

The new tests use case `3.1.8` (7 continuation bytes).  
```
Each unexpected continuation byte should be separately signaled as a         
malformed sequence of its own.                                                |
                                                                              
3.1.1  First continuation byte 0x80: "�"                                      
3.1.2  Last  continuation byte 0xbf: "�"                                      
                                                                              
3.1.3  2 continuation bytes: "��"                                             
3.1.4  3 continuation bytes: "���"                                            
3.1.5  4 continuation bytes: "����"                                           
3.1.6  5 continuation bytes: "�����"                                          
3.1.7  6 continuation bytes: "������"                                         
3.1.8  7 continuation bytes: "�������" 
```

This seems quite similar to what @ricmoo suggests. Not certain how to proceed here.

Am going to update the file with a note on the test case source for now...

[1]:  https://www.w3.org/2001/06/utf-8-wrong/UTF-8-test.html

cgewecke:
This needs a rebase, there are irrelevant changes in the diff....

ryanio (approved):
LGTM, thanks for investigating and adding the additional tests!

PR #3493, opened 2020-04-28
Fixes #3491 - missing events while resubscribing.

VladoDemcak:
## Description

For inspiration. I didnt run tests and dont have test for this. It's just my very quick test and fix for missing events while resubscribing. 

Maybe the checkpoint field is not necessary and can be included in args. Dont know but I think this could be helpful.
## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue) #3491
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I have selected the correct base branch.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules.
- [ ] I ran ```npm run dtslint``` with success and extended the tests and types if necessary.
- [ ] I ran ```npm run test:unit``` with success.
- [ ] I have executed ``npm run test:cov`` and my test cases do cover all lines and branches of the added code.
- [ ] I ran ```npm run build-all``` and tested the resulting file/'s from  ```dist``` folder in a browser.
- [ ] I have updated the ``CHANGELOG.md`` file in the root folder.
- [ ] I have tested my code on the live network.

PR #3502, opened 2020-05-06
Backfill logs when resubscribing

### LINKED ISSUE #3491: WebSocket reconnect - ws provider will miss some events.
Web3 v1.2.7

We are testing new web3 reconnect feature which has been added in v1.2.7. It works fine but there is one problem we have found.

**Setup.** WS provider with `auto reconnect = true` and `3sec delay`.
1. We send events 1,2,3,4
2. close WS, now wait delay 3 sec for reconnect
3. Within the delay gap - let's send events 5 and 6
4. WS will reconnect after 3 sec delay and start subscription again. However we will miss events 5 and 6.

We can set  small delay but I think there might be still a time gap where we can miss some events.

We tried to workaround this with keeping fromBlock once we processed an event and when ws is reconnected we recreated subscriptions. but web3's requestManager will create it's own subscriptions https://github.com/ethereum/web3.js/blob/v1.2.7/packages/web3-core-requestmanager/src/index.js#L113 and we will have 2 separated subscriptions. so we would process one event twice.

Image below is showing what was happening in the test case. 

`npm run server` -> without WS reconnection we have all events
`npm run listen` -> WS has been closed and reconnected. we missed events from blocks: 3255 and 3256
`npm run send` -> you can see when we sent BasicEvents and compare timestamps.

The red rectangles show the "black hole". 

image

You can use again my https://github.com/VladoDemcak/eth-subscribe-debug repository with steps how to reproduce the issue.

cgewecke:
## Description

As reported in #3491, when a connection is dropped and reconnection is successful, events subscriptions may have gaps for the blocks they missed. 

This PR is a continuation of the block tracking strategy proposed in #3493 with some small changes.

+ preserves pre-existing params
+ added a line to unset the tracker var when `unsubscribe` is called. 
+ renamed "checkpoint" to "lastBlock"

[EDIT: added a test per review advice]
~~This change is hard to test unfortunately. We've been simulating dropped connections by turning ganache on/off and that's won't work for this case since all log state is lost.~~

(Have pinged @VladoDemcak, the original author, for further review).

Fixes #3491 

## Type of change

## Checklist:


VladoDemcak:
@cgewecke Alright. I quickly checked “files changed” but didnt see test. Will take a look at it tomorrow or during the upcoming weekend.

For dropping connection of current ws we have different solution. not by turning ganache on/off but with closing ws. onnection. With turning ganache off or on we can also lost a state and deployed contract or addresses for instance.

So for testing tje reconnection feature we just simulate close of ws with a custom error code. in the example below code 4000.
``` 
web3.currentProvider.connection.close(4000);
```
example: https://github.com/VladoDemcak/eth-subscribe-debug/blob/cdd8c915dcf3ba4ceeb9a38c1813306ad230da71/src/listen.js#L31


I think this could be useful for testing the gap which could happen while reconnecting. since web3 has configurable maxAttempts, timedelay the gap can me 0ms but also 1min, you know... 

also this “closing ws” simulation might be useful for the tests where the reconnec simulation is done by turning ganache on/off.

cgewecke:
@VladoDemcak Ah! Thanks for the tip, I'll try a test with connection.close.

cgewecke:
@VladoDemcak 

Added a test in 7bdf68b per your advice - `connection.close` is perfect for this. 

Glad you suggested that because the changes I made to your idea in the first commit needed a fix. 🙂

VladoDemcak (approved):
LGTM. 

Cool! Dont see any problem and the test you added definitely reproduces the original issue.

test/e2e.contract.events.js
```diff
@@ -279,5 +279,65 @@ describe('contract.events [ @E2E ]', function() {
         assert.equal(childEvents[1].event, 'Identical');
         assert.equal(typeof childEvents[1].returnValues.childA, 'string');
     });
+
+    // This test only runs against the ganache client launched in scripts/e2e.ganache.sh
+    // It's too complicated for geth auto-mining (we'd have to poll for blocks)
+    // and geth instamine's websockets connection is too fragile for the tests in this file.
+    it('backfills missed events when auto-reconnecting', function(){
+        if(!process.env.GANACHE) return;
```
cgewecke, line 287: Per discussion in chat, added a note about this line in b6dce9c

ryanio (approved):

ryanio (approved):
LGTM!

PR #3517, opened 2020-05-12
Improve check if a provider supports subscriptions

### LINKED ISSUE #3432: Checking if a provider supports subscriptions by checking for "on" property often breaks
When a transaction is sent, web3 chooses to either poll for a receipt, or subscribe to new blocks. However, it makes this decision by checking if `on` is a property of the provider:

https://github.com/ethereum/web3.js/blob/af6472976c23cb61c05e925eebe5734d141384f4/packages/web3-core-method/src/index.js#L515-L521

This means if a project is using a provider that includes `on`, such as web3-provider-engine, it will incorrectly try to subscribe. This can cause errors when sending a transaction:

image

The workaround I've used is to remove `on` from a provider, however I imagine this may break some providers or apps.

```javascript
const provider = new ProviderEngine();
provider.on = null;
const web3 = new Web3(provider)
```

I'm not sure what the correct way is to determine whether a provider supports subscriptions is, maybe this needs to be addressed in the provider spec/API?

ryanio:
## Description

This PR aims to fix #3432 where @dmihal outlined some problems where if a given provider has the method `on` it is assumed to be capable of handling subscriptions.

The first commit adds a failing test validating the issue, and the second commit should resolve it.

In general there is no real way of detecting a provider's capabilities and developing some sort of standard would be of great benefit to the ecosystem. One solution suggested by @nivida that has been well received is a [`provider_capabilities`](https://github.com/ethereum/EIPs/issues/2319#issuecomment-600602580) endpoint.

## Type of change

## Checklist:


cgewecke (approved):
Looks perfect! 💯

dmihal:
Looks good to me, basically if `subscribe()` fails, it falls back to polling.

Thanks for the help!

PR #3550, opened 2020-05-29
Add events `sending` and `sent` to tx submission, add arg `latestBlockHash` to event `confirmation`

### LINKED ISSUE #3438: Add events to transaction submission in support of performance testing
It would be helpful if the PromiEvent that is returned when sending transactions fired off events that provided finer-grained detail on the transaction submission process. There are numerous benefits to this:

- Makes it easier to ensure that transactions are submitted in order (see #1846)
- Makes transaction timing in performance benchmarking tools more precise
- Makes it easier to monitor the transaction submission process for debugging purposes

#### Expected behavior
PromiEvents for transactions would fire the following events

| Event | Description |
| -------- | --------------- |
| `sending` | Fired immediately before transmitting the transaction request - connection to provider must be open when this is fired. Intermediate requests such as `eth_chainId`, `eth_estimateGas`, or `eth_gasPrice` must be completed prior this event being fired. |
| `sent` | Fired immediately after the request body has been written to the client, but before the transaction hash is received |
| `transactionHash` | Behaves as it does today |
| `confirmation` | Behaves as it does today, but includes the block hash for the latest block in the event handler arguments |
| `receipt` | Behaves as it does today |
| `error` | Behaves as it does today |

#### Actual behavior

Events are only fired for `transactionHash`, `receipt`, `confirmation`, and `error`. `Confirmation` events don't include the current block hash, meaning that applications must fetch it if they need that information.

#### Versions
Applies to all versions in existence today (`<=v1.2.6`, `v2.0.0-alpha`, and `v2.0.0-alpha.1`).

#### Related Issues

- #1846 
- hyperledger/caliper#776

ryanio:
This PR:

1. Adds events to the send tx promievent: `sending` and `sent`.
1. Adds third arg to `confirmation`: `latestBlockHash`.

Fixes #3438

## Type of change

## Checklist:


packages/web3-core-method/src/index.js
```diff
@@ -755,13 +766,19 @@ Method.prototype.buildCall = function () {
...
             });
 
         } else {
             sendRequest(payload, method);
         }
 
+        if (isSendTx) {
+            setTimeout(() => {
+                defer.eventEmitter.emit('sent');
+            }, 10);
```
cgewecke: Out of curiosity, does the test pass with a `0` length timeout? Would schedule the emit in the next tick...is that enough to resolve the race condition here?

packages/web3-core-method/src/index.js
```diff
@@ -738,9 +742,16 @@ Method.prototype.buildCall = function () {
...
 
+
             return method.requestManager.send(payload, sendTxCallback);
         };
 
+        if (isSendTx) {
+            setTimeout(() => {
+                defer.eventEmitter.emit('sending');
+            }, 10);
+        }
```
cgewecke: WDYT about the idea in the initial design that this always be fired after all the preparatory calls (like getGasPrice below here?) 

> Fired immediately before transmitting the transaction request - connection to provider must be open when this is fired. Intermediate requests such as eth_chainId, eth_estimateGas, or eth_gasPrice must be completed prior this event being fired.

Wonder if it might be useful to pass the payload as an argument to the handler for both `sending` and `sent`. Another idea floated in #3550 is that this event could help w/ debugging.

cgewecke (commented):
This is super cool! Just left a couple questions.

packages/web3-core-method/src/index.js
ryanio: ah good idea I will give it a try with 0, and also see what happens with process.nextTick()

packages/web3-core-method/src/index.js
ryanio: sending payload with handler sounds like a nice idea, I will add that.

ah right sure I’ll have to add the emitter in a few places then to satisfy branching conditional logic specific to the send flow but that’s no problem.

ryanio: both resolved in https://github.com/ethereum/web3.js/pull/3550/commits/34bdd39aad540ccf03bd22f812ed2ec8584291b0 👍

packages/web3-core-method/src/index.js
ryanio: set to `0` in https://github.com/ethereum/web3.js/pull/3550/commits/34bdd39aad540ccf03bd22f812ed2ec8584291b0 👍

packages/web3-core/types/index.d.ts
```diff
@@ -59,9 +59,19 @@ export interface Providers {
...
+
+    once(
+        type: 'sent',
+        handler: (payload: object) => void
+    ): PromiEvent<T>;
+
     once(
         type: 'transactionHash',
-        handler: (receipt: string) => void
+        handler: (transactionHash: string) => void
```
cgewecke, line 74: 👍  Nice catch.

cgewecke (approved):
This looks perfect to me 🎉 

The payload object seems good. The formatting's raw which is good for seeing what actually gets sent over the wire. 

An example from the deploy test:
```json
{
 "method": "eth_sendTransaction",
 "params": [
  {
   "from": "0x4f861a99cf97d053ccb97f9fd5acd5ce2f570171",
   "data": "0x6080604.....",
   "gasPrice": "0x1",
   "gas": "0x3d0900"
  }
 ]
}
```

ryanio:
great, thanks!

PR #3567, opened 2020-06-04
Ensure `toBlock` is not set when calling myContract.events.<eventName> or myContract.events.allEvents

### LINKED ISSUE #3207: Undesired behaviour and unexpected error when `toBlock` is passed to `myContract.events.<eventName>` or `myContract.events.allEvents`
### Intro

Right now if you want to use `myContract.events.<eventName>` or `myContract.events.allEvents` with the option `toBlock`, **which is not in the documentation**, different from `latest`. The call ends up in an undesired behaviour.

If you set `fromBlock` different from `latest`, a [`eth_getLogs` call will be done](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-core-subscriptions/src/subscription.js#L221) to get the past events from `fromBlock` to `latest` without using `toBlock` -> Almost ok!. But later, `fromBlock` is [removed from the options](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-core-subscriptions/src/subscription.js#L246) and a subscription is made with `{ fromBlock: undefined, toBlock: 100}`.  Based on the [subscribe implementation](https://github.com/ethereum/go-ethereum/blob/d42ce0f2c1b52e26cce475e411d97165fb975577/eth/filters/filter_system.go#L200) of _go-ethereum_, if `fromBlock === null` (bacically, if it is not set), will set `fromBLock` as `later`.

This will finally check `fromBlock: 'latest'` and `toBlock: 100` returning, obviously, [invalid from and to block combination: from > to](https://github.com/ethereum/go-ethereum/blob/d42ce0f2c1b52e26cce475e411d97165fb975577/eth/filters/filter_system.go#L231)

#### Expected behavior

Throw an error saying that `toBlock` is not a valid option.

#### Actual behavior

Error with `invalid from and to block combination: from > to` which is not an error from web3. But a  [go-ethereum](https://github.com/ethereum/go-ethereum/blob/d42ce0f2c1b52e26cce475e411d97165fb975577/eth/filters/filter_system.go#L231) one

1. Call `myContract.events.MyEvent({ fromBlock: 0,  toBlock: 'latest' })` or `myContract.events.allEvents({ fromBlock: 0,  toBlock: '100' })`

#### Versions
Web3.js 1.X

Thanks @fmiras for the heads up.

nachomazzara:
## Description

Warn if `toBlock` is set when calling  `myContract.events.<eventName>` or `myContract.events.allEvents`

Fixes #3207

## Type of change

- [x] bug fix (non-breaking change)


nachomazzara:
Tests failed before so I try again, not sure why. Local they are working. @ryanio What can I do to fix the ganache one? As being a breaking change, I think that their repo should be updated too.

Let me know what do you think.

ryanio:
@nachomazzara thanks, looking good! hm can we do this without any breaking changes so we can include this in our patch release? perhaps a warning instead of an error. it would also be good to include a small note in the docs as you mentioned in #3207 that it is not talked about.

---

```
should be able to get logs across the fork boundary:
     Error: Invalid option: toBlock. Use getPastEvents for specific range.
```

hm it looks like they are using `toBlock: "latest"` no?

```javascript
const event = example.events.ValueSet({ fromBlock: 0, toBlock: "latest" });
```
https://github.com/trufflesuite/ganache-core/blob/master/test/forking.js#L354

nachomazzara:
> @nachomazzara thanks, looking good! hm can we do this without any breaking changes so we can include this in our patch release? perhaps a warning instead of an error. it would also be good to include a small note in the docs as you mentioned in #3207 that it is not talked about.
> 
> ```
> should be able to get logs across the fork boundary:
>      Error: Invalid option: toBlock. Use getPastEvents for specific range.
> ```
> 
> hm it looks like they are using `toBlock: "latest"` no?
> 
> ```js
> const event = example.events.ValueSet({ fromBlock: 0, toBlock: "latest" });
> ```
> 
> https://github.com/trufflesuite/ganache-core/blob/master/test/forking.js#L354

Exactly, I can log a warning if you want. I can include also an small note in the docs.

I will do after your confirmation

nachomazzara:
@ryanio done! the note at the docu seems to be obsolete. They are not mention `toBlock` as a valid option nowadays :) I think that we are good. WDYT?

packages/web3-eth-contract/src/index.js
```diff
@@ -713,6 +713,10 @@ Contract.prototype.once = function(event, options, callback) {
 Contract.prototype._on = function(){
     var subOptions = this._generateEventOptions.apply(this, arguments);
 
+    if (subOptions.params && subOptions.params.toBlock) {
+        delete subOptions.params.toBlock;
+        console.warn('Invalid option: toBlock. Use getPastEvents for specific range.');
```
ryanio, line 718: great!

ryanio:
thanks!

looks like some more lint/format stuff got in (removed semicolons, spaces between parenthesis) if you could clean up, we are not too worried about that stuff right now, would prefer a clean diff.

i don't understand what you mean about the docs. I'm thinking we can add a small note under ["Example"](https://web3js.readthedocs.io/en/v1.2.8/web3-eth-contract.html#id52) here saying "Note: If you would like to use property `toBlock`, please use `getPastEvents`."

packages/web3-eth-contract/src/index.js
ryanio: maybe we can add a test to ensure this warning fires?

nachomazzara:
> thanks!
> 
> looks like some more lint/format stuff got in (removed semicolons, spaces between parenthesis) if you could clean up, we are not too worried about that stuff right now, would prefer a clean diff.
> 
> i don't understand what you mean about the docs. I'm thinking we can add a small note under ["Example"](https://web3js.readthedocs.io/en/v1.2.8/web3-eth-contract.html#id52) here saying "Note: If you would like to use property `toBlock`, please use `getPastEvents`."

Oh sorry! didn't realize the lint again. My bad! I have two commands
`cmd + s => format & save`
`cmd + shift + s => save`

It should be ready now 🙏

packages/web3-eth-contract/src/index.js
nachomazzara: Done

nachomazzara:
@ryanio Do I have old code again? is the hardest PR I've ever made haha. It seems that a test was changed at mosaic

ryanio:
haha no worries really appreciate you sticking with it.

that is a known error with mosaic so don’t worry about it, it should have been fixed downstream already though.

ryanio:
ok @cgewecke fixed downstream and is rerunning 👍

test/e2e.contract.events.js
```diff
@@ -74,6 +73,60 @@ describe('contract.events [ @E2E ]', function() {
...
+                    fromBlock: 0,
+                    toBlock: 'latest'
+                }).on('data', function(event) {
+                    assert.equal(event.event, 'BasicEvent');
+                    this.removeAllListeners();
+                    resolve();
+                });
+            
+            assert.equal(message, 'Invalid option: toBlock. Use getPastEvents for specific range.');
+            console.warn = originalWarn
```
cgewecke, line 95: 👍

cgewecke (approved):
LGTM! Really nice 🥇

ryanio:
thanks @nachomazzara and @cgewecke!

PR #3571, opened 2020-06-08
Handle revert reason errors for geth >= 1.9.5

### LINKED ISSUE #3520: Revert reason missing when using geth 1.9.14
cf: [geth 21082][1]

[1]: https://github.com/ethereum/go-ethereum/issues/21082 

In geth 1.9.14 no data is returned when replaying failing transactions with `eth_call` which means Web3 cannot extract the reason.

There's a reproduction case at: https://github.com/cgewecke/geth-revert-reason-repro

cgewecke:
## Description

Adds logic to handle a new error response from geth when `eth_call` executes revert with reason.

Beginning with v1.9.15, geth's response for this case looks like:
```
{ 
  jsonrpc: '2.0',
  id: 7,
  error: { 
     code: 3,
     message: 'execution reverted: REVERTED WITH REQUIRE',
     data: '0x08c379a00000....' 
  }
}
```

Previously, it didn't error and the data was an outer level key. 

Fixes #3520

## Type of change

## Checklist:

- [ ] The browser visual inspection check looks ok: [sudden-playground.surge.sh][1]

[1]: http://sudden-playground.surge.sh/

### DIFF
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -623,6 +623,7 @@ Method.prototype.buildCall = function () {
 
                 // Ganache / Geth <= 1.9.13 return the reason data as a successful eth_call response
                 // Geth >= 1.9.15 attaches the reason data to an error object.
+                // Geth 1.9.14 is missing revert reason (https://github.com/ethereum/web3.js/issues/3520)
                 if (!err && method.isRevertReasonString(result)){
                     reasonData = result.substring(10);
                 } else if (err && err.data){
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -550,7 +550,7 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
         const startInterval = () => {
             intervalId = setInterval(checkConfirmation.bind(null, existingReceipt, true), 1000);
         }
-        
+
         if (!this.requestManager.provider.on) {
             startInterval()
         } else {
@@ -618,22 +618,34 @@ Method.prototype.buildCall = function () {
 
         // CALLBACK function
         var sendTxCallback = function (err, result) {
-            if (method.handleRevert && !err && isCall && (method.isRevertReasonString(result) && method.abiCoder)) {
-                var reason = method.abiCoder.decodeParameter('string', '0x' + result.substring(10));
-                var signature = 'Error(String)';
+            if (method.handleRevert && isCall && method.abiCoder) {
+                var reasonData;
 
-                utils._fireError(
-                    errors.RevertInstructionError(reason, signature),
-                    defer.eventEmitter,
-                    defer.reject,
-                    payload.callback,
-                    {
-                        reason: reason,
-                        signature: signature
-                    }
-                );
+                // Ganache / Geth <= 1.9.13 return the reason data as a successful eth_call response
+                // Geth >= 1.9.15 attaches the reason data to an error object.
+                if (!err && method.isRevertReasonString(result)){
+                    reasonData = result.substring(10);
+                } else if (err && err.data){
+                    reasonData = err.data.substring(10);
+                }
 
-                return;
+                if (reasonData){
+                    var reason = method.abiCoder.decodeParameter('string', '0x' + reasonData);
+                    var signature = 'Error(String)';
+
+                    utils._fireError(
+                        errors.RevertInstructionError(reason, signature),
+                        defer.eventEmitter,
+                        defer.reject,
+                        payload.callback,
+                        {
+                            reason: reason,
+                            signature: signature
+                        }
+                    );
+
+                    return;
+                }
             }
 
             try {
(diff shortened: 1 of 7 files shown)


test/e2e.method.call.js
```diff
@@ -62,8 +62,14 @@ describe('method.call [ @E2E ]', function () {
...
-                assert(err.message.includes("Returned values aren't valid"));
-                assert(err.message.includes('the correct ABI'));
+                // ganache | geth <= 1.9.13
+                const nullDataResponse = err.message.includes("Returned values aren't valid") &&
+                                         err.message.includes('the correct ABI');
+
+                // geth >= 1.9.15
+                const gethErrResponse = err.message.includes("revert");
+
+                assert(nullDataResponse || gethErrResponse);
```
cgewecke, line 72: There's another behavioral change in 1.9.15. Geth now responds to this ABI mismatch case with:
```
{ 
  jsonrpc: '2.0',
  id: 8,
  error: { code: -32000, message: 'execution reverted' } 
}
```

Previously it returned 
```
{ 
  jsonrpc: '2.0', 
  id: 8, 
  result: '0x' 
}
```
which Web3 interprets as an invalid response, erroring with the message.
```
Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using 
the correct ABI for the contract you are retrieving data from, requesting data from a block number 
that does not exist, or querying a node which is not fully synced
``` 

We *can* detect this case and preserve the existing behavior but am unsure what to do here. 

**Questions**
+ Should we pass client errors as they are and let the clients manage their own error UX? 
+ ..or should we try to correct what seems like substantial reduction in the error message's clarity/usefulness?

packages/web3-core-method/src/index.js
```diff
@@ -618,22 +618,34 @@ Method.prototype.buildCall = function () {
...
-                    payload.callback,
-                    {
-                        reason: reason,
-                        signature: signature
-                    }
-                );
+            if (method.handleRevert && isCall && method.abiCoder) {
+                var reasonData;
+
+                // Ganache / Geth <= 1.9.13 return the reason data as a successful eth_call response
```
ryanio, line 624: great, what do you think about adding something like this to help document for historical reasons:
`// Geth 1.9.14 has revert reason missing (https://github.com/ethereum/web3.js/issues/3520)`

Might save someone some time somewhere down the line 😄

test/e2e.method.call.js
ryanio: I like the idea of preserving the existing error, although maybe it could be split up into more specific errors if possible, or at least reworded with better grammar.

I also like the idea of passing on client errors when we can, as I imagine it could be helpful in debugging.

cgewecke: One concern I have is that the new geth message doesn't seem right in this context and they might change it. The additional handling on our side basically involves grepping their message and routing the error handling back through the old path.

In this case, we've sent the client something that can't be executed...not positive but I don't think this is "reverting" per se. Will ask for further guidance over there.

cgewecke: For the mismatched ABI case, clients are responding as follows:

### "0x" data / non-error responses
**geth <= 1.9.14**
```
{
 "jsonrpc": "2.0",
 "id": 1002,
 "result": "0x"
}
```

**ganache-cli --noVMErrorsOnRPCResponse** (geth emulator)
```
{
 "id": 1002,
 "jsonrpc": "2.0",
 "result": "0x"
}
```

### Error responses
**geth >= 1.9.15**
```
{
 "jsonrpc": "2.0",
 "id": 1002,
 "error": {
  "code": -32000,
  "message": "execution reverted"
 }
}
```

**buidler node**
```
{
 "jsonrpc": "2.0",
 "id": 1002,
 "error": {
  "code": -32003,
  "message": "Transaction reverted without a reason"
 }
}
```

**ganache-cli (default)**
```
{
 "id": 1002,
 "jsonrpc": "2.0",
 "error": {
  "message": "VM Exception while processing transaction: revert",
  "code": -32000,
  "data": {
   "0xe60e50d8bb9a9190c651d6ae30bc44bedc217c10ac83ef3251a4d121f09a8d54": {
    "error": "revert",
    "program_counter": 77,
    "return": "0x"
   },
   "stack": "o: VM Exception while processing transaction: revert\n    at Function ..."
  }
}
```

**Parity/OpenEthereum** (via kovan)
```
{
 "jsonrpc": "2.0",
 "id": 1002,
 "error": {
  "code": -32015,
  "data": "Reverted 0x",
  "message": "VM execution error."
 }
}
```

So,the majority of responses report this as a `revert`. And the formats have a lot of variety. 

Am inclined towards forwarding errors as they are after seeing this. It's not clear how we'd distinguish wrong ABI from an intentional revert invocation in the code.

packages/web3-core-method/src/index.js
cgewecke: 1c905c3

ryanio (approved):
LGTM! Thanks for working on this :)

Issue #3573, opened 2020-06-08
eth.getBlock sometimes does not return at all; sometimes produces "TypeError: Cannot create property 'gasLimit' on string"


wbt:
This is a follow-up to #3568 in which I attempt to provide a reproducible example for that Issue and wind up creating a reproducible example for a different issue which seems like it might be related.

#### Steps to reproduce the behavior (commands for command-line shell)
1. Start Ganache (steps developed in v2.4.0) running in a **new Ethereum workspace** (so that block numbers line up). 
1. `mkdir blockArrayTest`
1. `cd blockArrayTest`
1. `truffle init`
1. `npm init` [OK to accept defaults]
1. Edit truffle-config.js networks.development to point to Ganache (change host/port/network_id as appropriate)
1. Add file contracts/Counter.sol:
```` solidity
//This is a simplified bug-demo contract with no security
pragma solidity ^0.5.0;

contract Counter {
    int public value;
    event ValueUpdated(int oldVal, int newVal);

    constructor() public {
        value = 0;
    }

    function increment() public {
        emit ValueUpdated(value, ++value);
    }
}
````
8. Add file migrations/2_adjust_value.js:
````javascript
var Counter = artifacts.require("Counter");
module.exports = function(deployer) {
    let contractInstance;
    deployer.deploy(Counter).
    then(function(instance) {
        contractInstance = instance;
        return contractInstance.increment();
    }).then(function() {
        return contractInstance.increment();
    }).then(function() {
        return contractInstance.increment();
    }).then(function() {
        return contractInstance.increment();
    }).then(function() {
        return contractInstance.increment();
    }).then(function() {
        return contractInstance.value();
    }).then(function(value) {
        console.log("Incremented counter to "+value); //should say 5
    }).catch(function(err) {
        console.log(err.message);
    });
};
````
9. `truffle migrate --reset --network development` 
1. `npm i @truffle/contract`
1. `npm i web3`
1. Add file getEvents.js:
````javascript
const TruffleContract= require("@truffle/contract");
const Web3 = require("web3");
let counterContract = TruffleContract(require("./build/contracts/Counter.json"));
let counterInstance;
//Connect to Ganache:
let provider = new Web3.providers.WebsocketProvider('ws://127.0.0.1:7545');
let web3 = new Web3(provider);
console.log("Using web3 version "+web3.version);
counterContract.setProvider(provider);
counterContract.deployed().
then(function(instance) {
    counterInstance = instance;
[...]

PR #3576, opened 2020-06-09
Check for duplicate queue entries in WebsocketProvider

### LINKED ISSUE #3568: eth.getBlock sometimes returns an array
#### Expected behavior
[web3.eth.getBlock](https://web3js.readthedocs.io/en/v1.2.8/web3-eth.html#getblock) "Returns a block matching the block number or block hash."

#### Actual behavior
In most cases, the expected behavior, but in some cases not.  When I call 
`web3.eth.getBlock(blockNumber, function(error, blockData) {`...
where blockNumber is a number (e.g. 2577), `blockData` typically contains an Object more or less as described in the documentation.   However, *sometimes*, I get back an ARRAY containing Objects, but where the "blockNumber" field does not match the requested block number.  These arrays range in length from 0 (empty), including 1, to 17 (probably higher, but this is what I've observed in limited testing).  

The object keys also look a little different, containing keys `address` (40-hex-digit string), `topics` (an array of a couple 64-hex-digit strings), `data` (320-hex-digit-string), `blockNumber` (hex string without leading zeros padding it to a specific length, notably different from `number` as specified in the Block object), `transactionHash` (64-hex-digit string), `transactionIndex` (usually '0x0'), `blockHash` (64-hex-digit string), `logIndex` (usually '0x0'), `removed` (usually false), `gasLimit` (undefined), `gasUsed` (undefined), `size` (undefined), `timestamp` (undefined), and `number` (undefined).   

There is no `nonce`, `hash`, `from`, `to`, `value`, `gas`, `gasPrice` or `input` fields as in a [Transaction object](https://web3js.readthedocs.io/en/v1.2.8/web3-eth.html#gettransaction).

My inspection of the object is within the callback, and the callback is defined in the call.  The call is the primary/only contents of a Promise constructor in a wrapper function, which is called several times in different asynchronous chains kicked off by arrival of an event in the block for which data (specifically, the timestamp) is being requested.  In repeated testing, I am getting the arrival of the exact same set of past events on the same contract on the same private chain running on Geth 1.9.1-stable.  In these tests, the set of blocks for which an array is returned instead of a Block object seems random, with the probability of an array being returned somewhere around 1.5%. 

Unfortunately, I don't have reliable reproduction steps as the effect appears to be somewhat random.

#### Environment
npm 6.14.5, node v9.3.0, web3 1.2.7, Windows 10

wbt:
## Description
This is intended to address the issue found in #3568 by detecting the condition causing that issue, and emitting an error instead of sending data to incorrect formatters and callbacks.  This error will not be emitted in cases where intended operation can continue because only ids that have already been fully used and deleted from queues are being reused, as the existing behavior is not specifically problematic in that case, though this may lead some bugs to go undiscovered in light testing.  

I have marked this as a draft because in limited testing with the example code of #3573, this does not produce helpful output such as the text of the error message; execution just silently stops and it's not clear why.  If someone else can do more testing or edits to make that clearer, that would be helpful. 

A corresponding change in 2.x might be a good idea. 

Fixes #3568

## Type of change
I am not sure if this should be considered a bug fix or breaking change.  It would only break code in cases where the code is already broken but where an existing issue might not be evident, making the issue more evident.

## Checklist:
This is as complete as of initial PR submission as the original submitter will get it:

### DIFF
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -74,6 +74,11 @@ module.exports = {
 
         return error;
     },
+    DuplicateRequestIDError: function(id){
+        return new Error('Duplicate request id '+id+
+                         ' sent to provider. Try setting a unique provider object for each copy of'+
+                         ' jsonrpc.js used in the project instead of sharing a single Provider.');
+    },
     RevertInstructionError: function(reason, signature) {
         var error = new Error('Your request got reverted with the following reason string: ' + reason);
         error.reason = reason;
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -76,7 +76,7 @@ module.exports = {
     },
     DuplicateRequestIDError: function(id){
         return new Error('Duplicate request id '+id+
-                         ' sent to provider. Try setting a unique provider object for each copy of'+
+                         ' sent to Provider. Try setting a unique provider object for each copy of'+
                          ' jsonrpc.js used in the project instead of sharing a single Provider.');
     },
     RevertInstructionError: function(reason, signature) {
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -76,7 +76,7 @@ module.exports = {
     },
     DuplicateRequestIDError: function(id){
         return new Error('Duplicate request id '+id+
-                         ' sent to Provider. Try setting a unique provider object for each copy of'+
+                         ' sent to Provider. Try setting a unique Provider object for each copy of'+
                          ' jsonrpc.js used in the project instead of sharing a single Provider.');
     },
     RevertInstructionError: function(reason, signature) {
--- a/packages/web3-core-helpers/types/index.d.ts
+++ b/packages/web3-core-helpers/types/index.d.ts
@@ -73,6 +73,7 @@ export class errors {
     static MaxAttemptsReachedOnReconnectingError(): Error;
     static PendingRequestsOnReconnectingError(): Error;
     static ConnectionError(msg: string, event?: WebSocketEvent): ConnectionError;
+    static DuplicateRequestIDError(id: string | number): Error;
     static RevertInstructionError(reason: string, signature: string): RevertInstructionError
     static TransactionRevertInstructionError(reason: string, signature: string, receipt: object): TransactionRevertInstructionError
     static TransactionError(message: string, receipt: object): TransactionError
--- a/packages/web3-providers-ws/src/index.js
+++ b/packages/web3-providers-ws/src/index.js
@@ -308,6 +308,11 @@ WebsocketProvider.prototype.send = function (payload, callback) {
         id = payload[0].id;
     }
 
+    if(this.requestQueue.has(id) || this.responseQueue.has(id)) {
+        this.emit(this.ERROR, errors.DuplicateRequestIDError(id));
+        return;
+    }
+    
     if (this.connection.readyState === this.connection.CONNECTING || this.reconnecting) {
         this.requestQueue.set(id, request);
 


cgewecke:
@wbt 

Tangentially, had a thought that your use case might also be helped by using yarn's [selective version resolutions][1] feature. This lets you coerce all instances of a package in the dependency tree to a single version, flat-packed once at the outer level. We use it here in E2E testing with projects that consume Web3 via truffle-contract. 

You add the following key to your package.json and install with yarn:
```json
"resolutions": {
  "*/**/web3": "1.2.8",
  "*/**/web3-bzz": "1.2.8",
  "*/**/web3-core-helpers": "1.2.8",
  "*/**/web3-core-method": "1.2.8",
  "*/**/web3-core-promievent": "1.2.8",
  "*/**/web3-core-requestmanager": "1.2.8",
  "*/**/web3-core-subscriptions": "1.2.8",
  "*/**/web3-core": "1.2.8",
  "*/**/web3-eth-abi": "1.2.8",
  "*/**/web3-eth-accounts": "1.2.8",
  "*/**/web3-eth-contract": "1.2.8",
  "*/**/web3-eth-ens": "1.2.8",
  "*/**/web3-eth-iban": "1.2.8",
  "*/**/web3-eth-personal": "1.2.8",
  "*/**/web3-eth": "1.2.8",
  "*/**/web3-net": "1.2.8",
  "*/**/web3-providers-http": "1.2.8",
  "*/**/web3-providers-ipc": "1.2.8",
  "*/**/web3-providers-ws": "1.2.8",
  "*/**/web3-shh": "1.2.8",
  "*/**/web3-utils": "1.2.8"
},
```

Nested Web3 dependencies will be installed at 1.2.8. You can check this by running `yarn list web3` and verifying that only a single version is listed for your project.

Another detail is that @truffle/contract automatically attaches a confirmation handler to each request which fires 24X by default. Setting `web3.eth.transactionConfirmationBlocks` to a lower number should address this. 

Side-note: if you're interested in running the ganache tests locally you can with
```
npm run test:e2e:ganache
```

To run against geth w/ websocket tests included (takes ~3 min)
```
npm run test:e2e:geth:auto
```

[1]: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

wbt:
@cgewecke Thanks for the pointers! 
I am not quite comfortable coercing truffle packages to use a version of web3 they aren't yet designed to work with, largely due to past experience about the interaction between those projects and due to having spent too many weeks in "dependency hell."

cgewecke:
> largely due to past experience about the interaction between those projects and due to having spent too many weeks in "dependency hell."

Understandable, although the interaction bug you discovered here with multiple versions installed is pretty bad. 

FWIW we use resolutions in testing here to inject the latest state of the lib into another project and run their units tests with it. So there's at least some ongoing validation that Web3 works with @truffle/contract. Would not pretend it's exhaustive though.

wbt:
>So there's at least some ongoing validation that Web3 works with @truffle/contract. Would not pretend it's exhaustive though.

That comes as a bit of a surprise, looking at the failure of [this PR](https://github.com/trufflesuite/truffle/pull/3078), where a using project breaks absent changes in the using project following an attempt to upgrade a patch-level web3 change.  (The interaction issues are not all old ones!)  However, it's nice to hear there's some attempts to continue compatibility checking.

cgewecke:
@wbt 

That PR is not getting past a typescript compilation step on install. Think it's something that could be resolved if they reworked their TS a bit. 

We've continued to make sure the core logic doesn't break @truffle/contract since that module is widely used on it's own. We're also trying to make sure that Web3's types changes are non-breaking by testing the latest state against a production TS project (gnosis/dex-react) [here][2].

Lastly, there's a tracking issue at [truffle 2688][1] that documents the main things Truffle would need to address to upgrade Web3.  

[1]: https://github.com/trufflesuite/truffle/issues/2688 
[2]: https://github.com/ethereum/web3.js/pull/3562

wbt:
I don't think this should be autoclosed; I think it could benefit from some collaborative effort to get it over the finish line.

wbt:
I still don't think this should be marked as stale.

wbt:
I still don't think this should be autoclosed; I think it could benefit from some collaborative effort to get it over the finish line.

wbt:
I don't remember the CLA requirement being part of the deal when taking over from the Ethereum Foundation.

GregTheGreek:
Please see #3846 . CLA is not required.

jdevcs:
This PR is still WIP and there is no activity, so closing . @wbt feel free to reopen this when you have time for contributions. Thanks

PR #3566, opened 2020-06-03
Fix WS clientConfig type and clarify docs

ryanio:
## Description

This PR fixes the WS clientConfig type and adds to the docs for the option `keepalive`.

Closes #3563 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] Any dependent changes have been merged and published in downstream modules.
- [x] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [x] I ran `npm run test:unit` with success.
- [x] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code.
- [x] I ran `npm run build-all` and tested the resulting files from `dist` folder in a browser.
- [x] I have updated the `CHANGELOG.md` file in the root folder.
- [x] I have tested my code on the live network.
- [x] The browser visual inspection check looks ok: [sudden-playground.surge.sh][1]

[1]: http://sudden-playground.surge.sh/

PR #3578, opened 2020-06-10
Remove webocket native dependency.

nebojsa94:
## Description

This PR removes websocket dependency and instead uses ws since it doesn't require native dependency compiling.

## Type of change

## Checklist:

- [X] I ran `npm run test:unit` with success.

### DIFF
--- a/packages/web3-providers-ws/package.json
+++ b/packages/web3-providers-ws/package.json
@@ -17,7 +17,7 @@
     "eventemitter3": "4.0.4",
     "underscore": "1.9.1",
     "web3-core-helpers": "1.2.11",
-    "websocket": "^1.0.31"
+    "@nomiclabs/websocket": "^1.0.31"
   },
   "devDependencies": {
     "dtslint": "^3.4.1",
--- a/packages/web3-providers-ws/src/index.js
+++ b/packages/web3-providers-ws/src/index.js
@@ -25,7 +25,7 @@
 var EventEmitter = require('eventemitter3');
 var helpers = require('./helpers.js');
 var errors = require('web3-core-helpers').errors;
-var Ws = require('websocket').w3cwebsocket;
+var Ws = require('@nomiclabs/websocket').w3cwebsocket;
 
 /**
  * @param {string} url
--- a/packages/web3-providers-ws/package-lock.json
+++ b/packages/web3-providers-ws/package-lock.json
@@ -1,1227 +1,2132 @@
 {
-	"name": "web3-providers-ws",
-	"version": "1.2.11",
-	"lockfileVersion": 1,
-	"requires": true,
-	"dependencies": {
-		"@types/color-name": {
-			"version": "1.1.1",
-			"resolved": "https://registry.npmjs.org/@types/color-name/-/color-name-1.1.1.tgz",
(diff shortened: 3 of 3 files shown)


packages/web3-providers-ws/package.json
```diff
@@ -16,7 +16,7 @@
     "eventemitter3": "^4.0.0",
     "underscore": "1.9.1",
     "web3-core-helpers": "1.2.8",
-    "websocket": "^1.0.31"
+    "ws": "^7.3.0"
```
alcuadrado: According to the documentation, this module doesn't work in the browser.

I think we need two implementations of this provider, one for node using `ws`, and another one for the browser using the native `WebSocket` implementation. They can share most of their code.

The browser version should be declared in the package.json's `browser` field: https://docs.npmjs.com/files/package.json#browser

packages/web3-providers-ws/src/index.js
```diff
@@ -101,7 +101,10 @@ WebsocketProvider.prototype.constructor = WebsocketProvider;
  * @returns {void}
  */
 WebsocketProvider.prototype.connect = function () {
-    this.connection = new Ws(this.url, this.protocol, undefined, this.headers, this.requestOptions, this.clientConfig);
+    // TODO: include this.requestOptions and this.clientConfig
```
cgewecke: This TODO is important. Relevant Web3 API docs are [here][1] plus recent additions in #3566

[1]: https://web3js.readthedocs.io/en/v1.2.9/web3.html#configuration

test/websocket.ganache.js
```diff
@@ -64,9 +62,8 @@ describe('WebsocketProvider (ganache)', function () {
         await web3.eth.getBlockNumber();
 
         await new Promise(async function(resolve){
-            web3.currentProvider.once('error', function(err){
-                assert(err.message.includes('1012'));
-                assert(err.message.includes('restart'));
+            web3.currentProvider.once('close', function(event){
+                assert.strictEqual(event.code, 1000);
```
cgewecke: Believe this test should continue work as expected. The disconnect uses an error code. Test was added in #3486.

cgewecke:
There's an (abandoned) PR converting to N-API in the current Websocket repo fwiw.

[theturtle/Websocket PR 302][1]

[1]: https://github.com/theturtle32/WebSocket-Node/pull/302

cgewecke:
For background - this PR is part of [a NomicLabs led initiative][1] to purge the JS Eth stack of dependencies that require node-gyp to install. It's the source of many problems as new Node versions come out and has bad installation UX.

In some cases this is straightforward and the substitution is painless. 

In others there are tradeoffs to consider and this looks like one of them...

So far the proposed replacement's issues include:
+ Configurability not obviously equivalent.
+ Disconnection behavior not identical
+ Browser compatibility needs work.

cc: @GregTheGreek 

[1]: https://github.com/ethereum/js-organization/issues/18

alcuadrado:
Thanks for raising these points, @cgewecke! We are now exploring fixing this in `websocket` instead. We'll report back when we have any updates.

GregTheGreek:
This as been superseded by #3686

PR #3581, opened 2020-06-11
Fix buffer incompatibility issue when used inside React (issue/3580)

### LINKED ISSUE #3580: Decrypting a keystore fails on 1.2.9 (browser / React )
#### Expected behavior

in the browser (chromium) I can decrypt an encrypted wallet keystore structure.

#### Actual behavior

it shows me an the error `TypeError: "list" argument must be an Array of Buffers`

call `web3.eth.accounts.wallet.decrypt(jsonKeystore, password);`

_I've put a demo repo here:_
https://github.com/cod1ng-earth/demo-web3-129-breaks-keystore-decryption

#### Environment

node 13 / 14 for building, chrome 83

#### Background

I debugged until the failing call that occurs on 1.2.9 on `web3-eth-accounts`, Line 384: https://github.com/ethereum/web3.js/commit/2f366c7a7925be59d6de8cadafa21a441d8b08ae#diff-7bb2a20126193b9ecfe4723f83429c49R384

The result of the newly introduced `derivedKey = scrypt.syncScrypt` seems to be related somehow, even though `[derivedKey.slice(16, 32), ciphertext]` is actually an array of two Uint8 arrays (`[Uint8Array(16), Uint8Array(32)]`). After having a look at the scrypt line (`https://github.com/ricmoo/scrypt-js/blob/master/scrypt.js#L462`) I'm absolutely not sure what could go wrong here, so I leave it up to you guys :( 

Screenshot from 2020-06-11 16-52-01

elmariachi111:
**NOTE: PR targets #3583**
 
fixes a type issue between UInt8Arrays and Buffer.concat in browsers.

## Description

Optional if an issue is fixed:
Fixes #3580 

## Type of change

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x ] I have selected the correct base branch.
- [ x ] I have performed a self-review of my own code.
- [ o ] I have commented my code, particularly in hard-to-understand areas.
- [ o ] I have made corresponding changes to the documentation.
- [ o ] My changes generate no new warnings.
- [ o ] Any dependent changes have been merged and published in downstream modules.
- [ x ] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [ x ] I ran `npm run test:unit` with success.
- [ x ] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code.
- [ x ] I ran `npm run build-all` and tested the resulting files from `dist` folder in a browser.
- [ x ] I have updated the `CHANGELOG.md` file in the root folder.
- [ - ] I have tested my code on the live network.
- [ x ] I have checked the Deploy Preview and it looks correct.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -236,4 +236,4 @@ Released with 1.0.0-beta.37 code base.
 ### Fixed
 
 - Extend `_txInputFormatter` with hex prefix check (#3317)
-- Fixed `Buffer.concat` breaking the new scrypt responses in browsers (#3580)
+- Fixed `Buffer.concat` incompatibility to the Buffer impl in React (#3580)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -236,3 +236,4 @@ Released with 1.0.0-beta.37 code base.
 ### Fixed
 
 - Extend `_txInputFormatter` with hex prefix check (#3317)
+- Fixed `Buffer.concat` breaking the new scrypt responses in browsers (#3580)
--- a/packages/web3-eth-accounts/src/index.js
+++ b/packages/web3-eth-accounts/src/index.js
@@ -381,7 +381,7 @@ Accounts.prototype.decrypt = function(v3Keystore, password, nonStrict) {
 
     var ciphertext = Buffer.from(json.crypto.ciphertext, 'hex');
 
-    var mac = utils.sha3(Buffer.concat([derivedKey.slice(16, 32), ciphertext])).replace('0x', '');
+    var mac = utils.sha3(Buffer.from([...derivedKey.slice(16, 32), ...ciphertext])).replace('0x', '');
     if (mac !== json.crypto.mac) {
         throw new Error('Key derivation failed - possibly wrong password');
     }
--- a/test/e2e.account.wallet.js
+++ b/test/e2e.account.wallet.js
@@ -1,0 +1,44 @@
+var assert = require('assert');
+var utils = require('./helpers/test.utils');
+var Web3 = utils.getWeb3();
+
+describe('decrypt keystore [ @E2E ]', function () {
+    var web3;
+    var keystore = {
+        "json": [
+        {
+            "version": 3,
+            "id": "cd3f55c7-b84d-4708-956a-23725971c7e3",
+            "address": "cc9967fa5b8185354492da1ac2d40da092366a8f",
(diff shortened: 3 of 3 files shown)


ryanio:
hey @elmariachi111 thanks so much for looking into this and opening a PR.

do you think you could try to add a test case that fails without the fix?

elmariachi111:
that's what I also thought about because actually: according to all tests that shouldn't happen at all and they all cover that use case. It only occurs in a browser context, on node the decryption just runs fine...  so, if you can point me to the repo location where e2e tests are written, I'd try my very best ;) So something close to https://github.com/ethereum/web3.js/blob/1.x/test/e2e.method.send.js would to best, right? Is the ci picking up everything that starts with `e2e` automatically for e2e testing?

cgewecke:
@elmariachi111 Yes! If you put a test in any of those files it will run in Chrome / Firefox.

elmariachi111:
ok, added an e2e test that covers these lines. Interesting: The test won't detect that behaviour, since as I in the meanwhile figured out, the error only occurs when running inside React (or inside code that has been bundled) and thereofre cannot be reproduced when ran in plain JS. The core reason is that the "global" `Buffer` object behaves differently in web3 and in React, maybe due to the transitive dependency on `node_modules/node-libs-browser/node_modules/buffer/index.js:399`. I added files to reproduce that behaviour in my demo repo: https://github.com/cod1ng-earth/demo-web3-129-breaks-keystore-decryption

the "fix" I'm proposing here doesn't seem to have adverse effects on anything else, so I think we can go with it without any issues..

elmariachi111:
oh, btw 2, I noticed that the root package-lock.json on the 1.2.9 release hasn't been committed. Since this is my very first contribution I'm quite surprised that I ended up with a changed package-lock.json but I didn't commit it (this is the one on the tag: https://github.com/ethereum/web3.js/blob/v1.2.9/package-lock.json referring to v1.2.6) and that's the one from the 1.x branch (https://github.com/ethereum/web3.js/blob/1.x/package-lock.json) which carries v1.2.9 in its package.json...

cgewecke (approved):
@elmariachi111 Thanks so much for this! Just great.

Am going to merge this into #3583 for further triage. 

Unfortunately decrypt is not the only affected method and it looks the feross/buffer polyfill might be incompatible with something in ethereumjs-util as well.

PR #3594, opened 2020-06-21
updating docs: genesis is not a valid block identifier any longer

### LINKED ISSUE #3584: "Genesis" still referenced in docs but not valid as per  #2068
#### Expected behavior
Remove 'genesis' and keep 'earliest'

#### Actual behavior
'Genesis' is still referenced in the docs.

You can go to version 1.2.9 getPastEvents and still see referenced 'genesis'

Error can be reproduced:

truffle(develop)> await instance.getPastEvents('Transfer',{fromBlock:'genesis', toBlock:'latest'}).then(function(ev){console.log(ev)})
Thrown:
Error: Given input "genesis" is not a number.

FelixMohr:
## Description

Updated the documentation to not mention "genesis" as a valid block identifier any longer. 
Closes #3584

## Type of change

- Documentation

### DIFF
--- a/docs/web3-eth-contract.rst
+++ b/docs/web3-eth-contract.rst
@@ -141,10 +141,9 @@ Property
 The default block parameters can be one of the following:
 
 - ``Number|BN|BigNumber``: A block number
-- ``"genesis"`` - ``String``: The genesis block
+- ``"earliest"`` - ``String``: The genesis block
 - ``"latest"`` - ``String``: The latest block (current head of the blockchain)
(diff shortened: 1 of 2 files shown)


cgewecke (approved):
LGTM! Thanks @FelixMohr!

**Reference**
JSON-RPC spec for [default block parameter][1].

> The following options are possible for the defaultBlock parameter:

> HEX String - an integer block number
> + String "earliest" for the earliest/genesis block
> + String "latest" - for the latest mined block
> + String "pending" - for the pending state/transactions

[1]: https://eth.wiki/json-rpc/API#the-default-block-parameter

PR #3596, opened 2020-06-22
format fromBlock param on resubscription

epiccoolguy:
## Description

Added use of `inputBlockNumberFormatter` to ensure `fromBlock` is formatted properly when resubscribing. Currently when resubscribing, the plain incremented number field is put into the RPC call.

The [JSON-RPC docs](https://eth.wiki/json-rpc/API#hex-value-encoding) mention that values like fromBlock should be hex encoded. I ran into an issue where Besu will return "Invalid Params" when web3 passes in the plain number value when reconnecting on a flaky WS connection: https://github.com/hyperledger/besu/issues/1088.

I'm not familiar with the web3 codebase, so it would be great if someone could point me in the right direction for adding/updating tests.

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-core-subscriptions/src/subscription.js
+++ b/packages/web3-core-subscriptions/src/subscription.js
@@ -228,7 +228,7 @@ Subscription.prototype.subscribe = function() {
     // a dropped connection may have resulted in gaps in the logs...
     if (this.lastBlock && _.isObject(this.options.params)){
         payload.params[1] = this.options.params
-        payload.params[1].fromBlock = this.lastBlock +1;
+        payload.params[1].fromBlock = formatters.inputBlockNumberFormatter(this.lastBlock + 1);
     }
 
     // if id is there unsubscribe first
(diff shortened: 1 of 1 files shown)


cgewecke (commented):
Thanks for catching this @epiccoolguy!

What do you think about moving the fix into the re-subscription logic that comes slightly earlier than where you've currently placed it?

https://github.com/ethereum/web3.js/blob/4558b27f63277e503aa4575b08c8fd99c6ebb7d5/packages/web3-core-subscriptions/src/subscription.js#L226-L231

It looks like the bug is introduced here, when `lastBlock` gets assigned an output formatted blockNumber

https://github.com/ethereum/web3.js/blob/4558b27f63277e503aa4575b08c8fd99c6ebb7d5/packages/web3-core-subscriptions/src/subscription.js#L292-L293

...so this only affects resubscriptions (as you've noted).

Re: tests - this looks ok to me without an additional test. There is a logs backfilling test [here][1] which hits these lines but it's only running vs. ganache (which must be less strict about the RPC spec than `besu`).

[1]: https://github.com/ethereum/web3.js/blob/4558b27f63277e503aa4575b08c8fd99c6ebb7d5/test/e2e.contract.events.js#L337

epiccoolguy:
Hey @cgewecke!

I initially added the formatting up there, but was worried about isFinite taking a number type value, but after reading the docs, I found out it'll convert it to number if required: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite

I'll revert it back to formatting on assignment.

If it's not necessary to test the output being formatted properly, I'd leave the PR as is.

cgewecke (approved):
Sweet! LGTM 💯

ryanio:
Thanks @epiccoolguy!

PR #3609, opened 2020-06-30
Add caching to ci

ryanio:
## Description

Thanks to @jjangga0214 in https://github.com/ethereum/web3.js/pull/3547, this PR adds to his work to add caching to our ci.

This PR also:

* Normalizes deps to use lerna [`--hoist`](https://github.com/lerna/lerna/blob/master/doc/hoist.md) option to reduce the size of node_modules cache
* Sets typescript devDep version from `latest` to `^3.9.5`
* Removes `puppeteer`
  * I don't think it was being used and made the dev ex on `npm install` really annoying by downloading a 110mb package of chrome unless you set `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true`
* Updates package-locks to latest

## Type of change

- [x] ci enhancements


.github/workflows/build.yml
```diff
@@ -5,8 +5,22 @@ on:
...
+        paths-ignore:
+            - "**.md"
+            - "**.rst"
+            - ".editorconfig"
+            - ".gitignore"
+            - ".npmignore"
+            - "LICENSE"
     pull_request:
         types: [opened, reopened, synchronize]
+        paths-ignore:
```
ryanio: I like the idea of the paths-ignore added here to be more efficient with the ci, but I'm not sure what happens if the checks are required and they don't get run - would the PR be blocked from merging?

@jjangga0214 do you know?

test/utils.toTwosComplement.js
```diff
@@ -7,7 +7,7 @@ var utils = require('../packages/web3-utils/');
 var BigNumber = require('bignumber.js');
 
 var tests = [
-    { value: 1, expected: utils.padLeft(new BigNumber(1).round().toString(16), 64) },
+    { value: 1, expected: utils.padLeft(new BigNumber(1).integerValue().toString(16), 64) },
```
GregTheGreek, line 10: Are we introducing some potential breaking features by bumpng to `bignumber.js: 9.0.0` ?

GregTheGreek: ah, i see it was being used previously at two different versions

ryanio: `bignumber.js` in `web3-core` was already on `9.0.0`, this PR just bumps the one in the root node_modules so it's normalized for hoisting. I think it only affects `bignumber.js` as used in the test suite and the above (round => integerValue) is the only breaking change I stumbled upon, but I will keep an eye out for other side effects.

karma.conf.js
```diff
@@ -1,9 +1,5 @@
 process.env.MOZ_HEADLESS = 1;
 
-if (!process.env.GITHUB_ACTION) {
-    process.env.CHROME_BIN = require('puppeteer').executablePath();
-}
-
```
cgewecke, line 6: Tiny detail but perhaps we can mark the browser tests `CI only` in [TESTING.md][1] as well. Think this change might make running those locally fail. 

[1]: https://github.com/ethereum/web3.js/blob/1.x/TESTING.md#required-tests

ryanio: @cgewecke thx, I think I have it working locally now with https://github.com/ethereum/web3.js/pull/3609/commits/c5430edbfc3d79a932b43422d7a589a7398f7c50

.github/workflows/build.yml
ryanio: I tried this on my fork and the status checks are still required thus the PR cannot be merged in, so I will remove these `paths-ignore` for now, although it is a nice idea.

cgewecke (approved):
This all looks good to me! Nice work. 💯  💯

PR #3618, opened 2020-07-03
Upgrade build process from gulp to webpack

ryanio:
## Description

This PR upgrades the build process from gulp/browserify to webpack.

This PR also:

* Removes `npm run build-all` for creating sub-package browser builds.
  * These builds weren't being published, and we aren't aware of many people using them.
  * They didn't have proper test coverage to ensure integrity of the builds (`web3.min.js` is tested in both chrome and firefox with karma)
  * It is now easy for anyone to create a browser build of a sub-package with this PR. Just modify the root `webpack.config.js` and update the `entry` and `library` to point to the sub-package you would like to build, then run `npm run build`.
* Updates some root deps and removes old/unneeded ones
* Updates readmes `var` => `const`/`let`

## Type of change

- [x] build process enhancements (non-breaking change)

## Checklist:

- [ ] I ran `npm run build` and tested the resulting files from `dist` folder in a browser.


package.json
```diff
@@ -1,17 +1,12 @@
 {
     "name": "web3.js",
     "private": true,
-    "version": "1.2.9",
```
cgewecke, line 4: 👍

cgewecke (approved):
Looks good! Great improvement 🥇 

Wrt `dist/web3.min.js`...do we usually merge changes to that on release?

ryanio:
@cgewecke thanks! hm yes that makes sense to me, so i can remove it from this PR. edit: done

package.json
GregTheGreek: why are we removing the version?

package.json
```diff
@@ -20,14 +15,11 @@
         }
     ],
     "scripts": {
-        "version": "npm run build-all",
+        "version": "npm run build",
```
GregTheGreek, line 18: is this really needed?

package.json
ryanio: There is no need for the package version in root node_modules as lerna doesn’t use it and doesn’t update it with `lerna version`

package.json
ryanio: Yes this lifecycle hook creates, commits, and tags the new build with the updated package version numbers as part of running `lerna version`

To learn more see  https://github.com/lerna/lerna/tree/master/commands/version#lifecycle-scripts

GregTheGreek (approved):

PR #3639, opened 2020-07-13
Linter fixes

### LINKED ISSUE #3632: Redundant newHeads Subscriptions when waiting on multiple outstanding requests.
#### Expected behavior
When I'm using a websocket or other subscription-capable provider, I'd assume that a single `newHeads` subscription would be all that's necessary for web3 to be notified when my transaction is confirmed.

#### Actual behavior
When I'm using a websocket or other subscription-capable provider, web3.js sets up one `newHeads` subscription for every outstanding transaction.

1. Send a transaction.
2. Quickly send another one before the previous one is mined.
3. Count the number of `newHeads` subscriptions active on the client (it'll have increased by two).

#### Environment

I'm running into this primarily in a NodeJS environment while performance testing Hyperledger Besu.

Admittedly this likely isn't a major issue for public Ethereum, but it becomes quite substantial when submitting a high volume of transactions to a private network. For example, on a network with a 2s block time (typical for IBFT 2 networks), if the user is sending 100 transactions per second, there will be up to 200 `newHeads` subscriptions active on the client, all from a single user.

benjamincburns:
## Description

In order to fix #3632, I'm planning on doing some work to refactor the transaction execution flow. When I got started on that I noticed a bunch of linter errors when running `npm test`. These make it hard to check that my changes haven't broken anything, so I'm submitting a PR to fix them.

Note that this doesn't fix the linter issues in `web3-core-method` pertaining to deeply nested blocks. Those will be fixed in the PR for #3632.

Note that some tests appear to be failing on my machine when I run `npm run test:unit` and `npm run test:cov`. These tests were failing in `1.x` before I made this change.

## Type of change

## Checklist:

- [x] I have tested my code on the live network. (N/A - I'm not introducing new logic changes - it passes all automated client tests)
- [x] I have updated the `CHANGELOG.md` file in the root folder. (N/A - I'm not introducing new logic changes)


packages/web3-eth-accounts/src/index.js
```diff
@@ -149,29 +149,7 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
-            error = new Error('"gas" is missing');
-        }
-
-        if (tx.nonce < 0 ||
-            tx.gas < 0 ||
-            tx.gasPrice < 0 ||
-            tx.chainId < 0) {
-            error = new Error('Gas, gasPrice, nonce or chainId is lower than 0');
-        }
+        _validateTransactionForSigning(tx);
```
benjamincburns: Forgot to handle the returned error here. Will push a fix shortly.

benjamincburns:
Update: if I switch back to node v10.21.0, tests pass just fine.

benjamincburns:
>I ran npm run build and tested dist/web3.min.js in a browser.

I'd like to check this off, but is there an existing test harness? The application that I'm working on doesn't run in the browser.

If the `e2e_browsers` CI test is passing, I'd assume we can call this good?

GregTheGreek:
The `netlify/web3-staging/deploy-preview` as a part of the CI actually runs web3 in the browser, so it seems to work!

packages/web3/src/index.js
```diff
@@ -52,7 +52,7 @@ var Web3 = function Web3() {
 
     // overwrite package setProvider
     var setProvider = this.setProvider;
-    this.setProvider = function (provider, net) {
```
GregTheGreek, line 55: line 56 `arguments` intakes `net`

GregTheGreek (changes requested):

packages/web3/src/index.js
benjamincburns: The `arguments` keyword contains all arguments passed, regardless of whether args are bound to variable names in the function signature.

A lot of best practices advice written today suggests that [the `arguments` keyword should be avoided](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments#Description). Perhaps this code would be more readable if it were rewritten to use a [rest parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters)?

GregTheGreek: I would agree with this, I think I'll be tackling this issue next week (or this weekend). Theres unfortunately a lot of usages of `arguments` throughout the subpackages which inherintely use `net`. It got implemented like `super` unfortunately...

GregTheGreek (approved):

PR #3649, opened 2020-07-19
Improve RequestManager send method

ryanio:
## Description

This PR improves RequestManager.send in a few ways:

1. Only sends `{ method, params }` to `provider.request`, instead of whole jsonrpc shape, to be more conformant with spec.
1. Simplifies the code by extracting the jsonrpc result callback to its own method on RequestManager.

I also removed the bind I had in callbackify, after reading more about callbackify and bind I determined it is not needed in this case. If anyone thinks otherwise I would very much appreciate some reasoning.

## Type of change

- [x] Refactoring

## Checklist:


GregTheGreek (commented):
LGTM! @ryanio just need to update the changelog. 

But this is very interesting... im wondering what the impact might be on #3656 and related issues. I'm wondering if previously there may have been a weird collision with how jsonRPC `id` was being incremented and other providers being used. What im really curious to know is if the bind:
```
callbackify(this.provider.request.bind(this.provider))(payload, callback);
```
was causing the provider to miss increment because of [setting _this](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-providers-ws/src/index.js#L305) in the websocket provider might pull the wrong instance of the provider?

anyway g2g and I'll monitor this in the future.

GregTheGreek (approved):

frankiebee (approved):
looks good to me :call_me_hand:

sapjax:
@ryanio @GregTheGreek 
This PR removed the `bind` with `callbackify` from:
```js
callbackify(this.provider.request.bind(this.provider))(payload, callback);
```
to:
```js
callbackRequest(requestArgs, callback);
```

I can confirm this change caused some problems.
Consider a wallet provider implement `window.ethereum` like:

```js
//....
EthereumProvider.prototype.request = (payload) => {
    return this._eip1193Send(payload.method, payload.params)
}
window.ethereum = new EthereumProvider()
```

Now, web3.js call `RequestManager.prototype.send`, the `callbackify` changed the `this` context of `provider.request` method.
`this` in `this._eip1193Send` will be `Window` Object and can't found `_eip1193Send` method.

GregTheGreek:
@frankiebee Hmmm We should add a way to test this out, maybe we can add some "buttons" to the netflify preview to allow us to test out that type of functionality?

sebsobseb:
👋 I can confirm the issue @sapjax is explaining.

I'm currently trying to use WalletConnect as web3 provider, but when i try to call `web3.eth.getAccounts()` for example, it throws the following error:

image

When i check with the debugger, `this` is referring to the Window object:
image

This is how my code +- looks:

```javascript
import WalletConnectProvider from "@walletconnect/web3-provider";
import Web3 from "web3";

const provider = new WalletConnectProvider({ infuraId, chainId });

// Enable session (triggers QR Code modal)
await provider.enable(); 

//  Create Web3
web3 = new Web3(provider);

// Get accouts
await vm.web3.eth.getAccounts(); // => throws the error!
```

Would appreciate the fix!

PR #3678, opened 2020-08-08
Fix misleading comment

junderw:
When looking through the source code to find the address checksum method, I was confused by this comment.

Basically you are making it upper case if the most significant bit of the 4-bit set is 1. (which for a 4 bit number would mean > 7)

9 to f threw me off for a bit.


GregTheGreek (approved):
great catch!

PR #3721, opened 2020-09-21
Fix: ensure using this.provider as the context `this`

chendatony31:
## Description

The bug is brought in this PR: https://github.com/ethereum/web3.js/pull/3649 

It may cause some problems when calling `provider.request`. As the context `this` maybe `window` sometimes, so we should make sure the context is the `provider`.

## Type of change

## Checklist:


ryanio:
Is it really necessary for the context to be `provider`? It could be, but I don't think it should be relied on like that. Could you adjust your provider code to bind to the scope as explicitly needed?

chendatony31:
> Is it really necessary for the context to be `provider`? It could be, but I don't think it should be relied on like that. Could you adjust your provider code to bind to the scope as explicitly needed?

Yeah, We've got a problem, and now I have to detect the context whether is the `window` or the `provider`. it may work, but not cool. 

More discussion may check here: https://github.com/ethereum/web3.js/pull/3649#issuecomment-696533463

sebsobseb (approved):

GregTheGreek:
This makes sense but will need to test it out later today and probably add a testcase for this

Velenir:
Here's a [codesandbox](https://codesandbox.io/s/wc-web3-callbackify-issue-1beb2?file=/src/App.js:661-805) with a reproduction of this issue with `@walletconnect/web3-provider`

frankiebee (approved):
this looks good to me, `bind`ing provider context is relatively harmless in this situation and good practice for this style code. it's a little strange though because callbackifiy should preserve the `this` but maybe due to the changes in the pr mentioned that has change.

vikmeup (changes requested):
Did you consider to write a test to avoid such issues in the future?

sebsobseb:
Any idea when this will get merged and released with tag 1.3.1 ?

GregTheGreek (approved):

PR #3732, opened 2020-09-28
fix eth packages typing issue, not export default #3198

ariesjia:
## Description

typing issue, export is not correct.

## Type of change

## Checklist:

- [  ] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [  ] I have updated the `CHANGELOG.md` file in the root folder.


packages/web3-eth-ens/package.json
```diff
@@ -26,6 +26,7 @@
     },
     "devDependencies": {
         "dtslint": "^3.4.1",
-        "typescript": "^3.9.5"
+        "typescript": "^3.9.5",
+        "web3-eth": "1.3.0"
```
Spacesai1or, line 30: Should this (`web3-eth`) be a dependency?

Spacesai1or:
So I'm a little confused on why we would want to do something like:

```javascript
import Contract, { ContractOptions} from 'web3-eth-contract';
```

If we update the packages, in this case `web3-eth-contract`, to have the default export be an object containing all the exports such as:

```javascript
export default {
    Contract,
    ContractOptions
}
```

You would still be able to do:

```javascript
import { Contract, ContractOptions} from 'web3-eth-contract';
```

Am I mistaken, is there a reason you wouldn't want to do it that way?

ariesjia:
@spacesailor24 
The current version the declaration tell us should use  `import { Contract, ContractOptions } from 'web3-eth-contract';` , but the  `Contract` is got `undefined`
Because source code is
```js
module.exports = Contract;
```
for ES module it equals  `export default`.  So the typing is not match the source.
If we use  `import  Contract, { ContractOptions } from 'web3-eth-contract';`, we will got the correct object ,  but  we will get Typescript compile err due to the declaration.

If we change the export will bring API broken change， this commit is only for quick fix

BryanLaura (approved):

BryanLaura (approved):
Fix #3732

BryanLaura (commented):
`export` `import` `default `

PR #3743, opened 2020-10-13
Add eth2 provider

Spacesai1or:
## Description

- Adds `web3-eth2-core` package
- Adds `web3-eth2-beaconchain` package

### Notes

- `web3-eth2-beaconchain` has a file located in `web3-eth2-beaconchain/src/schema.ts` where all the supported API  methods are defined

## Why is this a draft

- [ ] Lodestar types are not properly implemented for API wrapper methods
- ~[ ] Some methods listed in `web3-eth2-beaconchain/src/schema.ts` are not implemented in Lodestar yet (need to verify with Lodestar team)~
- ~[ ] Some implemented methods don't return expected values based off [official spec](https://ethereum.github.io/eth2.0-APIs/#/)~

Created [an issue](https://github.com/ChainSafe/lodestar/issues/1662) to track the following:
- Some methods listed in `web3-eth2-beaconchain/src/schema.ts` are not implemented in Lodestar yet (need to verify with Lodestar team)
- Some implemented methods don't return expected values based off [official spec](https://ethereum.github.io/eth2.0-APIs/#/)

## Type of change

## Checklist:


packages/web3-eth2-core/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "version": "1.3.0",
+    "description": "Web3 ETH2 core tools for sub-packages. This is an internal package.",
+    "repository": "https://github.com/ethereum/web3.js/tree/1.x/packages/web3-eth2-core",
+    "license": "LGPL-3.0",
+    "engines": {
+        "node": ">=8.0.0"
+    },
+    "types": "types/index.d.ts",
+    "scripts": {
+        "dtslint": "dtslint --localTs ../../node_modules/typescript/lib types",
```
GregTheGreek, line 12: Do we have to enter the node_modules?

packages/web3-eth2-core/package.json
```diff
@@ -0,0 +1,28 @@
...
+        "test": "jest",
+        "tsc": "tsc -b tsconfig.json"
+    },
+    "main": "lib/index.js",
+    "dependencies": {
+        "@types/node": "^12.12.6",
```
GregTheGreek: I think this can be a dev dependency?

packages/web3-eth2-core/package.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "web3-eth2-core",
+    "version": "1.3.0",
+    "description": "Web3 ETH2 core tools for sub-packages. This is an internal package.",
+    "repository": "https://github.com/ethereum/web3.js/tree/1.x/packages/web3-eth2-core",
+    "license": "LGPL-3.0",
```
GregTheGreek, line 6: I think this has to change to MIT, should match repo licensing

packages/web3-eth2-core/src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    [ key: string ]: ETH2Function | any;
+    
+    name: string
+    provider: string
+    protectProvider: boolean
+
+    constructor(provider: string, schema: IBaseAPISchema, opts: ETH2BaseOpts = {}) {
+        this.name = schema.packageName
+        this.setProvider(`${provider}${schema.routePrefix}`)
```
GregTheGreek, line 17: Might be worth adding a `/` between the two just in-case its missing that. `http://example.com//validator` is still a valid url even with a double `//`

packages/web3-eth2-core/src/index.ts
```diff
@@ -0,0 +1,89 @@
+import Axios, {AxiosInstance} from 'axios'
+
+import { ETH2BaseOpts, ETH2Function } from '../types/index'
+import { IBaseAPISchema } from './schema'
+
+export class ETH2Core {
+    private _httpClient: AxiosInstance
+
+    [ key: string ]: ETH2Function | any;
```
GregTheGreek, line 9: Do we need `any`?

packages/web3-eth2-core/src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    setProvider(provider: string) {
+        try {
+            if (!provider || typeof provider !== 'string' || !/^http(s)?:\/\//i.test(provider)) {
+                throw new Error(`Invalid HTTP(S) provider: ${provider}`)
+            }
+
+            const result = ETH2Core.createHttpClient(provider)
+            this._httpClient = result
```
GregTheGreek: Combine the two lines :)

packages/web3-eth2-core/src/index.ts
GregTheGreek: Looking at ETH2Function, since its a generic I think we can drop `any`. Thoughts?

packages/web3-eth2-beaconchain/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "devDependencies": {
+        "@chainsafe/lodestar-types": "^0.11.0",
+        "@types/jest": "^26.0.14",
+        "@types/node": "^14.11.2",
+        "dtslint": "^3.4.1",
+        "jest": "^26.4.2",
+        "ts-jest": "^26.4.1",
+        "typescript": "^4.0.3"
+    }
+}
```
GregTheGreek, line 29: Same comments from the `eth2-core` package.json

packages/web3-eth2-beaconchain/src/index.ts
```diff
@@ -0,0 +1,11 @@
+import { ETH2Core } from '../../web3-eth2-core/src/index'
+import { IETH2BeaconChain } from '../types/index'
+import { DefaultSchema } from './schema'
+import { IBaseAPISchema } from '../../web3-eth2-core/src/schema'
+
+// @ts-ignore
```
GregTheGreek: What error were you getting?

packages/web3-eth2-core/src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+                baseURL: baseUrl
+            })
+        } catch (error) {
+            throw new Error(`Failed to create HTTP client: ${error.message}`)
+        }
+    }
+
+    setProvider(provider: string) {
+        try {
+            if (!provider || typeof provider !== 'string' || !/^http(s)?:\/\//i.test(provider)) {
```
GregTheGreek, line 34: Do you want to add a check for `this.protectProvider` here?

packages/web3-eth2-beaconchain/src/schema.ts
```diff
@@ -0,0 +1,223 @@
...
+            name: 'getGenesis',
+            route: 'genesis',
+            restMethod: 'get',
+            inputFormatter: null,
+            outputFormatter: null,
+            errors: null,
+            errorPrefix: 'Failed to get genesis block:'
+        },
+        {
+            notImplemented: true,
```
GregTheGreek, line 17: `notImplemented` is defaulted true in the interface, perhaps you might want to change the logic surrounding how this works

packages/web3-eth2-beaconchain/test/schemaMethods.test.ts
```diff
@@ -0,0 +1,270 @@
...
+        genesis_validators_root: expect.stringMatching(/0x[a-f|A-F|\d]{64}/),
+        genesis_time: expect.stringMatching(/\d{10}/),
+        genesis_fork_version: expect.stringMatching(/0x[a-f|A-F|\d]{8}/)
+    }
+
+    const response = await eth2BeaconChain.getGenesis()
+    expect(response).toMatchObject(expectedResponse)
+})
+
+it('getHashRoot - NOT IMPLEMENTED IN LODESTAR', async () => {
```
GregTheGreek: Since these are Lodestar specific, perhaps renaming the file might be helpful

GregTheGreek (changes requested):
Really excited for this! Added a few comments below but nothing major!

packages/web3-eth2-core/package.json
Spacesai1or: I'm not sure, in this case I just copied from the previously existing packages (I think all current packages do this)

packages/web3-eth2-core/package.json
Spacesai1or: Totally! Must've been a mistype

packages/web3-eth2-core/package.json
Spacesai1or: Copied from other packages again, I think all the packages have `LGPL-3.0` as their license, should we create an issue to update this across the entire project?

packages/web3-eth2-core/src/index.ts
Spacesai1or: I understand your point, and am on the fence about this. I was trying to maintain zero assumptions on how a package author wants to structure their URL, but I could see how for most causes, it could lead to a typo. Looking at the [beacon chain API schema](https://github.com/ethereum/web3.js/blob/wyatt/eth2-refactor/packages/web3-eth2-beaconchain/src/schema.ts#L5) you can see that my intentions were for the provider to be a basic URI e.g. `http://localhost:1234` while the `routePrefix` would contain all slashes and such e.g. `/eth/v1/beacon/`.

I lean a little more towards keeping it how it is, only because it forces a package author to be completely explicit, instead of relying on any underlying assumptions from `ETH2Core`

packages/web3-eth2-core/src/index.ts
Spacesai1or: This was taken from Ethers.js, I believe. Removing it causes typescript to complain that class properties such as `name`, `provider`, and `protectProvider` are not of type `ETH2Function`, which they correctly aren't, and also that some class methods such as `setProvider` aren't of type `ETH2Function` because they don't return promises. If there's a better way to do this, I'm all for it, but I believe the `any` might be a necessary evil here

packages/web3-eth2-beaconchain/src/index.ts
Spacesai1or: `Class 'ETH2BeaconChain' incorrectly implements interface 'IETH2BeaconChain'.`

This was a bandaid solution for the TODO: `Lodestar types are not properly implemented for API wrapper methods` I mentioned above - I need help with this

packages/web3-eth2-core/src/index.ts
Spacesai1or: So `protectProvider` is for the global instance of `web3` to check if it should overwrite the provider for a particular sub package when using the future `.use` functionality. So if a user was to do:

```javascript
const web3 = new Web3(providerURL)
web3.use(Eth2Beaconchain(eth2ProviderURL, schema, { protectProvider: true }))

// Because the base web3 instance is probably going to use a different provider than the ETH2 packages,
// we want the use to have a control to dictate whether a certain subpackage should be
// affected by ther global use of:
web3.setProvider(newProviderURL)

// In this code example, web3.provider = newProviderURL
// while web3.eth2Beaconchain.provider = eth2ProviderURL
```

Spacesai1or: I did accidentally leave the call to `super` in the `ETH2BeaconChain` class as:

`super(provider, schema, { protectProvider: true })` instead of `super(provider, schema, opts)` where `opts` is passed by the user - will commit the changes for this soon

packages/web3-eth2-core/package.json
GregTheGreek: nvm we good

packages/web3-eth2-beaconchain/test/schemaMethods.test.ts
Spacesai1or: These routes aren't Lodestar specific, I've just been using Lodestar because it implements the official spec. I've take out the `...IN LODESTAR` from the test comment and just left it as `...NOT IMPLEMENTED`

frankiebee (approved):
THANK YOU! :partying_face:

GregTheGreek (approved):

PR #3766, opened 2020-10-29
Fix: mkdir directory already exists

RakibFiha:
## Description

mkdir directory already exists


scripts/e2e.windows.sh
```diff
@@ -7,7 +7,7 @@
 set -o errexit
 
 # Setup mock project to install web3 from virtual registry
-mkdir windows_test
+mkdir -p windows_test
```
GregTheGreek, line 10: @koraykoska was this an issue with you?

koraykoska (changes requested):
Merge the newest 1.x back into this PR and we can approve this.

Issue #3762, opened 2020-10-24
Wrong doc about the result of "new BN('0xea').toString();"


hihidev:
Following 
https://web3js.readthedocs.io/en/v1.3.0/web3-utils.html?highlight=BN#bn

Run:
new BN('0xea').toString();

#### Expected behavior
"234"

#### Actual behavior
"3450"

PS: web3.utils.toBN('0xea') returns "234" correctly

jdevassi:
When removing the single quotation marks -> expected behaviour.

On:
new BN(0xea).toString();

We get:
234

Same (previous comment's) conditions and results when using 'bn.js'  directly.

Spacesai1or:
Thank you for bringing this to our attention! This should be an easy one to fix :slightly_smiling_face:

PR #3770, opened 2020-11-03
fix: edit example error on utils-bn

cokia:
## Description

corrected the error in the web3-utils document, in the example of BN.

```
Run:
new BN('0xea').toString();

Expected behavior
"234"

Actual behavior
"3450"

PS: web3.utils.toBN('0xea') returns "234" correctly
```

The issue was generated here. #3762


cokia:
this is not stale PR.. :( 
please recheck it 👍

GregTheGreek:
~@cokia since we migrated the repo to chainsafe, you'll actually need to fill out our CLA~ 

Actually it looks like cla was added by automatically by our org rules. We will be removing that for web3.js

cokia:
In conclusion, do I not have to agree to the CLA up there?

GregTheGreek:
No please ignore it, ever since the repo transfer things got a little weird

GregTheGreek:
@cokia Is there a reason this was closed?

cokia:
I mistook this for a merged Pull Request.

vicnaum:
Actually, the correct value is 234.

It should had been corrected to:
`new BN('ea').toString();`

Spacesai1or:
> Actually, the correct value is 234.
> 
> It should had been corrected to:
> `new BN('ea').toString();`

This returns `NaN` without the `0x` prefix

I'm going to close this because running:

```javascript
const BN = require('bignumber.js')

console.log(new BN('0xea').toString())
```

returns `234`

vicnaum:
With bignumber.js - probably yes (I didn't check).
But I thought the default in web3 is BN.js?

With these two big number libraries - there's always some confusion...

PR #3773, opened 2020-11-04
Rename close event to disconnect

### LINKED ISSUE #3699: Metamask-deprecated .on({data, close}) used in constructor
Substantive details are described [here](https://github.com/MetaMask/metamask-extension/issues/9301), but this Issue is filed in web3 for updates which may need to be made in web3 for no longer relying on deprecated `data` and `close` events if they are, in fact, actually deprecated.

koraykoska:
Fixes #3699 

We will need to make sure that the close event is still added for old providers. Either by comparing versions or by registering both, which would lead to the same warning Metamask users experience right now for newer providers.

### DIFF
--- a/scripts/e2e.ganache.core.sh
+++ b/scripts/e2e.ganache.core.sh
@@ -11,6 +11,7 @@ set -o errexit
 # Install ganache-core
 git clone https://github.com/trufflesuite/ganache-core
 cd ganache-core
+git checkout tags/v2.13.0
 
 # Install via registry and verify
 echo ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
--- a/test/3_givenProvider-ganache.js
+++ b/test/3_givenProvider-ganache.js
@@ -30,7 +30,10 @@ describe('web.providers.givenProvider (ganache)', function(){
         assert.equal(1, web3.currentProvider.listenerCount('data'))
         assert.equal(1, web3.currentProvider.listenerCount('connect'))
         assert.equal(1, web3.currentProvider.listenerCount('error'))
-        assert.equal(1, web3.currentProvider.listenerCount('disconnect'))
+        // TODO: Remove close once the standard allows it
+        assert(
+            web3.currentProvider.listenerCount("disconnect") === 1 || web3.currentProvider.listenerCount("close") === 1
+        );
     });
 
     it('deploys a contract', async function(){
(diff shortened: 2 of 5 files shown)


koraykoska:
As described in [the EIP](https://eips.ethereum.org/EIPS/eip-1193#backwards-compatibility) this change has to be a strict superset, which means we are required to support the close event as well. This is what it does right now

frankiebee:
in order to be backwards compatible we probably still need to keep around the close event listener right?

koraykoska:
@frankiebee Yeah that's why I included the close event.

koraykoska:
@GregTheGreek

test/3_givenProvider-ganache.js
```diff
@@ -30,7 +30,7 @@ describe('web.providers.givenProvider (ganache)', function(){
         assert.equal(1, web3.currentProvider.listenerCount('data'))
         assert.equal(1, web3.currentProvider.listenerCount('connect'))
         assert.equal(1, web3.currentProvider.listenerCount('error'))
-        assert.equal(1, web3.currentProvider.listenerCount('close'))
```
GregTheGreek, line 33: can we add both the events to ensure we're backwards compatible?
Perhaps add a todo to remove `close` in the future

GregTheGreek (changes requested):

scripts/e2e.ganache.core.sh
```diff
@@ -11,6 +11,7 @@ set -o errexit
 # Install ganache-core
 git clone https://github.com/trufflesuite/ganache-core
 cd ganache-core
+git checkout tags/v2.13.0
```
frankiebee: is this because we need that version for `disconnect` event?

test/3_givenProvider-ganache.js
frankiebee: this is just tests

scripts/e2e.ganache.core.sh
koraykoska: @frankiebee Not really. First of all I don't see why we should take the newest develop branch commit as they have stable releases. Second, and more importantly, I stumbled upon this because the e2e Tests with ganache failed. The reason is that they added a patch file to their package.json after install script which should change a version of a dependency. But weirdly this patch file is broken. So rolling back to the release 2.13.0 fixed the issue.

I will make a PR at ganache-core to fix this issue and after that we can use the newest version again. But I would always add an explicit version to prevent tests from failing at some point with the same commit they passed before.

test/3_givenProvider-ganache.js
koraykoska: Yeah I can add the close event to the tests too.

koraykoska: @GregTheGreek Done

GregTheGreek (approved):

frankiebee (approved):
Looks good to me thanks

PR #3757, opened 2020-10-23
Release/1.3.1

GregTheGreek:
### Description

This release candidate fixes some errors throughout the codebase, and patches some security vulnerabilities.

We're also extremely excited to finally announce that web3.js has support for the eth2 beaconchain API! It should be noted that the eth2 API will follow the web3.js versioning and start at v1.3.1 (upon formally release next week) and should be considered an alpha release. It also will not bloat the existing web3 object, and was not imported into it. In order to use the eth2 beaconchain API please install it separately using `npm install --save web3-eth2-beaconchain`!

### Added

- Add `web3-eth2-core` package (#3743)
- Add `web3-eth2-beaconchain` package (#3743)

### Changed

- bump utils 0.10.0^ -> 0.12.0 (#3733)

### Removed

- Removed post-install script in `packages/web3`. Added documentation to root README (#3717)

### Fixed

- Fix possible unhandled promise rejection when sending a transaction (#3708)
- Fixed decoding bytes and string parameters for logs emitted with solc 0.4.x (#3724, #3738)
- Grammar changes to inputAddressFormatter error message
- Fixed vulnerable dependencies

PR #3814, opened 2020-12-11
increase timeouts

frankiebee:
Current tests on (rc)[#3757] are failing only in github actions

this fixes that by increasing the timeouts

With the new version of Firefox 83 the time to run the tests have increased thank you @cgewecke for finding this!

### DIFF
--- a/test/e2e.method.signing.js
+++ b/test/e2e.method.signing.js
@@ -394,22 +394,26 @@ describe('transaction and message signing [ @E2E ]', function() {
     // Smoke test to validate browserify's buffer polyfills (feross/buffer@5)
     // A companion regression test for Webpack & feross/buffer@4.9.2 exists at:
     // test/eth.accounts.webpack.js
-    it("encrypt then decrypt wallet", async function() {
-        this.timeout(10000);
+    it("encrypt then decrypt wallet", function(done) {
+        this.timeout(20000);
+        try {
+            const password = "qwerty";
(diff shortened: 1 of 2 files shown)


Spacesai1or (approved):

GregTheGreek (dismissed):

test/e2e.method.signing.js
```diff
@@ -394,22 +394,26 @@ describe('transaction and message signing [ @E2E ]', function() {
     // Smoke test to validate browserify's buffer polyfills (feross/buffer@5)
     // A companion regression test for Webpack & feross/buffer@4.9.2 exists at:
     // test/eth.accounts.webpack.js
-    it("encrypt then decrypt wallet", async function() {
-        this.timeout(10000);
-
-        const password = "qwerty";
-        const addressFromWallet = wallet[0].address;
+    it("encrypt then decrypt wallet", function(done) {
```
GregTheGreek, line 397: the test was changed to be async, you need to make sure that `done()` is called resolve the test

GregTheGreek (changes requested):

test/e2e.method.signing.js
frankiebee: done is called on line 413
it possible the timeout is not long enough :{
but hopefully the rerun will fix that.. something in the new update of firefox changed execution time of encrypt

GregTheGreek (approved):

PR #3817, opened 2020-12-15
add nvmrc file

frankiebee:
currently web3js will not successfully install deps with node version 15~ and npm version 7

please feel free to debate the node version am very willing to change this 

or weather we should specify in package.json file


Spacesai1or (approved):
I've tested installing web3 as a dependency using node `15.5.1`, and it works fine. However, installing the `devDependencies` with version `15.5.1` is broken, so I think adding the `.nvmrc` file with version `14.15.1` makes sense for developers of web3

koraykoska (dismissed):
Tested it and works for me.

Spacesai1or (changes requested):
Needs changelog update :)

Spacesai1or (approved):

Spacesai1or (approved):

PR #3818, opened 2020-12-16
Prevent ABI objects from being mutated

### LINKED ISSUE #3748: ABI objects passed into web3 functions may end up mutated
#### Expected behavior

ABI objects (or ABI item objects or ABI parameter objects) that are passed into web3 functions should not be mutated by web3, and should remain valid for use elsewhere without having to worry that web3 has altered them.

#### Actual behavior

If an ABI parameter object has type `function`, it will be mutated to have type `bytes24` instead.  As in, the actual object will be altered by a function that definitely shouldn't mutate it.

```
const Web3 = require("web3");
const web3 = new Web3();
const type = { type: "function", name: "", internalType: "function () external", indexed: false };
web3.eth.abi.decodeParameter(type, '0x0000000000000000000000000000000000000000000000000000000000000000');
console.log(type);
```

You will observe that `type` has been mutated, and thus will produce incorrect results if used anywhere else.

#### Further commentary and suggested fix

The problem seems to have been introduced in PR #3586, which attempts to add support for parameters of type `function`, by translating them to `bytes24`.  However, in doing so, it accidentally mutates the input.  The problem can be seen on [this line](https://github.com/ethereum/web3.js/pull/3586/files#diff-478b9def3657ee29af8ee7de117e8dcfe5f63f701fb773a8fc4eaada4aa00838R163), which mutates `type` rather than creating a modified clone.

So, my suggested fix is to change that line to clone rather than mutate.

(The context here, btw, is that we're trying to update Truffle to use Web 1.3.0, and this issue is causing test failures in tests involving `function` ABI parameters.)

#### Environment

Node: v10.22.1
Web3: 1.3.0

haltman-at:
## Description

Fixes #3748.  Prevents ABI parameter objects from being changed when used as inputs to encoding/decoding functions.  Does this by replacing `t.type = "bytes24"` with `t = Object.assign({}, t, { type: "bytes24" })`.

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -310,3 +310,4 @@ Released with 1.0.0-beta.37 code base.
 - Fixed decoding bytes and string parameters for logs emitted with solc 0.4.x (#3724, #3738)
 - Grammar changes to inputAddressFormatter error message
 - Fixed vulnerable dependencies
+- Fixed mutation of inputs to encoding and decoding functions (#3748)
--- a/packages/web3-eth-abi/src/index.js
+++ b/packages/web3-eth-abi/src/index.js
@@ -155,7 +155,7 @@ ABICoder.prototype.mapTypes = function (types) {
         // recognize former type. Solidity docs say `Function` is a bytes24
         // encoding the contract address followed by the function selector hash.
         if (typeof type === 'object' && type.type === 'function'){
-            type.type = "bytes24";
+            type = Object.assign({}, type, { type: "bytes24" });
         }
         if (self.isSimplifiedStructFormat(type)) {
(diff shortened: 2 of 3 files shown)


haltman-at:
I had trouble completing some of the checklist on my machine (and am still running some of it which I'll check off later), but since this is so simple I hope that's OK.

haltman-at:
Btw tagging @cgewecke and @GregTheGreek to make sure you two see this. :)

GregTheGreek:
@spacesailor24 cant seem to assign you...

GregTheGreek (approved):

Spacesai1or (approved):

PR #3824, opened 2020-12-23
Fix Web3 provider typings

daffl:
## Description

This pull request fixes the typings of the `web3-provider-*` packages. The modules use the default (`module.exports =`) export instead of a named export. The typing tests were only passing because the configuration aliased to the deprecated `web3-providers` module. This means the provider modules could not be used as intended in TypeScript because the named export does not exist and caused a runtime error and the default export was not typed.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/packages/web3-providers-http/types/index.d.ts
+++ b/packages/web3-providers-http/types/index.d.ts
@@ -43,7 +43,7 @@ export interface HttpProviderOptions {
     keepAlive?: boolean;
 }
 
-export class HttpProvider extends HttpProviderBase {
+export default class HttpProvider extends HttpProviderBase {
     host: string;
 
     withCredentials: boolean;
--- a/packages/web3-providers-http/types/tests/web3-provider-http-tests.ts
+++ b/packages/web3-providers-http/types/tests/web3-provider-http-tests.ts
@@ -22,8 +22,8 @@
 
 import * as http from 'http';
 import * as https from 'https';
-import { HttpProvider } from 'web3-providers';
 import { JsonRpcResponse } from 'web3-core-helpers';
+import HttpProvider from 'web3-providers-http';
 
 const httpProvider = new HttpProvider('http://localhost:8545', {
     timeout: 20000,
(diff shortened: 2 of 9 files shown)


Spacesai1or:
Hi there! Thank you for opening the PR!

I'm not incredibly familiar with Typescript, so forgive my ignorance, but it seems to be that named exports are [more preferable](https://basarat.gitbook.io/typescript/main-1/defaultisbad) and should work all the same?

daffl:
They are but the corresponding JavaScript code is assigning `module.exports =` which is a default export, see e.g. https://github.com/ChainSafe/web3.js/blob/134b34c568f0585497746861a7ac7562b52d189e/packages/web3-providers-http/src/index.js#L142

Changing those to a named export would be a breaking change.

Spacesai1or:
@daffl Can you update this with your latest changes? This lgtm

@GregTheGreek Do we need to remove the CLA bot from this?

GregTheGreek:
Cla is disabled so we should be able to merge without it

daffl:
Rebased with latest, I don't think any other changes were necessary.

Spacesai1or:
If you could just update the CHANGELOG.md, this would be good to go

koraykoska:
Correct me if I am wrong but if you change it to default exports now the changes are breaking for people using TS already right? I am not sure if we can do this in a minor release.

daffl:
Right now you can't use these modules directly with TS at all without ignoring the typings:

```ts
import WebsocketProvider from 'web3-providers-ws';

new WebsocketProvider(connect, options)
```

Gives you a compile time error like

```
This expression is not constructable.
  Type 'typeof import("web3-providers-ws/types/index")' has no construct signatures.
```

And

```ts
import { WebsocketProvider } from 'web3-providers-ws';

new WebsocketProvider(connect, options)
```

Throws a runtime error with

```
web3_providers_ws_1.WebsocketProvider is not a constructor
```

thebeanespinoza83 (approved):

thebeanespinoza83 (approved):

thebeanespinoza83 (approved):

thebeanespinoza83 (approved):

spb-web:
Temporary solution
`const Web3Eth = require('web3-eth') as typeof import('web3-eth')['Eth']`

mooo-ooo:
hi guys, is this solved ?

PR #3864, opened 2021-01-19
1.x: Fix EIP-1193 provider subscriptions

### LINKED ISSUE #3651: RequestManager depends on deprecated event when used with MetaMask
Refs: https://github.com/ethereum/web3.js/blob/1.x/packages/web3-core-requestmanager/src/index.js#L101-L108

`web3-core-requestmanager` depends on the provider `data` event, and makes assumptions about its state. I'm not sure if this even was ever intended for consumer usage in the misty past, but by now, it's superfluous.

For MetaMask providers (`ethereum.isMetaMask && typeof ethereum.request === 'function'`), please use the EIP-1193 `message` event, described in our docs: https://docs.metamask.io/guide/ethereum-provider.html#message

The `message` event emits with an object of the form `{ method: string, data: any }`. For `method === 'eth_subscription'`, the result is given as `data.result`.

rekmarks:
Fixes #3651 

Implements subscription support per [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193#subscriptions).

Should be tested manually by maintainers with current MetaMask (^9.0.0) on Firefox.

### DIFF
--- a/packages/web3-core-requestmanager/src/index.js
+++ b/packages/web3-core-requestmanager/src/index.js
@@ -98,14 +98,25 @@ RequestManager.prototype.setProvider = function (provider, net) {
 
     // listen to incoming notifications
     if (this.provider && this.provider.on) {
-        this.provider.on('data', function data(result, deprecatedResult) {
-            result = result || deprecatedResult; // this is for possible old providers, which may had the error first handler
+        if (typeof provider.request === 'function') { // EIP-1193 provider
+            this.provider.on('message', function (payload) {
+                if (payload && payload.type === 'eth_subscription' && payload.data) {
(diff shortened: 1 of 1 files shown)


frankiebee:
Cool thanks for this will take a look at this!

rekmarks:
FYI: I'm receiving reports that using `patch-package` with this PR fixes the issue.

GregTheGreek (approved):
LGTM

Spacesai1or (approved):
Installed this patch locally and tested using Firefox and MetaMask v9.0.1

frankiebee (approved):
purely from a code stand point this looks good!

PR #3908, opened 2021-02-09
Update soliditySha3 docs

### LINKED ISSUE #3907: Update docs for soliditySha3 to include security risk warning
As mentioned and discussed [here](https://github.com/ethereum/go-ethereum/pull/22273#issuecomment-775019023), a warning should be given to a user that multiple inputs can compute to the same hash using this method

Spacesai1or:
Fixes #3907


koraykoska (approved):
Totally something which should be fixed in terms of best practices. We should even think about deprecating it. I just tested it and the examples are correct.

koosy79 (commented):
Why can't I use js?

PR #3917, opened 2021-02-22
use posinstall to do deprecation warnings

frankiebee:
sorry for all the spam 😅


packages/web3/package.json
```diff
@@ -20,7 +20,9 @@
     "types": "types/index.d.ts",
     "scripts": {
         "tsc": "tsc -b tsconfig.json",
-        "dtslint": "dtslint --localTs ../../node_modules/typescript/lib types"
+        "dtslint": "dtslint --localTs ../../node_modules/typescript/lib types",
+        "preinstall": "echo \"WARNING: the web3-shh and web3-bzz api will be deprecated in the next version\"",
```
Spacesai1or: Do we need both pre and post? I think just post could suffice, no?

frankiebee: it was mentioned in slack that that was wanted can remove though if it's overkill

Spacesai1or (approved):

Spacesai1or (changes requested):
We should add warnings to individual packages (`bzz` and `shh`)

Spacesai1or (approved):

PR #3919, opened 2021-02-23
Upgrade use of ethereumjs-tx and ethereumjs-common

Spacesai1or:
## Why

In order to have support for [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718), [this PR](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1048) would need to be merged, and Web3.js would need to update to the latest version of `ethereumjs-X` pacakges. This PR handles updating the use of `ethereumjs-X` packages for signing transactions

## Changes made

- Update use of `ethereumjs-tx` to `@ethereumjs/tx` in `web3-eth-accounts`
- Update use of `ethereumjs-common` to `@ethereumjs/common` in `web3-eth-accounts`
- Update `web3-eth-accounts` tests accordingly

### ~Possible breaking change~

EDIT: `gasLimit` is defaulted to provided `gas` if not explicitly set by user

#### Change

`@ethereumjs/tx` requires the `gasLimit` property be set when [validating the tx](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/transaction.ts#L300), so `web3-eth-accounts` tests had to be updated to include `gasLimit` property and the `messageHash`was updated accordingly

#### Why it's breaking

I believe this to be a breaking change, as any current use of the library where `gasLimit` was not set when calling `web3.accounts.signTransaction` will now return the error: `gasLimit is too low. given ${this.gasLimit}, need at least ${this.getBaseFee()}` [link](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/transaction.ts#L300)

#### Workaround

I think the breaking change could be avoided by defaulting `gasLimit` to provided `gas` if `gasLimit === undefined`


Spacesai1or:
> This pull request **introduces 1 alert** when merging [2e94801](https://github.com/ChainSafe/web3.js/commit/2e948011bc457eeaecf58a4e7ab7a4707dce21f1) into [f196996](https://github.com/ChainSafe/web3.js/commit/f1969964d8905da5bac22b0646d5a742f9c0ab7d) - [view on LGTM.com](https://lgtm.com/projects/g/ChainSafe/web3.js/rev/pr-fde8637baf66afafee24be64c1d7193bb97fada3)
> 
> **new alerts:**
> 
> * 1 for Expression has no effect

Fixed this

Spacesai1or:
The failing tests reflect the breaking change I mentioned, so I believe we can either:

A. Bump the major version of the library
B. Default `gasLimit` to provided `gas` if `gasLimit === undefined`

test/eth.accounts.signTransaction.js
```diff
@@ -27,6 +27,7 @@ var tests = [
             nonce: 0,
             gasPrice: "20000000000",
             gas: 21000,
+            gasLimit: 21000,
```
Spacesai1or: If we default `gasLimit` to `gas`, then these changes should be reverted

frankiebee: revert this and a test that if gasLimit is passed that it is used and not the default

frankiebee (changes requested):
lets not do a breaking change and autofill gasLimit

Spacesai1or:
### Issue

So the tests failing because of (e.g. [here](https://github.com/ChainSafe/web3.js/pull/3919/checks?check_run_id=2063236242#step:6:467)):
```javascript
1) typical usage
       should deploy a contract, sign transaction, and return contract instance:

      AssertionError [ERR_ASSERTION]: Invalid address length
      + expected - actual

      -false
      +true
```
Are a result of the library currently setting the `to` property of a transaction to `0x` if not provided [here](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-eth-accounts/src/index.js#L161). The failing test is a result of `@ethereumjs/tx` checking if the `to` field is truthy, and if so it tries to parse an address from the value provided [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/baseTransaction.ts#L32)

### Possible Fix

Remove the line `transaction.to = transaction.to || '0x';` so `transaction.to` stays `undefined` and the `@ethereumjs/tx` object gets created as expected, however I'm not sure what the side effects are of this change

frankiebee:
Hmm yeah a null to should not default to 0x and should just be undefined or 0x0000...

packages/web3-eth-accounts/src/index.js
```diff
@@ -201,30 +201,32 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
-            ethTx.sign(Buffer.from(privateKey, 'hex'));
-
-            var validationResult = ethTx.validate(true);
-
-            if (validationResult !== '') {
-                throw new Error('Signer Error: ' + validationResult);
+            
+            var ethTx = Transaction.fromTxData(transaction, transactionOptions);
+            var signedTx = ethTx.sign(Buffer.from(privateKey, 'hex'));
+            var validationResult = signedTx.validate(true);
```
GregTheGreek: might make sense to rename this to `validationErrors` since the return values look like [this](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/baseTransaction.ts)

packages/web3-eth-accounts/src/index.js
```diff
@@ -201,30 +201,32 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
+
+            if (!validationResult || validationResult.length > 0) {
```
GregTheGreek: Since we're using `validate(true)` it will always return the array of errors `[]string` so I think we want to do:
```
if (validationResult.length > 0) { ... }
```

GregTheGreek (changes requested):

PR #3949, opened 2021-03-12
Bump geth-dev-assistant to 0.1.9

Spacesai1or:
This patch update includes a CLI flag bug fix that was causing some CI tests to fail


package.json
```diff
@@ -136,6 +136,5 @@
         "wait-port": "^0.2.9",
         "webpack": "^4.44.2",
         "webpack-cli": "^3.3.12"
-    },
-    "dependencies": {}
```
Spacesai1or, line 140: this was automatically removed by `npm i`

PR #3955, opened 2021-03-16
Rewrite web3-eth-iban

loredanacirstea:
## Description

Part of https://github.com/ChainSafe/web3.js/issues/3932

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules.
- [x] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [x] I ran `npm run test:unit` with success.
- [x] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code.
- [ ] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
- [ ] I have tested my code on the live network.
- [ ] I have checked the Deploy Preview and it looks correct.
- [ ] I have updated the `CHANGELOG.md` file in the root folder.

PR #3965, opened 2021-03-18
Feature/web3 eth iban es6 (#3964)

Spacesai1or:
* web3-eth-iban es6 rewrite - replace `var` with `const` and `let`

* web3-eth-iban es6 rewrite - replace object prototype with class

* web3-eth-iban code style fix

Co-authored-by: Loredana <loredana.cirstea@gmail.com>

Original PR: #3955


frankiebee (approved):
Sweeet

PR #3948, opened 2021-03-11
Fix: BSC - get transaction with gasLimit over 53 bits

alstjd0921:
## Description

Fix outputTransactionFormatter to format gas limit hex string to decimal string.

There are a lot of transactions in the BSC network that have set gas limit by 9,223,372,036,854,775,807 and I think it is necessary to switch to string format for look up these transactions.

The links below are examples of such transactions.
[0xb48fbbab4feee817b18efdbde134dba83ba47ea479ab85a2b3ff068bf7c8f5d8](https://bscscan.com/tx/0xb48fbbab4feee817b18efdbde134dba83ba47ea479ab85a2b3ff068bf7c8f5d8)
[0x02a3e52e32d1433f2d356ddf1408c3d9fb916c0a244f18e8a90ba742f45ae32a](https://bscscan.com/tx/0x02a3e52e32d1433f2d356ddf1408c3d9fb916c0a244f18e8a90ba742f45ae32a)
[0x65f46d95e70b1906694dbaaccaad12c2ffa9ae5c2a6513719837f50c9f97940a](https://bscscan.com/tx/0x65f46d95e70b1906694dbaaccaad12c2ffa9ae5c2a6513719837f50c9f97940a)

Fixes #3936
Fixes #3912
Fixes #3745

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Issue #3936, opened 2021-03-06
BSC - certain tx values break web3


eyooooo:
#### Expected behavior

this tx: https://bscscan.com/vmtrace?txhash=0xea6d2f025d6c2135b93d2c74a2a992c2ed80aac780298c1288b1addc2b6eb4d2&type=gethtrace2

i expect no errors when calling web3.eth.getTransaction(txhash)

#### Actual behavior

```
> web3.eth.getTransaction('0xea6d2f025d6c2135b93d2c74a2a992c2ed80aac780298c1288b1addc2b6eb4d2')
Promise { <pending> }
> (node:1992) UnhandledPromiseRejectionWarning: Error: Number can only safely store up to 53 bits
    at assert (/home/user/doodoocrap/node_modules/number-to-bn/node_modules/bn.js/lib/bn.js:7:21)
    at BN.toNumber (/home/user/doodoocrap/node_modules/number-to-bn/node_modules/bn.js/lib/bn.js:508:7)
    at Object.hexToNumber (/home/user/doodoocrap/node_modules/web3-utils/lib/utils.js:209:24)
    at Method.outputTransactionFormatter [as outputFormatter] (/home/user/doodoocrap/node_modules/web3-core-helpers/lib/formatters.js:204:20)
    at Method.formatOutput (/home/user/doodoocrap/node_modules/web3-core-method/lib/index.js:147:54)
    at sendTxCallback (/home/user/doodoocrap/node_modules/web3-core-method/lib/index.js:523:33)
    at Object.callback (/home/user/doodoocrap/node_modules/web3-core-requestmanager/lib/index.js:308:9)
    at /home/user/doodoocrap/node_modules/web3-providers-ws/lib/index.js:114:45
    at Array.forEach (<anonymous>)
    at WebsocketProvider._onMessage (/home/user/doodoocrap/node_modules/web3-providers-ws/lib/index.js:102:69)
    at W3CWebSocket._dispatchEvent [as dispatchEvent] (/home/user/doodoocrap/node_modules/yaeti/lib/EventTarget.js:115:12)
    at W3CWebSocket.onMessage (/home/user/doodoocrap/node_modules/websocket/lib/W3CWebSocket.js:234:14)
    at WebSocketConnection.<anonymous> (/home/user/doodoocrap/node_modules/websocket/lib/W3CWebSocket.js:205:19)
    at WebSocketConnection.emit (events.js:315:20)
[...]

Issue #3912, opened 2021-02-17
web3.eth.getBlock() throw Error: Number can only safely store up to 53 bits


kevin-leptons:
#### Expected behavior

Call function `web3.eth.getBlock(number, true)` without error.

#### Actual behavior

Call function `web3.eth.getBlock(number, true)` throw error `Error: Number can only safely store up to 53 bits`.

#### Steps to reproduce the behavior

```js
let provider = new Web3.providers.WebsocketProvider('ws://somewhere.com')
let web3 = new Web3(provider)
let block = await web3.eth.getBlock(595091, true)
```
#### Logs

```bash
Error: Number can only safely store up to 53 bits
    at assert (/home/kevin/package/bsc-txs-crawler/node_modules/number-to-bn/node_modules/bn.js/lib/bn.js:6:21)
    at BN.toNumber (/home/kevin/package/bsc-txs-crawler/node_modules/number-to-bn/node_modules/bn.js/lib/bn.js:506:7)
    at Object.hexToNumber (/home/kevin/package/bsc-txs-crawler/node_modules/web3-utils/lib/utils.js:208:24)
    at outputTransactionFormatter (/home/kevin/package/bsc-txs-crawler/node_modules/web3-core-helpers/lib/formatters.js:205:20)
    at /home/kevin/package/bsc-txs-crawler/node_modules/web3-core-helpers/lib/formatters.js:269:24
    at Array.forEach (<anonymous>)
    at Method.outputBlockFormatter [as outputFormatter] (/home/kevin/package/bsc-txs-crawler/node_modules/web3-core-helpers/lib/formatters.js:267:28)
    at Method.formatOutput (/home/kevin/package/bsc-txs-crawler/node_modules/web3-core-method/lib/index.js:147:54)
    at sendTxCallback (/home/kevin/package/bsc-txs-crawler/node_modules/web3-core-method/lib/index.js:523:33)
    at Object.callback (/home/kevin/package/bsc-txs-crawler/node_modules/web3-core-requestmanager/lib/index.js:308:9)
    at /home/kevin/package/bsc-txs-crawler/node_modules/web3-providers-ws/lib/index.js:114:45
    at Array.forEach (<anonymous>)
    at WebsocketProvider._onMessage (/home/kevin/package/bsc-txs-crawler/node_modules/web3-providers-ws/lib/index.js:102:69)
[...]

Issue #3829, opened 2021-01-02
number overflow


guotie:
#### Steps to reproduce the behavior

1. [First step]
2. [Second step]
3. [and so on...]

#### Logs

```
"stack":"Error: overflow (fault=\"overflow\", operation=\"toNumber\", value=\"29552863752176386703968378516918486365868501552295265006012618974855157514240\", code=NUMERIC_FAULT, version=bignumber/5.0.12)\n    at Logger.makeError 
(e:\\guotie\\project\\clover\\expoler\\node_modules\\@ethersproject\\logger\\lib\\index.js:179:21)\n
    at Logger.throwError (e:\\guotie\\project\\clover\\expoler\\node_modules\\@ethersproject\\logger\\lib\\index.js:188:20)\n
    at throwFault (e:\\guotie\\project\\clover\\expoler\\node_modules\\@ethersproject\\bignumber\\lib\\bignumber.js:295:19)\n
    at BigNumber.toNumber (e:\\guotie\\project\\clover\\expoler\\node_modules\\@ethersproject\\bignumber\\lib\\bignumber.js:152:13)\n
    at e:\\guotie\\project\\clover\\expoler\\node_modules\\web3-eth-abi\\node_modules\\@ethersproject\\abi\\lib\\coders\\array.js:89:60\n
    at Array.forEach (<anonymous>)\n
    at Object.unpack (e:\\guotie\\project\\clover\\expoler\\node_modules\\web3-eth-abi\\node_modules\\@ethersproject\\abi\\lib\\coders\\array.js:85:12)\n
    at TupleCoder.decode (e:\\guotie\\project\\clover\\expoler\\node_modules\\web3-eth-abi\\node_modules\\@ethersproject\\abi\\lib\\coders\\tuple.js:39:49)\n
    at AbiCoder.decode (e:\\guotie\\project\\clover\\expoler\\node_modules\\web3-eth-abi\\node_modules\\@ethersproject\\abi\\lib\\abi-coder.js:93:22)\n
    at ABICoder.decodeParametersWith (e:\\guotie\\project\\clover\\expoler\\node_modules\\web3-eth-abi\\lib\\index.js:304:30)\n
    at ABICoder.decodeParameters (e:\\guotie\\project\\clover\\expoler\\node_modules\\web3-eth-abi\\lib\\index.js:285:17)\n
[...]

PR #3976, opened 2021-03-25
Number Overflow Issue

### LINKED ISSUE #3745: Error while decode transaction gas: "Number can only safely store up to 53 bits"
I tried decode transaction from block but got a error 

```
Number can only safely store up to 53 bits
	at assert ./node_modules/number-to-bn/node_modules/bn.js/lib/bn.js:6
	at BN.toNumber ./node_modules/number-to-bn/node_modules/bn.js/lib/bn.js:506
	at Object.hexToNumber ./node_modules/web3-core-helpers/node_modules/web3-utils/lib/utils.js:208
	at Method.outputTransactionFormatter [as outputFormatter] ./node_modules/web3-core-helpers/lib/formatters.js:204
	at Method.formatOutput ./node_modules/web3-core-method/lib/index.js:147
	at sendTxCallback ./node_modules/web3-core-method/lib/index.js:523
	at ./node_modules/web3-core-requestmanager/lib/index.js:293
	at XMLHttpRequest.request.onreadystatechange ./node_modules/web3-providers-http/lib/index.js:98
	at XMLHttpRequestEventTarget.dispatchEvent ./node_modules/xhr2-cookies/xml-http-request-event-target.ts:44
	at XMLHttpRequest._setReadyState ./node_modules/xhr2-cookies/xml-http-request.ts:219
	at XMLHttpRequest._onHttpResponseEnd ./node_modules/xhr2-cookies/xml-http-request.ts:345
	at IncomingMessage.<anonymous> ./node_modules/xhr2-cookies/xml-http-request.ts:311
	at IncomingMessage.emit events.js:326
	at endReadableNT _stream_readable.js:1252
	at processTicksAndRejections internal/process/task_queues.js:80
```

error occur with tx id `0xe0db32759810271eec05326a4c96fbb42a185139134e16b4581f77456df226a9`

if look closely on https://bscscan.com/tx/0xe0db32759810271eec05326a4c96fbb42a185139134e16b4581f77456df226a9

we can see that gas equal `9,223,372,036,854,775,807`, so web3 js can't convert this string to number 

Anyone has idea how resolve this problem? 

```
import Web3 from 'web3';

async function main() {
  const web3 = new Web3('https://bsc-dataseed1.binance.org:443');

  const res = await web3.eth.getBlock(1365710);

  await Promise.all(
    res.transactions.map(async (txId) => {
      try {
        const tx = await web3.eth.getTransaction(txId);

        console.log(`decode ${txId}`);
      } catch (error) {
        console.error(`error decode ${txId}`);
      }
    }),
  );
}

main();
```

Spacesai1or:
Updates the use of `hexToNumber` to `outputBigNumberFormatter` (which returns a number string instead of a number)

This currently doesn't update other potentially problematic places such as [here](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-eth/src/index.js#L517) and [here](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-eth-contract/src/index.js#L915)

Fixes #3745, #3912, #3829, and #3936
Supersedes #3789 and #3948

## Notes

- This is a **breaking change** since it updates returns values from numbers to number strings, hence the base branch is `3.x`
- A testing repo that uses problematic code from #3745 and #3912 can be found [here[(https://github.com/spacesailor24/web3-3976)


alstjd0921:
Currently, no errors are being seen anywhere other than the process of formatting the gas value. I think it's too radical a change.

Spacesai1or:
@alstjd0921 Thank you for commenting/reviewing the code, this was actually a point of discussion I wanted to have:

I've tested your #3948 against the test repo I made, and it does pass using your branch, so presumably it fixes all the mentioned issues. 

However, even though changing one field is more desirable that forcing users to update multiple fields, I want to avoid having to make another breaking change in the future where some field related to gas overflows. So, I would like to keep any updates I've made to fields pertaining to gas, but revert other changes like `block.number` or `transactionIndex`, as they probably won't come near exceeding JavaScript's max integer.

To be specific, the fields I want to remain returning number strings are: `gas`, `cumulativeGasUsed`, `gasUsed`, and `gasLimit` (I will be updating this PR accordingly after this comment)

How do you (and anyone else reading this) feel about this?

P.s. in the rewrite we've started, we will be using the native BigInt, so this won't be a problem going forward

Spacesai1or:
~Blocked by #3982~

alstjd0921:
@spacesailor24 As you said, it seems desirable to limit the problem to gas-related things.
I think it makes sense now. I was just concerned about the change of unrelated with gas overflow (e.g., timestamp, index, or other things) in commit a687dde.

Spacesai1or:
@alstjd0921 if you want to merge [this](https://github.com/alstjd0921/web3.js/pull/1), then we could merge your #3948

alstjd0921:
@spacesailor24 
I merged [this](https://github.com/alstjd0921/web3.js/pull/1) but seems like some checks were broken. Should i handle it?

Spacesai1or:
@alstjd0921 Thought I fixed all the tests, but missed the CI ones. I'm going to work on getting those fixed today

Spacesai1or:
Am now just realizing that the failing tests are a result of pulling the [ganache-core](https://github.com/trufflesuite/ganache-core) repo and running their tests, so I'm trying to figure out what to do about that because of the breaking change

Spacesai1or:
So the plan will be to merge [this](https://github.com/ChainSafe/ganache-core/pull/1), branch off `develop`, update the tests, then target the branch for failing check

frankiebee (approved):
Wool woop

Spacesai1or:
The was merged with failing `ganache-core` tests. I updated most of the failing tests in [this](https://github.com/ChainSafe/ganache-core/tree/2.13.2-overflow-fix), but there were 3 remaining that seem to need actual core code to be updated

sebsobseb:
Am I correct that the actual `hexToNumber` util method remains the same? It will still return a number?

For example, this check with triple equals will not be impacted? :
`web3.util.hexToNumber("0x4") === 4`

PR #3989, opened 2021-04-05
Fix #3425

### LINKED ISSUE #3425: XMLHttpRequest errors are not handled correctly

koraykoska:
Added the custom error

## Description

Fixes #3425 with a custom error.

## Type of change

## Checklist:


Spacesai1or:
@GregTheGreek I think a really good point was made [here](https://github.com/ChainSafe/web3.js/issues/3425#issuecomment-619650271) and it would be irresponsible to merge this in as a patch version

Spacesai1or (changes requested):
`CHANGELOG.md` needs to be updated, but we might need to think about releasing this as apart of `3.x` or something

PR #4015, opened 2021-04-30
web3-providers-base package

### LINKED ISSUE #4019: [Rewrite] Breakout web3-providers-base package

Spacesai1or:
Includes base class expected to be inherited by future `web3-providers-X` packages

Closes #4019


packages/web3-providers-base/src/index.ts
```diff
@@ -0,0 +1,28 @@
+// TODO Fix eslint issues
+// eslint-disable-next-line import/no-unresolved, import/extensions
+import { ProviderOptions } from '../types';
+
+export default class Web3ProviderBase {
+  providerString = ''
```
GregTheGreek: It's good practice with classes parameters to always add the type.
Also not sure if it applies to any of these but you can also set modifiers to them `protected`, `static`, `readonly`

i.e providerString: string = ""

packages/web3-providers-base/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  protectProvider = false
```
GregTheGreek: Thoughts on setting this to default to true?

packages/web3-providers-base/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  supportsSubscriptions = false
+
+  constructor(options: ProviderOptions) {
+    this.providerString = options.providerString;
+    this.protectProvider = options.protectProvider;
+    this.supportsSubscriptions = options.supportsSubscriptions;
+  }
+
+  setProvider(providerString: string) {
```
GregTheGreek: Perhaps this can be simplified to the `if` statement? Any reason to throw twice?

GregTheGreek: Another note, this actually doesn't quite solve the problem of protecting it, you'll need to actually use `get` and `set` to achieve those results.

I made a little demo for you :) [here](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgApQPYDdgBNoDyADmMBiAM7IDeAUMskZjvlAMphSgDmAXMhU48A3PUaZICMOmx5o-AEYYMAGwhwQohhQCuRIhihgKbHQooIuJMpUXK1G0QF9atBCrgUqAdQgKAzDIs0ABCnih0DEzAWHCQyAD6TLKsHFwg3MgAvMgA5LliYsmS0sxyUNnI8CoUEK4MCOQgEFIQuJXVtfUCegZGJmYWVqTkVDmddWKNlJw6UoYAFBjWo-xB5cQjlACUNGIMYAAWwBQAdEllqUIZlctbZ8nB7NfcWgzIRyenxS2lKdC3FaUb4SX7rVhvA7HM66fSGYymcyWYBAsbIO42GG9eEDJHDTFvFxiWpgcT-KALR7lNI8fiCdLcXaRd6cACee3e72AMGQC0+D1BUnB0F2R0wAHdkABRKCYCm5YUVE5kkptXLbSEs6HnKlXBmVXXQGkZTVOZAIOIIQ686Bypn7LUS6WyxYAAxlcoEEDApBuhqg-AAJNRbYZTgBbCBeODcCBOV0ah0uBhEhix0n+hb2zlQb06KAgD7ai7k42vMQuIkAeiryAAKhhkOLLdajihcNyYNAICAkAAac2HDSx5Cun6tXCK13IUBFlDTelzMCGD6N12zCCutyjUkKcKVZqS3wBRVhWoLajjv5PfgTAf+sv8ABEjVUT4HsL6CMGyNR-A3TjbNulCqBApwqBg3ALE+e61AAjGSTxljOFDPgOsFgf6iYLqB4GQdBGEIQ+LwocgIAYKSiBIF4wAKGo77IBhOqXEapAFoytBMf6lRPsoRAUE+yA1vW0JNuQuSkuKhgANbARQuEQVBMHhAhAAieBNvOQ4ZCgwBgAA-AxXEsVA2xAA)

packages/web3-providers-base/test/unit/setProvider.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        expect(web3ProviderBase).toMatchObject(
+            {...providerOptions, providerString: updatedProviderString})
+    })
+
+    it('should error because providerOptions.protectProvider = true', () => {
+        providerOptions.protectProvider = true
+        const web3ProviderBase = new Web3ProviderBase(providerOptions)
+        expect(() => web3ProviderBase.setProvider(updatedProviderString))
+            .toThrow(`Error setting provider: Provider is protected`)
+    })
```
GregTheGreek: Should add a test case where you attempt to change the provider by explicitly calling 
```
web3ProverBase.providerString = "woopsies"
```

packages/web3-providers-base/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface ProviderOptions {
+    providerString: string
+    protectProvider: boolean
+    supportsSubscriptions: boolean
```
GregTheGreek: Since you set defaults in the constructor you should consider adding the optional type `?` to the types with defaults:
```
supportsSubscriptions?: boolean
```

packages/web3-providers-base/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import { ProviderOptions } from '../types';
+
+export default class Web3ProviderBase {
+  providerString = ''
+
+  protectProvider = false
+
+  connected = false
+
+  supportsSubscriptions = false
```
GregTheGreek: Both properties aren't being used, are they needed right now?

GregTheGreek (changes requested):

packages/web3-providers-base/src/index.ts
Spacesai1or: Really? Seems redundant no? And I could definitely make use of the modifiers

packages/web3-providers-base/src/index.ts
Spacesai1or: So this is intended to be used with the eventual `.use` functionality, to give the a user an option to not overwrite the provider when using `combinedWeb3Packages.setProvider`

e.g. if `combinedWeb3Packages` consists of many eth1 packages and a couple eth2, the user would want to protect the eth2 packages when updating the provider for the eth1 packages - however, this does seem to be a bit tricky to manage, so maybe there's a better way to handle this?

packages/web3-providers-base/src/index.ts
Spacesai1or: Right now I've only written the HTTP package, so no, but they will be needed for WS and maybe IPC

packages/web3-providers-base/src/index.ts
Spacesai1or: Oof, okay, this makes sense - will update

packages/web3-providers-base/src/index.ts
GregTheGreek: its really just a readability thing

packages/web3-providers-base/src/index.ts
GregTheGreek: Maybe we can break this into an issue an add it in later? Just prefer not having unused vars floating

packages/web3-providers-base/src/index.ts
GregTheGreek: So maybe let those PR's introduce them, since you haven't written the ws you're making an assumption on the architecture. just a nit

packages/web3-providers-base/src/index.ts
Spacesai1or: Updated accordingly

packages/web3-providers-base/src/index.ts
Spacesai1or: Okay I removed unused vars, won't create issues, they'll just get added back as needed

packages/web3-providers-base/src/index.ts
Spacesai1or: `setProvider` now has no `protectProvider` check and uses get/set

packages/web3-providers-base/test/unit/setProvider.test.ts
Spacesai1or: No longer relevant - `protectProvider` removed

packages/web3-providers-base/types.ts
Spacesai1or: No longer relevant - `supportsSubscriptions` removed

packages/web3-providers-base/src/index.ts
Spacesai1or: You're right, they've been removed :)

packages/web3-providers-base/src/index.ts
```diff
@@ -0,0 +1,17 @@
+import { ProviderOptions } from '../types';
+
+export default class Web3ProviderBase {
+  private _providerString: string
```
GregTheGreek: nit: `_providerUrl` makes more sense?

packages/web3-providers-base/test/unit/providerString.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    it('should return providerString', () => {
+        const web3ProviderBase = new Web3ProviderBase(providerOptions)
+        expect(web3ProviderBase.providerString).toBe(providerOptions.providerString)
+    })
+
+    it('should update providerString', () => {
+        const web3ProviderBase = new Web3ProviderBase(providerOptions)
+        web3ProviderBase.providerString = updatedProviderString
+        expect(web3ProviderBase.providerString).toBe(updatedProviderString)
+    })
```
GregTheGreek, line 18: Add a test case to check if it will throw when you query for `web3Prodiver._providerString` its a sanity, but ensures no one removes the `private` from it in the future

GregTheGreek (dismissed):

packages/web3-providers-base/src/index.ts
Spacesai1or: Would you call it a URL for IPC as well?

packages/web3-providers-base/test/unit/providerString.test.ts
Spacesai1or: I tried, but tsc won't compile it :(

packages/web3-providers-base/src/index.ts
GregTheGreek: Technically you can think of the filesystem as a url `ipc:///` or `file:///`

packages/web3-providers-base/test/unit/providerString.test.ts
GregTheGreek: oh funny lol. That makes sense becasue its not allowed ahhaha

packages/web3-providers-base/src/index.ts
Spacesai1or: Updated `providerString` to `providerUrl`

GregTheGreek (approved):

PR #4020, opened 2021-05-03
web3-providers-http

### LINKED ISSUE #4018: [Rewrite] Breakout web3-providers-http package

Spacesai1or:
Extends `web3-providers-base` to provide HTTP client for submitting requests

Depends on #4015 
Closes #4018


packages/web3-providers-http/src/index.ts
```diff
@@ -0,0 +1,58 @@
+import axios, { AxiosInstance } from 'axios';
+import Web3ProviderBase from 'web3-providers-base';
+import { ProviderOptions } from 'web3-providers-base/types';
```
GregTheGreek: We could export the types in the root, and allow you to pull them all in the same import statement. I don't have a real preference over this though so up to you!

packages/web3-providers-http/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { HttpRpcOptions, HttpRpcResponse, IWeb3Provider } from '../types';
+
+export default class Web3ProvidersHttp extends Web3ProviderBase implements IWeb3Provider {
+  private _httpClient: AxiosInstance | undefined
```
GregTheGreek: is null better in this case?

packages/web3-providers-http/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  static validateProviderString(providerString: string): boolean {
+    try {
+      return (typeof providerString !== 'string'
+          || /^http(s)?:\/\//i.test(providerString));
+    } catch (error) {
+      throw Error(`Failed to validate provider string: ${error.message}`);
+    }
+  }
+
+  static createHttpClient(baseUrl: string): AxiosInstance {
```
GregTheGreek, line 24: Would users want to create an httpclient outside of this scope of the class itself?

packages/web3-providers-http/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+    try {
+      if (!Web3ProvidersHttp.validateProviderString(baseUrl)) throw Error('Invalid HTTP(S) URI provided');
+      return axios.create({ baseURL: baseUrl });
+    } catch (error) {
+      throw Error(`Failed to create HTTP client: ${error.message}`);
+    }
+  }
+
+  setProvider(providerString: string) {
```
GregTheGreek: This seems a bit confusing, because in the baseclass you protected the providerString, but in this class you're allowing the user to skip the bounds set on providerString. 

I need to revisit this, but I can't immediately think of a better architecture for this

packages/web3-providers-http/types.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export interface IWeb3Provider {
+    providerString: string
+    protectProvider: boolean
+    connected: boolean
+    supportsSubscriptions: boolean
+    setProvider: (providerString: string) => void
+    // TODO get rid of anys
+    send: (options: any) => Promise<any>
```
GregTheGreek: Can't you just use `HttpRpcOptions`?

GregTheGreek (changes requested):
Very good overall! 

I think the refactor on providerString from the base class PR will basically automatically clean this up!

packages/web3-providers-http/src/index.ts
Spacesai1or: I don't see why, but I did this so the same method would be copied in every instance of the class, but that probably doesn't matter?

packages/web3-providers-http/src/index.ts
Spacesai1or: Yea, `super.setProvider(providerString);` should come first, I was just battling an architecture issue where the protected provider error was being thrown. I'm going to try rethinking this with getters/setters

packages/web3-providers-http/types.ts
Spacesai1or: `IWeb3Provider` should be in the `web3-providers-base` package, as any provider package should conform to this interface, but `options` could be anything depending on how the provider package works. We could do `HttpRpcOptins | WsRpcOptions | IpcOptions| any` to be a little more explicit though

packages/web3-providers-http/src/index.ts
GregTheGreek: In my mind its only ever called in the constructor? I'm trying to think of other use cases

packages/web3-providers-http/src/index.ts
GregTheGreek: ahh! Maybe when you adjust the get/set in the other pr this one will solve itself

packages/web3-providers-http/types.ts
GregTheGreek: ahh ok, well in that case, you could have something like:
```
interface IWeb3Provider {
    providerString: string
    protectProvider: boolean
    connected: boolean
    supportsSubscriptions: boolean
    setProvider: (providerString: string) => void
    // TODO get rid of anys
    send: (options: any) => Promise<any>
 }
 
interface IHttpProvider extends IWeb3Provider {
  // if needed custom logic
}

File: packages/web3-providers-http/src/index.ts (hunk 1 above)
Comment (Spacesai1or) [reply packages/web3-providers-http/src/index.ts]: Mmm I'm not sure on this one

File: packages/web3-providers-http/src/index.ts (hunk 2 above)
Comment (Spacesai1or) [reply packages/web3-providers-http/src/index.ts]: Removed `undefined` all together, no reason this shouldn't be set - should error otherwise (think it was just a temp fix to make tsc happy)

File: packages/web3-providers-http/src/index.ts (hunk 4 above)
Comment (Spacesai1or) [reply packages/web3-providers-http/src/index.ts]: Not relevant anymore - `protectProvider` was removed

File: packages/web3-providers-http/types.ts (hunk 5 above)
Comment (Spacesai1or) [reply packages/web3-providers-http/types.ts]: I moved `IWeb3Provider` into `web3-providers-base/types`

File: packages/web3-providers-http/src/index.ts (hunk 4 above)
Comment (GregTheGreek) [reply packages/web3-providers-http/src/index.ts]: Thinking about the user flow
```
const provider = new Web3ProvidersHttp({providerString: "http://localhost"})
provider.setProvider("http://123.123.123") // actually creates new http instance
provider.providerString("http://444.444.444") // doesn't do anything
```
What benefit do we have having the provider string in the base class?

GregTheGreek (changes requested):

packages/web3-providers-http/types.ts
Spacesai1or: Didn't see you last comment when I made mine, `IHttpProvider` doesn't need to exist...probably

packages/web3-providers-http/src/index.ts
Spacesai1or: Actually, I tried this out, and I think I like separating type imports because it's explicit that they're types - going to revert my changes and keep it how it is, unless someone believes otherwise is better

packages/web3-providers-http/src/index.ts
Spacesai1or: So by making `validateProviderUrl` and `createHttpClient` `private`, we loose the ability to test those methods. I'm going to revert these changes and keep the use of `static` as these methods don't belong to instances, but rather the class itself

GregTheGreek (approved):

PR #4022, opened 2021-05-03
web3-core-requestmanager

### LINKED ISSUE #4021: [Rewrite] Breakout web3-core-requestmanager

Spacesai1or:
Wraps `web3-providers-http` to provide generic interface for submitting requests

depends on #4015 and #4020
closes #4021


packages/web3-core-requestmanager/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    constructor(providerOptions: ProviderOptions) {
+      switch (Web3RequestManager.detectProviderProtocol(providerOptions.providerString)) {
+        case ProviderProtocol.HTTP:
+          this.provider = new Web3ProviderHttp(providerOptions);
+          this.providerProtocol = ProviderProtocol.HTTP;
+          break;
+        case ProviderProtocol.WS:
+          this.providerProtocol = ProviderProtocol.WS;
+          // TODO
```
GregTheGreek, line 21: might be good practice to `throw` here so that we remember it doesn't exist

packages/web3-core-requestmanager/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+          break;
+        default:
+          // TODO figure out if possible to support generic provider
```
GregTheGreek, line 28: Probably throw for unknown format

packages/web3-core-requestmanager/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+          return ProviderProtocol.WS;
+        }
+        return ProviderProtocol.UNKNOWN;
+      } catch (error) {
+        throw Error(`Error detecting provider protocol: ${error.message}`);
+      }
+    }
+
+    // TODO get rid of anys
+    async send(options: any): Promise<any> {
```
GregTheGreek, line 49: I'm assuming since we're not supporting the callback scheme, and forcing promises we can get away with removing all the old boilerplate from the request manager found [here](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-core-requestmanager/src/index.js#L167-L190)

packages/web3-core-requestmanager/test/unit/constructor.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    it('should have set providerProtocol to WS', () => {
+        providerOptions.providerString = 'wss://127.0.0.1:8545'
+
+        const web3RequestManager = new Web3RequestManager(providerOptions)
+        // providerProtocol is an enum declared in src/index.ts
+        expect(web3RequestManager.providerProtocol).toBe(2) // 2 === WS
+    })
+
+    it('should have set providerProtocol to UNKNOWN', () => {
```
GregTheGreek, line 39: You can also test for ipc, i believe it must begin wtih `ipc://`

GregTheGreek (changes requested):

packages/web3-core-requestmanager/src/index.ts
Spacesai1or: Code now throws for `WS`, `IPC`, and `UNKNOWN`

packages/web3-core-requestmanager/src/index.ts
Spacesai1or: Yes, but I'm not so sure we can actually get away with this. I really hope so, because it's a heck of a lot of code that just clutters everything, but I imagine projects that use older JS aren't going to be happy

GregTheGreek: 1.x will be LTS for long enough that they can upgrade. :)

GregTheGreek (approved):

PR #4024, opened 2021-05-03
web3-eth

### LINKED ISSUE #4023: [Rewrite] web3-eth
- [x] Pull package into separate branch and open PR into `4.x`

Some of the following may not be implemented, but I'm listing all available calls for the sake of completion:

- [x] web3_clientVersion
- [x] web3_sha3
- [x] net_version
- [x] net_peerCount
- [x] net_listening
- [x] eth_protocolVersion
- [x] eth_syncing
- [x] eth_coinbase
- [x] eth_mining
- [x] eth_hashrate
- [x] eth_gasPrice
- [x] eth_accounts
- [x] eth_blockNumber
- [x] eth_getBalance
- [x] eth_getStorageAt
- [x] eth_getTransactionCount
- [x] eth_getBlockTransactionCountByHash
- [x] eth_getBlockTransactionCountByNumber
- [x] eth_getUncleCountByBlockHash
- [x] eth_getUncleCountByBlockNumber
- [x] eth_getCode
- [x] eth_sign
- [x] eth_signTransaction
- [x] eth_sendTransaction
- [x] eth_sendRawTransaction
- [x] eth_call
- [x] eth_estimateGas
- [x] eth_getBlockByHash
- [x] eth_getBlockByNumber
- [x] eth_getTransactionByHash
- [x] eth_getTransactionByBlockHashAndIndex
- [x] eth_getTransactionByBlockNumberAndIndex
- [x] eth_getTransactionReceipt
- [x] eth_getUncleByBlockHashAndIndex
- [x] eth_getUncleByBlockNumberAndIndex
- [x] eth_getCompilers
- [x] eth_compileLLL
- [x] eth_compileSolidity
- [x] eth_compileSerpent
- [x] eth_newFilter
- [x] eth_newBlockFilter
- [x] eth_newPendingTransactionFilter
- [x] eth_uninstallFilter
- [x] eth_getFilterChanges
- [x] eth_getFilterLogs
- [x] eth_getLogs
- [x] eth_getWork
- [x] eth_submitWork
- [x] eth_submitHashrate

### Not being implemented

- [ ] db_putString
- [ ] db_getString
- [ ] db_putHex
- [ ] db_getHex
- [ ] shh_post
- [ ] shh_version
- [ ] shh_newIdentity
- [ ] shh_hasIdentity
- [ ] shh_newGroup
- [ ] shh_addToGroup
- [ ] shh_newFilter
- [ ] shh_uninstallFilter
- [ ] shh_getFilterChanges
- [ ] shh_getMessages

Spacesai1or:
Uses `web3-core-requestmanager` to wrap Ethereum RPC calls

depends on #4015 and #4020
closes #4023


packages/web3-eth/types.ts
```diff
@@ -0,0 +1,160 @@
...
+export interface Web3EthOptions {
+    packageName?: string;
+    providerUrl: string;
+}
+
+export interface EthCallTransaction extends EthTransaction {
+    to: string;
+}
+
+export interface EthStringResult extends HttpRpcResponse {
```
Spacesai1or, line 131: `extends HttpRpcResponse` will probably be changed to `RpcResponse` when writing `web3-providers-ws`, but leaving it for now

packages/web3-eth/test/unit/constructor.test.ts
```diff
@@ -0,0 +1,23 @@
...
+            providerUrl: 'http://127.0.0.1:8545',
+        };
+    });
+
+    it('should construct with expected providerUrl', () => {
+        const web3Eth = new Web3Eth(web3EthOptions);
+        expect(web3Eth.packageName).toBe('eth');
+    });
+
+    it('should construct with expected providerUrl', () => {
```
Spacesai1or: Update test string

packages/web3-eth/types.ts
```diff
@@ -0,0 +1,160 @@
...
+    result: string[];
+}
+
+export interface EthBooleanResult extends HttpRpcResponse {
+    result: boolean;
+}
+
+export interface EthSyncingResult extends HttpRpcResponse {
+    result:
+        | {
```
Spacesai1or, line 145: Unnecessary pipe?

Spacesai1or: Was added by prettier

packages/web3-eth/types.ts
```diff
@@ -0,0 +1,160 @@
+import { HttpRpcResponse } from 'web3-providers-http/types';
+
+export type BlockIdentifier = number | 'latest' | 'earliest' | 'pending';
```
GregTheGreek: Does this make more sense as an enum?
for example:
```
enum BlockTags {
  latest: "latest",
  earliest: "earliest",
  pending: "pending"
}
export type BlockIdentifier = number | BlockTags
```
So now users could do something like
```
const block = await eth.getBlock(BlockTags.latest)
```

packages/web3-eth/types.ts
```diff
@@ -0,0 +1,160 @@
...
+};
+
+export type EthTransaction = {
+    from: string;
+    to?: string;
+    gas?: BigInt;
+    gasPrice?: BigInt;
+    value?: BigInt;
+    data?: string;
+    nonce?: number;
```
GregTheGreek, line 39: Aren't these always going to be present?

packages/web3-eth/types.ts
```diff
@@ -0,0 +1,160 @@
...
+        };
+    };
+};
+
+export type EthFilter = {
+    fromBlock?: BlockIdentifier;
+    toBlock?: BlockIdentifier;
+    address?: string;
+    topics?: string | null | string[][];
+    blochHash?: string;
```
GregTheGreek, line 119: Similar here, won't we have some sort of default? Don't we require at least one of these?

packages/web3-eth/types.ts
```diff
@@ -0,0 +1,160 @@
...
+          }
+        | false;
+}
+
+export interface EthAccountsResult extends HttpRpcResponse {
+    result: string[];
+}
+
+export interface EthBlockResult extends HttpRpcResponse {
+    result: EthBlock | null;
```
GregTheGreek, line 158: Do we want to return null or some sort of special error type?

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,1254 @@
...
+    EthTransactionResult,
+    EthTransactionReceiptResult,
+    EthStringArrayResult,
+    EthCompiledSolidityResult,
+    EthLogResult,
+    EthFilter,
+} from '../types';
+
+export default class Web3Eth {
+    private _packageName: string;
```
GregTheGreek: what exactly is this for?

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,1254 @@
...
+                ...rpcOptions,
+                method: 'eth_submitHashRate',
+                jsonrpc: rpcOptions?.jsonrpc || this._DEFAULT_JSON_RPC_VERSION,
+                params: [hashRate, clientId],
+            });
+        } catch (error) {
+            throw Error(`Error submitting hash rate: ${error.message}`);
+        }
+    }
+}
```
GregTheGreek, line 1246: It's a lot of code, but its also set it once, and forget it.

Generator would be nice, but as you even said having type safety (in a readable way) is better. Maybe we can revisit the generator that Nick showed us later!

GregTheGreek (changes requested):

packages/web3-eth/types.ts
Spacesai1or: This seems a bit redundant though, no? Like the user still has to type out `latest`, `earliest`, or `pending` to access the enum value, and using the enum means the user has to have an additional import to get access to the `BlockTags` enum

packages/web3-eth/types.ts
Spacesai1or: For the fields marked optional, they are actually defaulted by the node the request is sent to. Here is an example excerpt from the [ETH spec](https://eth.wiki/json-rpc/API):

> gas: QUANTITY - (optional, default: 90000) Integer of the gas provided for the transaction execution. It will return unused gas.

In the doc blocks I have this default behavior mentioned [e.g.](https://github.com/ChainSafe/web3.js/blob/wyatt/rewrite/eth/packages/web3-eth/src/index.ts#L588)

However, I can add the same doc blocks to `types.ts` in case users are only looking at the types

packages/web3-eth/types.ts
Spacesai1or: See previous comment, but stemming from this is the idea of adding validation: #4076

packages/web3-eth/types.ts
Spacesai1or: Taken from the ETH spec:

> Object - A block object, or null when no block was found:

So I think `null` makes sense here, also `null` is returned when transactions and such haven't been mined and aren't found yet

packages/web3-eth/src/index.ts
Spacesai1or: This was for the `.use` functionality, but it's not being used right now, so I will remove it

Spacesai1or: Addressed

packages/web3-eth/types.ts
Spacesai1or: Addressed

packages/web3-eth/types.ts
GregTheGreek: the main difference is you guarantee the user uses the correct value.

A raw string like "latest" isn't typesafe, but an enum would be (wrt to Typescript)

packages/web3-eth/types.ts
Spacesai1or: Updated this in the `wyatt/rewrite/http-subscriptions` branch, but this needs to be merged first

packages/web3-eth/types.ts
Spacesai1or: Code was updated accordingly

GregTheGreek (approved):

PR #4061, opened 2021-05-13
Add unsubscribeByID

### LINKED ISSUE #4007: Unsubscring Using Subscription ID
In the docs https://web3js.readthedocs.io/en/v1.3.4/web3-eth-subscribe.html#subscribe
a return of the subscribe function is `subscribtion.id` , the docs mentions as follows
```
subscription.id: The subscription id, used to identify and unsubscribing the subscription.
```
however i have been unable to find any information relating to the same.

How can `subscription.id` be used to unsubscribe a subscription

Spacesai1or:
closes #4007 
Addressing this feature, 
call ```web3.eth.removeSubscriptionById(id)``` to unsubscribe by ID


test/eth.subscribe.ganache.js
```diff
@@ -63,6 +63,20 @@ describe('subscription connect/reconnect', function () {
...
     });
 
+    it('unsubscribes given an id', function (done) {
+        assert.equal(0, web3.eth._requestManager.subscriptions.size);
+        subscription = web3.eth
+            .subscribe('newBlockHeaders')
+            .on('connected', function (result) {
+                assert(result)
+                assert.equal(1, web3.eth._requestManager.subscriptions.size);
+                subscription.unsubscribeById(subscription.id); // Stop listening..
```
Spacesai1or: I don't think adding `unsubscribeById` to an instance of the `Subscription` class makes sense, since [unsubscribe](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-core-subscriptions/src/subscription.js#L178) already exists - this would just be adding a semi duplicate method to the class (bloat). Instead, if we can make `unsubscribeById` a static method, or similar, and we could have it unsubscribe with just the `id`, and not have to have an instance available to call the method from, I could see this being a useful method to add

Spacesai1or: I put this comment here to reference what I'm talking about

Instead of doing:

```javascript
subscription = web3.eth.subscribe(...')
subscription.unsubscribeById(...)
```

It would be cool if you could find a way to do

```javascript
Subscription.unsubscribeById(...)
```

Or maybe it would be

```javascript
web3.eth.subscribe.unsubscribeById(...)
```

so you don't have to have access to an instance of `Subscription`. But maybe what you have already works like that too? Idk, can you tell me? :)

Spacesai1or (commented):
Yea, I shouldn't have created that for you, now it's my PR and I can't request changes or approve xD

test/eth.subscribe.ganache.js
GregTheGreek: can probably make use of looking through `web3.eth._requestManager.subscriptions` and loop through the `id`s! Great suggestion

luu-alex: Cool! the request manager was helpful :). 
It can now be called with
```
web3.eth. removeSubscriptionById( id)
```

GregTheGreek (approved):

adnandothussain:
Seems like the typescript is update for this method

PR #4069, opened 2021-05-17
remove underscore

### LINKED ISSUE #4066: Remove underscore
Currently the [underscore](https://www.npmjs.com/package/underscore) package is used very lightly withing web3.js, and the associated methods are not hat complex. By removing underscore we can reduce the package size by ~1mb .

### Solution
Remove underscore and replace the methods with a set of custom helper methods instead.

luu-alex:
## Description
closes #4066

## Type of change

## Checklist:


packages/web3-bzz/src/index.js
```diff
@@ -22,7 +22,7 @@
 
 "use strict";
 
-var _ = require('underscore');
+
```
GregTheGreek:
```suggestion
```

packages/web3-core-method/src/index.js
```diff
@@ -180,6 +179,8 @@ Method.prototype.formatOutput = function (result) {
 Method.prototype.toPayload = function (args) {
     var call = this.getCall(args);
     var callback = this.extractCallback(args);
+    //check if last args is a function, if so pop it out.
+
```
GregTheGreek: ?

packages/web3-core-method/src/index.js
```diff
@@ -206,8 +207,8 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
         intervalId = null,
         lastBlock = null,
         receiptJSON = '',
-        gasProvided = (_.isObject(payload.params[0]) && payload.params[0].gas) ? payload.params[0].gas : null,
-        isContractDeployment = _.isObject(payload.params[0]) &&
+        gasProvided = ((typeof payload.params[0] === 'object' && !!payload.params[0]) && payload.params[0].gas) ? payload.params[0].gas : null,
```
GregTheGreek:
```suggestion
        gasProvided = ((typeof payload.params[0] === 'object' && payload.params[0]) && payload.params[0].gas) ? payload.params[0].gas : null,
```

Double bang operator not necessary, @spacesailor24 cab we run into weird falsey values?

packages/web3-core-method/src/index.js
```diff
@@ -206,8 +207,8 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
...
+        isContractDeployment = (typeof payload.params[0] === 'object' && !!payload.params[0]) &&
```
GregTheGreek: Mimic above change

packages/web3-core-subscriptions/src/subscription.js
```diff
@@ -22,16 +22,19 @@
 
 "use strict";
 
-var _ = require('underscore');
 var errors = require('web3-core-helpers').errors;
 var EventEmitter = require('eventemitter3');
 var formatters = require('web3-core-helpers').formatters;
 
+function identity(value) {
```
GregTheGreek: Please add comment explaining why this exists

test/abi.decodeParameter.js
```diff
@@ -349,10 +348,11 @@ describe('lib/solidity/coder', function () {
...
                 var resultArray = [];
-                _.each(result, function (res, key) {
-                    if(_.isFinite(key))
-                        resultArray.push(res);
-                });
+                for (var key in result) {
+                       if(isFinite(key)) 
+                            resultArray.push(result[key]);
+                }
+
```
GregTheGreek, line 354: consider converting to a Array.map

```
Object.keys(result).map(x => result[x])
```

GregTheGreek (dismissed):

packages/web3-core-method/src/index.js
Spacesai1or: I don't know if it's specifically needed here, but usually the `!!` is used to cast the value to a boolean, but it does seem redundant

GregTheGreek: wait you might be right, i think we should keep this

packages/web3-core-method/src/index.js
```diff
@@ -706,12 +706,13 @@ Method.prototype.buildCall = function () {
...
                 if (payload.method === 'eth_sendTransaction') {
                     var tx = payload.params[0];
-                    wallet = getWallet((_.isObject(tx)) ? tx.from : null, method.accounts);
+                    wallet = getWallet((typeof tx === 'object' && !!tx) ? tx.from : null, method.accounts);
 
 
                     // If wallet was found, sign tx, and send using sendRawTransaction
                     if (wallet && wallet.privateKey) {
-                        var txOptions = _.omit(tx, 'from');
+                        var txOptions = JSON.parse(JSON.stringify(tx));
```
Spacesai1or:
```suggestion
```

`_.omit` [takes an object](https://underscorejs.org/#omit), so we can assume `tx` is an object

packages/web3-core-method/src/index.js
```diff
@@ -706,12 +706,13 @@ Method.prototype.buildCall = function () {
...
+                        delete txOptions['from'];
```
Spacesai1or:
```suggestion
                        delete tx.from;
```

This is the usual syntax for deleting a property. You'd only want to do `[property]` if `property` was a variable or some function/if statement that resolves to a string. See [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete)

packages/web3-eth-accounts/src/index.js
```diff
@@ -334,7 +333,7 @@ Accounts.prototype.recover = function recover(message, signature, preFixed) {
     var args = [].slice.apply(arguments);
 
 
-    if (_.isObject(message)) {
+    if (typeof message === 'object' && !!message) {
```
Spacesai1or: ~I don't understand why `!!message` is necessary, an empty object is truthy (applies to all other occurrences of `!!`)~

Nvm, I see that `underscore.js` uses this for their `isObject` [check](https://github.com/jashkenas/underscore/blob/master/modules/isObject.js)

packages/web3-eth-accounts/src/index.js
```diff
@@ -468,7 +467,7 @@ function Wallet(accounts) {
 
 Wallet.prototype._findSafeIndex = function(pointer) {
     pointer = pointer || 0;
-    if (_.has(this, pointer)) {
+    if (this != null && this.hasOwnProperty(pointer)) {
```
Spacesai1or:
```suggestion
    if (this.hasOwnProperty(pointer)) {
```
You don't need to check if `this` is defined, I'm pretty sure it was always be defined within a function

packages/web3-eth-accounts/src/index.js
```diff
@@ -37,7 +36,7 @@ var Common = require('@ethereumjs/common').default;
 
 
 var isNot = function(value) {
-    return (_.isUndefined(value) || _.isNull(value));
+    return (typeof value === 'undefined') || value == null;
```
Spacesai1or:
```suggestion
    return (typeof value === 'undefined') || value === null;
```

packages/web3-eth-accounts/src/index.js
```diff
@@ -497,7 +496,7 @@ Wallet.prototype.create = function(numberOfAccounts, entropy) {
 
 Wallet.prototype.add = function(account) {
 
-    if (_.isString(account)) {
+    if (typeof account == 'string') {
```
Spacesai1or:
```suggestion
    if (typeof account === 'string') {
```

packages/web3-eth-contract/src/index.js
```diff
@@ -78,11 +76,11 @@ var Contract = function Contract(jsonInterface, address, options) {
     this.options = {};
 
     var lastArg = args[args.length - 1];
-    if(_.isObject(lastArg) && !_.isArray(lastArg)) {
+    if(typeof lastArg == 'object' && !!lastArg && !Array.isArray(lastArg)) {
```
Spacesai1or:
```suggestion
    if(typeof lastArg === 'object' && !!lastArg && !Array.isArray(lastArg)) {
```

packages/web3-eth-contract/src/index.js
```diff
@@ -78,11 +76,11 @@ var Contract = function Contract(jsonInterface, address, options) {
...
         options = lastArg;
 
-        this.options = _.extend(this.options, this._getOrSetDefaultOptions(options));
-        if(_.isObject(address)) {
+        this.options = { ...this.options, ...this._getOrSetDefaultOptions(options)};
+        if(typeof address == 'object' && !!address) {
```
Spacesai1or:
```suggestion
        if(typeof address === 'object' && !!address) {
```

packages/web3-utils/src/utils.js
```diff
@@ -232,7 +231,7 @@ var hexToNumber = function (value) {
         return value;
     }
 
-    if (_.isString(value) && !isHexStrict(value)) {
+    if (typeof value == 'string' && !isHexStrict(value)) {
```
Spacesai1or:
```suggestion
    if (typeof value === 'string' && !isHexStrict(value)) {
```

packages/web3-utils/src/utils.js
```diff
@@ -249,7 +248,7 @@ var hexToNumber = function (value) {
 var hexToNumberString = function (value) {
     if (!value) return value;
 
-    if (_.isString(value) && !isHexStrict(value)) {
+    if (typeof value == 'string' && !isHexStrict(value)) {
```
Spacesai1or:
```suggestion
    if (typeof value === 'string' && !isHexStrict(value)) {
```

Spacesai1or (changes requested):
Once my suggestions are addressed, this looks g2g!

Thank you @luu-alex :)

packages/web3-bzz/src/index.js
```diff
@@ -52,7 +51,7 @@ if (typeof ethereum !== 'undefined' && ethereum.bzz) {
 
 Bzz.prototype.setProvider = function(provider) {
     // is ethereum provider
-    if(_.isObject(provider) && _.isString(provider.bzz)) {
+    if(typeof provider === 'object' && !!provider && typeof provider.bzz === 'string') {
```
jdevcs: It will be more efficient to first check that provider is valid (not undefined etc ) and then move on with other checks as its short circuit and. like:

```  if(!!provider && typeof provider === 'object' && typeof provider.bzz === 'string')  ```

jdevcs: There are couple of more _same_ improvements possible with !! and && in this PR.

luu-alex:
@spacesailor24  @jdevcs  Thank you for the reviews, I changed the files accordingly to the feedback

Spacesai1or (approved):

PR #4100, opened 2021-06-03
Fixed web3-eth-abi typing

marcj:
The actual [code exports](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-eth-abi/src/index.js#L442) as default a AbiCoder instance. In the previous version the class itself was exported via its name, which breaks runtime code.

Correctly used it looks like that:

```typescript
import abiCoder from 'web3-eth-abi';

const data = abiCoder.encodeParameters(...);
```

The new typings reflects that now correctly.

## Description

## Type of change

## Checklist:


GregTheGreek:
Thanks for the PR, I haven't tested this specifically yet, but from my understanding this doesn't quite solve it and we have a deeper issue that requires breaking the api.

AFAIIK Typescript doesn't have a way to export a instantiated classes. Specifically, abicoder looks like:
```
const coder = new AbiCoder();
module.exports = coder;
```

marcj:
For me this solved it. We can not use the class AbiCoder in user land as you indicated since it isn't exported in the javascript file.

> AFAIIK Typescript doesn't have a way to export a instantiated classes

My PR provides precisely that.

> but from my understanding this doesn't quite solve it and we have a deeper issue that requires breaking the api

I don't see a deeper issue here. The published TS api is broken and unusable, so fixing it is the only way. No need to change the JS.

GregTheGreek:
> For me this solved it. We can not use the class AbiCoder in user land as you indicated since it isn't exported in the javascript file.
> 
> > AFAIIK Typescript doesn't have a way to export a instantiated classes
> 
> My PR provides precisely that.
> 
> > but from my understanding this doesn't quite solve it and we have a deeper issue that requires breaking the api
> 
> I don't see a deeper issue here. The published TS api is broken and unusable, so fixing it is the only way. No need to change the JS.

Let me run this tomorrow, last time someone brought this up, it actually didn't solve the problem. But perhaps we were all mistaken!

marcj:
👋 Alright, let me know if you need anything else.

onchainguy-btc:
Would be awesome to see this merged, having the same issue 👐

chancehudson:
I'd like to see this merged as well.

```
import AbiCoder from 'web3-eth-abi'
const encodeFunctionSignature = (AbiCoder as any).encodeFunctionSignature.bind(AbiCoder)
const decodeParameters = (AbiCoder as any).decodeParameters.bind(AbiCoder)
```

Ideally the export would simply be the functions, not a class. These are stateless utilities, why am I having to bind them?

I'll definitely take this PR as a fix though, it will let me remove the `any` cast.

sambacha:
this is deff. not stale issue

sambacha:
> This PR has been automatically marked as stale beacause it has not had recent activity. It will be closed in 7 days if no further activity occurs. Thank you for your contributions. If you believe this was a mistake, please comment.

lgtm

marcj:
can you please merge that?

Lucid36:
Fixed

marcj:
Can you please merge that?

packages/web3-eth-abi/types/index.d.ts
```diff
@@ -40,3 +40,6 @@ export class AbiCoder {
         topics: string[]
     ): { [key: string]: string };
 }
+
+declare var coder: AbiCoder;
```
avkos, line 44:
```suggestion
declare const coder: AbiCoder;
```
I think you need to use `const` here

avkos (changes requested):

nazarhussain (approved):
Apart from feedback from @avkos LGTM

packages/web3-eth-abi/types/index.d.ts
marcj: Why need? If it doesn't match you coding style, you can change it afterwards, or close and commit yourself. For me its only important that this bug gets fixed, not how.

avkos: just fix it, please. and also check tests. They do not pass after your changes.

PR #4118, opened 2021-06-18
Release v1.4.0

Spacesai1or:
### Added

- Berlin Transaction Support (#4083)
- When signing a transaction, common object now defaults to berlin instead of petersburg

### Changed

- Changed Geth Docker verision from `stable` to `1.10.3` in `e2e.geth.instamine.sh` and `scripts/e2e.geth.automine.sh` (#4154)

PR #4149, opened 2021-06-24
1.4.0 Geth version downgrade

Spacesai1or:
Downgrades Geth docker version from `stable` to `1.10.3` because `stable` includes [EIP1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md) support

**NOTE** This change needs to be reverted when support for EIP1559 is added (tracked via #4150)


luu-alex (approved):

Spacesai1or:
Status checks seem to be broken, as they no workflow action has been triggered to run tests. I'm going to merge this with just an approval since the tests will get ran in #4118

PR #4155, opened 2021-06-25
EIP1559

### LINKED ISSUE #4105: Implement EIP-1559
[EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) is a transaction type that targets fee market changes

### LINKED ISSUE #4150: Revert Geth Docker version change after EIP1559 support is added
The pulled Docker version of Geth in `scripts/e2e.geth.instamine.sh` needs to be reverted from `1.10.3` to `stable` as mentioned in #4149

Spacesai1or:
Uses [ethers.js' implementation/defaults reasoning](https://github.com/ethers-io/ethers.js/blob/c5bca7767e3f3d43e3d0bd3c9e9420321ee9907a/packages/abstract-signer/src.ts/index.ts#L199) as inspiration

closes #4150

~**NOTE** Extra file changes are a result of branch off #4118, since it contains the needed EIP-1559 changes~


packages/web3-eth-accounts/src/index.js
```diff
@@ -287,6 +320,30 @@ function _validateTransactionForSigning(tx) {
...
+                    resolve({
+                        maxFeePerGas: tx.maxFeePerGas || utils.toHex(utils.toBN(block.baseFee).mul(2)),
+                        maxPriorityFeePerGas: tx.maxPriorityFeePerGas || '0x1'
+                    });
+                } else {
+                    resolve({ gasPrice });
+                }
+            })
+        } catch (error) {
+            reject(e)
```
kchia: Looks like a typo here. Should say: ```reject(error)```

ricmoo (commented):
Keep in mind that EIP-1559 transactions also have an accessList, so [this section](https://github.com/ChainSafe/web3.js/pull/4155/files#diff-cf46f2521b1de659430e09e654742a41c3696dbdddc3cd563ca4e2c02d18ffacR172) might need some re-ordering.

There are also lots of decisions to make about incompatible properties (e.g. `maxFeePerGas` and `gasPrice` specifies).

packages/web3-eth-accounts/src/index.js
```diff
@@ -162,6 +172,9 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
             if (transaction.accessList) {
                 // EIP-2930
                 transaction.type = "0x01"
+            } else if (transaction.maxPriorityFeePerGas || transaction.maxFeePerGas) {
```
GregTheGreek: I think this should be the first case, since in practice 1559 can be appended to any transaction type.

packages/web3-eth-accounts/src/index.js
```diff
@@ -241,21 +254,41 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
             throw new Error('One of the values "chainId", "networkId", "gasPrice", or "nonce" couldn\'t be fetched: ' + JSON.stringify(args));
         }
-        return signed(_.extend(tx, {chainId: args[0], gasPrice: args[1], nonce: args[2], networkId: args[3]}));
+        return signed(_.extend(
+            tx,
+            {
+                chainId: args[0],
+                nonce: args[2],
+                networkId: args[3],
+                ...args[1] // Will either be gasPrice or maxFeePerGas and maxPriorityFeePerGas
```
GregTheGreek: ocd killing me on this, can you push `_handleTxPricing` to the end of `Promise.all(..)` so it inherits the 3rd index? Or just re-order this

packages/web3-eth-accounts/src/index.js
GregTheGreek:
```suggestion
            reject(error)
```

GregTheGreek (changes requested):
First glance, need to take a more thorough look

packages/web3-eth-accounts/src/index.js
ricmoo: <strike>But what happens if they specify `accessList` and `maxFeePerGas`?</strike>

Sorry, misread. Yes, I agree. That’s what I meant by re-ordering. :)

packages/web3-eth-accounts/src/index.js
Spacesai1or: Thank y'all!

packages/web3-eth-accounts/src/index.js
```diff
@@ -241,21 +254,41 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
+        hasTxSigningOptions
+    ) {
         return Promise.resolve(signed(tx));
     }
 
     // Otherwise, get the missing info from the Ethereum Node
     return Promise.all([
         isNot(tx.chainId) ? _this._ethereumCall.getChainId() : tx.chainId,
-        isNot(tx.gasPrice) ? _this._ethereumCall.getGasPrice() : tx.gasPrice,
+        // isNot(tx.gasPrice) ? _this._ethereumCall.getGasPrice() : tx.gasPrice,
```
Spacesai1or: Remove commented code

Spacesai1or:
Thank you @kchia , @ricmoo , and @GregTheGreek for taking a look at this

I've
- added a `_handleTxType` method that includes property checking as @ricmoo mentioned (copies [checks from Ethers](https://github.com/ethers-io/ethers.js/blob/2a7ce0e72a1e0c9469e10392b0329e75e341cf18/packages/abstract-signer/src.ts/index.ts#L215))
- updated `maxFeePerGas` calculation as per [Ethers](https://github.com/ethers-io/ethers.js/blob/ba6854bdd5a912fe873d5da494cb5c62c190adde/packages/abstract-provider/src.ts/index.ts#L247)
- updated `_handleTxPricing` to use `gasPrice` for `maxPriorityFeePerGas` and `maxFeePerGas` if network supports EIP-1559, but user provided `gasPrice` as per [Ethers](https://github.com/ethers-io/ethers.js/blob/2a7ce0e72a1e0c9469e10392b0329e75e341cf18/packages/abstract-signer/src.ts/index.ts#L246)

Thank you @ricmoo for your R&D :)

packages/web3-eth-accounts/src/index.js
```diff
@@ -81,7 +81,17 @@ var Accounts = function Accounts() {
...
+        new Method({
+            name: 'getBlockByNumber',
+            call: 'eth_getBlockByNumber',
+            params: 2,
+            inputFormatter: [function(blockNumber) {
+                return blockNumber ? utils.toHex(blockNumber) : 'latest'
+            }, function() {
+                return false
+            }]
+        }),
```
GregTheGreek, line 93: is this a new method?

test/eth.accounts.signTransaction.js
```diff
@@ -500,6 +557,52 @@ var tests = [
...
+            accessList: accessList
+        },
+        // signature from eth_signTransaction
+        rawTransaction: "0x02f8ca0180843b9aca00843b9aca0e826a4094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca0080f85bf859940000000000000000000000000000000000000101f842a00000000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000060a780a0e3a2e10c7d3af3407ec2d38c64788d6673926e9b28d6d2e7df3c94cdf0548233a00ad3e5faafaf3a9350ab16c1be0198ce9ff3c6bef0b91e05488d757f07de9557",
+        oldSignature: "0x02f8ca0180843b9aca00843b9aca0e826a4094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca0080f85bf859940000000000000000000000000000000000000101f842a00000000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000060a780a0e3a2e10c7d3af3407ec2d38c64788d6673926e9b28d6d2e7df3c94cdf0548233a00ad3e5faafaf3a9350ab16c1be0198ce9ff3c6bef0b91e05488d757f07de9557",
+        transactionHash: "0xbc2c9edab3d4e3a795fa402b52d6149e874de15f0cc6c0858eb34e1fe1ef31fe",
+        messageHash: "0xa3a2cdc45e9cefb9a614ead90ce65f68bcf8a90dbe0ccbd84c1b62403bd05346"
+    },
```
GregTheGreek, line 651: Perhaps we can add a few tests:
- 1559 without `accessList`
- 1559 without the gas values

GregTheGreek (changes requested):

packages/web3-eth-accounts/src/index.js
Spacesai1or: Sorta, this package doesn't make use of the RPC wrappers found in `web3-eth`, instead needed methods are redefined in the package. I added this method to be able to get a block to find the `baseFeePerGas`

test/eth.accounts.signTransaction.js
Spacesai1or: A test without `gasPrice` or fee related properties exists [here](https://github.com/ChainSafe/web3.js/blob/4ebd78a771a046dd287ecf1cfdedc37986162a71/test/eth.accounts.signTransaction.js#L585)

I'll add a test that excludes `accessList`, and tests that verify the `maxFeePerGas` calculation needs to be added as well

Spacesai1or: Added test [without](https://github.com/ChainSafe/web3.js/pull/4155/commits/43fc7aa08ea128a000f6be1d425170b62c62dac1) AccessList

Spacesai1or:
I created a [manual test repo](https://github.com/spacesailor24/web3-eip1559-test) for this, since existing `web3-eth-accounts` tests don't allow for testing internal methods, don't give access to constructed transaction object, and I wanted to test with a live network

OhArthits (approved):

liamaharon:
Hi there, I get an error calling `web3.eth.getTranasction` with a type 2 transaction hash using a web3.js WS provider.

The error occurs here https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-core-helpers/src/formatters.js#L242, it looks like this function also needs updating for web3.js to be compatible with new transaction types introduced with London. I'm using v1.4.0.

Is support for type 2 transactions planned to be added in this PR?

Thanks!

Spacesai1or:
@liamaharon Hi there! Thank you bringing this to our attention. This PR is looking to add support for type 2 transactions (EIP1559), but it does look like we still need to update the output formatter code you provided to account for the additional fields

packages/web3-eth-accounts/src/index.js
AndreMiras: You mean we can't leverage the [getBlock](https://github.com/ChainSafe/web3.js/blob/v1.4.0/packages/web3-eth/src/index.js#L420-L426) (also refs [blockCall](https://github.com/ChainSafe/web3.js/blob/v1.4.0/packages/web3-eth/src/index.js#L44-L46)) because it's defined in the `web3-eth` package and not accessible from `web3-eth-accounts`?
In this case don't we have a dependency or at least a DRY issue that could be addressed somehow?

corymsmith:
> @liamaharon Hi there! Thank you bringing this to our attention. This PR is looking to add support for type 2 transactions (EIP1559), but it does look like we still need to update the output formatter code you provided to account for the additional fields

@spacesailor24 Is someone actively implementing those updates at the moment? If not I could look into that right now.

corymsmith:
@spacesailor24 I created this PR with this one as the base branch: https://github.com/ChainSafe/web3.js/pull/4167

Spacesai1or:
> @spacesailor24 I created this PR with this one as the base branch: https://github.com/ChainSafe/web3.js/pull/4167

Thank you @corymsmith! I'll give this a more thorough ￼review, as well as test it against my test repo today, but at first glance it looks good!

packages/web3-eth-accounts/src/index.js
```diff
@@ -287,6 +313,93 @@ function _validateTransactionForSigning(tx) {
...
+                        let maxPriorityFeePerGas, maxFeePerGas;
+    
+                        if (tx.gasPrice) {
+                            // Using legacy gasPrice property on an eip-1559 network,
+                            // so use gasPrice as both fee properties
+                            maxPriorityFeePerGas = tx.gasPrice;
+                            maxFeePerGas = tx.gasPrice;
+                            delete tx.gasPrice;
+                        } else {
+                            maxPriorityFeePerGas = tx.maxPriorityFeePerGas || '0x3B9ACA00'; // 1 Gwei
```
jdevcs, line 392: Its not part of the EIP ( specs )  to must have default 'priority fee per gas' 1 Gwei. Its only suggested to set this default.

so I suggest It should always be coming from web3 user only, or if we are setting default then must notify / inform user as well.

packages/web3-eth-accounts/src/index.js
```diff
@@ -287,6 +313,93 @@ function _validateTransactionForSigning(tx) {
...
+                // Legacy transaction, return provided gasPrice
+                resolve({ gasPrice: tx.gasPrice })
+            } else {
+                Promise.all([
+                    _this._ethereumCall.getBlockByNumber(),
+                    _this._ethereumCall.getGasPrice()
+                ]).then(responses => {
+                    const [block, gasPrice] = responses;
+                    if (
+                        (tx.type !== '0x0' &&
```
jdevcs: Its already checked https://github.com/ChainSafe/web3.js/pull/4155/files#diff-cf46f2521b1de659430e09e654742a41c3696dbdddc3cd563ca4e2c02d18ffacR352 for legacy Tx types. May be we can remove this double check or is there any use of additional check?

packages/web3-eth-accounts/src/index.js
GregTheGreek: This was something that was suggested on a call with wallet/library implementers, but we should make a note in the docs

Spacesai1or: We'll add `maxPriorityFeePerGas` being defaulted to `1 Gwei` to the docs, but I think the understanding was that a tx without a priority fee wouldn't get processed/be excluded from the mem pool

packages/web3-eth-accounts/src/index.js
Spacesai1or: @AndreMiras Yes, and we plan to address the problematic architecture in `4.x` (our rewrite), so for now it's just about making the least amount of change to `1.x` as possible, to refrain from introducing additional bugs

packages/web3-eth-accounts/src/index.js
Spacesai1or: So while this is double checking `tx.type`, it's for different reasons:

[Here](https://github.com/ChainSafe/web3.js/blob/572e416e197156032b33a767e1457933afc2a68e/packages/web3-eth-accounts/src/index.js#L357) we are short circuiting and returning `gasPrice` if it's apart of the `tx` object and `tx.type` is `0x0` or `0x1` since these txs will not have the EIP1559 fields and will be using the user provided `gasPrice`

And [here](https://github.com/ChainSafe/web3.js/blob/572e416e197156032b33a767e1457933afc2a68e/packages/web3-eth-accounts/src/index.js#L365) we are verifying we aren't of type `0x0` or `0x1` before we add EIP1559 fields

I admit that this code is a bit convoluted, but it's essentially just a condensed version of Ether's EIP1559 implementation

Spacesai1or:
What remains to be done before this is ready for final review:

- [x] #4150 
- [x] Implement [these](https://github.com/spacesailor24/web3-eip1559-test/blob/main/index.js) tests within the repo using `geth-dev-assistant`

packages/web3-eth-accounts/src/index.js
```diff
@@ -287,6 +313,93 @@ function _validateTransactionForSigning(tx) {
...
+                            // so use gasPrice as both fee properties
+                            maxPriorityFeePerGas = tx.gasPrice;
+                            maxFeePerGas = tx.gasPrice;
+                            delete tx.gasPrice;
+                        } else {
+                            maxPriorityFeePerGas = tx.maxPriorityFeePerGas || '0x3B9ACA00'; // 1 Gwei
+                            maxFeePerGas = tx.maxFeePerGas ||
+                                utils.toHex(
+                                    utils.toBN(block.baseFeePerGas)
+                                        .mul(utils.toBN(2))
```
jdevcs, line 396: Also will be imp to highlight that web3 is doing it and why "baseFeePerGas **times 2**" to web3 users in doc.

OhArthits (dismissed):

test/eth.accounts.signTransaction.js
```diff
@@ -24,6 +24,63 @@ var commonBerlin = {
     hardfork: 'berlin'
 };
 
+var commonLondon = {
```
kchia, line 27: I'd recommend using `let` instead of `var` to declare variables. `let` is a more modern syntax than `var` and it aims to resolve some weird hoisting behaviors that `var` has that could lead to bugs.

Spacesai1or: Hey @kchia, thank you for the input! Unfortunately most of Web3.js was written a while back, before modern JavaScript techniques were widely available/supported. We're currently working on a `4.x` version this uses TypeScript and modern JS features all the way up to ES2020. So for `1.x` and `3.x` the concern is to limit the amount of changes made to reduce introducing any new bugs

corymsmith:
Any updates on this @spacesailor24, need any help?

Spacesai1or:
> Any updates on this @spacesailor24, need any help?

Hey there! Thanks for the offer, looks like we just got the [failing tests to pass](https://github.com/ChainSafe/web3.js/pull/4171). The team is going to have a sync call around this tomorrow as a sanity check, then this PR will most likely get merged and released as an RC tomorrow

corymsmith:
> > Any updates on this @spacesailor24, need any help?
> 
> Hey there! Thanks for the offer, looks like we just got the [failing tests to pass](https://github.com/ChainSafe/web3.js/pull/4171). The team is going to have a sync call around this tomorrow as a sanity check, then this PR will most likely get merged and released as an RC tomorrow

Awesome, once its released I can test it out in our staging environment, thanks for the update!

packages/web3-eth-accounts/src/index.js
```diff
@@ -272,20 +297,117 @@ function _validateTransactionForSigning(tx) {
...
         return new Error('"gas" is missing');
     }
 
-    if (tx.nonce < 0 ||
-        tx.gas < 0 ||
-        tx.gasPrice < 0 ||
-        tx.chainId < 0) {
+    if (tx.gas && tx.gasPrice) {
+        if (tx.gas < 0 || tx.gasPrice < 0) {
+            return new Error('Gas, gasPrice, nonce or chainId is lower than 0');
```
GregTheGreek:
```suggestion
            return new Error('Gas or gasPrice is lower than 0');
```

packages/web3-eth-accounts/src/index.js
```diff
@@ -272,20 +297,117 @@ function _validateTransactionForSigning(tx) {
...
+        }
+    } else {
+        if (tx.maxPriorityFeePerGas < 0 || tx.maxFeePerGas < 0) {
+            return new Error('maxPriorityFeePerGas or maxFeePerGas is lower than 0');
+        }
+    }
+
+    if (tx.nonce < 0 || tx.chainId < 0) {
         return new Error('Gas, gasPrice, nonce or chainId is lower than 0');
```
GregTheGreek:
```suggestion
        return new Error('Nonce or chainId is lower than 0');
```

GregTheGreek (changes requested):
Partial review.

packages/web3-eth-accounts/src/index.js
Spacesai1or: I wanted to change these, but technically it's a breaking change since we'll be returning a different error message to the user

GregTheGreek: It's not breaking the api or changing the behaviour, its still returning an error, so it would be caught with a `.catch()`. In general its not a great practice to be filtering errors based on exact matches...

I think in this case we would consider this to be a bug, since its now a misleading error message. Just my two cents.

packages/web3-eth-accounts/src/index.js
```diff
@@ -272,20 +297,117 @@ function _validateTransactionForSigning(tx) {
...
     return;
 }
 
+function _handleTxType(tx) {
+    let txType = tx.type !== undefined ? utils.toHex(tx.type) : '0x0';
+    // Taken from https://github.com/ethers-io/ethers.js/blob/2a7ce0e72a1e0c9469e10392b0329e75e341cf18/packages/abstract-signer/src.ts/index.ts#L215
+    const hasEip1559 = (tx.maxFeePerGas !== undefined || tx.maxPriorityFeePerGas !== undefined);
+    if (tx.gasPrice !== undefined && (tx.type === '0x2' || hasEip1559))
+        throw Error("eip-1559 transactions don't support gasPrice");
+    if ((tx.type === '0x1' || tx.type === '0x0') && hasEip1559)
```
GregTheGreek: here you use `tx.type` but everywhere else you use `txType` from line 325

packages/web3-eth-accounts/src/index.js
```diff
@@ -272,20 +297,117 @@ function _validateTransactionForSigning(tx) {
...
+        (
+            (tx.common && tx.common.hardfork && tx.common.hardfork.toLowerCase() === 'london') ||
+            (tx.hardfork && tx.hardfork.toLowerCase() === 'london')
+        )
+    ) {
+        txType = '0x2';
+    } else if (
+        tx.accessList ||
+        (
+            (tx.common && tx.common.hardfork && tx.common.hardfork.toLowerCase() === 'berlin') ||
```
GregTheGreek: The following is possible:
```
const {Hardfork} = require("@ethereumjs/common");

(tx.common && tx.common.hardfork && tx.common.hardfork === Hardfork.Berlin)
```

packages/web3-eth-accounts/src/index.js
```diff
@@ -272,20 +297,117 @@ function _validateTransactionForSigning(tx) {
...
+}
+
+function _handleTxPricing(_this, tx) {
+    return new Promise((resolve, reject) => {
+        try {
+            if (
+                (
+                    tx.type === undefined ||
+                    tx.type === '0x0' ||
+                    tx.type === '0x1'
```
GregTheGreek: this can be simplified to be 
```
tx.type < "0x2"
```

packages/web3-eth-accounts/src/index.js
```diff
@@ -272,20 +297,117 @@ function _validateTransactionForSigning(tx) {
...
+                ) && tx.gasPrice !== undefined)
```
GregTheGreek: This is checking that `tx.gasPrice` is equal to something, if its already set why are we setting it again?

packages/web3-eth-accounts/src/index.js
```diff
@@ -272,20 +297,117 @@ function _validateTransactionForSigning(tx) {
...
+                resolve({ gasPrice: tx.gasPrice })
+            } else {
+                Promise.all([
+                    _this._ethereumCall.getBlockByNumber(),
+                    _this._ethereumCall.getGasPrice()
+                ]).then(responses => {
+                    const [block, gasPrice] = responses;
+                    if (
+                        (tx.type !== '0x0' && 
+                        tx.type !== '0x1') &&
```
GregTheGreek: Simplified version?
```
if (tx.type == '0x2' || (block && block.baseFeePerGas)) { ... }
```

GregTheGreek (changes requested):

CHANGELOG.md
```diff
@@ -392,7 +392,7 @@ Released with 1.0.0-beta.37 code base.
 
 ### Changed
 
-- Changed Geth Docker verision from `stable` to `1.10.3` in `e2e.geth.instamine.sh`
+- Changed Geth Docker verision from `stable` to `1.10.3` in `e2e.geth.instamine.sh` and `scripts/e2e.geth.instamine.sh` (#4154)
```
Goldselleruk: Gdzie gold

Goldselleruk (commented):
Gdzie gold

Goldselleruk (dismissed):
Gdzie gold

packages/web3-eth-accounts/src/index.js
Spacesai1or: You're right that `tx.gasPrice` is already set, so returning the value here is redundant, but returning nothing will cause [this](https://github.com/ChainSafe/web3.js/pull/4155/checks?check_run_id=3119556049#step:6:420) error as it fails the check [here](https://github.com/ChainSafe/web3.js/blob/b3ab709f48fea959938c9d24845b5a7dd7bb1f1c/packages/web3-eth-accounts/src/index.js#L272). We could refactor the code to not call `_handleTxPricing` if `tx.gasPrice` is set and `tx.type < '0x2'`, but this doesn't really buy us anything, and moves some tx pricing logic out of `_handleTxPricing`. So I think it makes the most sense to keep how it is, and short circuit the `_handleTxPricing`

packages/web3-eth-accounts/src/index.js
Spacesai1or: Updated

packages/web3-eth-accounts/src/index.js
Spacesai1or: Updated

packages/web3-eth-accounts/src/index.js
Spacesai1or: Updated

packages/web3-eth-accounts/src/index.js
Spacesai1or: Updated

packages/web3-eth-accounts/src/index.js
```diff
@@ -272,20 +297,120 @@ function _validateTransactionForSigning(tx) {
...
+        txType = '0x1';
+    }
+    
+    return txType
+}
+
+function _handleTxPricing(_this, tx) {
+    return new Promise((resolve, reject) => {
+        try {
+            if (tx.type < '0x2' && tx.gasPrice !== undefined) {
```
luu-alex, line 367: when `tx.type = undefined` and is checked under `tx.type < "0x2"` it'll result into false
perhaps:
```
if (( tx.type === 'undefined' || tx.type) < '0x2' && ...
 ```

Spacesai1or: It's intended to be false if `tx.type` is `undefined`. We only want to resolve the provided `tx.gasPrice` if `tx.type` is `0x0` or `0x1`. If `undefined`, we don't know if the network supports `EIP-1559`, and whether we should use `tx.gasPrice` or the 1559 properties

luu-alex (approved):

GregTheGreek (approved):

Spacesai1or:
@corymsmith [v1.5.0-rc.0](https://github.com/ChainSafe/web3.js/releases/tag/v1.5.0-rc.0) has been released

ricmoo:
FYI, if you need more test cases, here are the ethers generated (with EthereumJS) testcases: https://github.com/ethers-io/ethers.js/blob/master/packages/testcases/testcases/typed-transactions.json.gz

PR #4167, opened 2021-07-06
Handling EIP1559 transactions in outputTransactionFormatter

corymsmith:
## Description

Adds support for EIP1559 fields in the `outputTransactionFormatter`, this includes a null `gasPrice` which causes the current web3 version to error out calling `getTransaction`

Relates to https://github.com/ChainSafe/web3.js/issues/4105

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-core-helpers/src/formatters.js
+++ b/packages/web3-core-helpers/src/formatters.js
@@ -239,7 +239,7 @@ var outputTransactionFormatter = function (tx) {
         tx.transactionIndex = utils.hexToNumber(tx.transactionIndex);
     tx.nonce = utils.hexToNumber(tx.nonce);
     tx.gas = utils.hexToNumber(tx.gas);
-    if (tx.gasPrice !== null)
+    if (tx.gasPrice)
         tx.gasPrice = outputBigNumberFormatter(tx.gasPrice);
     if (tx.maxFeePerGas)
         tx.maxFeePerGas = outputBigNumberFormatter(tx.maxFeePerGas);
--- a/packages/web3-core-helpers/src/formatters.js
+++ b/packages/web3-core-helpers/src/formatters.js
@@ -239,7 +239,14 @@ var outputTransactionFormatter = function (tx) {
         tx.transactionIndex = utils.hexToNumber(tx.transactionIndex);
     tx.nonce = utils.hexToNumber(tx.nonce);
     tx.gas = utils.hexToNumber(tx.gas);
-    tx.gasPrice = outputBigNumberFormatter(tx.gasPrice);
+    if (tx.gasPrice !== null)
+        tx.gasPrice = outputBigNumberFormatter(tx.gasPrice);
+    if (tx.maxFeePerGas)
+        tx.maxFeePerGas = outputBigNumberFormatter(tx.maxFeePerGas);
+    if (tx.maxPriorityFeePerGas)
+        tx.maxPriorityFeePerGas = outputBigNumberFormatter(tx.maxPriorityFeePerGas);
+    if (tx.type)
+        tx.type = utils.hexToNumber(tx.type);
     tx.value = outputBigNumberFormatter(tx.value);
 
     if (tx.to && utils.isAddress(tx.to)) { // tx.to could be `0x0` or `null` while contract creation
--- a/test/formatters.outputTransactionFormatter.js
+++ b/test/formatters.outputTransactionFormatter.js
@@ -38,7 +38,7 @@ describe('formatters', function () {
                 to: null,
                 value: '0x3e8',
                 gas: '0x3e8',
-                gasPrice: '0x3e8',
+                gasPrice: null,
                 nonce: '0xb',
                 transactionIndex: null,
                 blockNumber: null,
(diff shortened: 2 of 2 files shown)


corymsmith:
Let me know if I should update the base branch to `1.x` since thats what I originally branched off of and the tests all pass.

packages/web3-core-helpers/src/formatters.js
```diff
@@ -239,7 +239,14 @@ var outputTransactionFormatter = function (tx) {
...
     tx.nonce = utils.hexToNumber(tx.nonce);
     tx.gas = utils.hexToNumber(tx.gas);
-    tx.gasPrice = outputBigNumberFormatter(tx.gasPrice);
+    if (tx.gasPrice !== null)
+        tx.gasPrice = outputBigNumberFormatter(tx.gasPrice);
+    if (tx.maxFeePerGas)
+        tx.maxFeePerGas = outputBigNumberFormatter(tx.maxFeePerGas);
+    if (tx.maxPriorityFeePerGas)
+        tx.maxPriorityFeePerGas = outputBigNumberFormatter(tx.maxPriorityFeePerGas);
+    if (tx.type)
```
corymsmith, line 248: @spacesailor24 Should we be inferring the `type` based on the presence of `maxFeePerGas` for legacy transactions? My thinking is if connected to a node that is not running the fork (pre-fork) then type will just not be there for legacy transactions.

OhArthits (approved):

packages/web3-core-helpers/src/formatters.js
```diff
@@ -239,7 +239,14 @@ var outputTransactionFormatter = function (tx) {
         tx.transactionIndex = utils.hexToNumber(tx.transactionIndex);
     tx.nonce = utils.hexToNumber(tx.nonce);
     tx.gas = utils.hexToNumber(tx.gas);
-    tx.gasPrice = outputBigNumberFormatter(tx.gasPrice);
+    if (tx.gasPrice !== null)
```
GregTheGreek:
```suggestion
    if (tx.gasPrice !== null || tx.gasPrice !== undefined)
```

packages/web3-core-helpers/src/formatters.js
Spacesai1or: @corymsmith We can't infer it's existence purely off of the presence of `maxFeePerGas` because `type` will exist for [EIP2930](https://eips.ethereum.org/EIPS/eip-2930) as well as future `types`

packages/web3-core-helpers/src/formatters.js
GregTheGreek: do we even need this check?

packages/web3-core-helpers/src/formatters.js
```diff
@@ -239,7 +239,14 @@ var outputTransactionFormatter = function (tx) {
...
+        tx.gasPrice = outputBigNumberFormatter(tx.gasPrice);
+    if (tx.maxFeePerGas)
+        tx.maxFeePerGas = outputBigNumberFormatter(tx.maxFeePerGas);
+    if (tx.maxPriorityFeePerGas)
```
GregTheGreek, line 246: Could this be equal to `0`? If so this check will fail

packages/web3-core-helpers/src/formatters.js
GregTheGreek: Nvm I think the original is correct the node will return bull on an empty value.

packages/web3-core-helpers/src/formatters.js
corymsmith: To clarify I didn't mean just `maxFeePerGas` but with some logic similar to what you have in `_handleTxType` in your PR

packages/web3-core-helpers/src/formatters.js
corymsmith: Correct, that was the main reason I put together this PR, its coming back `null` in certain instances and throwing errors.

packages/web3-core-helpers/src/formatters.js
corymsmith: I assumed if it was `0` it would be returned as `"0x0"` so that check would still be accurate.

packages/web3-core-helpers/src/formatters.js
Spacesai1or:
```suggestion
    if (tx.gasPrice)
```

After discussing with @GregTheGreek, we think that this `!== null` check is unnecessary, and should match the other if statements

corymsmith: Agreed, I think my initial reasoning was if it was `0` then that check would fail and it wouldn't be formatted but it would be the string `"0x0"` in that case so your suggestion makes sense, will update.

packages/web3-core-helpers/src/formatters.js
Spacesai1or: I see, I don't find the extra checks necessary though, as `if (tx.type)` will cover if it's defined or `undefined`

corymsmith: Yeah I was thinking `type` wasn't being returned at all from pre-fork nodes so was curious if we needed to add it, I was incorrect, its already there :)

Spacesai1or (approved):

PR #4190, opened 2021-07-22
eth-feehistory docs

luu-alex:
## Description

Please include a summary of the changes and be sure to follow our [Contribution Guidelines](../CONTRIBUTIONS.md).

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I have selected the correct base branch.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules.
- [ ] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [ ] I ran `npm run test:unit` with success.
- [ ] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code.
- [ ] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
- [ ] I have tested my code on the live network.
- [ ] I have checked the Deploy Preview and it looks correct.
- [ ] I have updated the `CHANGELOG.md` file in the root folder.

PR #4191, opened 2021-07-23
eth_feeHistory (EIP 1559)

Spacesai1or:
Add missing `getFeeHistory` call that's made available via EIP 1559

**NOTE**

- ~[RPC docs](https://playground.open-rpc.org/) say that `blockCount` should be a hex string, but in my testing using Infura URLs on Ropsten and Rinkeby, only integers seem to be supported. Additionally, for `newestBlock`, block tags (e.g. `earliest`, `latest`, etc) don't seem to be supported, only hex strings~ To combat this, I added input formatters to `blockCount` and `lastBlock`
- Should be merged alongside #4190

### DIFF
--- a/packages/web3-eth/src/index.js
+++ b/packages/web3-eth/src/index.js
@@ -389,6 +389,7 @@ var Eth = function Eth() {
             name: 'getFeeHistory',
             call: 'eth_feeHistory',
             params: 3,
+            inputFormatter: [utils.toNumber, utils.toHex, function(value) {return value}]
         }),
         new Method({
             name: 'getAccounts',
--- a/packages/web3-eth/src/index.js
+++ b/packages/web3-eth/src/index.js
@@ -386,6 +386,11 @@ var Eth = function Eth() {
             outputFormatter: formatter.outputBigNumberFormatter
         }),
         new Method({
+            name: 'getFeeHistory',
+            call: 'eth_feeHistory',
+            params: 3,
+        }),
+        new Method({
             name: 'getAccounts',
             call: 'eth_accounts',
             params: 0,
--- a/packages/web3-eth/types/index.d.ts
+++ b/packages/web3-eth/types/index.d.ts
@@ -146,8 +146,15 @@ export class Eth {
     ): Promise<number>;
 
     getGasPrice(
+        blockCount: number,
+        newestBlock: string,
+        rewardPercentiles: number[],
         callback?: (error: Error, gasPrice: string) => void
     ): Promise<string>;
+
+    getFeeHistory(
+        callback?: (error: Error, feeHistory: FeeHistoryResult) => void
+    ): Promise<FeeHistoryResult>;
 
     getAccounts(
         callback?: (error: Error, accounts: string[]) => void
@@ -438,4 +445,11 @@ export interface StorageProof {
     key: string;
     value: string;
     proof: string[];
+}
+
+export interface FeeHistoryResult {
+    oldestBlock: string;
+    baseFeePerGas: string[];
+    gasUsedRatio: number[];
+    reward: string[][];
 }
--- a/packages/web3-eth/types/index.d.ts
+++ b/packages/web3-eth/types/index.d.ts
@@ -146,13 +146,13 @@ export class Eth {
     ): Promise<number>;
 
     getGasPrice(
-        blockCount: number,
-        newestBlock: string,
-        rewardPercentiles: number[],
         callback?: (error: Error, gasPrice: string) => void
     ): Promise<string>;
 
     getFeeHistory(
+        blockCount: number,
+        newestBlock: string,
+        rewardPercentiles: number[],
         callback?: (error: Error, feeHistory: FeeHistoryResult) => void
     ): Promise<FeeHistoryResult>;
 
--- a/packages/web3-eth/types/index.d.ts
+++ b/packages/web3-eth/types/index.d.ts
@@ -150,8 +150,8 @@ export class Eth {
     ): Promise<string>;
 
     getFeeHistory(
-        blockCount: number,
-        newestBlock: string,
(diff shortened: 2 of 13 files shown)


packages/web3-eth/src/index.js
```diff
@@ -385,6 +385,11 @@ var Eth = function Eth() {
             params: 0,
             outputFormatter: formatter.outputBigNumberFormatter
         }),
+        new Method({
+            name: 'getFeeHistory',
+            call: 'eth_feeHistory',
+            params: 3,
```
GregTheGreek, line 391:
```suggestion
            params: 4,
```

see comment below

packages/web3-eth/types/index.d.ts
```diff
@@ -146,9 +146,16 @@ export class Eth {
...
 
     getGasPrice(
+        blockCount: number,
+        newestBlock: string,
+        rewardPercentiles: number[],
         callback?: (error: Error, gasPrice: string) => void
     ): Promise<string>;
 
+    getFeeHistory(
+        callback?: (error: Error, feeHistory: FeeHistoryResult) => void
```
GregTheGreek, line 156: This should contain `blockCount`, `newestBlock` and `rewardPercentiles`?

GregTheGreek: I think you mixed up `getGasPrice` and `getFeeHistory`?

GregTheGreek (changes requested):

packages/web3-eth/src/index.js
corymsmith: Pretty sure 3 is correct, its just `blockCount`, `newestBlock` and `rewardPercentiles`, from looking at the other methods the callback function doesn't count as a param

packages/web3-eth/types/index.d.ts
corymsmith: Also `rewardPercentiles` should be  marked as optional

corymsmith:
> Add missing `getFeeHistory` call that's made available via EIP 1559
> 
> **NOTE**
> 
> * [RPC docs](https://playground.open-rpc.org/) say that `blockCount` should be a hex string, but in my testing using Infura URLs on Ropsten and Rinkeby, only integers seem to be supported. Additionally, for `newestBlock`, block tags (e.g. `earliest`, `latest`, etc) don't seem to be supported, only hex strings
> * Should be merged alongside [eth-feehistory docs #4190](https://github.com/ChainSafe/web3.js/pull/4190)

From looking here that looks correct, https://github.com/ethereum/go-ethereum/blob/9624f92edef5e0a76a97efd302e983077acb6e35/eth/gasprice/feehistory.go#L200

`blocks` which you named as `blockCount` is an int and `lastBlock` which you named as `newestBlock` is a UInt64 but is formatted that way using the `inputBlockNumberFormatter`

packages/web3-eth/types/index.d.ts
corymsmith: Actually I think the docs here are misleading https://github.com/ChainSafe/web3.js/pull/4190/files#diff-b1b5b8a010cd61ad800e528887f49730db04064d59d3ee5ed14de0f49f2ba11cR740

Looks like you still have to pass an empty array so I'm not sure if web3 should just do that as the default if that param isn't set? Would be more user friendly than having to pass an empty array.

packages/web3-eth/types/index.d.ts
```diff
@@ -439,3 +446,10 @@ export interface StorageProof {
     value: string;
     proof: string[];
 }
+
+export interface FeeHistoryResult {
+    oldestBlock: string;
```
corymsmith: This specifies `oldestBlock` is returned as a string but the tests are showing its returned as a number here: 
https://github.com/ChainSafe/web3.js/pull/4191/files#diff-99b1655ae7b89f474d7f663a07e45ccaa75c9270bb15bf852f1f0bdf2164f711R23

packages/web3-eth/types/index.d.ts
Spacesai1or: @GregTheGreek Whoops, forgot to push, thank you

@corymsmith I think web3 defaulting an empty array would be nice for the user, but we can't have two optional arguments in typescript, so I'm not sure how we could do this

Spacesai1or:
> > Add missing `getFeeHistory` call that's made available via EIP 1559
> > **NOTE**
> > 
> > * [RPC docs](https://playground.open-rpc.org/) say that `blockCount` should be a hex string, but in my testing using Infura URLs on Ropsten and Rinkeby, only integers seem to be supported. Additionally, for `newestBlock`, block tags (e.g. `earliest`, `latest`, etc) don't seem to be supported, only hex strings
> > * Should be merged alongside [eth-feehistory docs #4190](https://github.com/ChainSafe/web3.js/pull/4190)
> 
> From looking here that looks correct, https://github.com/ethereum/go-ethereum/blob/9624f92edef5e0a76a97efd302e983077acb6e35/eth/gasprice/feehistory.go#L200
> 
> `blocks` which you named as `blockCount` is an int and `lastBlock` which you named as `newestBlock` is a UInt64 but is formatted that way using the `inputBlockNumberFormatter`

So it appears that the docs are wrong :frowning_face:, I'll try to add some formatters to `blockCount` and `newestBlock` (which I want to rename to `lastBlock`) to make it easier for the user

packages/web3-eth/types/index.d.ts
Spacesai1or: You're right, I just blindly followed the RPC docs, thank you

packages/web3-eth/types/index.d.ts
corymsmith: True, I suppose the only way to do it would be to add a formatter that formats `null` into `[]`

GregTheGreek (approved):

PR #4213, opened 2021-07-30
Accept predefined block numbers for eth_feeHistory

brg8:
## Description

Allow the `eth_feeHistory` method to accept predefined block numbers, e.g. "latest". This is the appropriate schema type as defined in the ethereum spec: https://github.com/ethereum/eth1.0-specs/blob/master/json-rpc/spec.json.

The current behavior of forcing the parameter into `utils.toHex` actually does convert "latest" into a hex far beyond head resulting in the node response of "request beyond head block" (for geth).

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-eth/src/index.js
+++ b/packages/web3-eth/src/index.js
@@ -389,7 +389,7 @@ var Eth = function Eth() {
             name: 'getFeeHistory',
             call: 'eth_feeHistory',
             params: 3,
-            inputFormatter: [utils.toNumber, utils.toHex, function(value) {return value}]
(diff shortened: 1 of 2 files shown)


Spacesai1or:
~So the reason why this wasn't implemented with #4191 is because while the support for block tags is mentioned in the RPC spec, it doesn't seem to be supported via Geth. If you look [here](https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L90) it seems that only hex values are supported, and my testing confirms this~

~I've open [an issue](https://github.com/ethereum/go-ethereum/issues/23327) in Geth repo to track this~

I was mistaken on this, seems to be working for Geth

PR #4220, opened 2021-08-03
EIP-1559 support for web3.eth.sendTransaction

### LINKED ISSUE #4211: EIP-1559/London/0x2 transactions do not work due to default inclusion of `gasPrice`
#### Expected behavior

Including either of the `maxFeePerGas` or `maxPriorityFeePerGas` fields when sending a transaction should result in a transaction of type 0x2 with no `gasPrice` field.

#### Actual behavior

A `gasPrice` field is included by default, resulting in a nonsense transaction that's trying to be both EIP-1559 and pre-EIP-1559 at the same time, which Geth will reject.

Start Geth in development mode with HTTP enabled, then make a transaction on it using `web3.eth.sendTransaction`, specifying either of the `maxFeePerGas` or `maxPriorityFeePerGas` fields (make sure they're hex strings per #4210) (and making sure you use the unlocked account for your `from` address).

#### Logs

Example with Truffle Console:

```
truffle(development)> web3.eth.sendTransaction({ to: it.address, from: accounts[0], data: '0xc0406226', maxFeePerGas: "0x1", maxPriorityFeePerGas: "0x1" })
Uncaught:
Error: Returned error: both gasPrice and (maxFeePerGas or maxPriorityFeePerGas) specified
    at XMLHttpRequest._onHttpResponseEnd (/home/sniffnoy/truffle/truffle/node_modules/xhr2-cookies/xml-http-request.ts:345:8)
    at XMLHttpRequest._setReadyState (/home/sniffnoy/truffle/truffle/node_modules/xhr2-cookies/xml-http-request.ts:219:8)
    at XMLHttpRequestEventTarget.dispatchEvent (/home/sniffnoy/truffle/truffle/node_modules/xhr2-cookies/xml-http-request-event-target.ts:44:13)
    at XMLHttpRequest.request.onreadystatechange (/home/sniffnoy/truffle/truffle/node_modules/web3/node_modules/web3-providers-http/lib/index.js:98:13)
    at /home/sniffnoy/truffle/truffle/packages/provider/wrapper.js:107:9
    at /home/sniffnoy/truffle/truffle/node_modules/web3/node_modules/web3-core-requestmanager/lib/index.js:302:36
    at Object.ErrorResponse (/home/sniffnoy/truffle/truffle/node_modules/web3-core-helpers/lib/errors.js:28:19) {
  data: null
}
```

#### Environment

NPM: 6.14.11
Node: 12.21.0
web3.js: 1.5.0
Ethereum client: Geth 1.10.6-stable

Spacesai1or:
Adds EIP-1559 support to `web3.eth.sendTransaction`

closes #4211

### DIFF
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -707,7 +707,6 @@ Method.prototype.buildCall = function () {
                 // ETH_SENDTRANSACTION
                 if (payload.method === 'eth_sendTransaction') {
                     var tx = payload.params[0];
-                    console.log("payload", payload, "tx", tx)
                     wallet = getWallet((!!tx && typeof tx === 'object') ? tx.from : null, method.accounts);
 
 
@@ -749,7 +748,6 @@ Method.prototype.buildCall = function () {
                     // ETH_SIGN
                 } else if (payload.method === 'eth_sign') {
                     var data = payload.params[1];
-                    console.log("parload", payload,  "data", data) 
                     wallet = getWallet(payload.params[0], method.accounts);
 
                     // If wallet was found, sign tx, and send using sendRawTransaction
@@ -784,11 +782,15 @@ Method.prototype.buildCall = function () {
                 )
             )
         ) {
+            console.log('debug1', payload)
             if (typeof payload.params[0].type === 'undefined') 
                 payload.params[0].type = _handleTxType(payload.params[0]);
+            console.log('debug2', payload)
 
             _handleTxPricing(method, payload.params[0]).then(txPricing => {
                 payload.params[0] = {...payload.params[0], ...txPricing};
+
+                console.log('debug3', payload)
 
                 if (isSendTx) {
                     setTimeout(() => {
@@ -797,6 +799,8 @@ Method.prototype.buildCall = function () {
                 }
 
                 sendRequest(payload, method);
+
+                console.log('debug4', payload)
             })
         } else {
             if (isSendTx) {
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -786,7 +786,6 @@ Method.prototype.buildCall = function () {
                 payload.params[0].type = _handleTxType(payload.params[0]);
 
             _handleTxPricing(method, payload.params[0]).then(txPricing => {
-                console.log('txPricing', txPricing)
                 payload.params[0] = {...payload.params[0], ...txPricing};
 
                 if (isSendTx) {
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -786,13 +786,12 @@ Method.prototype.buildCall = function () {
             if (typeof payload.params[0].type === 'undefined') 
                 payload.params[0].type = _handleTxType(payload.params[0]);
             console.log('debug2', payload)
-            if (payload.params[0].type === '0x0') delete payload.params[0].type;
-            console.log('debug3', payload)
 
             _handleTxPricing(method, payload.params[0]).then(txPricing => {
                 payload.params[0] = {...payload.params[0], ...txPricing};
+                if (payload.params[0].type === '0x0') delete payload.params[0].type;
 
-                console.log('debug4', payload)
+                console.log('debug3', payload)
 
                 if (isSendTx) {
                     setTimeout(() => {
@@ -802,7 +801,7 @@ Method.prototype.buildCall = function () {
 
                 sendRequest(payload, method);
 
-                console.log('debug5', payload)
+                console.log('debug4', payload)
             })
         } else {
             if (isSendTx) {
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -771,63 +771,26 @@ Method.prototype.buildCall = function () {
         };
 
         // Send the actual transaction
-        // if (isSendTx
-        //     && !!payload.params[0]
-        //     && typeof payload.params[0] === 'object'
-        //     && typeof payload.params[0].gasPrice === 'undefined'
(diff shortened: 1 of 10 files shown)


packages/web3-core-method/src/index.js
```diff
@@ -819,6 +815,112 @@ Method.prototype.buildCall = function () {
     return send;
 };
 
+function _handleTxType(tx) {
```
GregTheGreek, line 833: This is duplicate code from web3-eth-accounts.

Why don't we push these two functions to web3-eth-helpers, since both web3-core-method and web3-eth-accounts import it?

Spacesai1or: We should definitely do this if there aren't any crazy refactors involved. I think adding it to `web3-eth-helpers` might solve the need for a provider, but idk if `web3-core-method` needing to passing in `method` [here](https://github.com/ChainSafe/web3.js/blob/b5dadf9ef2352a1378c26937d03d4109fd3319b0/packages/web3-core-method/src/index.js#L781) and `web3-eth-accounts` needing to pass in `_this` [here](https://github.com/ChainSafe/web3.js/blob/b5dadf9ef2352a1378c26937d03d4109fd3319b0/packages/web3-eth-accounts/src/index.js#L270) are going to be an issue. It looks like in both instances, `method` and `_this` are just used to make the RPC call, but they seem to go about it differently

packages/web3-core-method/src/index.js
```diff
@@ -770,19 +771,15 @@ Method.prototype.buildCall = function () {
         };
 
         // Send the actual transaction
-        if (isSendTx && !!payload.params[0] && typeof payload.params[0] === 'object' && typeof payload.params[0].gasPrice === 'undefined') {
+        if (isSendTx 
+            && !!payload.params[0]
+            && typeof payload.params[0] === 'object'
```
jdevcs, line 776: if we remove `typeof payload.params[0].gasPrice === 'undefined'` control never goes to https://github.com/ChainSafe/web3.js/blob/0299df8142006aa1cce28b567a42711ef2acb4bb/packages/web3-core-method/src/index.js#L794  for ganache tests and it fails. 

Plus I think if gas fileds are already with incoming object we can procede to `sendRequest()` without need of moving inside https://github.com/ChainSafe/web3.js/blob/0299df8142006aa1cce28b567a42711ef2acb4bb/packages/web3-core-method/src/index.js#L777 block.

so I suggest 
```suggestion
            && typeof payload.params[0] === 'object' 
            && typeof payload.params[0].gasPrice === 'undefined' 
            && typeof payload.params[0].maxPriorityFeePerGas === 'undefined'
            && typeof payload.params[0].maxFeePerGas === 'undefined'
```

packages/web3-core-method/src/index.js
```diff
@@ -770,19 +771,15 @@ Method.prototype.buildCall = function () {
...
+        ) {
+            if (typeof payload.params[0].type === 'undefined') 
+                payload.params[0].type = _handleTxType(payload.params[0]);
```
GregTheGreek, line 779: Shouldn't this be on line 773?

packages/web3-core-method/src/index.js
```diff
@@ -770,19 +771,15 @@ Method.prototype.buildCall = function () {
...
-                params: 0
-            })).createFunction(method.requestManager);
-
-            getGasPrice(function (err, gasPrice) {
-
-                if (gasPrice) {
-                    payload.params[0].gasPrice = gasPrice;
-                }
+            _handleTxPricing(method, payload.params[0]).then(txPricing => {
+                payload.params[0] = {...payload.params[0], ...txPricing};
```
GregTheGreek: Would it make more sense to move all this logic outside of the blocks?

Regardless of **how** its being sent, we should just update the TX accordingly

packages/web3-core-method/src/index.js
Spacesai1or: Made a modified version of you suggestion, since the user could not provide `gasPrice` or `maxPriorityFeePerGas` and `maxFeePerGas` (the latter two both be required if one is present)

packages/web3-core-method/src/index.js
Spacesai1or: I think in this case, we only want to add gas pricing info to the `payload` object if the info is not already provided for specifically [sendTxs](https://github.com/ChainSafe/web3.js/blob/b5dadf9ef2352a1378c26937d03d4109fd3319b0/packages/web3-core-method/src/index.js#L612), otherwise the code shouldn't be executed

williamsfu99:
Hi @spacesailor24, are we planning on closing this before the hardfork?

Spacesai1or:
> Hi @spacesailor24, are we planning on closing this before the hardfork?

@williamsfu99 yep, that's the plan, however we're having trouble getting the Ganache test suite to pass

gnidan:
> we're having trouble getting the Ganache test suite to pass

@spacesailor24 just so you know, Ganache doesn't support London yet (and won't til Ganache v7).

Spacesai1or:
This is just waiting approvals now

luu-alex (approved):

Spacesai1or:
[v1.5.1-rc.1](https://github.com/ChainSafe/web3.js/releases/tag/v1.5.1-rc.1) contains this fix FYI

Not sure when `v1.5.1` will be released, probably Soon™️

haltman-at:
Just saw the release candidate and tested this and can confirm that it works. :)

Edit: Also our CI with it is passing too. :)

Spacesai1or:
[v1.5.1](https://github.com/ChainSafe/web3.js/releases/tag/v1.5.1) has been released

gnidan:
Thank you for your tireless work getting this out, @spacesailor24

PR #4236, opened 2021-08-13
4.x - Rewrite eth2 beacon

### LINKED ISSUE #4028: [Rewrite] web3-eth2-beaconchain

luu-alex:
The rewrite for eth2 beacon
Will add and update existing testcases in next pr, this includes:
 - Init tests for new packages
-  Update tests for existing packages

closes #4028 
## Description
- Created the `web3-eth2-beacon` package
- updated `web3-core-types` package
- updated `web3-providers-http` now includes eth1 and eth2 requests

## ToDos
- [x] Update doc blocks in web3-eth2-beacon
- [x] Add doc blocks to remaining packages
- [x] Remove any unused types after refactor


packages/web3-eth2-beaconchain/src/types.ts
```diff
@@ -0,0 +1,36 @@
...
+    PrefixedHexString,
+    ValidTypesEnum,
+} from 'web3-utils/lib/types';
+
+export interface Web3EthOptions {
+    web3Client: string;
+    returnType?: ValidTypesEnum;
+}
+
+export type StateId =
```
jdevcs: will be more good if this is string enum

packages/web3-eth2-beaconchain/src/types.ts
```diff
@@ -0,0 +1,36 @@
...
+    | 'head'
+    | 'genesis'
+    | 'finalized'
+    | 'justified'
+    | PrefixedHexString;
+
+export type Status =
```
jdevcs: also convert this to enum

packages/web3-eth2-beaconchain/src/index.ts
```diff
@@ -0,0 +1,606 @@
...
+     */
+
+    async getStateRoot(
+        stateId: StateId,
+        requestArguments?: Partial<RequestArguments>
+    ): Promise<RpcStringResult> {
+        try {
+            return await this.provider.request({
+                ...requestArguments,
+                method: `states/${stateId}/root`,
```
jdevcs: these are restful endpoints, so should not be called `method`, as now HTTP and WS providers will not be eip1193 so it should be renamed accordingly in request method. this change will increase maintainability/readability. I suggest having `Eth2RequestArguments` type

packages/web3-providers-http/src/index.ts
```diff
@@ -138,16 +138,32 @@ export default class Web3ProvidersHttp
...
     }
 
-    /**
-     * Makes an Axios POST request using provided {args}
-     *
-     * @param args RPC options, request params, AxiosConfig
-     * @returns
-     */
-    async request(args: RequestArguments): Promise<RpcResponse> {
+    private async _eth2Request(args: RequestArguments): Promise<AxiosResponse> {
```
jdevcs: function documentation comments required, 
plus as suggested earlier you should define your own type instead of `RequestArguments` for better aligning with eth2 requests ( removing `method` and adding `endpoint` ..etc )

packages/web3-providers-http/src/index.ts
```diff
@@ -138,16 +138,32 @@ export default class Web3ProvidersHttp
...
+            if (error.code === 'ECONNREFUSED' && this._connected) {
+                this._connected = false;
+                // TODO replace with ProviderRpcError
+                this.emit(Web3ProviderEvents.Disconnect, { code: 4900 });
+            }
+            throw Error(error.message);
+        }
+    }
+
+    private async _eth1Request(args: RequestArguments): Promise<AxiosResponse> {
```
jdevcs: function documentation comments required,

packages/web3-eth2-beaconchain/src/types.ts
Spacesai1or: Fyi @luu-alex, you'll want an enum such as:

```javascript
export enum StateIds {
    Head = 'head',
    Genesis = 'genesis',
    ...
}
```

packages/web3-eth2-beaconchain/src/index.ts
Spacesai1or: If we make this change, then we can probably get rid of `ethVersion`, as now we detect via the presence of `endpoint` vs `method` in `Web3ProvidersHttp.request`

Spacesai1or:
Besides the comments, I think everything else looks good, awesome job!

@luu-alex Do you plan on including tests in this PR, or would you like to create an issue and work on that after this is merged?

packages/web3-eth2-beaconchain/src/index.ts
```diff
@@ -0,0 +1,606 @@
...
+     * Get Validator states from state
+     * @param {StateId} stateId State identifier, can be "head", "genesis", "finalized", "justified", slot or hex string encoded stateRoot with 0x prefix
+     * @param {object} requestArguments (Optional) rpcOptions, providerOptions, and desired returnType rpcOptions, providerOptions, and desired returnType
+     * @param {[string]} id (Optional) Array of either hex string encoded public keys or validator index
+     * @returns {Promise} Returns filterable array of validators with their balance, status and index.
+     */
+
+    async getValidatorBalances(
+        stateId: StateId,
+        requestArguments?: Partial<RequestArguments>,
```
Spacesai1or: `reqeustArguments` should be the last argument to follow the same pattern in `web3-eth` (this would need to be updated for every method)

Spacesai1or:
Input/output formatting for each applicable method is missing from `web3-eth2-beaconchain/src/index.ts`

packages/web3-core-types/src/types.ts
```diff
@@ -127,7 +127,7 @@ export type ProviderEventListenerMessage = (message: ProviderMessage) => void;
 export type Web3Client = string | Eip1193Provider;
 
 export interface Eip1193Provider {
-    request: (args: RequestArguments) => Promise<RpcResponse>;
+    request: (args: Eth1RequestArguments | any) => Promise<RpcResponse>;
```
luu-alex: Trying to debug this right now, running into a type error if I change this to `    request: (args: Eth1RequestArguments | Eth2RequestArguments) => Promise<RpcResponse>;` 

think this has to do with `interface IWeb3Provider extends Eip1193Provider`

Error: `web3-eth2-beaconchain: src/index.ts(43,17): error TS2345: Argument of type '{ endpoint: string; params?: object | readonly unknown[] | undefined; providerOptions?: HttpOptions | undefined; returnType?: ValidTypesEnum | undefined; }' is not assignable to parameter of type 'Eth1RequestArguments'.`

luu-alex: Will be addressed when eip1993 provider gets changed

jdevcs (dismissed):

.github/workflows/build.yml
```diff
@@ -52,7 +52,7 @@ jobs:
                   key: ${{ runner.os }}-node14-${{ hashFiles('**/yarn-lock.json') }}
 
             - run: yarn
-              if: steps.cache-deps.outputs.cache-hit != 'true'
+            #   if: steps.cache-deps.outputs.cache-hit != 'true'
```
Spacesai1or:
```suggestion
```

packages/web3-eth2-beaconchain/src/index.ts
```diff
@@ -0,0 +1,582 @@
...
+     * @param {StateId} stateId State identifier, can be "head", "genesis", "finalized", "justified", slot or hex string encoded stateRoot with 0x prefix.
+     * @param {object} requestArguments (Optional) rpcOptions, providerOptions, and desired returnType rpcOptions, providerOptions, and desired returnType
+     * @param {[string]} id (Optional) Array of ids, Either a hex string encoded public key (with 0x prefix) or validator index
+     * @param {Status} status (Optional) Validator status specification https://hackmd.io/ofFJ5gOmQpu1jjHilHbdQQ
+     * @returns {Promise} Array of the validators specified with balance, status and index
+     */
+
+    async getValidators(
+        stateId: StateId,
+        id?: [string],
```
Spacesai1or:
```suggestion
        id?: string[],
```

In typescript you denote an array of strings as `string[]`

`[string]` means a fixed size array with the first index being of type string

packages/web3-eth2-beaconchain/src/index.ts
```diff
@@ -0,0 +1,582 @@
...
+        status?: [Status],
```
Spacesai1or:
```suggestion
        status?: Status[],
```

packages/web3-eth2-beaconchain/src/index.ts
```diff
@@ -0,0 +1,582 @@
...
+     * Get Validator states from state
+     * @param {StateId} stateId State identifier, can be "head", "genesis", "finalized", "justified", slot or hex string encoded stateRoot with 0x prefix
+     * @param {object} requestArguments (Optional) rpcOptions, providerOptions, and desired returnType rpcOptions, providerOptions, and desired returnType
+     * @param {[string]} id (Optional) Array of either hex string encoded public keys or validator index
+     * @returns {Promise} Returns filterable array of validators with their balance, status and index.
+     */
+
+    async getValidatorBalances(
+        stateId: StateId,
+        id?: [string],
```
Spacesai1or:
```suggestion
        id?: string[],
```

packages/web3-providers-http/src/index.ts
```diff
@@ -161,6 +194,39 @@ export default class Web3ProvidersHttp
...
+
+    /**
+     * Makes an Axios POST request using provided {args}
+     *
+     * @param args RPC options, request params, AxiosConfig
+     * @returns
+     */
+    async request(
+        args: Eth1RequestArguments | Eth2RequestArguments
+    ): Promise<RpcResponse | any> {
```
Spacesai1or: We shouldn't have `| any` here

Spacesai1or: We can't just remove `| any` though, because the beaconchain responses don't match the `RpcResponse` that we see in ETH1
image

Spacesai1or: So I think we need to create a type of `eth2RpcResponse` (we'll rename `eth2` later) that looks like:

```javascript
export type eth2RpcResponse = { [key: string]: any} | { [key: string]: any}[]
```

Spacesai1or (approved):

PR #4241, opened 2021-08-14
Don't default type = 0x0 for eth_sendTransaction and eth_sendRawTransaction

Spacesai1or:
Previous to this PR, `eth_sendTransaction` and `eth_sendRawTransaction` would default to include `type = 0x0` when creating the tx object. Now, `type` will not be defaulted if EIP-1559 or EIP-2930 params aren't passed

Addresses issues mentioned [here](https://github.com/MetaMask/metamask-extension/issues/11824#issuecomment-898880026)

Supersedes #4240

### DIFF
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -782,9 +782,6 @@ Method.prototype.buildCall = function () {
                 )
             )
         ) {
-            if (typeof payload.params[0].type === 'undefined') 
-                payload.params[0].type = _handleTxType(payload.params[0]);
-
             _handleTxPricing(method, payload.params[0]).then(txPricing => {
                 if (txPricing.gasPrice !== undefined) {
                     payload.params[0].gasPrice = txPricing.gasPrice;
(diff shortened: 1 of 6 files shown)


packages/web3-core-method/src/index.js
```diff
@@ -889,47 +845,39 @@ function _handleTxPricing(method, tx) {
...
+                getGasPrice()
+            ]).then(responses => {
+                const [block, gasPrice] = responses;
+                if (block && block.baseFeePerGas) {
+                    // The network supports EIP-1559
+
+                    // Taken from https://github.com/ethers-io/ethers.js/blob/ba6854bdd5a912fe873d5da494cb5c62c190adde/packages/abstract-provider/src.ts/index.ts#L230
+                    let maxPriorityFeePerGas, maxFeePerGas;
+
+                    if (tx.gasPrice) {
```
GregTheGreek, line 859: Maybe we should check for eip1559 support and then default to that, else no support offer gasPrice?

Spacesai1or: This part of the code represents a user passing `gasPrice` on a network that supports EIP-1559, so we set the EIP-1559 related properties (`maxPriorityFeePerGas` and `maxFeePerGas`) to what the user gave us as `gasPrice`

We check for EIP-1559 support [here](https://github.com/ChainSafe/web3.js/blob/ce2de6dcbc8ebfa0228ebaaef25ee0acb62be598/packages/web3-core-method/src/index.js#L853) and if it's not supported we return `gasPrice` [here](https://github.com/ChainSafe/web3.js/blob/ce2de6dcbc8ebfa0228ebaaef25ee0acb62be598/packages/web3-core-method/src/index.js#L878)

GregTheGreek: Yep you're right

GregTheGreek (approved):
Lgtm, let's get this rc out 🚀🚀

PR #4248, opened 2021-08-17
Refactors from fixing tsc errors

Spacesai1or:


packages/web3-core-types/src/types.ts
```diff
@@ -164,14 +165,13 @@ export interface RequestArguments {
     rpcOptions?: RpcOptions;
     providerOptions?: HttpOptions;
     returnType?: ValidTypesEnum;
-    ethVersion?: number;
+    ethVersion?: 2;
```
Spacesai1or, line 168: if `ethVersion` is provided, it **has** to `2` since we're not passing it in `web3-eth`

packages/web3-providers-http/src/index.ts
```diff
@@ -140,13 +140,13 @@ export default class Web3ProvidersHttp
...
     private async _eth2Request(args: RequestArguments): Promise<AxiosResponse> {
         try {
-            // @ts-ignore tsc doesn't understand httpOptions.method || 'post'
             const response = await this._httpClient[
-                // @ts-ignore tsc will suggest that it can't be some of the methods like 'GET'
-                args?.providerOptions?.AxiosRequestConfig?.method || 'post'
-            ]('', args?.rpcOptions || {}, {
-                ...args?.providerOptions?.axiosConfig,
-            });
+                (args.providerOptions?.httpMethod as 'get' | 'post') || 'get'
```
Spacesai1or, line 144: Fixes the `// @ts-ignore tsc doesn't understand httpOptions.method || 'post'` error

packages/web3-providers-http/src/index.ts
```diff
@@ -140,13 +140,13 @@ export default class Web3ProvidersHttp
...
+            ](
+                args.method, // URL path
```
Spacesai1or, line 146: My suggestion is to remove relying on `axiosConfig` and use `args.method` as the URL path, and `args.params` as `POST` params

packages/web3-providers-http/src/index.ts
```diff
@@ -204,7 +204,7 @@ export default class Web3ProvidersHttp
                 throw Error('No HTTP client initiliazed');
 
             const response =
-                args?.ethVersion === 2
+                args.ethVersion === 2
```
Spacesai1or, line 207: In multiple places you have `args?.`, but for these methods, `args` is required, so there's no need for the `?`

Spacesai1or:
> This pull request **introduces 1 alert** when merging [975f0c1](https://github.com/ChainSafe/web3.js/commit/975f0c1542ad51a978eb5fde2229933cc3cbbd88) into [c79ae92](https://github.com/ChainSafe/web3.js/commit/c79ae920c93fede100e03ab1c802bc4158cf9046) - [view on LGTM.com](https://lgtm.com/projects/g/ChainSafe/web3.js/rev/pr-f39e16b690950b16c14b16c2c0f7649d6f5699ec)
> 
> **new alerts:**
> 
> * 1 for Unused variable, import, function or class

Can ignore this, I commented out a bunch of code in `web3-eth2-beaconchain/src.ts` so tsc would run successfully

packages/web3-providers-http/src/index.ts
luu-alex: ah i thought this was checking if ethVersion existed. got it

packages/web3-core-types/src/types.ts
luu-alex: Ah true, does it make more sense for it just to be called `ethVersion2` with type `boolean` instead of `ethVersion`?  Although it would hard to change that if any newer versions come out

luu-alex (approved):

Issue #4251, opened 2021-08-18
Remove EIP-1193 from provider packages


Spacesai1or:
Remove from

- [ ] `web3-providers-http` (@spacesailor24)
- [ ] `web3-providers-ws` (@jdevcs)
- [ ] `web3-core-types` Update the type Eip1193Provider and remove the `| any` type in request function (@luu-alex )

Spacesai1or:
Closing as it doesn't pertain to rewrite of 4.x

Issue #4268, opened 2021-08-24
web3-eth2-beacon - include format input/output and tests


luu-alex:
Create the testcases and format input/output for #4028 `web3-eth2-beaconchain` package

philknows:
Closing at request of Alex.

PR #4269, opened 2021-08-25
web3-core-logger

### LINKED ISSUE #4203: Refactor Error Handling for Rewrite
#### 4.x Improvement

In `Web3ProvidersHttp` following throw statements must be before / outside `try` block, it will be more standard and efficient as it will not (throw, catch and again throw), but instead directly throw.

https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-providers-http/src/index.ts#L43
https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-providers-http/src/index.ts#L69
https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-providers-http/src/index.ts#L87
https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-providers-http/src/index.ts#L132

Spacesai1or:
Inits `web3-core-logger` package and includes `web3-providers-http` related fixes for #4251

closes #4203 

Example usage of logger can be found [Here](https://github.com/ChainSafe/web3.js/blob/4072d22fbd6449e10287f1531ef44f565a5236d4/packages/web3-core-logger/README.md#how-to-use)


packages/web3-core-logger/src/index.ts
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    /**
+     * Creates a string with provided error parameters
+     *
+     * @param web3Error An object containing error details
+     * @returns Error string
+     */
+    private _makeErrorString(web3Error: Web3Error): string {
```
Spacesai1or, line 73: Currently the `error.message` (which is a `string`) looks like:

```javascript
loggerVersion: 1.0.0-alpha.0
packageName: web3-providers-http
packageVersion: 4.0.0-alpha.0
code: 1
name: invalidClientUrl
msg: Provided web3Client is an invalid HTTP(S) URL
params: {"web3Client":{}}
```

I was thinking it might be a good idea to `JSON.stringify` `web3Error` instead of piecing together an error string, so that `error.message` is parsable with `JSON.parse(error.message)` to allow the user to handle errors more effectively

packages/web3-eth/src/index.ts
```diff
@@ -59,13 +61,13 @@ export default class Web3Eth {
         requestArguments?: Partial<Eth1RequestArguments>
     ): Promise<RpcStringResult> {
         try {
-            return await this.provider.request({
+            return (await this.provider.request({
                 ...requestArguments,
                 method: 'web3_clientVersion',
                 params: [],
-            });
+            })) as RpcResponse;
```
Spacesai1or, line 68: This is extremely annoying and I believe is a symptom of improper typing

packages/web3-eth2-beaconchain/src/index.ts
```diff
@@ -24,7 +20,7 @@ export default class Web3Beacon {
     provider: IWeb3Provider;
 
     constructor(options: Web3EthOptions) {
-        this.provider = initWeb3Provider(options.web3Client);
+        this.provider = initWeb3Provider(options.web3Client) as IWeb3Provider;
```
Spacesai1or, line 23: Review this as well #4270

packages/web3-eth2-beaconchain/src/index.ts
```diff
@@ -36,7 +32,7 @@ export default class Web3Beacon {
      */
     async getGenesis(
         requestArguments?: Partial<Eth2RequestArguments>
-    ): Promise<RpcStringResult> {
+    ): Promise<{ [key: string]: any }> {
```
Spacesai1or, line 35: @luu-alex `RpcStringResult` was used and is incorrect. I applied this bandaid solution, but these need to be replaced with properly typed responses for each method's return value

packages/web3-providers-eip1193/src/errors.ts
```diff
@@ -0,0 +1,29 @@
...
+    errors: Record<Web3ProvidersEip1193ErrorNames, Web3Error>;
+}
+
+export const Web3ProvidersEip1193ErrorsConfig: Web3ProvidersEip1193ErrorsConfig =
+    {
+        packageName: 'web3-providers-eip1193',
+        packageVersion,
+        errors: {
+            invalidClient: {
+                code: 1,
```
Spacesai1or: As the error handling is currently, we don't really make use of `code`. I included it because I thought users may want to make use of it

@jdevcs @luu-alex

packages/web3-eth2-beaconchain/src/index.ts
luu-alex: This will be addressed in #4268

packages/web3-providers-eip1193/src/errors.ts
luu-alex: We check for duplicate error names, is it okay if we don't check for duplicate `code`s? users may get confuse themselves when creating `errors`

Spacesai1or: That's a good point, we should add a check, but first we must answer if we even need them

luu-alex: I'd say we should have codes. I think it'll provide developers convenience if they wanted to use logic or if statements when they create errors.
Edit:
I dont think we need `code`.  users can use `name` in the same object for the same purpose

packages/web3-eth/src/index.ts
Spacesai1or: #4270 tracks fixing this

packages/web3-core-logger/src/index.ts
Spacesai1or: Updated

packages/web3-providers-eip1193/src/errors.ts
Spacesai1or: Removed `code`

PR #4273, opened 2021-08-25
4.x ws tests and ws changes

### LINKED ISSUE #4255: Revisit types in web3-providers-ws
Can we use [RpcResponse](https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-core-types/src/types.ts#L169) here? Looks like we need to update `result` to optional and add `error`

_Originally posted by @spacesailor24 in https://github.com/ChainSafe/web3.js/pull/4209#r690892338_

### LINKED ISSUE #4256: Init web3-providers-ws tests
As discussed in call tests will be added in new PR (TODO).

_Originally posted by @jdevcs in https://github.com/ChainSafe/web3.js/pull/4209#r691973213_

jdevcs:
## Description

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-core-provider/test/unit/initWeb3Provider.test.ts
+++ b/packages/web3-core-provider/test/unit/initWeb3Provider.test.ts
@@ -48,15 +48,26 @@ describe('Instantiates correct provider for varying provided clients', () => {
         expect(web3ProvidersHttp.supportsSubscriptions).not.toBe(undefined);
     });
 
-    it('should instantiate WebSocket provider', () => {
+    it('should instantiate WebSocket provider', (done) => {
         const wsClient = 'ws://127.0.0.1:8545';
         const Web3ProviderWS: IWeb3Provider = initWeb3Provider(wsClient);
+        
+        let isConnected = (status: boolean) =>{
+            expect(status).toBeTruthy();
+            expect(Web3ProviderWS.web3Client).toBe(wsClient);
+            expect(Web3ProviderWS.setWeb3Client).not.toBe(undefined);
+            expect(Web3ProviderWS.on).not.toBe(undefined);
+            expect(Web3ProviderWS.request).not.toBe(undefined);
+            expect(Web3ProviderWS.supportsSubscriptions).not.toBe(undefined);
 
-        expect(Web3ProviderWS.web3Client).toBe(wsClient);
-        expect(Web3ProviderWS.setWeb3Client).not.toBe(undefined);
-        expect(Web3ProviderWS.on).not.toBe(undefined);
-        expect(Web3ProviderWS.request).not.toBe(undefined);
-        expect(Web3ProviderWS.supportsSubscriptions).not.toBe(undefined);
+            (Web3ProviderWS as Web3ProvidersWS).disconnect(0);
+            done();
+        }
+
+        Web3ProviderWS.on(Web3ProviderEvents.Connect, () => {
+            isConnected(true);
+        });
+        
     });
 
     it('should throw not implemented error for IPC client', () => {
--- a/packages/web3-core-provider/test/unit/initWeb3Provider.test.ts
+++ b/packages/web3-core-provider/test/unit/initWeb3Provider.test.ts
@@ -51,22 +51,12 @@ describe('Instantiates correct provider for varying provided clients', () => {
     it('should instantiate WebSocket provider', (done) => {
         const wsClient = 'ws://127.0.0.1:8545';
         const Web3ProviderWS: IWeb3Provider = initWeb3Provider(wsClient);
-
-        let isConnected = (status: boolean) => {
-            expect(status).toBeTruthy();
-            expect(Web3ProviderWS.web3Client).toBe(wsClient);
-            expect(Web3ProviderWS.setWeb3Client).not.toBe(undefined);
-            expect(Web3ProviderWS.on).not.toBe(undefined);
-            expect(Web3ProviderWS.request).not.toBe(undefined);
-            expect(Web3ProviderWS.supportsSubscriptions).not.toBe(undefined);
-
-            (Web3ProviderWS as Web3ProvidersWS).disconnect(0);
-            done();
-        };
-
-        Web3ProviderWS.on(Web3ProviderEvents.Connect, () => {
-            isConnected(true);
-        });
+        expect(Web3ProviderWS.web3Client).toBe(wsClient);
+        expect(Web3ProviderWS.setWeb3Client).not.toBe(undefined);
+        expect(Web3ProviderWS.on).not.toBe(undefined);
+        expect(Web3ProviderWS.request).not.toBe(undefined);
+        expect(Web3ProviderWS.supportsSubscriptions).not.toBe(undefined);
+        done();
     });
 
     it('should throw not implemented error for IPC client', () => {
--- a/packages/web3-core-provider/test/unit/initWeb3Provider.test.ts
+++ b/packages/web3-core-provider/test/unit/initWeb3Provider.test.ts
@@ -51,8 +51,8 @@ describe('Instantiates correct provider for varying provided clients', () => {
     it('should instantiate WebSocket provider', (done) => {
         const wsClient = 'ws://127.0.0.1:8545';
         const Web3ProviderWS: IWeb3Provider = initWeb3Provider(wsClient);
-        
-        let isConnected = (status: boolean) =>{
+
+        let isConnected = (status: boolean) => {
             expect(status).toBeTruthy();
             expect(Web3ProviderWS.web3Client).toBe(wsClient);
             expect(Web3ProviderWS.setWeb3Client).not.toBe(undefined);
@@ -62,12 +62,11 @@ describe('Instantiates correct provider for varying provided clients', () => {
 
             (Web3ProviderWS as Web3ProvidersWS).disconnect(0);
             done();
-        }
+        };
 
         Web3ProviderWS.on(Web3ProviderEvents.Connect, () => {
             isConnected(true);
         });
-        
     });
 
     it('should throw not implemented error for IPC client', () => {
--- a/packages/web3-core-types/src/types.ts
+++ b/packages/web3-core-types/src/types.ts
@@ -177,7 +177,8 @@ export interface Eth2RequestArguments extends RequestArguments {
 export interface RpcResponse {
     id: number;
     jsonrpc: string;
-    result: any;
+    result?: any;
+    error?: any;
 }
 
 export type Eth2RpcResponse = { [key: string]: any } | { [key: string]: any }[];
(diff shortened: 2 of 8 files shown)


packages/web3-core-provider/test/unit/initWeb3Provider.test.ts
```diff
@@ -48,15 +48,15 @@ describe('Instantiates correct provider for varying provided clients', () => {
         expect(web3ProvidersHttp.supportsSubscriptions).not.toBe(undefined);
     });
 
-    it('should instantiate WebSocket provider', () => {
+    it('should instantiate WebSocket provider', (done) => {
```
Spacesai1or, line 51: I'm assuming `done` was required here because [connect](https://github.com/ChainSafe/web3.js/blob/junaid/wstests/packages/web3-providers-ws/src/index.ts#L417) is called when the constructor calls `setWeb3Client` which is actually trying to connect to the `wsClient`. Instead of doing this and having an attempted connection, make use of the Jest mock functions as used [here](https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-providers-http/test/unit/setWeb3Client.test.ts#L18)

So you'd having something like:

```typescript
beforeAll(() => {
    Web3ProvidersWs.prototype.connect = jest.fn();
}
```

packages/web3-providers-ws/src/index.ts
```diff
@@ -315,9 +316,7 @@ export default class Web3ProviderWS
                 }
 
                 if (id && this.responseQueue.has(id)) {
-                    let requestItem = this.responseQueue.get(id);
```
Spacesai1or, line 318: Why is this not needed anymore?

packages/web3-providers-ws/test/unit/constructor.test.ts
```diff
@@ -1,10 +1,30 @@
...
 describe('Web3ProviderWS Tests', () => {
-    it('should error because invalid WS address', () => {
+    let provider: Web3ProviderWS;
+    it('should throw error for invalid WS address', () => {
         expect(() => {
             new Web3ProviderWS('http://127.0.0.1:8545');
         }).toThrowError('Invalid WebSocket URL provided');
     });
+
+    test('should instantiate with valid client URL', async (done) => {
```
Spacesai1or, line 11: Same `done` fixed as mentioned above

packages/web3-providers-ws/test/unit/constructor.test.ts
```diff
@@ -1,10 +1,30 @@
...
+        done();
+    });
+
+    test('should create instance of Web3ProviderWS and call connect', async (done) => {
+        const connectFunc = jest.spyOn(
+            Web3ProviderWS.prototype as any,
+            'connect'
+        );
+        connectFunc.mockImplementation(() => {
+            done();
```
Spacesai1or, line 21: Maybe this is the better way to do it, but wouldn't calling `done` here cause the test to finish executing before line 28 executes?

packages/web3-providers-ws/test/unit/request.test.ts
```diff
@@ -1,38 +1,38 @@
...
 } from 'web3-core-types/lib/types';
 
-describe('Web3ProvidersWS.request', () => {
-    const request: Eth1RequestArguments = {
-        rpcOptions: { id: 1, jsonrpc: '2.0' },
-        method: 'eth_blockNumber',
-        params: [],
-    };
-    jest.setTimeout(30000);
+describe.skip('Web3ProvidersWS.request', () => {
```
Spacesai1or, line 8: These are unit tests, so the return values for the web socket client should be mocked

CI and an actual WS endpoint will be introduced via integration or e2e tests

packages/web3-providers-ws/test/unit/supportsSubscriptions.test.ts
```diff
@@ -2,7 +2,7 @@ import Web3ProviderWS from '../../src/index';
 import { WebSocketOptions } from '../../src/types';
 
 describe('web3ProvidersWS.supportsSubscriptions', () => {
-    it('should return true', () => {
+    it('should return true', async () => {
```
Spacesai1or: I don't think you any `async` here, and the mocking of `connect` needs to be done here as well

Spacesai1or (changes requested):

packages/web3-core-provider/test/unit/initWeb3Provider.test.ts
nazarhussain: In general we during unit tests should not initiate any socket, any connection at all and always mock such behaviour.

nazarhussain (changes requested):
In the same file I observed some code like following. 

```javascript
        Web3ProvidersHttp.prototype.request = jest.fn();
        // Web3ProviderHttp makes a request to get chainId of
        // connected client upon instantiation, so we mock the response
        // @ts-ignore mockReturnValue added by jest
        Web3ProvidersHttp.prototype.request.mockReturnValue('0x1');
```

1. We should avoid assigning jest mock functions to any object, these are not restored in jest test cycle. 
2. Jest provide full TS support so you can use Jest types when calling mock values. 
3. Jest don't specify mock functions but prefer start with spy and then implement mock return values. 

The above code should be better to write as: 

```javascript 
jest.spyOn(Web3ProvidersHttp.prototype, 'request').mockReturnValue('0x1');
```

Also in the same test 

```
expect(web3ProvidersHttp.request).not.toBe(undefined);
```

Such expectations can lead to false positives. You must expect the exact value which you used in the mock, otherwise inner functionality may not be using correct mock and yet test will pass as its returning some value.

packages/web3-providers-ws/test/unit/constructor.test.ts
```diff
@@ -1,10 +1,30 @@
...
+    test('should instantiate with valid client URL', async (done) => {
+        let client = 'ws://127.0.0.1:8545';
+        provider = new Web3ProviderWS(client);
+        expect(provider.web3Client).toBe(client);
+        done();
+    });
+
+    test('should create instance of Web3ProviderWS and call connect', async (done) => {
+        const connectFunc = jest.spyOn(
+            Web3ProviderWS.prototype as any,
```
nazarhussain: Please avoid using `any`, if its the correct object and really have `connect` method it should work normal.

packages/web3-providers-ws/test/unit/supportsSubscriptions.test.ts
```diff
@@ -2,7 +2,7 @@ import Web3ProviderWS from '../../src/index';
...
         const web3ProvidersWS = new Web3ProviderWS('ws://127.0.0.1:8546');
         expect(web3ProvidersWS.supportsSubscriptions()).toBeTruthy();
```
nazarhussain: Please avoid using `toBeTruthy` matcher, and always match the exact type value the function is supposed to return.

nazarhussain (changes requested):

packages/web3-core-types/src/types.ts
```diff
@@ -177,7 +177,8 @@ export interface Eth2RequestArguments extends RequestArguments {
 export interface RpcResponse {
     id: number;
     jsonrpc: string;
-    result: any;
+    result?: any;
+    error?: any;
```
nazarhussain, line 181: I consider its **JSON PRC 2** spec, in that case this type `RpcResponse` is not correct. Both of the `result` and `error` can't be optional the same time. Also when `error` is present it have a particular and fixed structure which implies it MUST contain `code` attribute.  See reference https://www.jsonrpc.org/specification#response_object

nazarhussain (changes requested):

packages/web3-core-provider/test/unit/initWeb3Provider.test.ts
jdevcs: you are right but only in case: 
`Mock functions are also known as "spies", because they let you spy on the behavior of a function that is called indirectly by some other code, rather than only testing the output`  and we are not testing output of connect here.

This test is not for connect ( looking successful connection ) but init of few other items.

packages/web3-providers-ws/src/index.ts
jdevcs: its already mentioned in title of [this](https://github.com/ChainSafe/web3.js/pull/4273/commits/e11b514358f4c84ac4b47864166e42ab6b7b0e9a) commit

packages/web3-providers-ws/test/unit/constructor.test.ts
jdevcs: added comments for this above,

packages/web3-providers-ws/test/unit/constructor.test.ts
jdevcs: thats not the case line 28 is executed, its not in scope of test but in a mock function.

yes its better and mocked here because test is dependent on this  " ...and call connect"

packages/web3-providers-ws/test/unit/request.test.ts
jdevcs: This PR is init tests only so I skipped it, based on same assumption you shared here.

packages/web3-providers-ws/test/unit/supportsSubscriptions.test.ts
jdevcs: added comments for where to use mock above

packages/web3-core-types/src/types.ts
jdevcs: and as per specs these cant be mandatory at same time:

result
This member is REQUIRED on success. >> required
This member MUST NOT exist if there was an error invoking the method. >> optional 
The value of this member is determined by the method invoked on the Server.
error
This member is REQUIRED on error. >> required
This member MUST NOT exist if there was no error triggered during invocation. >> optional
The value for this member MUST be an Object as defined in section 5.1.

packages/web3-providers-ws/test/unit/supportsSubscriptions.test.ts
jdevcs: made change

packages/web3-providers-ws/test/unit/constructor.test.ts
jdevcs: made change, you are right

packages/web3-core-types/src/types.ts
nazarhussain: I am fully aware of JSON Rpc specs. Had implemented  it many times already. If you take a close look at specs you mentioned `result` and `error` can't be optional the same time. For any given RPC Response there MUST exists either the `result` or `error`. Its an important type so we have to fix it.

nazarhussain (changes requested):

jdevcs:
closing this as types changes are already in 4.x with new architecture and ws & its tests will be added.

PR #4277, opened 2021-08-27
EIP-1559 Fix Issue #4258

### LINKED ISSUE #4258: If network supports EIP-1559, then we cannot send a legacy transaction
#### Expected behavior
I would expect to be able to specify a legacy `tx.type` (i.e. "0x0" ) and allow for the gasPrice to be auto-filled. The motivation for this is that some hardware wallets do not yet support EIP1559 transactions, while the network provider will.

#### Actual behavior
Instead the `maxFeePerGas` and `maxPriorityFeePerGas` are set causing a transaction envelope rejection. This I imagine is due to the difference between [web3-core-method](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-core-method/src/index.js#L853-L878) `_handleTxPricing` and [web3-eth-accounts](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-eth-accounts/src/index.js#L376-L380) `_handleTxPricing`. `web3-eth-accounts` checks the `tx.type` while `web3-core-method` does not.

Note that this is the reverse of the issue mentioned in https://github.com/MetaMask/metamask-extension/issues/11824 where if we intend to do a legacy transaction, the `gasPrice` gets overwritten.

1. Create a `const ethContract = new web3.eth.Contract(...)`
2. Send a transaction via the contract methods `ethContract.methods["approve"].send({ type: "0x0" })`

#### Environment

web3 1.5.2

Spacesai1or:
- Checks `tx.type` before deicing what gas params to add to `tx`

Closes #4258

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -438,7 +438,7 @@ Released with 1.0.0-beta.37 code base.
 
 ### Fixed
 
-- Unable to send legacy transaction if network supported EIP-1559 (#4277w)
+- Unable to send legacy transaction if network supported EIP-1559 (#4277)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -434,6 +434,12 @@ Released with 1.0.0-beta.37 code base.
 
 ## [Unreleased]
 
+## [1.5.3]
+
+### Fixed
+
+- Unable to send legacy transaction if network supported EIP-1559 (#4277w)
+
 ### Changed
 
 - ethers from 5.1.4 to 5.4.4 (#4231)
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -850,7 +850,10 @@ function _handleTxPricing(method, tx) {
                 getGasPrice()
             ]).then(responses => {
                 const [block, gasPrice] = responses;
-                if (block && block.baseFeePerGas) {
+                if (
+                    (tx.type === '0x2' || tx.type === undefined) &&
+                    (block && block.baseFeePerGas)
+                ) {
                     // The network supports EIP-1559
 
                     // Taken from https://github.com/ethers-io/ethers.js/blob/ba6854bdd5a912fe873d5da494cb5c62c190adde/packages/abstract-provider/src.ts/index.ts#L230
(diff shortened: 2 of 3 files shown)


sambacha:
Just a note further down in the codebase here:

https://github.com/ChainSafe/web3.js/blob/154908bfb6fd335b409d88a1e00c1a5db931c423/packages/web3-core-method/src/index.js#L869

was updated recently, see https://github.com/ethers-io/ethers.js/issues/1817 / commit: https://github.com/ethers-io/ethers.js/commit/7175e2e99c2747e8d2314feb407bf0a0f9371ece

the value should read: `0x9502F900` (2500000000)

Spacesai1or:
> Just a note further down in the codebase here:
> 
> https://github.com/ChainSafe/web3.js/blob/154908bfb6fd335b409d88a1e00c1a5db931c423/packages/web3-core-method/src/index.js#L869
> 
> was updated recently, see [ethers-io/ethers.js#1817](https://github.com/ethers-io/ethers.js/issues/1817) / commit: [ethers-io/ethers.js@7175e2e](https://github.com/ethers-io/ethers.js/commit/7175e2e99c2747e8d2314feb407bf0a0f9371ece)
> 
> the value should read: `0x9502F900` (2500000000)

Thank you @sambacha for bringing this to our attention!

luu-alex (dismissed):

CHANGELOG.md
```diff
@@ -434,6 +434,12 @@ Released with 1.0.0-beta.37 code base.
 
 ## [Unreleased]
 
+## [1.5.3]
+
+### Fixed
+
+- Unable to send legacy transaction if network supported EIP-1559 (#4277w)
```
jdevcs: Plus add more description of current behavior for release notes and doc: 

"If `tx.type` is not specified web3 will consider it type `0x2` transaction ( provided NW supports this ) and will add  `maxFeePerGas`, `maxPriorityFeePerGas`. "

other thn that lgtm

Spacesai1or: I think the linked issue number provides the needed context. No other entry in this file has such detail

jdevcs (dismissed):

luu-alex (approved):

jdevcs (approved):

nazarhussain (approved):

PR #4293, opened 2021-09-06
add tx.chainId if tx.common.chainId not provided

### LINKED ISSUE #4208: Don't require `chainId` if `common` is present [eth.accounts.signTransaction]
https://github.com/ChainSafe/web3.js/blob/458a47af814982df6cfd111c0d8a8f9b17581937/packages/web3-eth-accounts/src/index.js#L266-L267
`transaction.chainId` is not used anyway if `transaction.common` is present:
https://github.com/ChainSafe/web3.js/blob/458a47af814982df6cfd111c0d8a8f9b17581937/packages/web3-eth-accounts/src/index.js#L185-L192

jdevcs:
## Description

https://github.com/ChainSafe/web3.js/issues/4208

Throw error if user provided `tx.chainId` and `tx.common.customChain.chainId`plus these are not matching.

Not considering `tx.chainId` if `tx.common.customChain.chainId` is provided for `web3.eth.accounts.signTransaction` function ,

## Type of change

## Checklist:


packages/web3-eth-accounts/src/index.js
```diff
@@ -264,18 +271,18 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
 
     // Otherwise, get the missing info from the Ethereum Node
     return Promise.all([
-        isNot(tx.chainId) ? _this._ethereumCall.getChainId() : tx.chainId,
+        ((isNot(tx.common) || isNot(tx.common.customChain.chainId)) ? ( isNot(tx.chainId) ? _this._ethereumCall.getChainId() : tx.chainId) : undefined ), //tx.common.customChain.chainId is not optional inside tx.common if tx.common is provided
```
Spacesai1or: This is confusing, but we are limited by possibly having to do a async call (`getChainId`), so I guess it's okay for `1.x`

Spacesai1or:
```suggestion
        isNot(tx.common) || isNot(tx.common.customChain.chainId) ?
            isNot(tx.chainId) ? _this._ethereumCall.getChainId() : tx.chainId
            : undefined, //tx.common.customChain.chainId is not optional inside tx.common if tx.common is provided
```

I think this formatting makes it a bit more readable

packages/web3-eth-accounts/src/index.js
```diff
@@ -264,18 +271,18 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
-        if (isNot(args[0]) || isNot(args[1]) || isNot(args[2]) || isNot(args[3])) {
+        if ( (isNot(args[0]) && isNot(tx.common) && isNot(tx.common.customChain.chainId)) || isNot(args[1]) || isNot(args[2]) || isNot(args[3])) {
             throw new Error('One of the values "chainId", "networkId", "gasPrice", or "nonce" couldn\'t be fetched: ' + JSON.stringify(args));
         }
-    
+
     return signed({
             ...tx,
-            chainId: args[0],
+            ... ((isNot(tx.common) || isNot(tx.common.customChain.chainId) ) ? {chainId: args[0]}:{}), // if common.customChain.chainId is provided no need to add tx.chainId
```
spacesailor24, line 285:
```suggestion
            ...tx // Will overwrite above chaindId if provided via tx.chainId or common.customChain.chainId
```

Spacesai1or (commented):
Looks good, just a formatting suggestion

packages/web3-eth-accounts/src/index.js
```diff
@@ -154,6 +154,13 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
         return Promise.reject(error);
     }
 
+    if (!isNot(tx.common) && !isNot(tx.common.customChain.chainId) && !isNot(tx.chainId) && tx.chainId !== tx.common.customChain.chainId) {
```
nazarhussain: There is no check I could find in the flow which enforces that if `tx.common` exists then `tx.common.customChain` must exists. So in case `tx.common` exists but `tx.common.customCain` does not code will throw a runtime error. So we must add an additional check here. 

Also using so many negative assertions reduce readability. So create a small utility function `isExists` and then use it here to avoid those double negations. 

```javascript
if(isExists(tx.common) && isExists(tx.common.customChain) && isExists(tx.common.customChain.chainId) && isExists(tx.chainId) && tx.chainId !== tx.common.customChain.chainId)) {
```

packages/web3-eth-accounts/src/index.js
```diff
@@ -264,18 +271,20 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
 
     // Otherwise, get the missing info from the Ethereum Node
     return Promise.all([
-        isNot(tx.chainId) ? _this._ethereumCall.getChainId() : tx.chainId,
+        ((isNot(tx.common) || isNot(tx.common.customChain.chainId)) ? //tx.common.customChain.chainId is not optional inside tx.common if tx.common is provided
+            ( isNot(tx.chainId) ? _this._ethereumCall.getChainId() : tx.chainId) 
+            : undefined ),
```
nazarhussain, line 294: Why we need to specify an `undefined` value in the promise chain? the `chainId` must exists either in the tx object or from the ethereum call.

packages/web3-eth-accounts/src/index.js
```diff
@@ -264,18 +271,20 @@ Accounts.prototype.signTransaction = function signTransaction(tx, privateKey, ca
...
-        isNot(tx.chainId) ? _this._ethereumCall.getChainId() : tx.chainId,
+        ((isNot(tx.common) || isNot(tx.common.customChain.chainId)) ? //tx.common.customChain.chainId is not optional inside tx.common if tx.common is provided
+            ( isNot(tx.chainId) ? _this._ethereumCall.getChainId() : tx.chainId) 
+            : undefined ), 
         isNot(tx.nonce) ? _this._ethereumCall.getTransactionCount(_this.privateKeyToAccount(privateKey).address) : tx.nonce,
         isNot(hasTxSigningOptions) ? _this._ethereumCall.getNetworkId() : 1,
         _handleTxPricing(_this, tx)
     ]).then(function(args) {
-        if (isNot(args[0]) || isNot(args[1]) || isNot(args[2]) || isNot(args[3])) {
+        if ( (isNot(args[0]) && isNot(tx.common) && isNot(tx.common.customChain.chainId)) || isNot(args[1]) || isNot(args[2]) || isNot(args[3])) {
```
nazarhussain: Please use the destruct for the args array to improve the code readability. 

```javascript
const [chainId, nonce, networkId, gasInformation] = args; 
```

nazarhussain (changes requested):

packages/web3-eth-accounts/src/index.js
jdevcs: yes I was aware of it already and there are 100s of places in 1.x for **readability** improvement, that why we will only focus our energies on only functional changes in 1.x and will put energy in 4.x

packages/web3-eth-accounts/src/index.js
jdevcs: Technically ChainId is optional. 

But this logic is already adding it, via tx.chainId or from common obj, PR was all about avoiding "the chainId must exists either in the tx object or from the ethereum call." in case common is there ;)

packages/web3-eth-accounts/src/index.js
jdevcs: There are no double negation operators in this function and its call, 
I even personally dnt like naming of `isNot` and there are 100s of places in 1.x for **readability** improvement, that why we will only focus our energies on only functional changes in 1.x and will not overhaul it.

nazarhussain: I didn't said there is double negation operator 😄 

I said dual negative assertions, once from `isNot` and then `!` operator makes it difficult to read. And such few minutes of refactoring which increase readability is not an effort we should avoid for the code base we touch upon. Earlier such practice was avoided end-up in poor readable code.

packages/web3-eth-accounts/src/index.js
nazarhussain: Yes I believe there could be more 100 more places, but the code base we touch upon must be touch up a little. Refactoring code is on-going process, there can't be a time where we say let's refactor all complex conditional checks in the library, rather we keep improving which ever code we touch.

luu-alex (dismissed):

Spacesai1or (dismissed):

nazarhussain (approved):

luu-alex (approved):

PR #3947, opened 2021-03-11
Feature/improve xml request error handling

uluhonolulu:
## Description

Adds explicit handling for XMLHttpRequest errors.

When an XMLHttpRequest fails, instead of throwing `Error: Invalid JSON RPC response: ""` throw `Error: HTTP request failed` with error code and other details.

Fixes #3425.

Details: HttpProvider uses `xhr2-cookies` for Node HTTP requests. This package catches and propagates HTTP errors but swallows the actual error. In other words, we know that something wrong happened but we don't know what exactly, see https://github.com/souldreamer/xhr2-cookies/issues/9. I'm accessing a private variable that references the underlying ClientRequest to get the error information and pass it upstream.

BTW, "xhr2-cookies" is not maintained anymore, the last commit was done 3 years ago. I think it would be safer to update to something else, maybe "node-xhr2" which has the same API.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation. _This is an internal fix so no documentation changes should be necessary._
- [x] My changes generate no new warnings.
- [x] Any dependent changes have been merged and published in downstream modules.
- [x] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [x] I ran `npm run test:unit` with success. _TBH I had a few unrelated tests failing, but they were failing before my changes, see #3946._
- [ ] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code. _Not applicable_
- [x] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
- [x] I have tested my code on the live network.
- [x] I have checked the Deploy Preview and it looks correct.
- [x] I have updated the `CHANGELOG.md` file in the root folder.

PR #4298, opened 2021-09-08
Feature/improve xml request error handling (#3947)

### LINKED ISSUE #3425: XMLHttpRequest errors are not handled correctly

Spacesai1or:
I took over #3947 to fix `CHANGELOG.md`

### DIFF
--- a/packages/web3-providers-http/src/index.js
+++ b/packages/web3-providers-http/src/index.js
@@ -97,7 +97,7 @@ HttpProvider.prototype.send = function (payload, callback) {
     var request = this._prepareRequest();
 
     request.onreadystatechange = function() {
-        if (request.readyState === 4 && request.timeout !== 1) {
+        if (request.readyState === 4 && request.timeout !== 1 && (request.status >= 200 && request.status < 300)) {
             var result = request.responseText;
             var error = null;
 
@@ -112,6 +112,16 @@ HttpProvider.prototype.send = function (payload, callback) {
         }
     };
 
+	//since XHR2._onHttpRequestError swallows the initial request error, we need to get it from the underlying request
+	request.addEventListener('loadstart', function() {
+		var clientRequest = request._request;
+		if (clientRequest) {
+			clientRequest.on('error', function (error) {
+				callback(error || errors.RequestFailed());
+			});
+		}		
+	});
+
     request.ontimeout = function() {
         _this.connected = false;
         callback(errors.ConnectionTimeout(this.timeout));
@@ -121,7 +131,7 @@ HttpProvider.prototype.send = function (payload, callback) {
         request.send(JSON.stringify(payload));
     } catch(error) {
         this.connected = false;
-        callback(errors.InvalidConnection(this.host));
+        callback(errors.InvalidConnection(this.host, { code: 'ECONNREFUSED', reason: 'ECONNREFUSED' }));
     }
 };
 
--- a/packages/web3-providers-http/src/index.js
+++ b/packages/web3-providers-http/src/index.js
@@ -97,7 +97,7 @@ HttpProvider.prototype.send = function (payload, callback) {
     var request = this._prepareRequest();
 
     request.onreadystatechange = function() {
-        if (request.readyState === 4 && request.timeout !== 1 && (request.status >= 200 && request.status < 300)) {
+        if (request.readyState === 4 && request.timeout !== 1 && request.status !== 0) {
             var result = request.responseText;
             var error = null;
 
--- a/test/helpers/FakeXHR2.js
+++ b/test/helpers/FakeXHR2.js
@@ -5,6 +5,7 @@ var assert = chai.assert;
 var FakeXHR2 = function () {
     this.responseText = undefined;
     this.readyState = 4;
+    this.status = 200;
     this.onreadystatechange = null;
     this.async = true;
     this.agents = {};
@@ -37,5 +38,9 @@ FakeXHR2.prototype.send = function (payload) {
         this.onreadystatechange();
     }
 };
+
+FakeXHR2.prototype.addEventListener = function(eventType, callback) {
+	callback();
+}
 
 module.exports = {XMLHttpRequest: FakeXHR2};
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -165,6 +165,7 @@ Released with 1.0.0-beta.37 code base.
 - Add missing subscription.on('connected') TS type definition (#3319)
 - Add missing bignumber.js dependency for TS types (#3386)
 - Upgrade swarm-js to 0.1.40 to remove npm vulnerability warning (#3399)
+- Fixed misleading error thrown on XMLHTTP request error
 - Upgrade devDeps to resolve security warnings (#3464)
   - dtslint 0.4.2 => 3.4.1
   - definitelytyped-header-parser 1.0.1 => 3.9.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -165,7 +165,6 @@ Released with 1.0.0-beta.37 code base.
 - Add missing subscription.on('connected') TS type definition (#3319)
 - Add missing bignumber.js dependency for TS types (#3386)
 - Upgrade swarm-js to 0.1.40 to remove npm vulnerability warning (#3399)
-- Fixed misleading error thrown on XMLHTTP request error
 - Upgrade devDeps to resolve security warnings (#3464)
   - dtslint 0.4.2 => 3.4.1
   - definitelytyped-header-parser 1.0.1 => 3.9.0
@@ -442,3 +441,4 @@ Released with 1.0.0-beta.37 code base.
 - lerna from 3.22.1 to 4.0.0 (#4231)
 - Dropped build tests in CI for Node v8 and v10, and added support for Node v14
 - Change default value for `maxPriorityFeePerGas` from `1 Gwei` to `2.5 Gwei` (#4284)
+- Fixed misleading error thrown on XMLHTTP request error (#4298)
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -44,6 +44,9 @@ module.exports = {
         var message = !!result && !!result.error && !!result.error.message ? result.error.message : 'Invalid JSON RPC response: ' + JSON.stringify(result);
         return new Error(message);
     },
+    RequestFailed: function (){
+        return new Error('HTTP request failed');
+    },
     ConnectionTimeout: function (ms){
         return new Error('CONNECTION TIMEOUT: timeout of ' + ms + ' ms achived');
     },


packages/web3-providers-http/src/index.js
```diff
@@ -97,7 +97,7 @@ HttpProvider.prototype.send = function (payload, callback) {
     var request = this._prepareRequest();
 
     request.onreadystatechange = function() {
-        if (request.readyState === 4 && request.timeout !== 1) {
+        if (request.readyState === 4 && request.timeout !== 1 && (request.status >= 200 && request.status < 300)) {
```
Spacesai1or: @uluhonolulu Do you think this change is necessary? I'm worried this could break someone code

Do you think we could change it to (as `request.status` is `0` mentioned [here](https://github.com/ChainSafe/web3.js/issues/3425)):
```suggestion
        if (request.readyState === 4 && request.timeout !== 1 && request.status !== 0) {
```

uluhonolulu: Good idea, I'll do that.

uluhonolulu: @spacesailor24 Did you mean that you would do it yourself? I tried to push my change but it doesn't show up here.

Spacesai1or: @uluhonolulu Yea, sorry about that, since I merged your PR into my branch you won't be able to push changes - just wanted your feedback

test/helpers/FakeXHR2.js
```diff
@@ -5,6 +5,7 @@ var assert = chai.assert;
 var FakeXHR2 = function () {
     this.responseText = undefined;
     this.readyState = 4;
+    this.status = 200;
```
Spacesai1or, line 8:
```suggestion
```

Spacesai1or:
For better understanding, we can see that on line `227` the `xhr2-cookies` package calls it's `_sendHxxpRequest`, which then has a `.on('error'...` that calls `_onHttpRequestError` on line `252`

image

And here we can see that `_onHttpRequestError` does nothing with `error`, so it does appear to be swallowing any Http errors

image

Spacesai1or:
@uluhonolulu Did you test that these changes solve #3425? Would it be possible for you to provide an example project where an `xhr2-cookies` is caught and emitted from the changes you made? I'm just having a hard time understanding this enough to merge it

packages/web3-providers-http/src/index.js
```diff
@@ -97,7 +97,7 @@ HttpProvider.prototype.send = function (payload, callback) {
     var request = this._prepareRequest();
 
     request.onreadystatechange = function() {
-        if (request.readyState === 4 && request.timeout !== 1) {
+        if (request.readyState === 4 && request.timeout !== 1 && request.status !== 0) {
```
nazarhussain, line 100: As per documentation this property returns (HTTP Status Codes)[https://developer.mozilla.org/en-US/docs/Web/HTTP/Status] and there is no status code `0` exists, so why are we checking it here? 

Do we want to check if the status code exists? if yes then condition should not use a constant zero. 

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/status

packages/web3-providers-http/src/index.js
```diff
@@ -97,7 +97,7 @@ HttpProvider.prototype.send = function (payload, callback) {
     var request = this._prepareRequest();
```
nazarhussain, line 97: @spacesailor24  The PR referred in mentioned to upgrade `xhr2-cookies ` to `node-xhr2`. I could not find that change incorporated here. Does the scope for this PR changed?

packages/web3-providers-http/src/index.js
```diff
@@ -112,6 +112,16 @@ HttpProvider.prototype.send = function (payload, callback) {
         }
     };
 
+	//since XHR2._onHttpRequestError swallows the initial request error, we need to get it from the underlying request
+	request.addEventListener('loadstart', function() {
+		var clientRequest = request._request;
```
nazarhussain, line 117: If `request` is an object of `XMLHttpRequest` then I could not find any documented reference of `_request` property. We should avoid using such private behaviour of a native object. 

Also the `request` object could be either `xhr2-cookies` or `XMLHttpRequest` so such private attributes can't provide confirmation of availability on different type of objects. 


Did you tried top level events `error` or `abort`? As per docs if there is any error `error` should be fired after `loadstart`

nazarhussain (changes requested):

packages/web3-providers-http/src/index.js
Spacesai1or: Mentioned in the [related issue](https://github.com/ChainSafe/web3.js/issues/3425#issue-585688207), `request.status` is `0`. Also mentioned [here](https://github.com/ChainSafe/web3.js/pull/3947#discussion_r670207027)

packages/web3-providers-http/src/index.js
Spacesai1or: This PR does not include the package switch. What do y'all think about updating the package used?

[Current package](https://github.com/souldreamer/xhr2-cookies) - Last release 4 years ago
[Suggested package](https://github.com/pwnall/node-xhr2) - Last release was Feb 9 2021

CC @jdevcs @luu-alex

packages/web3-providers-http/src/index.js
Spacesai1or: `_request` is a private property, and this PR would likely break if we switch out the package to use `node-xhr2`

This PR was originally created by @uluhonolulu, and I'm not really sure how to setup an environment to test that this works as intended. However, how we adjust this specific code is dependent on what package we choose to go with

jdevcs: why we are binding `clientRequest.on('error'...` when request started loading `loadstart` ? shouldnt it be done directly with `request` inside `HttpProvider.prototype.send` ?

packages/web3-providers-http/src/index.js
jdevcs: @spacesailor24  could you check xhr2-cookies vs node-xhr2 with internal **Lib Selection Checklist for Web3.js** . 
( build size impact, ..etc ) , lets open another PR for lib switch, if this lib update is not directly required for fix of xml request error handling ( this PR )

uluhonolulu:
> And here we can see that `_onHttpRequestError` does nothing with `error`, so it does appear to be swallowing any Http errors

Exactly. This is why I suggested switching to node-xhr2.

uluhonolulu:
> @uluhonolulu Did you test that these changes solve #3425? Would it be possible for you to provide an example project where an `xhr2-cookies` is caught and emitted from the changes you made? I'm just having a hard time understanding this enough to merge it

@spacesailor24 I tested it with my own project and I'm 100% sure it worked as intended, but it was half a year ago so I don't even remember how I invoked the error. Most probably I just tried to connect to a node that didn't exist.

I can provide a test project, but TBH I'm in a middle of moving to a new apt so it's going to be the next weekend at best.

packages/web3-providers-http/src/index.js
uluhonolulu: > If `request` is an object of `XMLHttpRequest` then I could not find any documented reference of `_request` property. We should avoid using such private behaviour of a native object.
> 
> Also the `request` object could be either `xhr2-cookies` or `XMLHttpRequest` so such private attributes can't provide confirmation of availability on different type of objects.
> 
> Did you tried top level events `error` or `abort`? As per docs if there is any error `error` should be fired after `loadstart`

@nazarhussain this is my hacky attempt to get the underlying error, because xhr2-cookies just swallows it (see https://github.com/souldreamer/xhr2-cookies/issues/9).

uluhonolulu: > why we are binding `clientRequest.on('error'...` when request started loading `loadstart` ? shouldnt it be done directly with `request` inside `HttpProvider.prototype.send` ?

@jdevcs because `request._request` doesn't exist at this point yet.

packages/web3-providers-http/src/index.js
uluhonolulu: @spacesailor24 BTW node-xhr2 has the same issue (https://github.com/pwnall/node-xhr2/blob/master/src/001-xml_http_request.coffee#L608), but at least it's developed more actively so there's a chance they fix it. 

As far as I see, the code looks pretty similar to xhr2-cookies, just rewritten in CoffeeScript.

Spacesai1or:
> > @uluhonolulu Did you test that these changes solve #3425? Would it be possible for you to provide an example project where an `xhr2-cookies` is caught and emitted from the changes you made? I'm just having a hard time understanding this enough to merge it
> 
> 
> 
> @spacesailor24 I tested it with my own project and I'm 100% sure it worked as intended, but it was half a year ago so I don't even remember how I invoked the error. Most probably I just tried to connect to a node that didn't exist.
> 
> 
> 
> I can provide a test project, but TBH I'm in a middle of moving to a new apt so it's going to be the next weekend at best.

@uluhonolulu okay! no worries on the test project, I'll try connecting to a node that doesn't exist. Thank you for your patience on getting this merged, apologies for it taking so long!

Spacesai1or:
I used the following code to produce the below errors (left terminal is `1.x`, right is with this patch):

```typescript
const Web3 = require('web3');

(async () => {
    const web3 = new Web3('http://127.0.0.1');
    try {
        const result = await web3.eth.getBlock(1);
    } catch (error) {
        console.log(error)
    }
})()
```

image

Spacesai1or:
My opinion on this is that we shouldn't even be using an XML HTTP package and should rely either on the native Node.js HTTP package or something like Fetch or Axios. However, as most of our tests make use of the `FakeHttpProvider`, and we don't have many tests actually executing HTTP requests, I don't think it would be wise to make changes to what's probably the most used protocol for Web3.js (unless it's for major bugs or security patches). Instead, I suggest waiting for the rewrite (`4.x`) to be completed, where this will be handled in a more modern way; if we're not merging this PR as is for the immediate benefit

valamidev:
> My opinion on this is that we shouldn't even be using an XML HTTP package and should rely either on the native Node.js HTTP package or something like Fetch or Axios. However, as most of our tests make use of the `FakeHttpProvider`, and we don't have many tests actually executing HTTP requests, I don't think it would be wise to make changes to what's probably the most used protocol for Web3.js (unless it's for major bugs or security patches). Instead, I suggest waiting for the rewrite (`4.x`) to be completed, where this will be handled in a more modern way; if we're not merging this PR as is for the immediate benefit

Well `xhr2-cookies` were last updated 4 years ago and clearly cause a lot of trouble with his error handling / 'hiding' . When I discovered that my `Invalid JSON RPC response: ""` error origin is coming from an expired CERT...
```
Error: certificate has expired
    at TLSSocket.onConnectSecure (_tls_wrap.js:1507:34)
    at TLSSocket.emit (events.js:376:20)
    at TLSSocket._finishInit (_tls_wrap.js:932:8)
    at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:706:12) {
  code: 'CERT_HAS_EXPIRED'
  ```
First it is a false error, because the response was not invalid it was throwed by an error, second I never expected that in case of HTTP error I won't get any notification.
I am happy to see that the issue is already have an ongoing PR, but IMO it should be in the mid priority stuffs or even high.

jdevcs:
Thanks for bringing this discussion in 1.x, our team will consider this request error handling improvement in web3.js 4.x.

jdevcs:
Closing this in favor of 4.x

> Thanks for bringing this discussion in 1.x, our team will consider this request error handling improvement in web3.js 4.x.

Issue #3816, opened 2020-12-14
CONNECTION ERROR: Couldn't connect to node on WS.


Wyctus:
This error started to happen to me regularly for a couple of weeks now. I can't tell the reason, seems to be totally random. Maybe it's a problem with Infura, since the error pops up on two different machines at the same time (they are running the same code).
My code is simply listening to some events on a some contracts. Sometimes it works for a day without any problem. But after it happens I can't reconnect for 30-60 minutes.
Looks pretty similar to #1646.

#### Expected behavior
The connection should stay open.

#### Actual behavior
It disconnects and can't reconnect for a long time.

#### Steps to reproduce the behavior

```javascript
const subscribe_to_contract_events = async contract => {    
    contract.events.SomeEventName().on("data", async event => {
           // process the event
    });
}

for (const contract of contracts) {
    try {
        await subscribe_to_contract_events(contract);
    } catch (err) {
        console.error(err);
    }
}
```
The  `contracts` contains around 400 elements. 

#### Logs
```
Error: CONNECTION ERROR: Couldn't connect to node on WS.
    at Object.ConnectionError (/home/wyctus/p1/node_modules/web3-core-helpers/lib/errors.js:66:23)
    at Object.InvalidConnection (/home/wyctus/p1/node_modules/web3-core-helpers/lib/errors.js:36:21)
    at /home/wyctus/p1/node_modules/web3-providers-ws/lib/index.js:161:37
    at Map.forEach (<anonymous>)
    at WebsocketProvider._onClose (/home/wyctus/p1/node_modules/web3-providers-ws/lib/index.js:160:28)
    at W3CWebSocket._dispatchEvent [as dispatchEvent] (/home/wyctus/p1/node_modules/yaeti/lib/EventTarget.js:115:12)
    at W3CWebSocket.onClose (/home/wyctus/p1/node_modules/websocket/lib/W3CWebSocket.js:228:10)
    at WebSocketConnection.<anonymous> (/home/wyctus/p1/node_modules/websocket/lib/W3CWebSocket.js:201:17)
    at WebSocketConnection.emit (node:events:329:20)
    at WebSocketConnection.handleSocketClose (/home/wyctus/p1/node_modules/websocket/lib/WebSocketConnection.js:389:14)
    at TLSSocket.emit (node:events:341:22)
    at node:net:655:12
    at TCP.done (node:_tls_wrap:564:7) {
  code: 1011,
  reason: 'Internal response error'
}
```

#### Environment

[...]

PR #4299, opened 2021-09-08
Add clientConfig to WS test

Spacesai1or:
We experience the issue mentioned in #3816 with our `e2e_windows` test suite, this aims to fix it

### DIFF
--- a/scripts/js/basic_usage.js
+++ b/scripts/js/basic_usage.js
@@ -57,7 +57,15 @@ async function main(){
   log('>>>>>>');
 
   // WebSockets
-  web3 = new Web3(process.env.INFURA_WSS);
+  web3 = new Web3(
+    process.env.INFURA_WSS,
+    {
+      clientConfig: {
+        maxReceivedFrameSize: 100000000,
+        maxReceivedMessageSize: 100000000,
+      }
+    }
+  );
   block = await getBlockWithRetry(web3);
   web3.currentProvider.disconnect();
   log(util.inspect(block));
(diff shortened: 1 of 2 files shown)


scripts/js/basic_usage.js
```diff
@@ -57,7 +57,15 @@ async function main(){
   log('>>>>>>');
 
   // WebSockets
-  web3 = new Web3(process.env.INFURA_WSS);
+  web3 = new Web3(
+    process.env.INFURA_WSS,
+    {
+      clientConfig: {
+        maxReceivedFrameSize: 100000000,
```
jdevcs, line 64: One message can fit into one or more(if message is fragmented) data frames as https://datatracker.ietf.org/doc/html/rfc6455#section-5.6 , 

The default values are `maxReceivedFrameSize` 1MiB and `maxReceivedMessageSize` 8MiB
https://github.com/theturtle32/WebSocket-Node/blob/a2cd3065167668a9685db0d5f9c4083e8a1839f0/docs/WebSocketClient.md#client-config-options 

Both `maxReceivedFrameSize` and `maxReceivedMessageSize` are made here around 95.3 MiB, which will hinder fragmentation and multiplexing https://datatracker.ietf.org/doc/html/rfc6455#section-5.4 in case of huge message.

Plus It will not make ws connection fail safe.

nazarhussain: As these are just the tests so be fine. But I agree with @jdevcs we should be using realistic values here.

nazarhussain (approved):
Apart form the point raised by @jdevcs LGTM.

luu-alex (approved):

Spacesai1or:
I'm going to close this until we start to see the error pop up again, so that we can test with values other than `100000000` to obtain a more reasonable value. Additionally, I'm not entirely convinced this is the solution

PR #4058, opened 2021-05-11
Fix unsubscribe and add a test

samlior:
## Description

Fix unsubscribe and add a test.
#3992 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules.
- [x] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [x] I ran `npm run test:unit` with success.
- [x] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code.
- [ ] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
- [x] I have tested my code on the live network.
- [ ] I have checked the Deploy Preview and it looks correct.
- [ ] I have updated the `CHANGELOG.md` file in the root folder.

PR #4300, opened 2021-09-08
Fix unsubscribe and add a test (#4058)

Spacesai1or:
Took over #4058 to update `CHANGELOG.md`


jdevcs (dismissed):

packages/web3-core-subscriptions/src/subscription.js
```diff
@@ -310,6 +309,7 @@ Subscription.prototype.subscribe = function() {
                     _this.emit('error', error);
                 }
             });
+            _this.emit('connected', result);
```
nazarhussain, line 312: I don't get it, why we have to move this emitter outside the block? The `connected` event just implies that client is connected and should not be related to `this.options.requestManager.send`. I think we should remove that emitter, connection state should be managed, some where else regardless of `.send` for the request manager.

nazarhussain, line 312: I don't get it, why we have to move this emitter outside the block? The `connected` event just implies that client is connected and should not be related to `this.options.requestManager.send`. I think we should remove that emitter, connection state should be managed, some where else regardless of `.send` for the request manager.

nazarhussain (changes requested):

packages/web3-core-subscriptions/src/subscription.js
Spacesai1or: This moves the emission to be after the subscription gets added via `_this.options.requestManager.addSubscription`, so that it is present when attempting to unsubscribe

Perhaps this could be refactored as you suggest, but I'd like to refrain from making big unnecessary changes in 1.x (to keep from causing new bugs), and just keep this in mind for the rewrite

nazarhussain (dismissed):

luu-alex (dismissed):

jdevcs (dismissed):

jdevcs (dismissed):

jdevcs (approved):

nazarhussain (approved):

PR #4303, opened 2021-09-09
Let toChecksumAddress throw for 'undefined'

sterlu:
## Description

`Web3.utils.toChecksumAddress` throws for invalid addresses, but for an unknown reason has a special case if input = `undefined`. This PR makes the behavior more predictable by making validation more strict - throwing if the input is not a string. 

## Type of change

Marking this as a **potentially breaking change** since it was previously explicitly allowed (https://github.com/ChainSafe/web3.js/commit/c6c344d7d892d3fa0f02db1f8f2aaebd800637e4), even though it does not seem like expected behavior?

### DIFF
--- a/packages/web3-utils/src/index.js
+++ b/packages/web3-utils/src/index.js
@@ -292,11 +292,8 @@ var toWei = function(number, unit) {
  * @return {String}
  */
 var toChecksumAddress = function (address) {
-    if (typeof address === 'undefined') return '';
-
-    if(!/^(0x)?[0-9a-f]{40}$/i.test(address))
+    if(typeof address !== 'string' || !/^(0x)?[0-9a-f]{40}$/i.test(address))
         throw new Error('Given address "'+ address +'" is not a valid Ethereum address.');
-
 
 
     address = address.toLowerCase().replace(/^0x/i,'');
@@ -350,7 +347,7 @@ var compareBlockNumbers = function(a, b) {
             return 1;
         } else {
             // b !== ("pending" OR "latest"), thus a > b
-            return -1 
+            return -1
         }
     } else if (a == "pending") {
         // b (== OR <) "latest", thus a > b
--- a/test/extend.js
+++ b/test/extend.js
@@ -77,12 +77,13 @@ describe('web3', function () {
                                 count++;
                         });
 
+                        const params = property.params === 2 ? ['0x11f4d0a3c12e86b4b5f39b213f7e19d048276dae', 'latest'] : [];
                         if(test.property) {
                             assert.isFunction(web3[test.property][property.name]);
-                            web3[test.property][property.name]();
+                            web3[test.property][property.name].apply(web3, params);
                         } else {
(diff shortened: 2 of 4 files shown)


luu-alex:
Thanks you for this PR! I'll look into this shortly

test/extend.js
```diff
@@ -77,12 +77,13 @@ describe('web3', function () {
                                 count++;
                         });
 
+                        const params = property.params === 2 ? ['0x11f4d0a3c12e86b4b5f39b213f7e19d048276dae', 'latest'] : [];
                         if(test.property) {
                             assert.isFunction(web3[test.property][property.name]);
-                            web3[test.property][property.name]();
+                            web3[test.property][property.name].apply(web3, params);
```
luu-alex, line 83: Why was this changed?

sterlu: If no params are passed to the function, it throws now (as it should after this PR) and the test fails

packages/web3-utils/src/index.js
```diff
@@ -292,13 +292,10 @@ var toWei = function(number, unit) {
  * @return {String}
  */
 var toChecksumAddress = function (address) {
-    if (typeof address === 'undefined') return '';
-
-    if(!/^(0x)?[0-9a-f]{40}$/i.test(address))
+    if(typeof address !== 'string' || !/^(0x)?[0-9a-f]{40}$/i.test(address))
```
nazarhussain, line 295: There is a function called `isAddress` I woudl prefeer using it instead of repeating the same regex here.

packages/web3-utils/src/index.js
```diff
@@ -350,7 +347,7 @@ var compareBlockNumbers = function(a, b) {
             return 1;
         } else {
             // b !== ("pending" OR "latest"), thus a > b
-            return -1 
+            return -1
```
nazarhussain, line 350: The change seems unncessary. Please format the code to avoid such changes.

test/formatters.inputAddressFormatter.js
```diff
@@ -24,7 +24,9 @@ var errorTests = [
     '0x11f4d0a3c12e86B4b5F39b213f7E19D048276DAe',
     {},
     [],
-    ''
+    '',
+    undefined,
```
nazarhussain, line 28: +1

nazarhussain (changes requested):
Thanks for your contribution. Here are my small feedback on it. Please check and resolve the conflits as well.

Spacesai1or:
This is technically a breaking change for any users expecting a return value of `''`

jdevcs:
> This is technically a breaking change for any users expecting a return value of `''`

@sterlu Thanks for your contributions. We are only doing necessary maintenance of 1.x and trying to not have breaking changes. Could you see and do contributions ( features, bugs, fixes ) in our [web3.js 4.x](https://github.com/ChainSafe/web3.js/tree/4.x) if you have time.

PR #4304, opened 2021-09-09
Not getting success callback/ receipt with walletconnect with web3.js 1.3.3 - Closes #3891

### LINKED ISSUE #3891: Not getting success callback/ receipt with walletconnect with web3.js 1.3.3
#### Expected behavior
- It should give success callback & receipt if used with WalletConnect

#### Actual behavior
- After 1.3.x update (tried with 1.3.3 & 1.3.4 rc.1), it is working upto below callback
```js
.on('transactionHash', function(hash){
    ...
}) 
```
on 'receipt' OR below one
```js
.then(function(receipt){
    // receipt can also be a new contract instance, when coming from a "contract.deploy({...}).send()"
});
```
 Never gets called but it is working fine with web3.js 1.2.11.

1. Connect wallet by scanning QR code with Metamask mobile or WalletConnect supported wallet app
2. Do any contract interaction Txn (using myMethod(123).send) 
3. See if its calling final success callback with receipt

#### Environment
WalletConnect web3 provider:  tried 1.3.1 & current latest 1.3.4
Node: 14.15.x
web3.js: 1.3.3 & 1.3.4.rc1
OS: Mac OS 11.1

If possible please release new 1.2.x version with Metamask hotfix, if it can't be fixed quickly with 1.3.x

nazarhussain:
## Description

We were using an assumption that if any provider contains `on` handler, it must also support `newBlockHeaders` event. Which was not the case for all providers and was not enforced by [EIP-1193](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md). So for the flow of waiting for the receipt we reverted back only to polling technique. 

Fixes: #3891 

## Type of change

## Checklist:


nazarhussain:
To come up with a solution which impact least on bandwidth usage and to providers which support `newBlockHeaders` I came up with this solution. 

1. If provider does not support `on` handler use polling anyway. 
2. If provider support `on` handler subscribe to `newBlockHeaders` event
3. Check after `blockHeaderTimeout` default to 10s, if block header event received or not 
4. If not, that's probably implies provider does not support that event and we fallback to polling. 

This approach will introduce no behaviour changes for providers which support `newBlockHeaders` events, and would introduce some wait time for others. 

@jdevcs  @spacesailor24 Let me know if that solutions seems feasible to you in reference to the comment posted on issue earlier https://github.com/ChainSafe/web3.js/issues/3891#issuecomment-916387467. If its ok to you guys then I will add few more tests as well. Also let me know if that change required to update some other places as well.

packages/web3-core-method/src/index.js
```diff
@@ -547,22 +548,35 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
...
+        // If provider do not support event subscription use polling
+        if(!this.requestManager.provider.on) {
+            return startInterval();            
         }
+
+        // Subscribe to new block headers to look for tx receipt
+        _ethereumCall.subscribe('newBlockHeaders', function (err, blockHeader, sub) {
+            blockHeaderArrived = true; 
+
+            if (err || !blockHeader) {
```
spacesailor24, line 566: I understand this is the original code, but kinda weird to revert to polling if the subscription returns an error, no?

test/method.buildCall.js
```diff
@@ -1117,6 +1117,67 @@ describe('lib/web3/method', function () {
                 gasPrice: '23435234234'
             })
         });
+
+        it('should fallback to polling if provider support `on` but `newBlockHeaders` does not arrive in `blockHeaderTimeout` seconds', function (done) {
```
Spacesai1or, line 1121: This doesn't test if `blockHeaderTimeout` works as expected if a response for `newBlockHeaders` isn't returned, correct?

packages/web3-core-method/src/index.js
Spacesai1or, line 566: If the provider doesn't support `newBlockHeaders`, does the callback never get called? I'm confused when `!blockHeader` would occur

packages/web3-core-method/src/index.js
```diff
@@ -56,6 +56,7 @@ var Method = function Method(options) {
     this.transactionBlockTimeout = options.transactionBlockTimeout || 50;
     this.transactionConfirmationBlocks = options.transactionConfirmationBlocks || 24;
     this.transactionPollingTimeout = options.transactionPollingTimeout || 750;
+    this.blockHeaderTimeout = options.blockHeaderTimeout || 10; // 10 seconds
```
Spacesai1or, line 59: Any issue with cutting this down to 5 seconds?

Spacesai1or (changes requested):
Couple questions, but otherwise changes look good

Could use at least one other test if my test related question is correct, also needs `CHANGELOG.md` update

packages/web3-core-method/src/index.js
nazarhussain: Personally I was not sure about this default value. I will update it to 5.

packages/web3-core-method/src/index.js
nazarhussain: If provider does not support `newBlockHeaders` then this callback never triggered. The check here is a safety check when `newBlockHeaders` event fired but with some error.

test/method.buildCall.js
nazarhussain: Based on the sequence of assertions provider will expect `eth_getTransactionReceipt` call from the code, that can only happen when `blockHeaderTimeout` triggers and fallback to polling..

nazarhussain:
I had tested it with @aksdevac [provided code repo](https://github.com/ChainSafe/web3.js/issues/3891#issuecomment-773148060) by linking `web3-eth` and `web3-core-method` packages and it worked. For the wallet connect provider is that it wait for 5 seconds for `newBlockHeaders` event and since it does not return then it fallback to polling and fetch the tx receipt and return full-filled promise.

packages/web3-core-method/src/index.js
```diff
@@ -56,6 +56,7 @@ var Method = function Method(options) {
     this.transactionBlockTimeout = options.transactionBlockTimeout || 50;
     this.transactionConfirmationBlocks = options.transactionConfirmationBlocks || 24;
     this.transactionPollingTimeout = options.transactionPollingTimeout || 750;
+    this.blockHeaderTimeout = options.blockHeaderTimeout || 5; // 10 seconds
```
Spacesai1or:
```suggestion
    this.blockHeaderTimeout = options.blockHeaderTimeout || 5; // 5 seconds
```

Spacesai1or: I'm just going to commit this since it's a comment change, and so I can provide my approval without it going stale

Spacesai1or (dismissed):
Thank you for updating the docs!

packages/web3-core-method/src/index.js
```diff
@@ -56,6 +56,7 @@ var Method = function Method(options) {
     this.transactionBlockTimeout = options.transactionBlockTimeout || 50;
     this.transactionConfirmationBlocks = options.transactionConfirmationBlocks || 24;
     this.transactionPollingTimeout = options.transactionPollingTimeout || 750;
+    this.blockHeaderTimeout = options.blockHeaderTimeout || 5; // 5 seconds
```
jdevcs: Good to make `this.blockHeaderTimeout` default > or 10 sec as normal mining time is above that https://etherscan.io/chart/blocktime

packages/web3-core-method/src/index.js
```diff
@@ -547,22 +548,35 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
...
-            _ethereumCall.subscribe('newBlockHeaders', function (err, blockHeader, sub) {
-                if (err || !blockHeader) {
-                    // fall back to polling
-                    startInterval();
-                } else {
-                    checkConfirmation(existingReceipt, false, err, blockHeader, sub);
-                }
-            });
+        // If provider do not support event subscription use polling
+        if(!this.requestManager.provider.on) {
```
jdevcs, line 558: Shouldnt we look specific event subscription support  in provider that we are interested in `newBlockHeaders` ,

packages/web3-eth/src/index.js
```diff
@@ -100,6 +100,7 @@ var Eth = function Eth() {
     var transactionBlockTimeout = 50;
     var transactionConfirmationBlocks = 24;
     var transactionPollingTimeout = 750;
+    var blockHeaderTimeout = 5;
```
jdevcs: same here should be 10 sec

imaksp:
it seems many files are not using es6 & es7 features, is there any plan to update it.?

packages/web3-core-method/src/index.js
nazarhussain: There is no way to check if a provider a particular event or not.

nazarhussain:
@aksdevac Yes for 1.x we are not refactoring the code a lot. As our focus is on the 4.x rewrite, which will cover best coding practices and Typescript language features.

xale76:
please help. I don't understand how to let it work. No events received after send calls

Spacesai1or (dismissed):

nazarhussain:
@xale76 Hey, Can you please explain your comment. Are you referring to this PR? or to the issue itself? And how did you subscribed to those events which you mentioned not receiving?

xale76:
@nazarhussain hello, and thanks. I'm using web3modal together with walletconnectprovider. My react app works perfectly from browser and metamask extension (connected by infura node on rinkeby). If I use walletconnectprovider, so I use QR code and I confirm actions on metamask app on my android mobile, after the confirmation (in example an "approve" action), my code remains stuck in "await". In example 

await mytoken.methods.approve(spender,amount).send({from:address});

I don't know what you mean with "subscribed" to events. I don't know, what events you're talking about, and I think it's not easy to list every events and register for them. Is there any other solution?

nazarhussain:
@xale76 This PR is about to fix that problem. Did you get the error even after compiling and linking this branch locally?

xale76:
@nazarhussain how to install this branch from npm?

jdevcs (approved):

Spacesai1or (approved):

nazarhussain:
@xale76 You can not install the branch from npm. But this issue is scheduled for next release soon it will be available for you to test.

hoductrong:
Any update for this issues, or which release contains this fix?, plz

jdevcs:
@hoductrong in 1.5.3 https://github.com/ChainSafe/web3.js/pull/4304/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR440

PR #4345, opened 2021-09-22
Update typescript configuration to make it more typesafe - Closes #4341

### LINKED ISSUE #4341: Update typescript configuration to make it more typesafe
### Motivation 

We need to configuration which is aligned with more type safe approach for all the packages. 

### Acceptance Criteria   

- Enforce no use of `any` type 
- Enforce no console usage 
- Enforce camel case usage 
- Align prettier configuration with the eslint 
- Create special lint configuration for Jest

nazarhussain:
## Description

Added TS linting and prettier support. 

Fixes #4341

## Type of change

## Checklist:


tools/eslint-config-web3-base/ts-jest.js
```diff
@@ -0,0 +1,31 @@
+module.exports = {
+	parser: '@typescript-eslint/parser',
+	plugins: ['@typescript-eslint', 'jest'],
+	extends: ['./ts', 'plugin:jest/recommended', 'plugin:jest/style'],
+	env: {
+		'jest/globals': true,
+	},
+	rules: {
```
Spacesai1or, line 8: I'm not sure about all these rules being turned off. I feel like we could turn most of them back on and toggle off the ones we need to while developing

tools/eslint-config-web3-base/ts.js
```diff
@@ -0,0 +1,86 @@
...
+		'airbnb-base',
+		'eslint:recommended',
+		'plugin:@typescript-eslint/recommended',
+		'plugin:@typescript-eslint/recommended-requiring-type-checking',
+		'prettier',
+		'plugin:import/errors',
+		'plugin:import/warnings',
+		'plugin:import/typescript',
+	],
+	rules: {
```
Spacesai1or, line 14: Same with these rules

nazarhussain: I made the rules for the src more strict.

tools/eslint-config-web3-base/ts-jest.js
nazarhussain: I updated the rules for the jest, but kept a bit flexibility which I learned with experience is required to to write Jest tests more quickly.

.eslintrc.js
```diff
@@ -0,0 +1,15 @@
...
+	root: true,
+	env: {
+		browser: true,
+		es2021: true,
+		node: true,
+	},
+	parserOptions: {
+		project: './tsconfig.json',
+		tsconfigRootDir: __dirname,
+		ecmaVersion: 12,
```
jdevcs: Line number 5 in this file: automatically sets the ecmaVersion parser option to 12, so this is duplicate.

If you want to keep both `es2021: true,` and `ecmaVersion: 12,` good to add comments so any future developer doesn't accidentally mis-configure

.eslintrc.js
```diff
@@ -0,0 +1,15 @@
...
+		sourceType: 'module',
+	},
+	extends: ['web3-base/ts'],
```
jdevcs, line 13: Is it referring to [this](https://github.com/ChainSafe/web3.js/tree/nh/4341-ts-setup/tools/eslint-config-web3-base) with correct package name ?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -23,9 +23,9 @@ Fixes #(issue)
...
 - [ ] My changes generate no new warnings.
 - [ ] Any dependent changes have been merged and published in downstream modules.
-- [ ] I ran `npm run dtslint` with success and extended the tests and types if necessary.
+- [ ] I ran `npm run lint` with success and extended the tests and types if necessary.
 - [ ] I ran `npm run test:unit` with success.
-- [ ] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code.
+- [ ] I ran `npm run test:coverage` and my test cases cover all the lines and branches of the added code.
 - [ ] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
 - [ ] I have tested my code on the live network.
 - [ ] I have checked the Deploy Preview and it looks correct.
```
jdevcs, line 31: As you touched this file so good to add following here instead of another PR for it:
`- [ ] I have linked Issue(s) with this PR in "Linked Issues" menu.`

.eslintrc.js
nazarhussain: Yes its the `eslint` convention when you extend it tries to prefix `eslint-config` automatically. So its actually extending `eslint-config-web3-base/ts`

jdevcs:
why there are two templates https://github.com/ChainSafe/web3.js/tree/nh/4341-ts-setup/tools/web3-packagetemplate and https://github.com/ChainSafe/web3.js/tree/nh/4341-ts-setup/templates . Could you clean this and create one template for all future packages.

nazarhussain:
@jdevcs  The `templates` directory contains source files for all configuration files which are symlinked to the packages. So if we have to change any configuration have to chance at one place. The `web3-packagetemplate` is a dummy package to just copy it with different name to create other web3 packages in `packages` directory`. This package also contains symlinks to the `templates` directory.

tools/eslint-config-web3-base/ts-jest.js
```diff
@@ -0,0 +1,31 @@
...
+		'jest/valid-title': ['error'],
+		'jest/no-conditional-expect': ['error'],
+		'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
+		'jest/consistent-test-it': ['error'],
+		'class-methods-use-this': ['error'],
+		'@typescript-eslint/no-magic-numbers': 'off',
```
Spacesai1or: I think this is a good rule to keep enabled

tools/eslint-config-web3-base/ts-jest.js
```diff
@@ -0,0 +1,31 @@
...
+		'@typescript-eslint/unbound-method': 'off',
+		'@typescript-eslint/no-require-imports': 'off',
```
Spacesai1or: Ideally we wouldn't have in `require` imports, yea?

tools/eslint-config-web3-base/ts-jest.js
```diff
@@ -0,0 +1,31 @@
...
+		'@typescript-eslint/no-explicit-any': 'off',
```
Spacesai1or, line 17: I think this is a good rule to keep enabled

tools/eslint-config-web3-base/ts-jest.js
```diff
@@ -0,0 +1,31 @@
...
+		'@typescript-eslint/no-unsafe-assignment': 'off',
+		'@typescript-eslint/no-unsafe-member-access': 'off',
+		'@typescript-eslint/no-unsafe-call': 'off',
+		'@typescript-eslint/no-unsafe-return': 'off',
```
Spacesai1or: I think these can all be enabled until we run into enough instances where writing code like this is necessary

tools/eslint-config-web3-base/ts.js
```diff
@@ -0,0 +1,86 @@
...
+		'class-methods-use-this': ['error'],
+		'no-unused-expressions': ['error'],
+		'no-continue': 'off',
+		'no-underscore-dangle': ['error'],
+		'import/prefer-default-export': 'off',
```
Spacesai1or, line 19: I think this is a good rule to keep enabled

tools/eslint-config-web3-base/ts-jest.js
nazarhussain: In tests we got a lot of cases to write expectations with numbers e.g. 

```js
expect(obj.value).toBeGreaterThan(10); 
```

This rule makes it difficult to write such expectations and its inconvenient for tests to always declare constants for such values and then write expectations.

tools/eslint-config-web3-base/ts-jest.js
nazarhussain: It's a tricky one was added because of Jest setup files. Will update the files to use imports.

tools/eslint-config-web3-base/ts-jest.js
nazarhussain: For tests its very redundant to declare types for all kind of custom mocked data so this makes it quick to write tests like. 

```js
expect((fetchMockObject() as any).myAttr).toEqual('223')
```

tools/eslint-config-web3-base/ts-jest.js
nazarhussain: I am switching these to `error` but these will become problem very quickly as we grow the tests.

tools/eslint-config-web3-base/ts.js
nazarhussain: I don't prefer enforcing default export in all the files, if its just one export. Named exports are much readable and stay consistent in the code. 


```

// file 1
export default class MyClass {}

// file 2
import AlphaClass from './my_class'
```

See in this example we lost the context and readability of the actual class reference.

tools/eslint-config-web3-base/ts-jest.js
Spacesai1or: Even in these cases I prefer to document what the number represents by giving is a constant/variable name. Additionally, it makes debugging/updating values easier when all test values are declared at the top of the test/test suite

tools/eslint-config-web3-base/ts-jest.js
Spacesai1or: I'd prefer if even tests were strongly typed (to aid with debugging tests), but I understand your point

Spacesai1or: I think `@typescript-eslint/no-require-imports` could still be enabled though

jdevcs (dismissed):

tools/eslint-config-web3-base/ts-jest.js
nazarhussain: I don't think its convenient for writing tests. Simple statements like `jest.setTimeout(5000);`  would require to declare a constant and use it. But I am changing it anyway, we can turn it off as soon we encounter that case.

jdevcs (approved):

Spacesai1or (approved):

PR #4377, opened 2021-09-24
Migrate all errors to web3-common as error constructors

### LINKED ISSUE #4359: Migrate all errors to web3-common as error constructors
### Description

Need to migrate all errors from `1.x` branch to `web3-common` package as **error constructors**. 

https://github.com/ChainSafe/web3.js/blob/40974e9fe9a8f9075fcbb4c002430c6480d83031/packages/web3-core-helpers/src/errors.js#L27-L136

These error constructors should follow following pattern and extend base class `Web3Error`. 

```js
class Web3Error extends Error {
	public name: string;
	public code = 'ERR_WEB3';
	public constructor(...args: unknown[]) {
		super(...unknown);
		this.name = this.constructor.name;
		Error.captureStackTrace(this, Web3Error);
	}
}

class  InvalidConnectionError extends Web3Error {
   constructor(host, event) {
      super(`CONNECTION ERROR: Couldn\'t connect to node ${host}.');  
      this.code = 'ERR_CONNECTION_ERR';
   } 
}

```

### Acceptance Criteria

- All errors should be migrated 
- All errors should have minimum unit tests 

### Additional Information

- Use string interpolation instead of concatenation

nazarhussain:
## Description

Fixes #4359

## Type of change

## Checklist:


packages/web3-common/README.md
```diff
@@ -22,27 +22,26 @@ yarn add web3-common
...
-| lint:fix | Uses `eslint` to check and fix any warnings
-| format | Uses `prettier` to format the code
-| test | Uses `jest` to run unit tests
-| test:integration | Uses `jest` to run tests under `/test/integration`
-| test:unit | Uses `jest` to run tests under `/test/unit`
-
+| Script           | Description                                         |
+| ---------------- | --------------------------------------------------- |
+| clean            | Uses `rimraf` to remove `dist/`                     |
+| build            | Uses `tsc` to build package and depedenent packages |
```
jdevcs: typo in depedenent so change to: `Uses `tsc` to build package and dependent packages`

packages/web3-common/src/errors.ts
```diff
@@ -0,0 +1,333 @@
...
+	ERR_CONTRACT_MISSING_DEPLOY_DATA,
+	ERR_CONTRACT_MISSING_ADDRESS,
+	ERR_CONTRACT_MISSING_FROM_ADDRESS,
+} from './constants';
+
+type ConnectionEvent = { code: string; reason: string };
+type Response<T = unknown> = { error?: { message?: string; data?: T } };
+type Receipt = Record<string, unknown>;
+
+export abstract class Web3Error extends Error {
```
jdevcs, line 36: will be nice to add one line doc comments for each class

jdevcs, line 36: are you referring here ts Error [interface error](https://github.com/microsoft/TypeScript/blob/main/lib/lib.es5.d.ts#L972) ? , if so then class should implement not extend.

tools/web3-packagetemplate/README.md
```diff
@@ -22,28 +22,26 @@ yarn add web3-packagetemplate
...
-| lint:fix | Uses `eslint` to check and fix any warnings
-| format | Uses `prettier` to format the code
-| test | Uses `jest` to run unit tests
-| test:integration | Uses `jest` to run tests under `/test/integration`
-| test:unit | Uses `jest` to run tests under `/test/unit`
-
+| Script           | Description                                         |
+| ---------------- | --------------------------------------------------- |
+| clean            | Uses `rimraf` to remove `dist/`                     |
+| build            | Uses `tsc` to build package and depedenent packages |
```
jdevcs: depedenent typo

jdevcs (changes requested):

packages/web3-common/src/errors.ts
nazarhussain: No its JS global `Error` object https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/error

nazarhussain: I follow the philosophy 

> The best comment is a good name for a method or class.

The names of the classes are very well described and clearly meaningful so don't think there is a need to add comments for it. 

https://belief-driven-design.com/your-comments-are-bad-74cee29b5a6/
https://refactoring.guru/smells/comments
https://fagnerbrack.com/code-comment-is-a-smell-4e8d78b0415b

jdevcs: in that case `public readonly name: string`; is already in [Error](https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/error#instance_properties)

nazarhussain: Yes its already in the `Error` object but not `readonly`. So why had to redeclare it here.

packages/web3-common/src/errors.ts
```diff
@@ -0,0 +1,333 @@
...
+
+	public toJSON() {
+		return { name: this.name, code: this.code, message: this.message };
+	}
+}
+
+export class InvalidNumberOfParamsError extends Web3Error {
+	public code = ERR_PARAM;
+
+	public constructor(public got: number, public expected: number, public method: string) {
```
Spacesai1or, line 54: I thought constructors were inherently `public`?

packages/web3-common/src/errors.ts
```diff
@@ -0,0 +1,333 @@
...
+	public code = ERR_CONTRACT_ABI_MISSING;
+
+	public constructor() {
+		super(
+			'You must provide the json interface of the contract when instantiating a contract object.',
+		);
+	}
+}
+
+export class ContractOnceRequiresCallbackError extends Web3Error {
```
Spacesai1or, line 279: Since we're not supporting callbacks, this isn't necessary?

packages/web3-common/src/errors.ts
```diff
@@ -0,0 +1,333 @@
...
+	public constructor() {
+		super('CONNECTION ERROR: Provider started to reconnect before the response got received!');
+		this.code = ERR_CONN_PENDING_REQUESTS;
+	}
+}
+
+export class InvalidProviderError extends Web3Error {
+	public code = ERR_INVALID_PROVIDER;
+
+	public constructor() {
```
Spacesai1or: I think this error refers to what should be called the `Client` based off the definitions defined [here](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#definitions). As we are constructing the `Provider` objects via class instantiation, I don't think we could have an invalid one? But perhaps there's a use case where a user provides their own instantiated `Provider`, and we throw this when their `Provider` doesn't match our defined interface?

Regardless, we're missing a `InvalidClientError` that should take the provided `clientUrl` as a constructor argument to be used in the error message

packages/web3-common/src/errors.ts
nazarhussain: We have to provide callback support on `on` and `once` or `addEventListener` functions. But we should avoid using composite support fo `Promise` and `callback` e.g. in `call()` and `send()` functions. 

I am not sure this error would be used or not, I just moved all existing error constructs. If not used then we can remove it afterwards.

packages/web3-common/src/errors.ts
nazarhussain: Yes the constructors can also be `protected` and in TS `private` as well. Where only class own's static members are allowed to call those private constructors. 

To improve the readability of the code our Eslint rules enforce to must specify the visibility..

packages/web3-common/src/errors.ts
nazarhussain: For the compiled code running in JS we can't enforce that every provider must be inherited by our abstract class interface. So would be nice to do also validate on runtime.

Spacesai1or: > we throw this when their Provider doesn't match our defined interface

I mean we would check that `request`, `setProvider`, etc. methods are available on whatever Provider they pass. If we did this, then we'd have separate errors for invalid providers and clients e.g.

```typescript
if (userProvider.request === undefined) throw InvalidProviderError()
```

My suggestion is to revert the `InvalidProviderError` to what you originally had since it exists in `1.x` (although I'm not entirely sure where we are going to use it), and create a new error called `InvalidClientError` that takes the `clientUrl`. `InvalidClientError` will be used when performing validation on the URL provided by the user when instantiating `Web3()` , `Web3.httpProvider()`, etc.

packages/web3-common/src/errors.ts
```diff
@@ -0,0 +1,333 @@
...
+	ERR_TX_REVERT_WITHOUT_REASON,
+	ERR_TX_OUT_OF_GAS,
+	ERR_CONTRACT_RESOLVER_MISSING,
+	ERR_CONTRACT_ABI_MISSING,
+	ERR_CONTRACT_REQUIRED_CALLBACK,
+	ERR_CONTRACT_EVENT_NOT_EXISTS,
+	ERR_CONTRACT_RESERVED_EVENT,
+	ERR_CONTRACT_MISSING_DEPLOY_DATA,
+	ERR_CONTRACT_MISSING_ADDRESS,
+	ERR_CONTRACT_MISSING_FROM_ADDRESS,
```
Spacesai1or, line 29: Shouldn't these errors be declared in their respective classes? If everything is pulling in `web3-common`, then we should limit what's in the package to what's being used by majority of the child packages

nazarhussain: These codes are `numerical constants` nice to keep these separate from the error construct implementation.

packages/web3-common/src/errors.ts
nazarhussain: I am also not sure where and how these errors are gonna used. So for now I just created what we already had. I suggest we move forward with it and as its a work in progress, if we ever need more particular error objects we can create.

Spacesai1or (approved):

jdevcs (approved):

PR #4390, opened 2021-09-26
Partially replace the use of eth-lib

Spacesai1or:
We're making an effort to replace the usage of one-off libraries with official [@ethereumjs](https://github.com/ethereumjs/ethereumjs-monorepo) libraries, and we're doing it one library at a time to reduce the chance of breaking changes

In this PR we're partially replacing the use of [eth-lib](https://github.com/MaiaVictor/eth-lib/tree/master/src), by updating the hashing methods, but keeping the methods used for account generation and signatures. We'll need to have further discussion around these since there's no replacement provided by `@ethereumjs`


packages/web3-utils/src/utils.js
```diff
@@ -490,10 +488,13 @@ var sha3 = function (value) {
...
     if (isHexStrict(value) && /^0x/i.test((value).toString())) {
-        value = hexToBytes(value);
+        value = ethereumjsUtil.toBuffer(value);
+    } else if (typeof value === 'string') {
+        // Assume value is an arbitrary string
+        value = Buffer.from(value, 'utf-8');
     }
 
-    var returnValue = Hash.keccak256(value); // jshint ignore:line
+    returnValue = ethereumjsUtil.bufferToHex(ethereumjsUtil.keccak256(value));
```
luu-alex: `var returnValue` , otherwise lgtm

Spacesai1or:
```suggestion
    var returnValue = ethereumjsUtil.bufferToHex(ethereumjsUtil.keccak256(value));
```

Spacesai1or: lol thanks, just fixed it

luu-alex (approved):

PR #4395, opened 2021-09-27
Create base provider and wallet interfaces 

### LINKED ISSUE #4361: Create `Web3BaseProvider` interface for web3-common
### Description

Define an **abstract class** named `Web3BaseProvider` which should be used to implement other HTTP, WS and other providers. 

That base provider should extract generic but consistent interfaces, some shared behaviour can also be implemented in that class. Some inspiration is given at: 

https://github.com/ChainSafe/web3.js/blob/40974e9fe9a8f9075fcbb4c002430c6480d83031/packages/web3-core-helpers/types/index.d.ts#L93

In addition to the class we need to provide a utility function to test any implementation of the providers.

### Acceptance Criteria

- The class `Web3BaseProvider` must be abstract 
- The class `Web3BaseProvider` should be exported from the package 
- There must exists a utility function to test the providers 
- Same utility function should cover the unit tests for that abstract class

### Additional Information

 For the test utility function the inspiration can be taken from.

https://github.com/libp2p/js-libp2p-interfaces/blob/2eff94a1bc67a741ada1c98bd57dc37c23acce3b/packages/compliance-tests/src/peer-discovery/index.js#L12-L90

### LINKED ISSUE #4362: Create `Web3BaseWallet` interface for web3-common
### Description

Define an **abstract class** named `Web3BaseWallet` which should be used to implement wallet provider for accounts. 

That provider should extract generic but consistent interfaces from existing implementation. It must not contain any implementation but just the interface. 

https://github.com/ChainSafe/web3.js/blob/40974e9fe9a8f9075fcbb4c002430c6480d83031/packages/web3-eth-accounts/src/index.js#L617

In addition to the class we need to provide a utility function to test any implementation of any wallet which is extended from that interface.

### Acceptance Criteria

- The class `Web3BaseWallet` must be abstract 
- The class `Web3BaseWallet` should be exported from the package 
- There must exists a utility function to test the wallet
- Same utility function should cover the unit tests for that abstract class

### Additional Information

 For that test function the inspiration can be taken from.

https://github.com/libp2p/js-libp2p-interfaces/blob/2eff94a1bc67a741ada1c98bd57dc37c23acce3b/packages/compliance-tests/src/peer-discovery/index.js#L12-L90

nazarhussain:
## Description

Created base provider and wallet interfaces 

Fixes #4361, #4362

## Type of change

## Checklist:


packages/web3-common/src/web3_base_provider.ts
```diff
@@ -0,0 +1,79 @@
...
+			.catch(e => callback(e, undefined));
+	}
+
+	abstract getStatus(): Web3BaseProviderStatus;
+	abstract supportsSubscriptions(): boolean;
+
+	// https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#request
+	abstract request<T = JSONRPCResult, T2 = unknown[]>(request: JSONRPCRequest<T2>): Promise<T>;
+
+	// https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#events
```
jdevcs, line 52: Its not necessary to enforce EIP1193 to all providers who implement via Web3BaseProvider  , so it should be removed. Discussed this in old 4.x with @spacesailor24

packages/web3-common/src/web3_base_provider.ts
```diff
@@ -0,0 +1,79 @@
...
+		type: 'accountsChanged',
+		callback: Web3BaseProviderCallback<{
+			readonly [key: string]: unknown;
+			readonly accountsChanged: string[];
+		}>,
+	): void;
+	abstract removeListener(type: string, callback: Web3BaseProviderCallback): void;
+
+	abstract once?<T = JSONRPCResult>(type: string, callback: Web3BaseProviderCallback<T>): void;
+	abstract removeAllListeners?(type: string): void;
```
jdevcs, line 74: As mentioned in related issue: https://github.com/ChainSafe/web3.js/issues/4361 the inspiration is taken from https://github.com/ChainSafe/web3.js/blob/40974e9fe9a8f9075fcbb4c002430c6480d83031/packages/web3-core-helpers/types/index.d.ts#L93  This WebsocketProviderBase is for websocket only and not for generic so

 in some providers like HTTP we dnt need to enforce removing listeners but its for WS.

packages/web3-common/src/types.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export interface JSONRPCResponseWithResult<T = JSONRPCResult> {
+	readonly id: JSONRPCId;
+	readonly jsonrpc: JSONRPCIdentifier;
+	readonly error?: never;
+	readonly result: T;
+}
+
+export type JSONRPCResponse<T = JSONRPCResult> =
```
jdevcs: it should `export type JSONRPCResponse<T = JSONRPCResult> =  JSONRPCResponseWithError<T> | JSONRPCResponseWithResult<T>;`

jdevcs (changes requested):

packages/web3-common/src/web3_base_provider.ts
nazarhussain: Yes its not necessary to enforce EIP1193, but its necessary to have one standard. This abstract class is not fully copy of EIP1193, it had its own extra behaviour as well. So you can consider it a `Web3BaseProvider` which has similarities with EIP1193 as its a common accepted standard.

packages/web3-common/src/web3_base_provider.ts
nazarhussain: Yes these are optional interfaces `once?` and `removeAllListeners?`. Key trick is `?` sign at the end of function name.

packages/web3-common/src/types.ts
nazarhussain: Yes its exactly same but formatted over multi-line syntax.

packages/web3-common/src/web3_base_provider.ts
jdevcs: yes these are optional, but I think lets define common specs here for all provides and not other way

nazarhussain: I think its common for all providers, if they want to implement these interfaces or not. It's upto each provider and the base interface is not enforcing it. Having optional interfaces in base class helps in passing base interface instead of an actual type.

packages/web3-common/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './types';
+export * from './web3_base_provider';
```
Spacesai1or, line 4: The repo already has spacing denoted using `-` (kebab casing), lets not mix in snake casing as well

packages/web3-common/src/types.ts
```diff
@@ -0,0 +1,37 @@
+export type JSONRPCId = string | number | null;
+export type JSONRPCResult = string | number | boolean | Record<string, unknown>;
+export type JSONRPCIdentifier = '2.0' | '1.0';
+
+export interface JSONRPCError<T = JSONRPCResult> {
```
Spacesai1or: If `JSONRPCResult` is only going to be used once in the type, why not just `readonly data?: JSONRPCResult` (this applies for all other instances)?

packages/web3-common/src/types.ts
```diff
@@ -0,0 +1,37 @@
+export type JSONRPCId = string | number | null;
```
Spacesai1or: Nit picky, but `JSONRPC` should be `JsonRpc` for both readability and to match Lodestar

packages/web3-common/src/web3_base_provider.ts
```diff
@@ -0,0 +1,79 @@
...
+export const JSONRPC_ERR_UNAUTHORIZED = 4001;
+export const JSONRPC_ERR_UNSUPPORTED_METHOD = 4200;
+export const JSONRPC_ERR_DISCONNECTED = 4900;
+export const JSONRPC_ERR_CHAIN_DISCONNECTED = 4901;
+
+// Provider interface compatible with EIP-1193
+// https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md
+export abstract class Web3BaseProvider {
+	// TODO: For legacy support, should be deprecated and removed in favor of `request`
+	public send<T = JSONRPCResult, T2 = unknown[]>(
```
Spacesai1or: And even here where `JSONRPCResult` and `unknown[]` are used multiple times, I believe this to decrease readability

packages/web3-common/src/web3_base_provider.ts
```diff
@@ -0,0 +1,79 @@
...
+			readonly chainId: string;
+		}>,
+	): void;
+	abstract on(
+		type: 'accountsChanged',
+		callback: Web3BaseProviderCallback<{
+			readonly [key: string]: unknown;
+			readonly accountsChanged: string[];
+		}>,
+	): void;
```
Spacesai1or: `.on` should only exist for web sockets and EIP1193 providers, no?

packages/web3-common/src/web3_base_provider.ts
```diff
@@ -0,0 +1,79 @@
...
+	abstract removeListener(type: string, callback: Web3BaseProviderCallback): void;
+
+	abstract once?<T = JSONRPCResult>(type: string, callback: Web3BaseProviderCallback<T>): void;
+	abstract removeAllListeners?(type: string): void;
+	abstract connect(): void;
+	abstract disconnect(code: number, reason: string): void;
+	abstract reset(): void;
+	abstract reconnect(): void;
```
Spacesai1or, line 78: I think these only make sense for web sockets. They aren't really useful for HTTP and are specified in EIP1193

packages/web3-common/src/web3_base_wallet.ts
```diff
@@ -0,0 +1,25 @@
...
+	remove: (addressOrIndex: string | number) => boolean;
+	clear: () => Web3BaseWallet<T>;
+	encrypt: (password: string, options?: Record<string, unknown>) => T[];
+	decrypt: (
+		encryptedWallet: Web3EncryptedWallet[],
+		password: string,
+		options?: Record<string, unknown>,
+	) => Web3BaseWallet<T>;
+	save: (password: string, keyName?: string) => boolean | never;
+	load: (password: string, keyName?: string) => Web3BaseWallet<T>;
```
Spacesai1or, line 24: `keyName` would be the name of the file? In which case I would suggest renaming to `fileName` to be explicit. Also, if it does refer to the file name, why are they optional?

packages/web3-common/src/types.ts
nazarhussain: `JSONRPCResult` represents basic probable types, but user may know exact `type` of `data` of the a particular request. So with this approach user can pass the type.

packages/web3-common/src/web3_base_provider.ts
nazarhussain: All the generics passed down for the extensibility. If user knows exact type while calling `send` he can just pass the type otherwise the default types will be used.

packages/web3-common/src/web3_base_provider.ts
nazarhussain: No on should exists on abstract class as its very basic behaviour of any provider. If some provider implementing this abstract class does not provide this behaviour it should still implement and throw `NotSupportedError`.

packages/web3-common/src/web3_base_provider.ts
nazarhussain: As I highlighted it couple of times earlier in discussion, EIP1193 standard or our provider abstract class both present a generic behaviour across all the providers regardless of any technology. 

As the EIP1193 mentions it as:

```
A JavaScript Ethereum Provider API for consistency across clients and applications.
```
So such protocol standards does not think of a particular technology implementation. 

e.g. If `HTTP` provider does not have ability to `reconnect`, it can still implement an empty function to make sure all the polymorphic interfaces using that abstract class works consistently.

packages/web3-common/src/web3_base_wallet.ts
nazarhussain: Its the default behaviour of the current implementation, that parameter is optional. And each wallet implementation can have some default `keyName | fileName` to use internally. 

And additionally `keyName` seems much better to me as for local storage wallet which is our current use case this makes more sense.

packages/web3-common/src/index.ts
nazarhussain: I am not sure of using `kebab` case for the file names. Due to common JS community convention of package names we had it for the package directory names. 

In `loadstar` its `kebab` case for the package directory names and then `camelCase` for the file names. I think we should keep `snake` case for the files name that's much better in readability and support across platforms.

Spacesai1or (dismissed):

jdevcs (dismissed):

jdevcs:
Approved. IMO, interfaces / base classes are supposed to define and enforce common specifications and not moving up individual internals specific methods of WS, 1193, ..etc and then making these `NotSupportedErrors` in other providers. 
Current approach will work but is not optimal design.

jdevcs (approved):

Spacesai1or (approved):

PR #4397, opened 2021-09-28
doc: fix example in web3.eth.personal

robertu7:


Spacesai1or:
Thank you for finding and fixing this!

PR #4405, opened 2021-09-29
Migrate data type conversion utils functions from 1.x

### LINKED ISSUE #4350: Migrate data type conversion utils functions from 1.x
### Description

Migrate the following functions form `web3-utils` 1.x branch. 

- bytesToHex
- hexToBytes
- hexToNumberString
- hexToNumber
- numberToHex
- toDecimal
- fromDecimal
- toHex
- toUtf8
- fromUtf8
- toAscii
- fromAscii
- asciiToHex
- hexToAscii
- hexToUtf8
- utf8ToHex
- hexToString
- stringToHex
- toWei
- fromWei
- toNumber
- toChecksumAddress

### Acceptance Criteria

- All the functions mentioned above should migrated 
- All the functions must be covered with their unit tests 

### Additional Information

- Don't use `BigNumber` where needed instead use native `BigInt`

nazarhussain:
## Description

Create conversion functions for `web3-utils` 

Fixes #4350

## Type of change

## Checklist:


packages/web3-utils/src/big_decimal.ts
```diff
@@ -0,0 +1,51 @@
...
+	private readonly decimals: number;
+
+	public constructor(value: bigint | string, decimals?: number) {
+		if (typeof value === 'string' && decimals) {
+			throw new Error('Can not specify decimal with string value');
+		}
+
+		if (typeof value === 'string') {
+			try {
+				const [integer, fraction] = String(value).split('.').concat('');
```
Spacesai1or: What is the purpose of `.concat('')`?

```typescript
'4.20'.split('.') // > (2) ['4', '20']
'4.20'.split('.').concat('') // (3) ['4', '20', '']
```

packages/web3-utils/src/big_decimal.ts
```diff
@@ -0,0 +1,51 @@
...
+				this.decimals = decimals ?? fraction.length;
+				this.bigint = BigInt(`${integer}${fraction.padEnd(this.decimals, '0')}`);
+			} catch {
+				throw new Error(`Invalid value given "${value}". Error: not a valid integer.`);
```
Spacesai1or: We should avoid mixing inline errors with the use of `errors.ts`, no?

packages/web3-utils/test/fixtures/converters.ts
```diff
@@ -0,0 +1,234 @@
...
+export const fromWeiInvalidData: [[any, any], string][] = [
+	[['123.34', 'kwei'], 'Invalid value given "123.34". Error: not a valid integer.'],
+	[[null, 'kwei'], 'Invalid value given "null". Error: not a valid integer.'],
+	[[undefined, 'kwei'], 'Invalid value given "undefined". Error: not a valid integer.'],
+	[[{}, 'kwei'], 'Invalid value given "[object Object]". Error: not a valid integer.'],
+	[['data', 'kwei'], 'Invalid value given "data". Error: not a valid integer.'],
+	[['1234', 'uwei'], 'Invalid value given "uwei". Error: invalid unit.'],
+];
+
+export const toWeiInvalidData: [[any, any], string][] = [
```
spacesailor24, line 228: Missing `toWeiValidData` tests

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+	grand: expo10(21),
+	mether: expo10(24),
+	gether: expo10(27),
+	tether: expo10(30),
+};
+
+const errorMessage = (value: unknown, error: string) =>
+	`Invalid value given "${String(value)}". Error: ${error}.`;
+
+export type EtherUnits = keyof typeof ethUnitMap;
```
Spacesai1or, line 54: What's the benefit of doing it this way as opposed to an enum?

packages/web3-utils/src/types.ts
```diff
@@ -0,0 +1,5 @@
+export type HexString = string;
+export type ValueTypes = 'address' | 'bool' | 'string' | 'int256' | 'uint256' | 'bytes' | 'bigint';
+export type Address = string;
+export type Bytes = Buffer | Uint8Array | ArrayBuffer | HexString;
+export type Numbers = number | bigint | string | HexString;
```
Spacesai1or, line 5: Wouldn't we want to support the use of `BigDecimal`?

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+		return value;
+	}
+
+	return hexToNumber(numberToHex(value));
+};
+
+/**
+ * Takes a number of wei and converts it to any other ether unit.
+ */
+export const fromWei = (number: Numbers, unit: EtherUnits): string => {
```
Spacesai1or, line 270: What are your thought around returning a `BigDecimal` instead of a `string` here and for `toWei`?

packages/web3-utils/src/big_decimal.ts
```diff
@@ -0,0 +1,51 @@
+export class BigDecimal {
+	private readonly bigint: bigint;
+	private readonly decimals: number;
+
+	public constructor(value: bigint | string, decimals?: number) {
```
Spacesai1or: Couldn't `value` (and even `decimals`) be of type `Numbers`?

packages/web3-utils/src/big_decimal.ts
```diff
@@ -0,0 +1,51 @@
...
+		if (typeof value === 'string' && decimals) {
+			throw new Error('Can not specify decimal with string value');
```
Spacesai1or: If we're padding the number with zeros here:

```typescript
this.bigint = BigInt(`${integer}${fraction.padEnd(this.decimals, '0')}`);
```

why can't we support strings and decimals? e.g.:

```typescript
new BigDecimal('4.2', 5) // '420000'
```

Spacesai1or (changes requested):
There aren't any tests for `BigDecimal`

packages/web3-utils/src/big_decimal.ts
```diff
@@ -0,0 +1,51 @@
...
+		if (typeof value === 'string') {
+			try {
+				const [integer, fraction] = String(value).split('.').concat('');
+				this.decimals = decimals ?? fraction.length;
+				this.bigint = BigInt(`${integer}${fraction.padEnd(this.decimals, '0')}`);
+			} catch {
+				throw new Error(`Invalid value given "${value}". Error: not a valid integer.`);
+			}
+		} else {
+			this.bigint = BigInt(value);
```
Spacesai1or: If `value` can only be `bigint | string` then it's already a `bigint` here and doesn't need to be re-casted

Spacesai1or: Unless this is handling the case where a user passes `42n` as a `bigint`?

packages/web3-utils/src/big_decimal.ts
```diff
@@ -0,0 +1,51 @@
...
+
+	public constructor(value: bigint | string, decimals?: number) {
+		if (typeof value === 'string' && decimals) {
+			throw new Error('Can not specify decimal with string value');
+		}
+
+		if (typeof value === 'string') {
+			try {
+				const [integer, fraction] = String(value).split('.').concat('');
+				this.decimals = decimals ?? fraction.length;
```
Spacesai1or: The error defined above:

```typescript
throw new Error('Can not specify decimal with string value');
```

means that `decimals` will never be defined

packages/web3-utils/src/big_decimal.ts
Spacesai1or: If we choose to support this, the padding logic:

```typescript
this.bigint = BigInt(`${integer}${fraction.padEnd(this.decimals, '0')}`);
```

needs to be updated to account for the number of decimals already present. So it should know that `'4.2'` already has a one decimal and the desired amount is `5`, so add `4` zeros. Currently it would add `5`

packages/web3-utils/src/big_decimal.ts
Spacesai1or: Also this error doesn't make sense since we're splitting on the `.` here:

```typescript
const [integer, fraction] = String(value).split('.').concat('');
```

I think the error message should say `...not a valid number.` as opposed to `integer`

packages/web3-utils/src/hash.ts
```diff
@@ -0,0 +1,3 @@
+// TODO: Implement later
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const sha3 = (data: string): string => data;
```
spacesailor24, line 3: Implement `sha3` in `web3-utils`

packages/web3-utils/test/unit/converters.test.ts
```diff
@@ -0,0 +1,210 @@
...
+	hexToUtf8ValidData,
+	numberToHexInvalidData,
+	numberToHexValidData,
+	toHexValidData,
+	toWeiInvalidData,
+	utf8ToHexInvalidData,
+	utf8ToHexValidData,
+} from '../fixtures/converters';
+
+describe('converters', () => {
```
Spacesai1or, line 46: There aren't tests for

- toDecimal
- fromDecimal
- toUtf8
- fromUtf8
- toAscii
- fromAscii
- hexToString
- stringToHex
- toChecksumAddress
- hexToNumberString

For the above that are simply aliases for other functions, there should still be tests to serve as sanity checks (remember we're maintaining this for the foreseeable future, but there will be developers after us that won't know the assumptions we make today)

spacesailor24:
Pointing out the missing util methods from [our docs](https://web3js.readthedocs.io/en/v1.5.2/web3-utils.html)

- isBN
- isBigNumber
- soliditySha3
- isHex
- isHexStrict
- isAddress
- checkAddressChecksum
- toBN
- padLeft
- padRight
- toTwosComplement

We probably don't need to support `isBN`, `isBigNumber`, and `toBN`, but we should make a decision now if we're going to support them otherwise we should make a note that we're dropping support for these methods in `4.x` (so we remember when making the migration guide)

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+	milli: expo10(15),
+	ether: expo10(18),
+	kether: expo10(21),
+	grand: expo10(21),
+	mether: expo10(24),
+	gether: expo10(27),
+	tether: expo10(30),
+};
+
+const errorMessage = (value: unknown, error: string) =>
```
Spacesai1or: I would think this should be defined in `errors.ts` to be used by all the inline errors below

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+ */
+export const hexToNumber = (value: HexString): bigint | number => {
+	if (!isHexStrict(value)) {
+		throw new Error(errorMessage(value, 'not valid hex string'));
+	}
+
+	const [negative, hexValue] = value.startsWith('-') ? [true, value.substr(1)] : [false, value];
+	const num = BigInt(hexValue);
+
+	if (num > BigInt(Number.MAX_SAFE_INTEGER) || num < BigInt(Number.MIN_SAFE_INTEGER)) {
```
Spacesai1or:
```suggestion
	if (num > BigInt(Number.MAX_SAFE_INTEGER)) {
```

`num < BigInt(Number.MIN_SAFE_INTEGER)` is never going to happen since the `-` is stripped off by `value.substr(1)`

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * Convert a hex string to a byte array
+ */
+export const hexToBytes = (bytes: HexString): Buffer => bytesToBuffer(bytes);
+
+/**
+ * Converts value to it's number representation
+ */
+export const hexToNumber = (value: HexString): bigint | number => {
```
Spacesai1or, line 91: Shouldn't we support non string hexadecimals as well?

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+/**
+ * Converts value to it's hex representation
+ */
+export const numberToHex = (value: Numbers): HexString => {
+	try {
+		if (typeof value === 'number' && !Number.isFinite(value)) {
+			throw new Error(errorMessage(value, 'contains invalid integer values'));
+		}
+
+		if ((typeof value === 'number' || typeof value === 'bigint') && BigInt(value) < BigInt(0)) {
```
Spacesai1or:
```suggestion
		if ((typeof value === 'number' || typeof value === 'bigint') && value < 0) {
```

Casting isn't necessary, if it's a number it's
`Number.MIN_SAFE_INTEGER <= value <= Number.MAX_SAFE_INTEGER` or else the user wouldn't be able to pass it

```typescript
-100 < 0 // true
BigInt(-100) < 0 // true
BigInt(Number.MIN_SAFE_INTEGER - 1) < 0 // true
```

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+			return `-0x${value.toString(16).substr(1)}`;
+		}
+
+		if (
+			(typeof value === 'number' || typeof value === 'bigint') &&
+			BigInt(value) >= BigInt(0)
```
Spacesai1or: Same as above

```suggestion
			value >= 0
```

```typescript
BigInt(Number.MAX_SAFE_INTEGER + 100) >= 0 // true
```

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+		) {
+			return `0x${value.toString(16)}`;
+		}
+
+		if (typeof value === 'string' && isHexStrict(value)) {
+			return numberToHex(hexToNumber(value));
+		}
```
Spacesai1or: Can't we just return here? If `value` passes `isHexStrict` than it's already a prefixed hex string

packages/web3-utils/test/fixtures/converters.ts
```diff
@@ -0,0 +1,234 @@
...
+	[1, '0x1'],
+	[255, '0xff'],
+	[256, '0x100'],
+	[54, '0x36'],
+	[BigInt(12), '0xc'],
+	['768', '0x300'],
+	['-768', '-0x300'],
+	[-255, '-0xff'],
+	['0xFF0', '0xff0'],
+	['-0xa0', '-0xa0'],
```
Spacesai1or, line 70: Add hex number test cases

```suggestion
	['-0xa0', '-0xa0'],
	[0xff, '0xff'],
	[-0xff, '-0xff'],
```

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+export const fromUtf8 = utf8ToHex;
+/**
+ * @alias `utf8ToHex`
+ */
+export const stringToHex = utf8ToHex;
+
+/**
+ * Should be called to get utf8 from it's hex representation
+ */
+export const hexToUtf8 = (str: HexString): string => bytesToBuffer(str).toString('utf8');
```
Spacesai1or, line 171: Could also support non string hex, just not in it's current form

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * @alias `asciiToHex`
+ */
+export const fromAscii = utf8ToHex;
+
+/**
+ * Should be called to get ascii from it's hex representation
+ */
+export const hexToAscii = (str: HexString): string => bytesToBuffer(str).toString('ascii');
```
Spacesai1or, line 200: Could also support non string hex, just not in it's current form

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * Auto converts any given value into it's hex representation.
+ */
+export const toHex = (
+	value: Numbers | Bytes | Address | boolean,
+	returnType?: boolean,
+): HexString | ValueTypes => {
+	if (typeof value === 'string' && isAddress(value)) {
+		return returnType ? 'address' : `0x${value.toLowerCase().replace(/^0x/i, '')}`;
```
Spacesai1or, line 215: `.replace(/^0x/i, '')` could probably get removed after `isAddress` is implemented

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+};
+
+/**
+ * Auto converts any given value into it's hex representation,
+ * then converts hex to number.
+ */
+export const toNumber = (value: Numbers): number | bigint => {
+	if (typeof value === 'number' || typeof value === 'bigint') {
+		return value;
+	}
```
Spacesai1or: What do you think about this?

```suggestion
	if (typeof value === 'number') return value;
	
        if (typeof value === 'bigint') {
                return value >= Number.MIN_SAFE_INTEGER || value <= Number.MAX_SAFE_INTEGER ?
                        Number(value) : value;
	}
```

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,340 @@
...
+	if (typeof value === 'bigint') {
+		return returnType ? 'bigint' : numberToHex(value);
+	}
+
+	if (typeof value === 'string') {
+		if (value.startsWith('-0x') || value.startsWith('-0X')) {
+			return returnType ? 'int256' : numberToHex(value);
+		}
+
+		if (value.startsWith('0x') || value.startsWith('0X')) {
```
luu-alex: should `value` also be checked with `isHexStrict`?

packages/web3-utils/src/converters.ts
luu-alex: `if (num > BigInt(Number.MAX_SAFE_INTEGER) || num < BigInt(Number.MIN_SAFE_INTEGER))`

Can be:

`if (num > Number.MAX_SAFE_INTEGER || num < Number.MIN_SAFE_INTEGER)`

luu-alex:
Since we are replacing `bn.js` with just using the type `bigint` we should probably make a note. 

As it looks like we are removing the `toBN` function

packages/web3-utils/src/converters.ts
nazarhussain: In this condition the variable `num` is a `bigint` so why we have to convert `Number. MAX_SAFE_INTEGER` and `Number. MIN_SAFE_INTEGER` to the biginteger first before comparing.

nazarhussain:
@luu-alex Yes we are replacing `bn.js` with native `bigint`. So any function related to it will also be removed.

packages/web3-utils/src/converters.ts
luu-alex: From [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt), A Number value and a BigInt value may be compared as usual.

Also works in console if i try 
`BigInt(3) < Number.MAX_SAFE_INTEGER //True`

nazarhussain: I tend not to mixup values of numbers and bigint. As even the docs says, notice the `may be`.

> A Number value and a BigInt value may be compared as usual:

But in this case I know the range is safer for the number so we can do it.

packages/web3-utils/src/converters.ts
nazarhussain: With the enum values, user has to pass the enum. With this approach user can pass matching `string` values. Both approaches are type safe, but this one seems more friendly for user.

packages/web3-utils/src/converters.ts
nazarhussain: I think we should not expose `BigDecimal` outside library. It should remain an internal behaviour.

packages/web3-utils/src/big_decimal.ts
nazarhussain: No as per the logic for divide/multiply. Value can be either `bigint` and `decimal` or just `string` value.

packages/web3-utils/src/big_decimal.ts
nazarhussain: If we support `string` with `decimals` it will be confusing.

packages/web3-utils/test/unit/converters.test.ts
nazarhussain: Sure I will add these tests.

packages/web3-utils/src/converters.ts
nazarhussain: I don't think so. The name suggest itself `hexToNumber` is clear indication that one must specify a hex string. And every other function involved in `hex` is treating it as string.

packages/web3-utils/src/converters.ts
nazarhussain: Yes we can this check make sure the sanity of the input. It its actually a number.

packages/web3-utils/src/big_decimal.ts
Spacesai1or: The purpose was to avoid have a `null` value for `fraction`. A comment is going to be added to clarify intention

packages/web3-utils/src/types.ts
Spacesai1or: `BigDecimal` is not going to be exposed to the user

packages/web3-utils/src/converters.ts
nazarhussain: Not sure, I copied it exactly from old implementation.

packages/web3-utils/src/big_decimal.ts
Spacesai1or: I think it would be a nice to have for users, but since this isn't going to be exposed publicly, than it's fine not to support it

luu-alex (approved):

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,375 @@
...
+		throw new Error(errorInvalidUnit(unit));
+	}
+
+	// count number of zeros in denomination
+	// 1000000 -> 6
+	const numberOfZerosInDenomination = denomination.toString().length - 1;
+
+	if (numberOfZerosInDenomination <= 0) {
+		return number.toString();
+	}
```
Spacesai1or: I don't understand what's happening here, could you clarify with a comment?

packages/web3-utils/src/converters.ts
```diff
@@ -0,0 +1,375 @@
...
+
+	// value in wei would always be integer
+	// 13456789, 1234
+	const value = String(toNumber(number));
```
Spacesai1or: Is all the casting necessary? Above you just do `number.toString()`

packages/web3-utils/src/converters.ts
Spacesai1or: Oh I think I understand, this handles if the provided `unit` is `wei` or `noether`?

packages/web3-utils/src/errors.ts
```diff
@@ -0,0 +1,22 @@
+export const errorMessage = (value: unknown, error: string) =>
```
Spacesai1or: I think errors should remain uniformed to how they were done [here](https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-common/src/errors.ts)

packages/web3-utils/test/fixtures/converters.ts
```diff
@@ -0,0 +1,241 @@
...
+	[undefined, 'Invalid value given "undefined". Error: can not parse as byte data.'],
+	[{}, 'Invalid value given "[object Object]". Error: can not parse as byte data.'],
+];
+
+export const numberToHexValidData: [Numbers, HexString][] = [
+	[1, '0x1'],
+	[255, '0xff'],
+	[256, '0x100'],
+	[54, '0x36'],
+	[BigInt(12), '0xc'],
```
Spacesai1or, line 64:
```suggestion
	[BigInt(12), '0xc'],
	[12n, '0xc'],
```

Add test case for alternative syntax for BigInt

packages/web3-utils/test/fixtures/converters.ts
```diff
@@ -0,0 +1,241 @@
...
+	['I ♥ data', '0x4920e299a52064617461'],
+	['I \u0000 data', '0x4920002064617461'],
+	['\u0000 null suffix', '0x206e756c6c20737566666978'],
+	['null prefix\u0000', '0x6e756c6c20707265666978'],
+	['\u0000', '0x'],
+];
+
+export const utf8ToHexInvalidData: [any, string][] = [
+	[12, 'Invalid value given "12". Error: not a valid string.'],
+	[BigInt(12), 'Invalid value given "12". Error: not a valid string.'],
```
Spacesai1or, line 121:
```suggestion
	[BigInt(12), 'Invalid value given "12". Error: not a valid string.'],
	[12n, 'Invalid value given "12". Error: not a valid string.'],
```

Add test case for alternative syntax for BigInt

packages/web3-utils/test/fixtures/converters.ts
```diff
@@ -0,0 +1,241 @@
...
+	['0x6e756c6c2070726566697800', 'null prefix\u0000'],
+	['0x00', '\u0000'],
+	['0x', ''],
+];
+
+export const toHexValidData: [Numbers | Bytes | Address | boolean, [HexString, ValueTypes]][] = [
+	[1, ['0x1', 'uint256']],
+	[255, ['0xff', 'uint256']],
+	[256, ['0x100', 'uint256']],
+	[BigInt(12), ['0xc', 'bigint']],
```
Spacesai1or, line 178:
```suggestion
	[BigInt(12), ['0xc', 'bigint']],
	[12n, ['0xc', 'bigint']],
```

Add test case for alternative syntax for BigInt

packages/web3-utils/test/fixtures/converters.ts
```diff
@@ -0,0 +1,241 @@
...
+	[true, ['0x01', 'bool']],
+	[false, ['0x00', 'bool']],
+	['0x123c', ['0x123c', 'bytes']],
+];
+
+export const fromWeiValidData: [[Numbers, EtherUnits], string][] = [
+	[[0, 'wei'], '0'],
+	[[123, 'wei'], '123'],
+	[['123', 'wei'], '123'],
+	[[BigInt(123), 'wei'], '123'],
```
Spacesai1or, line 197:
```suggestion
	[[BigInt(123), 'wei'], '123'],
	[[123n, 'wei'], '123'],
```

Add test case for alternative syntax for BigInt

packages/web3-utils/src/converters.ts
nazarhussain: Yes you are right.

packages/web3-utils/src/converters.ts
nazarhussain: The input `number` is of type `Numbers` which is better to be casted with `toNumber` as it could include `HexString` as well.

Spacesai1or (approved):

PR #4391, opened 2021-09-26
Release v1.6.0

Spacesai1or:
### Changed

- Partially replace usage of [eth-lib](https://github.com/MaiaVictor/eth-lib) with [ethereumjs-util](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/util) (#4390)

PR #4407, opened 2021-09-30
Release 1.6.0 (Manual)

Spacesai1or:
The`release/1.6.0` branch was made from the `release/1.5.3` branch to avoid including any changes that had been merged since then and are scheduled to be release now as `1.6.1` (#4391). This resulted in a bunch of merge conflicts and I'm not sure the best way to handle the situation, so I opted to create this PR that does everything #4391 is supposed to do (bump the version numbers and update `dist/`) manually, with the end result being as if we had released `1.6.0` before merging any of the changes schedule for `1.6.1`


nazarhussain (approved):

jdevcs (approved):

jdevcs:
one observation, seems `release/1.6.0` was branched from the `release/1.5.3` instead of `release/1.5.2`?

Issue #4381, opened 2021-09-25
Migrate errors as error constructors


Spacesai1or:
- [ ] `invalidClientUrl`
    - msg: `Provided web3Client is an invalid HTTP(S) URL`
- [ ] `noHttpClient`
    - msg: `No HTTP client has be initialized`
- [ ] `connectionRefused`
    - msg: `Unable to make connection with HTTP client`
- [ ] `unsupportedOperation`
- [ ] `networkError`

Issue #4380, opened 2021-09-25
web3.providers.http.supportsSubscriptions


Spacesai1or:
Current implementation
  ```typescript
  HttpProvider.prototype.supportsSubscriptions = function () {
      return false;
  };
  ```

Issue #4379, opened 2021-09-25
web3.providers.http.disconnect


Spacesai1or:
Current implementation
  ```typescript
  HttpProvider.prototype.disconnect = function () {
      //NO OP
  };
  ```

Issue #4374, opened 2021-09-24
web3.providers-http: ETH2 Style JSON RPC Schema support


Spacesai1or:
[ETH2 Style JSON RPC Schema](https://ethereum.github.io/beacon-APIs/?urls.primaryName=v1#/Beacon)

- `GET` requests with URL parameters
    ```typescript
    `http://127.0.0.1/eth/v1/beacon/states/${state_id}/validators/${validator_id}`
    ```
- `POST` requests with URL parameters and request body
    - [Publishing a signed block](https://ethereum.github.io/beacon-APIs/?urls.primaryName=v1#/Beacon/publishBlock)
    - [Submit attestation](https://ethereum.github.io/beacon-APIs/?urls.primaryName=v1#/Beacon/submitPoolAttestations)
    - [Submit attester slashing](https://ethereum.github.io/beacon-APIs/?urls.primaryName=v1#/Beacon/submitPoolAttestations)

philknows:
Outstanding issue to be discussed later. Need to further investigate provider architecture.

PR #4416, opened 2021-10-01
Init web3-providers-http

### LINKED ISSUE #4312: [Rewrite 4.x]  Http Provider
### Description  

Create package `web3-providers-http` to server JSON RPC via HTTP

### Related Issues / Feature Requests / Discussion

- [x] #4371
    - [x] #4373 
    - [x] #4374 
- [x] #4379
- [x] #4380
- [x] #4381
- [x] #4456

### Acceptance Criteria

- [ ] README is added with overview and examples 
- [ ] Required documentations are updated 
- [ ] Test coverage is greater than 80% 
    - [ ] unit tests
    - [ ] integration tests
    - [ ] e2e tests
- [ ] Verify that no function left to be migrated

### Dependencies / Related Modules

- `web3-common` #4340

### Issues / Feature Requests to be considered
https://github.com/ChainSafe/web3.js/issues/3425 XMLHttpRequest errors handling
https://github.com/ChainSafe/web3.js/issues/4696

Spacesai1or:
Epic: #4312

Child issues to be closed:
- #4371
- #4373
- #4379
- #4380
- #4381

Not included:

- #4374
- #4456


tools/eslint-config-web3-base/ts.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
 		'import/prefer-default-export': 'off',
 		'lines-between-class-members': 'off',
 		'no-use-before-define': ['error'],
-		'no-shadow': ['error'],
+		'no-shadow': ['off'],
```
Spacesai1or: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-shadow.md#how-to-use

Spacesai1or: This was incorrectly reporting an error for the enum defined [here](https://github.com/ChainSafe/web3.js/blob/ff716daf3f7b388bfc059b1c7b3c395fd7ef0372/packages/web3-common/src/types.ts#L5)

packages/web3-common/src/types.ts
```diff
@@ -1,37 +1,65 @@
...
 
-export interface JsonRpcResponseWithError<T = JsonRpcResult> {
+export interface ExecutionJsonRpcRequest<T = unknown[]> {
 	readonly id: JsonRpcId;
 	readonly jsonrpc: JsonRpcIdentifier;
-	readonly error: JsonRpcError<T>;
-	readonly result?: never;
+	readonly method: string;
+	readonly params?: T;
+	readonly protocol?: SupportedProtocolsEnum.EXECUTION;
```
nazarhussain: There can't exist any extra parameter in the JSON-RPC request. If you want to add some predefined params then it must go under `params`

https://www.jsonrpc.org/specification#request_object

packages/web3-common/src/types.ts
```diff
@@ -1,37 +1,65 @@
...
 }
 
-export interface JsonRpcResponseWithResult<T = JsonRpcResult> {
+export interface ConsensusJsonRpcRequest<T = unknown> {
+	readonly endpoint: string;
+	readonly requestBody?: T;
+	readonly protocol: SupportedProtocolsEnum.CONSENSUS;
+}
```
nazarhussain: Same applies here. JSON-RPC request can't have any of these parameters. If you want to build your own request object then don't call it `JSON-RPC` request. 

https://www.jsonrpc.org/specification#request_object

packages/web3-common/src/types.ts
```diff
@@ -1,37 +1,65 @@
...
+	readonly data: T;
+}
 
-export interface JsonRpcRequest<T = unknown[]> {
-	readonly method: string;
-	readonly params?: T;
+export interface JsonRpcError {
+	readonly code: number;
+	readonly message: string;
+}
```
nazarhussain: The existing `JsonRpcError` type was fully compliant with the specification. The Eth API is also following the specification, I don't see any reason to modify these objects. 

https://www.jsonrpc.org/specification#error_object

nazarhussain (changes requested):
The `eth1` API is fully complaint with the [JSON-RPC](https://github.com/ethereum/execution-apis) standard, and `beacon` is using [OpenAPI](https://github.com/ethereum/beacon-APIs) standard. 

Since both standard are standalone and fully independent of each other, so let's not mix up the types for these together. 

We should look into using our other [Eth2 API package ](https://www.npmjs.com/package/@chainsafe/eth2.0-api-wrapper) internally in our provider and don't redo the work. If there is any blocker using it let's discuss.

packages/web3-providers-http/src/index.ts
```diff
@@ -0,0 +1,122 @@
...
+		callback: (
+			error?: JsonRpcResponseWithError<T>,
+			result?: JsonRpcResponseWithResult<T>,
+		) => void,
+		providerOptions?: T3,
+	): void {
+		this.request<JsonRpcResponseWithResult<T>, T2, T3>(payload, providerOptions)
+			.then(d => callback(undefined, d))
+			.catch(e => callback(e, undefined));
+	}
```
nazarhussain: You don't need this as it will be handled by the `BaseProvider` for you. 

```suggestion
```

packages/web3-providers-http/src/index.ts
```diff
@@ -0,0 +1,122 @@
...
+
+	/* eslint-disable class-methods-use-this */
+	public getStatus(): Web3BaseProviderStatus {
+		throw new MethodNotImplementedError();
+	}
```
nazarhussain, line 32: May be would be better to return `connected` status as HTTP provider always connected and `request` is not dependent on existing connection.

nazarhussain (changes requested):

packages/web3-common/src/web3_base_provider.ts
```diff
@@ -29,25 +29,27 @@ export const JSONRPC_ERR_CHAIN_DISCONNECTED = 4901;
...
 			.catch(e => callback(e, undefined));
 	}
 
 	abstract getStatus(): Web3BaseProviderStatus;
 	abstract supportsSubscriptions(): boolean;
 
 	// https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#request
-	abstract request<T = JsonRpcResult, T2 = unknown[]>(request: JsonRpcRequest<T2>): Promise<T>;
+	abstract request<T = JsonRpcResponse, T2 = unknown[], T3 = unknown>(
+		request: JsonRpcPayload<T2>,
```
nazarhussain, line 50: I thought on it further and to align with the `EIP1339` it would be better to just use `JsonRpcRequest` here. User of the provider don't get into complexity of maintaining the request `id` and provider implementation should do it internally.

nazarhussain (changes requested):

packages/web3-providers-http/src/index.ts
Spacesai1or: > The Provider is said to be "disconnected" when it cannot service RPC requests to any chain at all.

[Source](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#connectivity)

Which means we'd need to add logic to toggle the status depending on if requests can be made. Without making extra network calls to check the liveliness of the endpoint, we could only update when the user attempts to make a request. The status could easily be outdated if a successful request is made, but the endpoint immediately stops serving requests afterwards. Because of this, I don't think the status can be relied upon, and I question it's usefulness

packages/web3-common/src/web3_base_provider.ts
Spacesai1or: The `1.x` implementation just passes whatever [payload](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-providers-http/src/index.js#L121) is given to it, so we should allow for the specification of `jsonrpc` and `id` to maintain backwards compatibility

packages/web3-providers-http/src/index.ts
Spacesai1or: However, based on [this](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-providers-http/src/index.js#L128), maybe we shouldn't throw errors and just do `// NO OP`

nazarhussain: My understanding to this is, the provider we have here is serving JSON-RPC over HTTP. And if we have a URL it means we can serve the JSON-RPC request. If there is an error during the request that should be treated as error.

packages/web3-common/src/web3_base_provider.ts
nazarhussain: For this point I think we should break the 1.x compatibility. It will still not be a breaking change in terms of public exposed interfaces of the `web3` but will bring us more aligned with EIP1139.

The idea is the users of the providers would have to deal with less and less complexity. e.g. 

```ts
import {Web3HttpProvider} from 'web3-http-provider`;

const provider = new Web3HttpProvider('infura url');

await provivder.request({method: 'eth_getBlockByNumber', params: ['0x111']}
```

With this approach user will be less worried of maintaining the request id and json-rpc identifier and focus only on the request method he is interested in. 

What do you think?

nazarhussain (changes requested):

packages/web3-common/src/web3_base_provider.ts
nazarhussain: We decided to keep the `id` and `jsonrpc` identifier as optional for the `request`.

nazarhussain (approved):

luu-alex (approved):

PR #4418, opened 2021-10-01
fix incorrectly versioned bn.js type export

mrice32:
## Description

The `@types/bn.js` package is not listed as a dependency in all packages that have an `index.d.ts` file that calls `require("bn.js")`. This causes downstream typescript users to resolve a type from a nested web3 dependency (`ethereumjs-utils`), which uses `bn.js@5.X.X` rather than `bn.js@4.X.X` as web3 does. The result is that web3 uses `bn.js@4.X.X` in js, but exports the `bn.js@5.X.X` type.

For users, this can result in compilation errors if using the `bn.js` library directly and using the objects to interact with web3. In the worst case, this can cause typescript users to make invalid calls into these BN objects due to the type mismatch.

## Type of change

## Checklist:


Spacesai1or:
Hey there! Thank you for opening this PR, it might actually be a solution to the failing `Build / e2e (e2e_gnosis_dex)` test

The `gnosis dex` repo which was renamed to `gp-v1-ui` uses the [toWei](https://github.com/gnosis/gp-v1-ui/blob/develop/src/api/wallet/WalletApiMock.ts#L33) method which supports a `bn` as input. It imports it from `dex-js`, and `dex-js` is just [re-exporting](https://github.com/gnosis/dex-js/blob/master/src/utils/ethereum.ts#L3) it from `web3-utils`. Even though both `gp-v1-ui` and `dex-js` import and use `bn.js`, I didn't find it listed as a dependency in their `package.json`s. However, `dex-js` does import `dex-contracts` which has `bn.js` listed as a [dependency](https://github.com/gnosis/dex-contracts/blob/master/package.json#L78) for version `^5.2.0`

I'm not sure why this error would pop up now, but it sounds similar to the problem you were describing by not including `bn.js` as a dependency

Spacesai1or (changes requested):
The files

- `dist/web3.min.js`
- `dist/web3.min.js.map`

shouldn't haven been updated, this only happens in release PRs

philknows:
As mentioned in today's call, we should try to merge this for 1.6.1 @jdevcs to fix the failing tests. 
@spacesailor24 will create a new issue if there's no response from PR author to request changes.

mrice32:
> As mentioned in today's call, we should try to merge this for 1.6.1 @jdevcs to fix the failing tests. @spacesailor24 will create a new issue if there's no response from PR author to request changes.

Sorry, I can make the changes requested now. Sorry for the delay!

mrice32:
> The files
> 
> * `dist/web3.min.js`
> * `dist/web3.min.js.map`
> 
> shouldn't haven been updated, this only happens in release PRs

@spacesailor24 Thanks! This should be fixed now. Let me know if there's anything else that needs to be done before merging.

CHANGELOG.md
```diff
@@ -472,3 +472,4 @@ Released with 1.0.0-beta.37 code base.
 - Format `block.baseFeePerGas` to number (#4330)
 - Correct `web3-eth-personal.sendTransaction` example in documentation (#4409)
 - Updated README to include Webpack 5 angular support instructions (#4174)
+- - Added @types/bn.js dependency to additional packages (notably web3-utils) so the type export aligns with the js.
```
Spacesai1or, line 475:
```suggestion
- Added @types/bn.js dependency to additional packages (notably web3-utils) so the type export aligns with the js (#4418)
```

Extra `-` and missing PR number

Spacesai1or:
I'm going to take over this, some changes need to be applied, namely `@types/bn.js` was added as a dependency instead of a dev dependency

@mrice32 You'll still be apart of the git history, we're just trying to get the next RC out and want to merge this in to see if it fixes #4436

mrice32:
@spacesailor24 

> I'm going to take over this

Sounds good!

> some changes need to be applied, namely @types/bn.js was added as a dependency instead of a dev dependency

I totally could be wrong, but it was my impression that the issue was that it _needed_ to be a regular dependency. If you are depending on this package and using typescript, you need the `@types/bn.js` package to be installed in the subtree below this package, and (as far as I know) dev deps are _only_ installed during local development within this repo, not for downstream users when they install via npm. If it's not installed the `require("bn.js")` call within the index.d.ts file will resolve to `ethereumjs-utils`'s version of `@types/bn.js` instead of the one web3 uses, which causes the issue we're seeing.

Does that make sense?

Spacesai1or:
@mrice32 

Yep, I think you're right, corrected myself [here](https://github.com/ChainSafe/web3.js/pull/4437#issuecomment-937379382). Just waiting for consensus around it from the rest of the team

Thank you for being so responsive and helpful with fixing this!

Issue #4401, opened 2021-09-28
[Rewrite 4.x] web3-providers-ws reconnect


jdevcs:
web3-providers-ws reconnect
- [ ]  try to reconnect for defined number of times and if connection successful send requests waiting in queue

Issue #4400, opened 2021-09-28
[Rewrite 4.x] web3-providers-ws disconnect


jdevcs:
web3-providers-ws disconnect:

- [x] removeSocketListeners
- [x] close connection / reset request queues

Issue #4399, opened 2021-09-28
[Rewrite 4.x] web3-providers-ws request 


jdevcs:
This function is used to send request if provider is connected or if the provider is connecting it will add request to the requests queue.

Issue #4398, opened 2021-09-28
[Rewrite 4.x] web3-providers-ws connect


jdevcs:
- [ ] web3-providers-ws connect

- [ ] addSocketListeners for events ( incoming message, open, close )

- on message incoming event parse response and give result to user ( via callback or events ? )
- on connection open event if there are requests in queue send requests
- on close event notify user if there are requests waiting and not sent

PR #4434, opened 2021-10-04
websocket provider

jdevcs:
Epic #4313 

Issues to be addressed in this PR:
- [x] #4398
- [x] #4399
- [x] #4400
- [x] #4401


packages/web3-common/src/errors.ts
```diff
@@ -32,7 +32,12 @@ import {
 	ERR_INVALID_CLIENT,
 } from './constants';
 
-type ConnectionEvent = { code: string; reason: string };
+export type ConnectionEvent = {
+	code: number;
+	reason: string;
+	wasClean?: boolean; // if WS connection was closed properly
+};
+
```
nazarhussain: Types should be moved to `types.ts` file.

packages/web3-common/src/types.ts
```diff
@@ -20,18 +20,36 @@ export interface JsonRpcResponseWithError<T = JsonRpcResult> {
 	readonly jsonrpc: JsonRpcIdentifier;
 	readonly error: JsonRpcError<T>;
 	readonly result?: never;
+	readonly method?: never;
+	readonly params?: never;
```
nazarhussain: As this type reflects the output of the JSON-RPC call which always be compliance with the JSON-RC standard. Which does not include these properties. 

https://www.jsonrpc.org/specification#response_object

```suggestion
```

packages/web3-common/src/types.ts
```diff
@@ -20,18 +20,36 @@ export interface JsonRpcResponseWithError<T = JsonRpcResult> {
...
 }
 
 export interface JsonRpcResponseWithResult<T = JsonRpcResult> {
 	readonly id: JsonRpcId;
 	readonly jsonrpc: JsonRpcIdentifier;
 	readonly error?: never;
 	readonly result: T;
+	readonly method?: never;
+	readonly params?: never;
```
nazarhussain: Same as above.

packages/web3-common/src/types.ts
```diff
@@ -20,18 +20,36 @@ export interface JsonRpcResponseWithError<T = JsonRpcResult> {
...
+	readonly subscription: string; // for subscription id
+	readonly result: T;
+}
+export interface JsonRpcResponseWithSubscriptionResult<T = JsonRpcResult> {
+	readonly id?: JsonRpcId;
+	readonly jsonrpc: JsonRpcIdentifier;
+	readonly method: string; // for subscription
+	readonly params: Params<T>; // for subscription results
+	readonly error?: never;
+	readonly result: never;
```
nazarhussain: Ethereum clients implement subscriptions with [JSON-RPC Notifications ](https://geth.ethereum.org/docs/rpc/pubsub)

> Geth v1.4 and later support publish / subscribe using JSON-RPC notifications. This allows clients to wait for events instead of polling for them.

There is a particular [standard](https://www.jsonrpc.org/specification#notification) for it. In JSON-RPC Notifications are considered as Requests rather Response. In Geth case these Notifications are sent from server to client. Please refactor the types and names accordingly.

packages/web3-providers-ws/src/deferredPromise.ts
```diff
@@ -0,0 +1,35 @@
+export class DeferredPromise<T> {
+	public realPromise: Promise<T>;
+
+	private resolveFunction!: (value: T | PromiseLike<T>) => void;
+	private rejectFunction!: (reason?: unknown) => void;
```
nazarhussain: Prefix all private members with `_`

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+	/* eslint-disable @typescript-eslint/no-explicit-any */
+	private readonly requestQueue: Map<JsonRpcId, WSRequestItem<any, any>>;
+	/* eslint-disable @typescript-eslint/no-explicit-any */
+	private readonly sentQueue: Map<JsonRpcId, WSRequestItem<any, any>>;
+
+	private providerStatus: Web3BaseProviderStatus;
+
+	private reconnectAttempts!: number;
+	private lastDataChunk!: string;
+	private lastChunkTimeout!: NodeJS.Timeout;
```
nazarhussain: Prefix all private members with `_`

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+
+	public constructor(clientUrl: string, wsProviderOptions?: WebSocketOptions) {
+		super();
+		if (!WebSocketProvider.validateProviderUrl(clientUrl))
+			throw new InvalidClientError(clientUrl);
+
+		this.clientUrl = clientUrl;
+		if (wsProviderOptions !== undefined) this.wsProviderOptions = wsProviderOptions;
```
nazarhussain: You can use typescript [parameter properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties) to avoid such assignments.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+	private providerStatus: Web3BaseProviderStatus;
+
+	private reconnectAttempts!: number;
+	private lastDataChunk!: string;
+	private lastChunkTimeout!: NodeJS.Timeout;
+
+	public constructor(clientUrl: string, wsProviderOptions?: WebSocketOptions) {
+		super();
+		if (!WebSocketProvider.validateProviderUrl(clientUrl))
+			throw new InvalidClientError(clientUrl);
```
nazarhussain: Consistently use `{` for code blocks. It's more readable.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+	PendingRequestsOnReconnectingError,
+	ConnectionNotOpenError,
+	InvalidClientError,
+	ConnectionEvent,
+} from 'web3-common/dist/errors';
+import { WebSocketOptions, WSRequestItem } from './types';
+import { DeferredPromise } from './deferredPromise';
+
+export default class WebSocketProvider extends Web3BaseProvider {
+	public wsEventEmitter: EventEmitter = new EventEmitter();
```
nazarhussain: Don't expose the emitter to public level. It cause unintended uses and bugs in future.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+import {
+	Web3BaseProvider,
+	Web3BaseProviderStatus,
+	JsonRpcId,
+	JsonRpcPayload,
+	JsonRpcResponse,
+	JsonRpcResult,
+	Web3BaseProviderCallback,
+} from 'web3-common';
+import { IMessageEvent, w3cwebsocket as W3WS } from 'websocket';
```
nazarhussain: I would suggest to use `ws` package instead. That is more performant, more stable and more popular. And we have `isomorphic-ws` package to provider cross platform support. 

https://www.npmjs.com/package/ws

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+	private init() {
+		this.lastDataChunk = '';
+		this.reconnectAttempts = 0;
+
+		if (this.lastChunkTimeout) {
+			clearTimeout(this.lastChunkTimeout);
+		}
+	}
+
+	private addSocketListeners(): void {
```
nazarhussain: Please prefix all private functions with `_`

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+		// eslint-disable-next-line
```
nazarhussain: Disable particular linting rules which are intended.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+		(this.webSocketConnection as any).addEventListener('message', this.onMessage.bind(this));
```
nazarhussain: Why we have to cast `any` here? If it does not have proper types, then we should look into other library.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+			this.reconnectAttempts < this.wsProviderOptions.reconnectOptions.maxAttempts
+		) {
+			setTimeout(() => {
+				this.reconnectAttempts += 1;
+
+				this.removeSocketListeners();
+				this.connect();
+			}, this.wsProviderOptions?.reconnectOptions?.delay);
+		}
+	}
```
nazarhussain: You removed `reconnect` from the abstract interface. If it's just the private reconnect attempt, then use `_attempReconnect` name.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+		this.init();
+		this.removeSocketListeners();
+		this.addSocketListeners();
+	}
+
+	public async request<T = JsonRpcResponse, T2 = unknown[]>(
+		request: JsonRpcPayload<T2>,
+	): Promise<T> {
+		if (this.webSocketConnection === undefined)
+			throw new Error('WebSocket connection is undefined');
```
nazarhussain, line 145: Always use web3 provided error constructors.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+
+		if (request.id === undefined) throw new Error('Request Id not defined');
```
nazarhussain, line 147: Yes its expected, and instead of throwing error, you should keep track of some local request ids.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+			this.webSocketConnection.readyState === this.webSocketConnection.CLOSING
+		) {
+			this.providerStatus = 'disconnected';
+
+			this.requestQueue.delete(id);
+
+			throw new ConnectionNotOpenError();
+		}
+
+		const requestItem = this.requestQueue.get(id);
```
nazarhussain: I believe this will always return undefined. As particular request is never been added to queue in above code.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+				const reqItem: WSRequestItem<T2, T> = {
+					payload: request,
+					deferredPromise: defPromise,
+				};
+
+				this.requestQueue.set(id, reqItem);
+				return defPromise.realPromise;
+			}
+
+			return requestItem.deferredPromise.realPromise as Promise<T>;
```
nazarhussain, line 172: If the types are correct, then you don't need to use `as` like this.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+
+	private readonly clientUrl: string;
+	private readonly wsProviderOptions: WebSocketOptions;
+
+	private webSocketConnection?: W3WS;
+
+	/* eslint-disable @typescript-eslint/no-explicit-any */
+	private readonly requestQueue: Map<JsonRpcId, WSRequestItem<any, any>>;
+	/* eslint-disable @typescript-eslint/no-explicit-any */
+	private readonly sentQueue: Map<JsonRpcId, WSRequestItem<any, any>>;
```
nazarhussain: We don't need to maintain two queues here. Only one `pendingRequestsQueue` will work fine. 

1. User invoked `request` 
2. You sent to ws connection 
3. If request sent then add that request to `pendingRequestsQueue`
4. When receive message from ws, match id in the `pendingRequestsQueue`
5. If id match, resolved the deferred promise.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+	private parseResponse(dataReceived: string): JsonRpcResponse[] {
+		const returnValues: JsonRpcResponse[] = [];
+
+		// de chunk
+		const dechunkedData = dataReceived
+			.replace(/\}[\n\r]?\{/g, '}|--|{') // }{
+			.replace(/\}\][\n\r]?\[\{/g, '}]|--|[{') // }][{
+			.replace(/\}[\n\r]?\[\{/g, '}|--|[{') // }[{
+			.replace(/\}\][\n\r]?\{/g, '}]|--|{') // }]{
+			.split('|--|');
```
nazarhussain: Why we we have to deal with this logic? It must be handled by any particular web-socket library itself. If the `ws` library is not dealing with the streaming responses per message, then it's a good reason to not use that library.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+				this.removeSocketListeners();
+				this.connect();
+			}, this.wsProviderOptions?.reconnectOptions?.delay);
+		}
+	}
+
+	private onMessage(e: IMessageEvent): void {
+		this.parseResponse(typeof e.data === 'string' ? e.data : '').forEach(
+			(response: JsonRpcResponse) => {
+				if (response.method?.includes('_subscription')) {
```
nazarhussain: Use `endsWith` instead https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+					this.wsEventEmitter.emit('message', null, response);
+					return;
+				}
+
+				let { id } = response;
+
+				if (Array.isArray(response)) {
+					// eslint-disable-next-line
+					id = response[0].id;
+				}
```
nazarhussain: When particularly we are gonna get an array response? And why we only taking `id` of the first response?

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+
+				if (id && this.sentQueue.has(id)) {
+					const requestItem = this.sentQueue.get(id);
+					if (response.result !== undefined) {
+						this.wsEventEmitter.emit('message', null, response);
+						requestItem?.deferredPromise.resolve(response);
```
nazarhussain: It should resolve with the `response.result` not the full response object.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+					} else if (response.error !== undefined) {
+						this.wsEventEmitter.emit('message', response, null);
+						requestItem?.deferredPromise.reject(response);
```
nazarhussain: It should be rejected with `response.error` not the full response object.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,391 @@
...
+	public reset(): void {
+		this.sentQueue.clear();
+		this.requestQueue.clear();
+
+		this.init();
+		this.removeSocketListeners();
+		this.addSocketListeners();
+	}
+
+	public async request<T = JsonRpcResponse, T2 = unknown[]>(
```
nazarhussain: I believe this change is missed during the review. The user of the providers are intended the result of the request, not the wrappers. 

e.g. 

```ts
const result = await provider.request({method: 'eth_getBlockByNumber', params: ['0x12', true]});
```

In the above example I would be looking for the block **result** not the json-rpc **response** object.

nazarhussain (changes requested):

packages/web3-providers-ws/src/index.ts
Spacesai1or: This is a tricky feature where a developer would easily get confused why the properties aren't getting explicitly set if they don't understand this is a feature of TypeScript

packages/web3-providers-ws/src/index.ts
Spacesai1or: I think prettier will also get rid of curly braces for one line if statements - I am okay with this as I never want us to edit prettier rules (makes it easier to stay consistent with packages outside of Web3.js)

packages/web3-providers-ws/src/deferredPromise.ts
jdevcs: imo its outdated convention , good to follow some conventions of [ts project](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines) it self ( 6. Do not use _ as a prefix for private properties. )

packages/web3-providers-ws/src/index.ts
jdevcs: I like `ws` as it have 0 dependencies, but it doesn't support browsers out of box.

packages/web3-providers-ws/src/index.ts
jdevcs: It gives good set of functionality to users of this class, if they need.

packages/web3-providers-ws/src/index.ts
jdevcs: Its not necessary that request is always sent to server from request function, like: in case of `this.webSocketConnection.readyState === this.webSocketConnection.CONNECTING` request will be added to first requestQueue, and on connect event this queue will be flushed.

packages/web3-providers-ws/src/index.ts
jdevcs: Transport must return actual response, so users are able to track it with id

packages/web3-providers-ws/src/index.ts
jdevcs: Its [issue](https://github.com/ChainSafe/web3.js/issues/4495) already created

packages/web3-providers-ws/src/index.ts
jdevcs: request object is provided by user, and transport should not modify & manage its internals.

packages/web3-providers-ws/src/index.ts
jdevcs: it will not return undefined always, : in case request is already waiting in queue and there is reconnect event it will give request object.

packages/web3-providers-ws/src/index.ts
jdevcs: you are right, but its for linting rules as: 

`web3-providers-ws:   179:4  error  Unsafe return of type `Promise<any>` from function with return type `Promise<T>`  @typescript-eslint/no-unsafe-return
web3-providers-ws:   207:3  error  Unsafe return of type `Promise<any>` from function with return type `Promise<T>`  @typescript-eslint/no-unsafe-return
`

packages/web3-providers-ws/src/index.ts
jdevcs: naming convention explained above

packages/web3-providers-ws/src/index.ts
jdevcs: It doesn't have type defined.

packages/web3-providers-ws/src/index.ts
jdevcs: naming convention explained above

packages/web3-providers-ws/src/index.ts
jdevcs: I haven't used or tested this scenario, it is supported in 1.x , I think for batch response. Plz inform me if you found any doc on this.

packages/web3-providers-ws/src/index.ts
jdevcs: informed above, transport must share response instead of result or error, so user can use id

packages/web3-providers-ws/src/index.ts
jdevcs: same as above

packages/web3-providers-ws/src/deferredPromise.ts
Spacesai1or: I prefer using `_`, it categorizes the properties the IDE suggests when you start typing: `this.`

nazarhussain: @jdevcs As the document you referred says. 

> This is NOT a prescriptive guideline for the TypeScript community

And the majority of the Typescript community tend to follow the standard of having `_` prefix for the `private` and `protected`  attributes and functions. And the reason is the value addition to the readability. While writing the code first time you may not notice a difference. But when you will be reading the same code after few months, you will notice that just knowing by convention which is not public adds a lot of context to the code line. 


The other major reason why Typescript itself suggest to not use `_` prefix is the availability of [native private properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) in Javascript and was introduced in [Typescript 3.8](https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/#ecmascript-private-fields). But that feature is yet on Stage 3, where further discussion and refinement is needed. 

So until we have that native private fields feature adopted, we should stick to using `_` as prefix for private and protected fields.

packages/web3-providers-ws/src/index.ts
nazarhussain: It should not need to support to browser. Because browsers have native support for [Websocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API). You just need to use `isomorphic-ws` package and it will switch the context between native Websocket and `ws` implementation without any code change. 


The [comparison](https://npmcompare.com/compare/websocket,ws) between `ws` and `websocket` packages are not questionable. So we should stick to using `ws` package in our provider.

packages/web3-providers-ws/src/index.ts
nazarhussain: As I said it exposes the API which endup creating unintended uses. As a library author, remember a rule only expose some API for which you are sure about the use case. 

e.g. one can inject a code which use `provider.wsEventEmitter.emit` to emit fraudulent events and the listeners can lead to some intended behaviours. 

As our library is widely used in many different products, so we have to be very careful in exposing any API to public interfaces.

packages/web3-providers-ws/src/index.ts
nazarhussain: From user perspective `provider.request` is the only interface to send the request. If you keep pushing the requests to the queue, then how will user be notified that his requests are not sent?

packages/web3-providers-ws/src/index.ts
nazarhussain: The input of function is the request, and request contains only `method` and `parameters`. The `id` and other fields are part of the payload and which are optional in the parameters. The transport had to sent the payload and for that it needed to be created and validated before sending. So if user didn't provided the `id` it's not an issue and provider should handle creating the payload internally.

packages/web3-providers-ws/src/index.ts
nazarhussain: Dont' use any untyped code in the `src`. If it does not have a type defined then define it by your self so you know what are you expecting form that object.

packages/web3-providers-ws/src/index.ts
nazarhussain: As we don't know exact reason for it, then let's remove that code block.

nazarhussain (changes requested):
In addition to pending review items. There is one important thing which is missing in the implementation. Which is the handling of the timeout scenario. How long user would wait for the response? In current implementation, even if the provider is not connected and user made a request, it will be hanged forever.

packages/web3-providers-ws/src/deferredPromise.ts
jdevcs: Its not good convention to use (manual) naming prefix or post-fix for differentiating variables access modifier traits.
[Google guidelines](https://google.github.io/styleguide/tsguide.html#naming-style)

packages/web3-providers-ws/src/index.ts
jdevcs: For that user has to wait for promise reject or Error.

packages/web3-providers-ws/src/index.ts
jdevcs: GRASP High cohesion, transport should only focus its function 'transport', instead of other things: managing and modifying param obj internals.

packages/web3-providers-ws/src/index.ts
jdevcs: by support it means: auto taken care/switched in browsers by lib. That's not done in ws, and its done in websocket. so as you mentioned for ws need switching via isomorphic . Opened [issue](https://github.com/ChainSafe/web3.js/issues/4515) for it,

packages/web3-providers-ws/src/index.ts
jdevcs: https://github.com/ChainSafe/web3.js/issues/4515

packages/web3-providers-ws/src/index.ts
jdevcs: if switched https://github.com/ChainSafe/web3.js/issues/4515 this will go away

packages/web3-common/src/errors.ts
```diff
@@ -33,9 +33,9 @@ import {
 	ERR_PROVIDER,
 	ERR_SUBSCRIPTION,
 } from './constants';
-import { JsonRpcResponse } from './types';
 
-type ConnectionEvent = { code: string; reason: string };
+import { ConnectionEvent, JsonRpcResponse } from './types';
+
 type Receipt = Record<string, unknown>;
```
Spacesai1or: Should `type Receipt = Record<string, unknown>;` be defined in `./types`?

packages/web3-common/src/types.ts
```diff
@@ -29,6 +29,17 @@ export interface JsonRpcResponseWithResult<T = JsonRpcResult> {
 	readonly result: T;
 }
 
+export interface Params<T = JsonRpcResult> {
```
Spacesai1or: Rename this to `SubscriptionParams`? Otherwise it's too generic being that it's apart of the `web3-common` package

packages/web3-common/src/web3_base_provider.ts
```diff
@@ -75,5 +75,4 @@ export abstract class Web3BaseProvider {
 	abstract connect(): void;
 	abstract disconnect(code: number, reason: string): void;
 	abstract reset(): void;
-	abstract reconnect(): void;
```
Spacesai1or, line 78: I'm confused why this was removed?

Spacesai1or:
Should remove unused files such as: `packages/web3-providers-ws/src/errors.ts`?

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+	public constructor(clientUrl: string, wsProviderOptions?: WebSocketOptions) {
+		super();
+		if (!WebSocketProvider.validateProviderUrl(clientUrl))
+			throw new InvalidClientError(clientUrl);
+
+		this.clientUrl = clientUrl;
+		if (wsProviderOptions !== undefined) this.wsProviderOptions = wsProviderOptions;
+		else
+			this.wsProviderOptions = {
+				customTimeout: 1000 * 15,
```
spacesailor24, line 52: Should probably make the `1000 * 15` a `const DEFAULT_TIMEOUT`

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+
+		if (!this.wsProviderOptions.customTimeout) this.wsProviderOptions.customTimeout = 1000 * 15;
+
+		if (!this.wsProviderOptions.reconnectOptions)
+			this.wsProviderOptions.reconnectOptions = {
+				auto: false,
+				delay: 5000,
+				maxAttempts: 5,
+				onTimeout: false,
+			};
```
spacesailor24, line 63: Does all this need to be defined? Because if a user provided `wsProviderOptions` as:

```typescript
{
    reconnectOptions: {}
}
```

none of this is going to get set. This also doesn't guarentee all those properties are set if the user does:

```typescript
{
    reconnectOptions: {
        delay: 5000
    }
}
```

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+			throw new InvalidClientError(clientUrl);
+
+		this.clientUrl = clientUrl;
+		if (wsProviderOptions !== undefined) this.wsProviderOptions = wsProviderOptions;
+		else
+			this.wsProviderOptions = {
+				customTimeout: 1000 * 15,
+			};
+
+		if (!this.wsProviderOptions.customTimeout) this.wsProviderOptions.customTimeout = 1000 * 15;
```
spacesailor24, line 55:
```suggestion		
		this.wsProviderOptions = wsProviderOptions ?? { customTimeout: 1000 * 15 };
		this.wsProviderOptions.customTimeout = this.wsProviderOptions.customTimeout ?? 1000 * 15;
```

packages/web3-providers-ws/src/index.ts
Spacesai1or: The way this is currently written doesn't validate that all these properties are going to get set. Instead I'd do something like:

```typescript
const DEFAULT_WS_PROVIDER_OPTIONS = {
    customTimeout: 1000 * 15,
    reconnectOptions: {
        auto: false,
        delay: 5000,
        maxAttempts: 5,
        onTimeout: false,
    }
}

this.wsProviderOptions = {
    ...DEFAULT_WS_PROVIDER_OPTIONS,
    ...wsProviderOptions
    reconnectOptions: {
        ...DEFAULT_WS_PROVIDER_OPTIONS.reconnectOptions,
        ...wsProviderOptions?.reconnectOptions
    }
}
```

This will default all the needed properties regardless of how the user passes `wsProviderOptions`

packages/web3-providers-ws/src/types.ts
```diff
@@ -0,0 +1,29 @@
...
+	onTimeout: boolean;
+};
+
+export interface WebSocketOptions {
+	protocol?: string;
+	headers?: OutgoingHttpHeaders;
+	/* eslint-disable @typescript-eslint/ban-types */
+	requestOptions?: Object;
+	clientConfig?: IClientConfig;
+	customTimeout: number;
```
Spacesai1or: What's the difference between `customTimeout` and `timeout`?

packages/web3-providers-ws/src/types.ts
```diff
@@ -0,0 +1,29 @@
...
+	delay: number;
+	maxAttempts: number;
+	onTimeout: boolean;
+};
+
+export interface WebSocketOptions {
+	protocol?: string;
+	headers?: OutgoingHttpHeaders;
+	/* eslint-disable @typescript-eslint/ban-types */
+	requestOptions?: Object;
```
Spacesai1or:
```suggestion
	requestOptions?: Record<string, unknown>;
```

[Typescript docs](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html) say to not use `Object` as a type. Additionally, we should know what the options are and make this strongly typed (instead of using `Record<string, unknown>`)

packages/web3-providers-ws/src/index.ts
Spacesai1or: The http provider [generates an ID](https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-providers-http/src/index.ts#L55) if it's not provided by the user, and I think WS and IPC providers should do the same. However, I can't find how IDs are generated in `1.x`

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+			throw new Error('WebSocket connection is undefined');
+
+		if (request.id === undefined) throw new Error('Request Id not defined');
+
+		if (this.requestQueue.has(request.id) || this.sentQueue.has(request.id))
+			throw new Error(
+				'Duplicate request Id. Another request is already in Queue with same request Id.',
+			);
+
+		const { id } = request;
```
Spacesai1or: Why not just continue to use `request.id` are you in the code above this?

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+
+		if (
+			this.webSocketConnection.readyState === this.webSocketConnection.CLOSED ||
+			this.webSocketConnection.readyState === this.webSocketConnection.CLOSING
+		) {
+			this.providerStatus = 'disconnected';
+
+			this.requestQueue.delete(id);
```
Spacesai1or: Shouldn't the `id` not exist in `this.requestQueue` here? The[ following code](https://github.com/ChainSafe/web3.js/blob/d3bc30822e132ebc1a9d94862e834f99f8d94936/packages/web3-providers-ws/src/index.ts#L145) even suggests that:

```typescript
if (this.requestQueue.has(request.id) || this.sentQueue.has(request.id))
    throw new Error(
        'Duplicate request Id. Another request is already in Queue with same request Id.',
    );
```

Spacesai1or:
What do you think about splitting the [request method](https://github.com/ChainSafe/web3.js/blob/d3bc30822e132ebc1a9d94862e834f99f8d94936/packages/web3-providers-ws/src/index.ts#L137) into helper methods with function names that describe what they're doing? Adding comments would also work, but having a 71 lined function broken up into fragments of code makes it easier to read/understand

packages/web3-providers-ws/src/index.ts
Spacesai1or: I think an explanation (like error message +  reasons why it's not being fixed) should be provided when disabling linting. Additionally, you should be able to disable the specific rule for the line instead of disabling linting entirely

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+
+	private onMessage(e: IMessageEvent): void {
+		this.parseResponse(typeof e.data === 'string' ? e.data : '').forEach(
+			(response: JsonRpcResponse | SubscriptionResultNotification) => {
+				if ('method' in response && response.method.endsWith('_subscription')) {
+					this.wsEventEmitter.emit('message', null, response);
+					return;
+				}
+
+				const { id } = response;
```
Spacesai1or: Why not just use `request.id`?

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+		);
+	}
+
+	private onConnect() {
+		this.providerStatus = 'connected';
+		this.reconnectAttempts = 0;
+
+		if (this.requestQueue.size > 0) {
+			for (const value of this.requestQueue.values()) {
+				// eslint-disable-next-line
```
Spacesai1or: I think an explanation (like error message + reasons why it's not being fixed) should be provided when disabling linting. Additionally, you should be able to disable the specific rule for the line instead of disabling linting entirely

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+
+		dechunkedData.forEach((data: string) => {
+			let dataToParse: string = data;
+
+			if (this.lastDataChunk !== '') dataToParse = this.lastDataChunk + dataToParse;
+
+			let result: JsonRpcResponse;
+
+			try {
+				// eslint-disable-next-line
```
Spacesai1or: I think an explanation (like error message + reasons why it's not being fixed) should be provided when disabling linting. Additionally, you should be able to disable the specific rule for the line instead of disabling linting entirely

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+			this.lastDataChunk = '';
+
+			if (result) returnValues.push(result);
+		});
+
+		return returnValues;
+	}
+
+	private removeSocketListeners(): void {
+		// eslint-disable-next-line
```
Spacesai1or: I think an explanation (like error message + reasons why it's not being fixed) should be provided when disabling linting. Additionally, you should be able to disable the specific rule for the line instead of disabling linting entirely

Spacesai1or (changes requested):
Some of my code comments require some sort of code change, additionally there are no tests

packages/web3-providers-ws/src/deferredPromise.ts
nazarhussain: I had explained it very briefly already. The reason to not use prefix just because of some published guide is not enough. There are two questions to answer: 

1. Does adding a prefix to private variables add some value? 
2. Does adding a prefix to private variables create some issue? 

The answer to first question is Yes, it add a huge value towards code readability. Reading code from color coding of the IDE is not effective (no one had such brain power). Reading code with a known convention is most effective pattern. 

The answer to second question is No. Adding prefix does not create any issue in the code. Neither of the guides provide any reason or problem this convention may cause. 

So with that said we should just focus on the value addition and follow same pattern in the code. In addition myself and @spacesailor24 agrees of adding prefix for the private/protected fields and functions of the classes, so if you follow it we will have consistent code in the library.

nazarhussain (changes requested):

packages/web3-common/src/errors.ts
jdevcs: (It wasnt added in this PR ), you are right it should be moved

packages/web3-common/src/web3_base_provider.ts
jdevcs: It is specific to ws only, so it shouldn't be enforced to all providers via baseprovider

packages/web3-providers-ws/src/index.ts
jdevcs: it was for id collision check, if user passes request with same id that is already in Queue. 
But clean approach will be that transport shouldn't manage id, it should be moved to request manage.

packages/web3-providers-ws/src/index.ts
jdevcs: I think generating id from http provider should also be removed, as it will add extra collision check responsibility in all providers: 
- managing internal collisions ( if id is generated inside provider ) 
- managing external collisions of user's passed requests with internal generated ones

packages/web3-providers-ws/src/types.ts
jdevcs: with new lib `isomorphic-ws` and `ws` its no longer required.

packages/web3-providers-ws/src/types.ts
jdevcs: with new lib isomorphic-ws and ws its no longer required.

packages/web3-providers-ws/src/index.ts
```diff
@@ -0,0 +1,308 @@
...
+	SubscriptionResultNotification,
+} from 'web3-common';
+import { ClientRequestArgs } from 'http';
+import { ClientOptions, WebSocket, MessageEvent, CloseEvent } from 'isomorphic-ws';
+import {
+	InvalidConnectionError,
+	PendingRequestsOnReconnectingError,
+	ConnectionNotOpenError,
+	InvalidClientError,
+} from 'web3-common/dist/errors';
```
nazarhussain: Don't import from `dist` folder. 
```suggestion
} from 'web3-common';
```

packages/web3-providers-ws/test/unit/constructor.test.ts
```diff
@@ -0,0 +1,20 @@
+// import Web3ProviderBase from '../../src/index'
+// import {ProviderOptions} from '../../types'
+
+describe('constructs a PLACEHOLDER instance with expected properties', () => {
```
nazarhussain: You can use `it.todo` to add any placeholder. 

https://jestjs.io/docs/api#testtodoname

nazarhussain (changes requested):

packages/web3-providers-ws/src/index.ts
jdevcs: good catch, earlier there was some issue in vs complaining import/export so i used this . now fixed.

nazarhussain (approved):

Spacesai1or (approved):

PR #4437, opened 2021-10-07
Fix incorrectly versioned bn.js type export (#4418)

Spacesai1or:
Took over the original PR to make some changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -472,3 +472,4 @@ Released with 1.0.0-beta.37 code base.
 - Format `block.baseFeePerGas` to number (#4330)
 - Correct `web3-eth-personal.sendTransaction` example in documentation (#4409)
 - Updated README to include Webpack 5 angular support instructions (#4174)
+- - Added @types/bn.js dependency to additional packages (notably web3-utils) so the type export aligns with the js.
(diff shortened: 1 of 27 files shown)


Spacesai1or:
Oh I think [I'm wrong](https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies#:~:text=In%20that%20particular%20case%20you,%22devDependencies%22%20are%20good%20enough.&text=If%20you're%20just%20generating,distinction%20between%20dependencies%20and%20devDependencies%20.) about including `@types/bn.js` as a `dev-dependency`

In this particular case, we want to provide users with the specific version of the types so that type mismatches don't occur

nazarhussain (changes requested):
As per the change log we have very minimal change.

> `@types/bn.js` as dev-dependency to additional packages (notably web3-utils)

But the files which are changed are 27, Is there a possibility to only include required changes in this PR. And if there is any other package update we can create specific PR for it.

Spacesai1or:
@nazarhussain 

> As per the change log we have very minimal change.
> 
> > `@types/bn.js` as dev-dependency to additional packages (notably web3-utils)
> 
> But the files which are changed are 27, Is there a possibility to only include required changes in this PR. And if there is any other package update we can create specific PR for it.

The files changes are all updates to `package.json` and `package-lock.json` (minus `CHANGELOG.md`). Updating these files all at once, and not trying to discern what `package-lock.json` changes are bare minimum necessary, is the easiest way to avoid breaking something, imo

nazarhussain (approved):

luu-alex (approved):

deauna (commented):
cheers

PR #4447, opened 2021-10-11
Migrate all formatters to web3-common 

### LINKED ISSUE #4360: Migrate all formatters to web3-common
### Description

Need to migrate all formatters from `1.x` branch to `web3-common` package as **error constructors**. 

https://github.com/ChainSafe/web3.js/blob/40974e9fe9a8f9075fcbb4c002430c6480d83031/packages/web3-core-helpers/src/formatters.js#L39-L58

### Acceptance Criteria

- All formatters should be migrated 
- The formatters should be exported as `formatters` namespace 

### Additional Information

- Use `BigInt` where appropriate.

nazarhussain:
## Description

Migrate existing formatters from 1.x

Fixes #4360

## Type of change

## Checklist:


packages/web3-common/src/constants.ts
```diff
@@ -37,3 +38,6 @@ export const ERR_CONN_PENDING_REQUESTS = 506;
 // Provider error codes
 export const ERR_PROVIDER = 600;
 export const ERR_INVALID_PROVIDER = 601;
+
+export const PREDEFINED_BLOCK_NUMBERS = ['latest', 'pending', 'earliest'];
```
Spacesai1or: Could this be an enum?

packages/web3-common/src/types.ts
```diff
@@ -1,7 +1,13 @@
+import { HexString } from 'web3-utils';
+
 export type JsonRpcId = string | number | null;
 export type JsonRpcResult = string | number | boolean | Record<string, unknown>;
 export type JsonRpcIdentifier = '2.0' | '1.0';
 
+export type Mutable<T> = {
+	-readonly [P in keyof T]: T[P];
+};
+
```
Spacesai1or, line 11: Could you add a small comment explaining what's going on here?

packages/web3-common/src/types.ts
```diff
@@ -35,3 +41,145 @@ export interface JsonRpcPayload<T = unknown[]> extends JsonRpcRequest<T> {
...
 }
+
+export interface Proof {
+	readonly address: HexString;
+	readonly nonce: string;
+	readonly balance: string;
+}
+
+export interface TransactionInput {
+	readonly [key: string]: unknown;
```
Spacesai1or, line 53: What is this handling?

packages/web3-common/src/types.ts
```diff
@@ -35,3 +41,145 @@ export interface JsonRpcPayload<T = unknown[]> extends JsonRpcRequest<T> {
...
+	readonly to?: HexString; // If its a contract creation tx then no address wil be specified.
+	readonly from?: HexString;
+	readonly data?: string;
+	readonly input?: string;
+	readonly gas: HexString;
+	readonly gasLimit?: string;
+	readonly gasPrice?: string;
+	readonly maxPriorityFeePerGas?: string;
+	readonly maxFeePerGas?: string;
```
Spacesai1or, line 62: How do you feel about have explicit interfaces for each transaction type so that the appropriate fields can be required?

packages/web3-common/src/types.ts
```diff
@@ -35,3 +41,145 @@ export interface JsonRpcPayload<T = unknown[]> extends JsonRpcRequest<T> {
...
+	readonly gas?: bigint | number;
+	readonly gasLimit?: string;
+	readonly nonce: bigint | number;
+	readonly value: bigint | number;
+	readonly blockNumber?: bigint | number;
+	readonly transactionIndex?: bigint | number;
+} & (
+	| { maxPriorityFeePerGas: bigint | number; maxFeePerGas: bigint | number; gasPrice?: never }
+	| { maxPriorityFeePerGas?: never; maxFeePerGas?: never; gasPrice: bigint | number }
+);
```
Spacesai1or, line 84: I'm not a fan of this, I think I'd rather have something like:

```typescript
export type TransactionOutput = LegacyTransactionOutput | Eip2930TransactionOutput | Eip1559TransactionOutput
```

packages/web3-common/src/types.ts
```diff
@@ -35,3 +41,145 @@ export interface JsonRpcPayload<T = unknown[]> extends JsonRpcRequest<T> {
...
+	readonly gasLimit: HexString;
+	readonly gasUsed: HexString;
+	readonly size: HexString;
+	readonly timestamp: HexString;
+	readonly number?: HexString;
+	readonly difficulty?: HexString;
+	readonly totalDifficulty?: HexString;
+	readonly transactions?: TransactionInput[];
+	readonly miner?: HexString;
+	readonly baseFeePerGas?: HexString;
```
Spacesai1or, line 115: If we decided to have more explicit interfaces, than this should also be update to `LegacyBlockInput` and `Eip1559BlockInput`

packages/web3-common/src/types.ts
```diff
@@ -35,3 +41,145 @@ export interface JsonRpcPayload<T = unknown[]> extends JsonRpcRequest<T> {
...
+	readonly gasLimit: bigint | number;
+	readonly gasUsed: bigint | number;
+	readonly size: bigint | number;
+	readonly timestamp: bigint | number;
+	readonly number?: bigint | number;
+	readonly difficulty?: bigint | number;
+	readonly totalDifficulty?: bigint | number;
+	readonly transactions?: TransactionOutput[];
+	readonly miner?: HexString;
+	readonly baseFeePerGas?: bigint | number;
```
Spacesai1or, line 128: Same as `BlockInput`

packages/web3-common/test/unit/formatters.test.ts
```diff
@@ -0,0 +1,674 @@
...
+} from '../../src/formatters';
+
+import * as formatters from '../../src/formatters';
+
+// TODO: The issue with export * causing issue with jest.spyOn
+// have to hack this way until find resolutions
+jest.mock('web3-utils', () => ({
+	__esModule: true,
+	...jest.requireActual('web3-utils'),
+}));
```
Spacesai1or:
```suggestion
jest.mock('web3-utils');
```

I've tested this works locally, opened a [PR](https://github.com/ChainSafe/web3.js/pull/4460) but forgot the tests wouldn't run there

packages/web3-common/src/errors.ts
```diff
@@ -331,3 +332,7 @@ export class ContractNoFromAddressDefinedError extends Web3Error {
 		super('No "from" address specified in neither the given options, nor the default options.');
 	}
 }
+
+export class FormatterError extends Web3Error {
+	public code = ERR_FORMATTERS;
+}
```
Spacesai1or, line 338: Doing all errors like this forces the user to use the error message to discern specific errors (instead of each error having a specific code). I thought the whole point of `errors.ts` is to have all errors defined there for uniformity and maintainability

packages/web3-common/src/constants.ts
nazarhussain: I always prefer not using `enum` for public facing APIs. Reason is that using string value for end user is more easy. Think of a user who store this value to some storage/object and then on some condition want to perform an action.

```
fetchBlock(user.preferences.blockToFetch)
```

Enum values only works with enum values there is no easy way to coherence a string value to respective enum. Take this as example. You will notice that on last line Typescript is not gonna detect any error. 

```
enum BlockNumbers {
    LATEST = 'latest',
    HEAD = 'head'
}

const bn1:BlockNumbers = BlockNumbers.HEAD;

const bn2:BlockNumbers = 'latest1' as BlockNumbers;
```

For the code which usage is mostly internal or at least very clear then I prefer using enum there.

packages/web3-common/src/types.ts
nazarhussain: This one is handling the extra values that could come in as input. As per `1.x` implementation the object was open to take any properties, so to keep it consistent I had to add this.

packages/web3-common/src/types.ts
nazarhussain: The union with limited properties don't work with dependant cases. You have to specify all attributes in each type and set to `never`, this way TS will detect if one property is present then other should not.

packages/web3-common/src/types.ts
nazarhussain: It depends upon the API usage. e.g. `web3.eth.sendTransaction` does it get all common input and decide on tx type internally then we have to proceed with above. Does it ask user to provide a particular TS type for the transaction type then we can split it.

packages/web3-common/src/errors.ts
nazarhussain: Yes errors are defined for user to know exactly what happened. But we have to group the errors in particular domains, otherwise we will endup having tons of error constructors. e.g. in JS we have `TypeError` that covers any kind of type errors, they don't have particular error constructor for each type. 

Same as we have here particular domain `FormatterError` which can tell user there is some error in formatters. And user can check exact error message to find details. 

If you there there is no domain segregation for other errors do let me know, we should do it.

luu-alex (approved):

packages/web3-common/src/constants.ts
Spacesai1or: In the current implementation, the user would have to know the index of the block number they'd like to use:

```typescript
fetchBlock(PREDEFINED_BLOCK_NUMBERS[1])
```

vs

```typescript
fetchBlock(PredefinedBlockNumbers.latest)
```

Additionally, in the latter example the user would have a suggestion from their IDE with all possible values:

image

By defining the enum as above, the user could still pass `blockToFetch` as a value

```typescript
const user = { preferences: { blockToFetch: PredefinedBlockNumbers.latest } };

function fetchBlock(blockNumber: PredefinedBlockNumbers | number) { ... }

fetchBlock(user.preferences.blockToFetch);
```

The biggest deficiency I can think of with using enums, is the the user would have to import them, but I think the benefits outweigh having to import such a small object

---

If a user is doing:

```typescript
const bn2:BlockNumbers = 'latest1' as BlockNumbers;
```

Then they are purposefully circumventing our defined types, and that's not our concern

packages/web3-common/src/types.ts
Spacesai1or: Hmm I wonder how necessary this is,.I suppose we could end up breaking someone's implementation by removing this, but using Ethereum as intended, there shouldn't be a need for this I don't think?

packages/web3-common/src/types.ts
Spacesai1or: Okay, that makes sense. `sendTransaction` does take any transaction type, and then should validate type based on provided properties (like we're doing now)

packages/web3-common/src/errors.ts
Spacesai1or: My motivation to have each error defined explicitly is to have unique error codes users can search our docs for to find a further explanation of the exact error they came across - something like [Hard Hat's errors](https://hardhat.org/errors/)

But I think in this case, with the formatters, you're probably right to group them with explicit error messages. I just don't like mixing inline errors with the constructors defined in `errors.ts`. It makes how to handle errors in the codebase more ambiguous for devs

nazarhussain: Our approach is similar to Hardhat, See this [reference](https://hardhat.org/errors/#HH8). In this error they cover any error related to configuration. They don't have individual error individual config. Similarly we have one error constructor for any error related to formatters.

packages/web3-common/src/types.ts
nazarhussain: Hopefully there should not be any need of it. But I am not confident considering the existing implementation. May be we revisit that topic while we are doing integration or functional tests.

packages/web3-common/src/types.ts
```diff
@@ -35,3 +42,151 @@ export interface JsonRpcPayload<T = unknown[]> extends JsonRpcRequest<T> {
...
+	readonly currentBlock: bigint | number;
+	readonly highestBlock: bigint | number;
+	readonly knownStates?: bigint | number;
+	readonly pulledStates?: bigint | number;
+}
+
+export enum PredefinedBlockNumbers {
+	LATEST = 'latest',
+	PENDING = 'pending',
+	EARLIEST = 'earliest',
```
Spacesai1or, line 191:
```suggestion
	latest = 'latest',
	pending = 'pending',
	earliest = 'earliest',
```

These need to be lower case to support indexing by enum value, and lowercase should be okay since [Lodestar is doing it](https://github.com/ChainSafe/lodestar/blob/456b50ee628eb60e3daba23cbb596cef41ea6113/packages/lodestar/src/sync/interface.ts#L68)

**NOTE** If you merge this suggestion, we need to update the use of the enums elsewhere

Spacesai1or: Created #4462 to track this

Spacesai1or (approved):
Approving so Nazar can move forward (we're 12hrs apart), but #4462 needs to be addressed

PR #4459, opened 2021-10-12
Create `LocalStorageWallet` class to handle the wallet interaction

### LINKED ISSUE #4431: Create `LocalStorageWallet` class to handle the wallet interaction
### Description

We need to create `LocalStorageWallet` class which implements `BaseWallet` interface. It should have all the logic implemented in the following. 

https://github.com/ChainSafe/web3.js/blob/50afcd0318112b5da12f4944d3fdbdbb2f862ca2/packages/web3-eth-accounts/src/index.js#L617-L621

### Acceptance Criteria

- The class `load` and `save` behaviour works only when `localStorage` is available. Else it should throw error. 
- The class should be exported from `web3-core` package
- Any utility structure for that class should not be exported from the package. 

### Additional Information

- The class should use `scrypt-js` package for any encryption requirement

nazarhussain:
## Description

Create a local storage wallet implementation. 

Fixes #4431

## Type of change

## Checklist:


packages/web3-core/src/local_storage_wallet.ts
```diff
@@ -0,0 +1,145 @@
...
+				? storage
+				: undefined;
+		}
+	}
+
+	public get length() {
+		return Object.keys(this._accounts).length;
+	}
+
+	public create(numberOfAccounts: number, entropy: string) {
```
luu-alex, line 56: Looking at [1.x](https://github.com/ChainSafe/web3.js/blob/50afcd0318112b5da12f4944d3fdbdbb2f862ca2/packages/web3-eth-accounts/src/index.js#L129) if an entropy isn't provided should be generated with `randomHex`.

Was looking at the wrong function oops

packages/web3-core/src/local_storage_wallet.ts
```diff
@@ -0,0 +1,145 @@
...
+	public remove(addressOrIndex: string | number): boolean {
+		if (typeof addressOrIndex === 'string') {
+			delete this._accounts[addressOrIndex];
+			return true;
+		}
+
+		const { address } = Object.values(this._accounts)[addressOrIndex];
+		delete this._accounts[address];
+
+		return true;
```
luu-alex: following [1.x](https://github.com/ChainSafe/web3.js/blob/50afcd0318112b5da12f4944d3fdbdbb2f862ca2/packages/web3-eth-accounts/src/index.js#L691) we should return false if the account does not exist

packages/web3-core/src/local_storage_wallet.ts
```diff
@@ -0,0 +1,147 @@
...
+> extends Web3BaseWallet<T> {
+	private readonly _accounts: { [key: string]: T };
+	private readonly _defaultKeyName = 'web3js_wallet';
+
+	public constructor(accountProvider: Web3AccountProvider<T>) {
+		super(accountProvider);
+		this._accounts = {};
+	}
+
+	public static getStorage(): Storage | undefined {
```
luu-alex, line 23: In [1.x](https://github.com/ChainSafe/web3.js/blob/50afcd0318112b5da12f4944d3fdbdbb2f862ca2/packages/web3-eth-accounts/src/index.js#L768) this function is named `storageAvailable`, perhaps we can make an alias so it's not breaking

luu-alex:
change `local_stroage_wallet.test.ts` -> `local_storage_wallet.test.ts`, otherwise lgtm

packages/web3-core/src/local_storage_wallet.ts
nazarhussain: Its a private function and never exposed outside. So its not a breaking change.

luu-alex (approved):

packages/web3-common/src/web3_base_wallet.ts
```diff
@@ -1,25 +1,39 @@
...
+	privateKeyToAccount: (privateKey: string) => T;
+	create: (entropy: string) => T;
+	decrypt: (keystore: string, password: string, options?: Record<string, unknown>) => T;
+}
+
+export abstract class Web3BaseWallet<T extends Web3BaseWalletAccount> {
+	protected readonly _accountProvider: Web3AccountProvider<T>;
+
+	public constructor(accountProvider: Web3AccountProvider<T>) {
+		// eslint-disable-next-line no-underscore-dangle
```
Spacesai1or, line 22: I prefer the use of `_` for protected/private properties in classes. To refrain from having one-offs where we disable the rule inline, can we just disable the rule for the whole project?

Spacesai1or: Created #4461 to track this

Spacesai1or (approved):
#4461 still needs to be addressed, but approving this so Nazar can move forward since we're 12hrs apart

PR #4478, opened 2021-10-19
Create RequestManager and SubscriptionManager classes

### LINKED ISSUE #4429: Create `RequestManager` class to handle internal web3 requests
### Description

We need to create `RequestManager` class to handle existing internal request handling. This package would migrate all the logic for existing `web3-core-requestmanager` package from `1.x`.

https://github.com/ChainSafe/web3.js/blob/50afcd0318112b5da12f4944d3fdbdbb2f862ca2/packages/web3-core-requestmanager/src/index.js#L40-L46

### Acceptance Criteria

- The class should cover existing logic available in `1.x` package 
- The class should be exported from `web3-core` package
- Any utility structure for that class should not be exported from the package. 

### Additional Information

- The class should not have any dependency on available providers, wherever necessary use `BaseProvider`

### LINKED ISSUE #4430: Create `SubscriptionManager` class to handle internal web3 requests
### Description

We need to create `SubscriptionManager` class to handle existing internal request handling. This package would migrate all the logic for existing `web3-core-subscriptions` package from `1.x`.

https://github.com/ChainSafe/web3.js/blob/50afcd0318112b5da12f4944d3fdbdbb2f862ca2/packages/web3-core-subscriptions/src/index.js#L28-L33

### Acceptance Criteria

- The class should cover existing logic available in `1.x` package 
- The class should be exported from `web3-core` package
- Any utility structure for that class should not be exported from the package. 

### Additional Information

- The class should not have any dependency on available providers, wherever necessary use `BaseProvider`
- The class should use `RequestManager` as dependency if required

nazarhussain:
## Description

Create the implementation of `RequestManager` and `SubscriptionManager`. 

Fixes #4429, #4430

## Type of change

## Checklist:


packages/web3-common/src/types.ts
```diff
@@ -29,9 +29,9 @@ export interface JsonRpcResponseWithResult<T = JsonRpcResult> {
 	readonly result: T;
 }
 
-export type JsonRpcResponse<T = JsonRpcResult> =
-	| JsonRpcResponseWithError
-	| JsonRpcResponseWithResult<T>;
+export type JsonRpcResponse<R = JsonRpcResult, E = JsonRpcResult> =
```
Spacesai1or, line 32: We should formalize how we're going to specify these generics

Currently the rest of the repo uses `T, T2, etc.`, above you use:

```typescript
export class InvalidResponseError<ErrorType = unknown>
```

and now here you're using `R and E`.

My preference would to be explicit and use `ReturnType, ErrorType, etc.`

packages/web3-common/src/web3_base_provider.ts
```diff
@@ -25,22 +19,21 @@ export const JSONRPC_ERR_UNSUPPORTED_METHOD = 4200;
...
+	public static isWeb3Provider(provider: unknown) {
+		return (
+			provider instanceof Web3BaseProvider ||
+			Boolean(provider && (provider as { [symbol]: boolean })[symbol])
+		);
+	}
+
+	// eslint-disable-next-line class-methods-use-this
+	public get [symbol]() {
+		return true;
```
Spacesai1or: Can you add documentation for what's going on here?

packages/web3-core/src/web3_subscription_manager.ts
```diff
@@ -0,0 +1,105 @@
...
+	): Promise<InstanceType<ST[T]>> {
+		if (!this.requestManager.provider) {
+			throw new InvalidProviderError('No provider set');
+		}
+
+		if (!this.supportsSubscriptions()) {
+			throw new Error('The current provider do not support subscriptions');
+		}
+
+		const Klass = this.registeredSubscriptions[name];
```
Spacesai1or, line 36: `Klass`?

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -0,0 +1,201 @@
...
+): provider is LegacySendAsyncProvider => 'sendAsync' in provider;
+
+export class Web3RequestManager extends Web3EventEmitter<{
+	[key in Web3RequestManagerEvent]: SupportedProviders;
+}> {
+	private _requestCounter = 0;
+	private _provider!: SupportedProviders;
+	private readonly _providers: { [key: string]: Web3BaseProviderConstructor };
+
+	public constructor(provider: Web3BaseProvider | string, net: Socket) {
```
Spacesai1or: Should `net` be optional, or `Socket` could be an additional type for `provider`, so:

```typescript
public constructor(provider: Web3BaseProvider | string | Socket) {
```

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -0,0 +1,201 @@
...
+
+export const isLegacySendProvider = (
+	provider: SupportedProviders,
+): provider is LegacySendProvider => 'send' in provider;
+
+export const isLegacySendAsyncProvider = (
+	provider: SupportedProviders,
+): provider is LegacySendAsyncProvider => 'sendAsync' in provider;
+
+export class Web3RequestManager extends Web3EventEmitter<{
```
Spacesai1or, line 25: This should have a public `request` method, no?

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -0,0 +1,201 @@
...
+	[key in Web3RequestManagerEvent]: SupportedProviders;
```
Spacesai1or, line 26: this resolves to type `any`

packages/web3-common/src/types.ts
nazarhussain: I earlier used `T1`, `T2` which is common in TS space. But later I observed with complex generics its better to use full names, so why latest code is using the full name and will be using similar onward.

packages/web3-core/src/web3_subscription_manager.ts
nazarhussain: It returns the `classes` not the the `objects`, so why used this name.

packages/web3-core/src/web3_request_manager.ts
nazarhussain: To use existing code logic made it optional.

packages/web3-core/src/web3_request_manager.ts
nazarhussain: For using existing name logic I used `send` if you prefer `request` we can rename it.

packages/web3-core/src/web3_request_manager.ts
nazarhussain: May be IDE is not loading up correctly. I tried in a test script and used tsc and it works fine.

packages/web3-common/src/types.ts
Spacesai1or: Okay, we need to update all other instances to use full name as well - created [an issue](https://github.com/ChainSafe/web3.js/issues/4499) to track this

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -0,0 +1,168 @@
...
+				throw new ProviderError(`Can't autodetect provider for "${provider}'"`);
+			}
+		}
+
+		this.emit(Web3RequestManagerEvent.BEFORE_PROVIDER_CHANGE, this.provider);
+		this._provider = newProvider ?? null;
+		this.emit(Web3RequestManagerEvent.PROVIDER_CHANGED, this.provider);
+	}
+
+	public async send<ResultType, RequestType = unknown>(request: JsonRpcRequest<RequestType>) {
```
Spacesai1or, line 85: What about the [sendBatch](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-core-requestmanager/src/index.js#L198) method?

nazarhussain: We have a separate issue for this https://github.com/ChainSafe/web3.js/issues/4432

Spacesai1or (approved):

luu-alex (approved):

packages/web3-common/src/web3_event_emitter.ts
```diff
@@ -0,0 +1,28 @@
...
+export interface Web3Emitter<T extends Web3EventMap> {
+	on<K extends Web3EventKey<T>>(eventName: K, fn: Web3EventCallback<T[K]>): void;
+	off<K extends Web3EventKey<T>>(eventName: K, fn: Web3EventCallback<T[K]>): void;
+	emit<K extends Web3EventKey<T>>(eventName: K, params: T[K]): void;
+}
+
+export class Web3EventEmitter<T extends Web3EventMap> implements Web3Emitter<T> {
+	private readonly _emitter = new EventEmitter();
+
+	public on<K extends Web3EventKey<T>>(eventName: K, fn: Web3EventCallback<T[K]>) {
```
jdevcs, line 16: It should return Emitter for chained subscriptions. 

also there must be support of `once, removeListener, removeAllListeners`, ... etc 

IMO, there should be more control to user on EventEmitter for making it flexible, and we can do that by extending EventEmitter by Web3EventEmitter

packages/web3-core/src/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const isLegacySendProvider = (
+	provider: SupportedProviders,
+): provider is LegacySendProvider => 'send' in provider;
+
+export const isLegacySendAsyncProvider = (
+	provider: SupportedProviders,
+): provider is LegacySendAsyncProvider => 'sendAsync' in provider;
+
+export const isSupportSubscriptions = (provider: SupportedProviders): boolean => {
```
jdevcs, line 24: `isSupportSubscriptions` should be enforced by baseProvider and implemented/handled in actual providers only. so I recommend to remove it from web3-core utils.

packages/web3-core/src/web3_config.ts
```diff
@@ -14,13 +15,18 @@ interface ConfigOptions {
...
 
+export enum Web3ConfigEvent {
+	CONFIG_CHANGE = 'CONFIG_CHANGE',
+}
+
+export abstract class Web3Config
+	extends Web3EventEmitter<{ [Web3ConfigEvent.CONFIG_CHANGE]: ConfigEvent<ConfigOptions> }>
+	implements ConfigOptions
+{
 	private _config: ConfigOptions = {
```
jdevcs, line 30: should be named as eth1Config or `executionLayerConfig`

jdevcs (changes requested):

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -0,0 +1,168 @@
...
+			} else if (/^ws(s)?:\/\//i.test(provider)) {
+				newProvider = new this.providers.WebsocketProvider(provider);
+
+				// IPC
+			} else if (typeof net === 'object' && typeof net.connect === 'function') {
+				newProvider = new this.providers.IpcProvider(provider, net);
+			} else {
+				throw new ProviderError(`Can't autodetect provider for "${provider}'"`);
+			}
+		}
```
jdevcs, line 78: if `provider` is not string but injected provider it must also set that.

packages/web3-common/src/web3_event_emitter.ts
nazarhussain: There should be lesser control for user and more for us. There should be more usability for the user in the provided API interfaces. 

e.g. For some of the use cases we don't want user to have `emit` on public interface (instance level) and we just want user to have event listener there. 

This above example provide the fact that we as library author need more control on what interfaces to make available for the users. So why we shouldn't extend from the `EventEmitter`, which will just expose unnecessary and sometime even vulnerabilities to the users.

packages/web3-core/src/utils.ts
nazarhussain: This logic is coupled with the `RequestManager` so why is part of `web3-core`. The `Web3BaseProvider` already have the public abstract interface for it. The function `isSupportSubscriptions` is using that interface to check.

packages/web3-core/src/web3_config.ts
nazarhussain: Its the the private variable, never been exposed to anywhere in the library, so no matter what name to use.

jdevcs: better readability

packages/web3-common/src/web3_event_emitter.ts
jdevcs: how will I as user can listen **once** for event, or remove listeners ?

nazarhussain: For removing listener we have `off` interface and I just added the support for `once` as well.

packages/web3-core/src/web3_config.ts
nazarhussain: I think as of now we have only this set of configuration so lest's just use `Web3Config` name as it is. Once we have some other set of configurations we can refactor.

jdevcs (approved):

PR #4481, opened 2021-10-20
Revert #4437 (@types/bn.js)

Spacesai1or:
Reverts the PR #4481 and we think it was causing dependency issues - will be included in a later release


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.6.0",
+  "version": "1.6.1-rc.0",
```
nazarhussain, line 2: Does this version not already published? `1.6.1-rc.0`, Should not we consider this as `1.6.1-rc.1`?

nazarhussain (changes requested):

jdevcs:
same question as Nazar, also github actions still 'Waiting for status to be reported' after long time.

nazarhussain (approved):

lerna.json
Spacesai1or: No, RC version will get bumped in the `release/1.6.1` branch after `1.x` branch is updated with these changes. Flow is:

New changes -> 1.x -> `release/1.6.1` -> Bump RC version -> release next RC

jdevcs (approved):

PR #4513, opened 2021-10-30
Strong type support for Eth API 

### LINKED ISSUE #4520: Strong type support for Eth API

### Feature Request

Strong types support for Eth JSON RPC API

### Use Cases

Use consistent types everywhere in the project. 

### Implementation Ideas

_No response_

_No response_

- [X] Yes

nazarhussain:
## Description

THis PR provides strong type support for Eth API with flexible API syntax for better maintenance. 

Fixes #4520

## Type of change

## Checklist:


packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+	) => boolean;
+	eth_submitHashrate: (hashRate: HexString32Bytes, id: HexString32Bytes) => boolean;
+
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/sign.json
+	eth_sign: (address: Address, message: HexStringBytes) => HexString256Bytes;
+	eth_signTransaction: (transaction: TransactionWithSender) => HexStringBytes;
+
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/state.json
+	eth_getBalance: (address: Address, block: BlockNumberOrTag) => Uint;
+	eth_getStorage: (
```
Spacesai1or:
```suggestion
	eth_getStorageAt: (
```

I can't link to the exact doc reference, but it's:

```bash
curl -X POST --data '{"jsonrpc":"2.0", "method": "eth_getStorageAt", "params": ["0x295a70b2de5e3953354a6a8344e616ed314d7251", "0x0", "latest"], "id": 1}' localhost:8545
```

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+// Hex encoded 1 byte
+type HexStringBytes = HexString;
+// Hex encoded 256 byte
+type HexString256Bytes = HexString;
+// Hex encoded unsigned integer
+type Uint = HexString;
+// Hex encoded unsigned integer 32 bytes
+type Uint256 = HexString;
+// Hex encoded address
+type Address = HexString;
```
Spacesai1or: These should probably be exported so we can use them elsewhere in the code (like in rpc wrappers)

Also we have [repeats](https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-utils/src/types.ts#L3)

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+
+/* eslint-disable camelcase */
+export type EthExecutionAPI = {
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/block.json
+	eth_getBlockByHash: (blockHash: HexString32Bytes, hydrated: boolean) => Block;
+	eth_getBlockByNumber: (blockNumber: Uint, hydrated: boolean) => Block;
+	eth_getBlockTransactionCountByHash: (blockHash: HexString32Bytes) => Uint[];
+	eth_getBlockTransactionCountByNumber: (blockNumber: Uint) => Uint[];
+	eth_getUncleCountByBlockHash: (blockHash: HexString32Bytes) => Uint[];
+	eth_getUncleCountByBlockNumber: (blockNumber: Uint) => Uint[];
```
Spacesai1or:
```suggestion
	eth_getBlockTransactionCountByHash: (blockHash: HexString32Bytes) => Uint;
	eth_getBlockTransactionCountByNumber: (blockNumber: Uint) => Uint;
	eth_getUncleCountByBlockHash: (blockHash: HexString32Bytes) => Uint;
	eth_getUncleCountByBlockNumber: (blockNumber: BlockNumberOrTag) => Uint;
	eth_getUncleByBlockHashAndIndex: (blockHash: HexString32Bytes, uncleIndex: Uint) => Block;
	eth_getUncleByBlockNumberAndIndex: (blockNumber: BlockNumberOrTag, uncleIndex: Uint) => Block;
```

These are returned as hex strings, I'm not sure why it's listed as an array [here](https://github.com/ethereum/execution-apis/blob/main/src/eth/block.json)

Missing `eth_getUncleByBlockHashAndIndex`, `eth_getUncleByBlockNumberAndIndex: (blockNumber: BlockNumberOrTag, uncleIndex: Uint) => Block;`

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+export type TransactionInfo = TransactionSigned & {
+	readonly blockHash: HexString32Bytes;
+	readonly blockNumber: Uint;
+	readonly from: Address;
+	readonly hash: HexString32Bytes;
+	readonly transactionIndex: Uint;
+};
+
+// https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.json#L24
+export type TransactionWithSender = TransactionUnsigned & { from: Address };
```
Spacesai1or, line 99: Some of the calls which this type is used with (e.g. `eth_call`) `from` is optional

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+export type TransactionHash = HexString;
+export type Uncles = HexString32Bytes[];
+export type BlockTag = 'earliest' | 'latest' | 'pending';
+export type BlockNumberOrTag = Uint | BlockTag;
+
+export interface BaseTransaction {
+	readonly to?: Address;
+	readonly type: HexStringSingleByte;
+	readonly nonce: Uint;
+	readonly gas: Uint;
```
Spacesai1or, line 39: Some of the calls which this type is used with (e.g. `eth_call`) `gas` is optional

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+}
+
+export interface Transaction1559Signed extends Transaction1559Unsigned {
+	readonly yParity: Uint;
+	readonly r: Uint;
+	readonly s: Uint;
+}
+
+export interface Transaction2930Unsigned extends BaseTransaction {
+	readonly gasPrice: Uint;
```
Spacesai1or, line 57: Some of the calls which this type is used with (e.g. `eth_call`) `gasPrice` is optional

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+}
+
+export interface Transaction2930Signed extends Transaction2930Unsigned {
+	readonly yParity: Uint;
+	readonly r: Uint;
+	readonly s: Uint;
+}
+
+export interface TransactionLegacyUnsigned extends BaseTransaction {
+	readonly gasPrice: Uint;
```
Spacesai1or, line 68: Some of the calls which this type is used with (e.g. `eth_call`) `gasPrice` is optional

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+
+// https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.json#L262
+export type TransactionSigned =
+	| Transaction1559Signed
+	| Transaction2930Signed
+	| TransactionLegacySigned;
+
+// https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.json#L269
+export type TransactionInfo = TransactionSigned & {
+	readonly blockHash: HexString32Bytes;
```
spacesailor24, line 89:
```suggestion
	readonly blockHash: HexString32Bytes | null;
```
Can be `null` if transaction is pending

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+	readonly blockNumber: Uint;
```
Spacesai1or:
```suggestion
	readonly blockHash: HexString32Bytes | null;
	readonly blockNumber: Uint | null;
```

Can be `null` if transaction is pending

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+	readonly from: Address;
+	readonly hash: HexString32Bytes;
+	readonly transactionIndex: Uint;
```
Spacesai1or:
```suggestion
	readonly transactionIndex: Uint | null;
```
Can be `null` if transaction is pending

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+	eth_getUncleCountByBlockNumber: (blockNumber: Uint) => Uint[];
+
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/transaction.json
+	eth_getTransactionByHash: (transactionHash: HexString32Bytes) => TransactionInfo;
+	eth_getTransactionByBlockHashAndIndex: (
+		blockHash: HexString32Bytes,
+		transactionIndex: Uint,
+	) => TransactionInfo;
+	eth_getTransactionByBlockNumberAndIndex: (
+		blockNumber: Uint,
```
Spacesai1or:
```suggestion
		blockNumber: BlockNumberOrTag,
```

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// https://github.com/ethereum/execution-apis/blob/main/src/schemas/filter.json#L2
+export type FilterResults = HexString32Bytes[] | Log[];
+
+/* eslint-disable camelcase */
+export type EthExecutionAPI = {
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/block.json
+	eth_getBlockByHash: (blockHash: HexString32Bytes, hydrated: boolean) => Block;
+	eth_getBlockByNumber: (blockNumber: Uint, hydrated: boolean) => Block;
```
Spacesai1or:
```suggestion
	eth_getBlockByNumber: (blockNumber: BlockNumberOrTag, hydrated: boolean) => Block;
```

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+	eth_getBlockTransactionCountByHash: (blockHash: HexString32Bytes) => Uint[];
+	eth_getBlockTransactionCountByNumber: (blockNumber: Uint) => Uint[];
```
Spacesai1or:
```suggestion
	eth_getBlockTransactionCountByNumber: (blockNumber: BlockNumberOrTag) => Uint[];
```

packages/web3-common/src/eth_execution_api.ts
nazarhussain: The reference to the specification says `eth_getStorage`. So I am not sure if the Eth specification are wrong `eth_getStorageAt` is just an extra alias from the implementation side. 

https://github.com/ethereum/execution-apis/blob/7bcb2f082eed6e786486765b8401dddee6ece339/src/eth/state.json#L29

packages/web3-common/src/eth_execution_api.ts
nazarhussain: I intended to not export them because of two reason 1) these are just alias for the native types 2) in our code we should be using concrete types like `Block`, `SyncStatus` etc.

packages/web3-common/src/eth_execution_api.ts
nazarhussain: I could not find `eth_getUncleByBlockHashAndIndex` and `eth_getUncleByBlockNumberAndIndex` anywhere in the specification. Please suggest some reference.

Spacesai1or: I created [an issue](https://github.com/ethereum/execution-apis/issues/96) to ask about the return types

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/client.json
+	eth_protocolVersion: () => string;
+	eth_syncing: () => SyncingStatus;
+	eth_coinbase: () => Address;
+	eth_accounts: () => Address[];
+	eth_blockNumber: () => Uint;
+
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/execute.json
+	eth_call: (transaction: TransactionWithSender) => HexStringBytes;
```
Spacesai1or: https://github.com/ethereum/execution-apis/issues/98

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+
+/* eslint-disable camelcase */
+export type EthExecutionAPI = {
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/block.json
+	eth_getBlockByHash: (blockHash: HexString32Bytes, hydrated: boolean) => Block;
+	eth_getBlockByNumber: (blockNumber: BlockNumberOrTag, hydrated: boolean) => Block;
+	eth_getBlockTransactionCountByHash: (blockHash: HexString32Bytes) => Uint[];
+	eth_getBlockTransactionCountByNumber: (blockNumber: BlockNumberOrTag) => Uint[];
+	eth_getUncleCountByBlockHash: (blockHash: HexString32Bytes) => Uint[];
+	eth_getUncleCountByBlockNumber: (blockNumber: Uint) => Uint[];
```
Spacesai1or: https://github.com/ethereum/execution-apis/issues/96

packages/web3-common/src/eth_execution_api.ts
Spacesai1or: [Issue](https://github.com/ethereum/execution-apis/issues/30) regarding missing uncle methods

packages/web3-common/src/eth_execution_api.ts
Spacesai1or: https://github.com/ethereum/execution-apis/issues/95

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/client.json
+	eth_protocolVersion: () => string;
+	eth_syncing: () => SyncingStatus;
+	eth_coinbase: () => Address;
+	eth_accounts: () => Address[];
+	eth_blockNumber: () => Uint;
+
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/execute.json
+	eth_call: (transaction: TransactionWithSender) => HexStringBytes;
+	eth_estimateGas: (transaction: TransactionWithSender) => Uint;
```
Spacesai1or: https://github.com/ethereum/execution-apis/issues/99

Spacesai1or:
`eth_getCompilers` is missing from new spec, created [an issue](https://github.com/ethereum/execution-apis/issues/100)

Spacesai1or:
`eth_compileSolidity`, `eth_compileLLL`, and `eth_compileSerpent` are missing, created [an issue](https://github.com/ethereum/execution-apis/issues/101)

packages/web3-common/src/eth_execution_api.ts
spacesailor24: https://github.com/ethereum/execution-apis/issues/102

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,266 @@
...
+	readonly from: Address;
+	readonly hash: HexString32Bytes;
+	readonly transactionIndex: Uint | null;
+};
+
+// https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.json#L24
+export type TransactionWithSender = TransactionUnsigned & { from: Address };
+
+// https://github.com/ethereum/execution-apis/blob/main/src/schemas/block.json#L2
+export interface Block {
```
Spacesai1or, line 102: [Issue](https://github.com/ethereum/execution-apis/issues/102) for incorrect spec

packages/web3-common/src/eth_execution_api.ts
nazarhussain: Fixed this for `eth_call`. Please let me know if there is any other explicit case.

packages/web3-common/src/eth_execution_api.ts
nazarhussain: Fixed this for eth_call. Please let me know if there is any other explicit case.

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+
+/* eslint-disable camelcase */
+export type EthExecutionAPI = {
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/block.json
+	eth_getBlockByHash: (blockHash: HexString32Bytes, hydrated: boolean) => Block;
+	eth_getBlockByNumber: (blockNumber: BlockNumberOrTag, hydrated: boolean) => Block;
+	eth_getBlockTransactionCountByHash: (blockHash: HexString32Bytes) => Uint;
+	eth_getBlockTransactionCountByNumber: (blockNumber: BlockNumberOrTag) => Uint;
+	eth_getUncleCountByBlockHash: (blockHash: HexString32Bytes) => Uint;
+	eth_getUncleCountByBlockNumber: (blockNumber: Uint) => Uint;
```
Spacesai1or:
```suggestion
	eth_getUncleCountByBlockNumber: (blockNumber: BlockNumberOrTag) => Uint;
```

Opened an [issue](https://github.com/ethereum/execution-apis/issues/114) for this

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+}
+
+export interface TransactionLegacySigned extends TransactionLegacyUnsigned {
+	readonly v: Uint;
+	readonly r: Uint;
+	readonly s: Uint;
+}
+
+// https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.json#L178
+export type TransactionUnsigned =
```
Spacesai1or, line 78: Possible optional parameters, creating an [issue](https://github.com/ethereum/execution-apis/issues/115) for this

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/client.json
+	eth_protocolVersion: () => string;
+	eth_syncing: () => SyncingStatus;
+	eth_coinbase: () => Address;
+	eth_accounts: () => Address[];
+	eth_blockNumber: () => Uint;
+
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/execute.json
+	eth_call: (transaction: TransactionCall, blockNumber: BlockNumberOrTag) => HexStringBytes;
+	eth_estimateGas: (transaction: TransactionWithSender, blockNumber: BlockNumberOrTag) => Uint;
```
Spacesai1or: Legacy spec mentions that all parameters for `transaction` are optional, created an [issue](https://github.com/ethereum/execution-apis/issues/116) for this

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+	readonly from?: Address;
+	readonly to: Address;
+	readonly gas?: Uint;
+	readonly gasPrice?: Uint;
+	readonly value?: Uint;
+	readonly data?: HexString;
+}
+
+export interface BaseTransaction {
+	readonly to?: Address;
```
Spacesai1or: For transaction object that are returned from calls like `eth_getTransactionByHash`, the `to` field can be `null` when the transaction creates a contract. We have `to` listed as optional, but that does not include the possible `null` value

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+	readonly transactionIndex: HexString32Bytes;
+	readonly blockHash?: HexString32Bytes;
+	readonly blockNumber?: Uint;
+	readonly from?: Address;
+	readonly to?: Address;
+	readonly cumulativeGasUsed?: Uint;
+	readonly gasUsed?: Uint;
+	readonly contractAddress?: Address;
+	readonly logs?: Log[];
+	readonly logsBloom?: HexString256Bytes;
```
Spacesai1or: These fields probably shouldn't be optional, created an [issue](https://github.com/ethereum/execution-apis/issues/117) for this

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+/* eslint-disable camelcase */
+export type EthExecutionAPI = {
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/block.json
+	eth_getBlockByHash: (blockHash: HexString32Bytes, hydrated: boolean) => Block;
+	eth_getBlockByNumber: (blockNumber: BlockNumberOrTag, hydrated: boolean) => Block;
+	eth_getBlockTransactionCountByHash: (blockHash: HexString32Bytes) => Uint;
+	eth_getBlockTransactionCountByNumber: (blockNumber: BlockNumberOrTag) => Uint;
+	eth_getUncleCountByBlockHash: (blockHash: HexString32Bytes) => Uint;
+	eth_getUncleCountByBlockNumber: (blockNumber: Uint) => Uint;
+	eth_getUncleByBlockHashAndIndex: (blockNumber: BlockNumberOrTag, uncleIndex: Uint) => Block;
```
Spacesai1or:
```suggestion
	eth_getUncleByBlockHashAndIndex: (blockHash: HexString32Bytes, uncleIndex: Uint) => Block;
```

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+export interface FeeHistoryResult {
+	readonly oldestBlock: Uint;
+	readonly baseFeePerGas: Uint;
+	readonly reward: number[][];
+}
+
+// https://github.com/ethereum/execution-apis/blob/main/src/schemas/filter.json#L28
+export interface Filter {
+	readonly fromBlock?: Uint;
+	readonly toBlock?: Uint;
```
Spacesai1or:
```suggestion
	readonly fromBlock?: BlockNumberOrTag;
	readonly toBlock?: BlockNumberOrTag;
```

Created an [issue](https://github.com/ethereum/execution-apis/blob/bea0266c42919a2fb3ee524fb91e624a23bc17c5/src/schemas/filter.json#L32) for this

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+	readonly address?: Address | Address[];
+	readonly topics?: Topic[];
```
Spacesai1or:
```suggestion
	readonly topics?: Topic | Topic[] | null [];
```

Example from legacy spec:

```bash
"topics": ["0x000000000000000000000000a94f5374fce5edbc8e2a8697c15331677e6ebf0b", null, ["0x000000000000000000000000a94f5374fce5edbc8e2a8697c15331677e6ebf0b", "0x0000000000000000000000000aff3454fce5edbc8cca8697c15331677e6ebccc"]]
```

Created an [issue](https://github.com/ethereum/execution-apis/issues/118) for this

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+export type Topic = HexString256Bytes;
+
+// https://github.com/ethereum/execution-apis/blob/main/src/schemas/receipt.json#L2
+export interface Log {
+	readonly removed?: boolean;
+	readonly logIndex?: Uint;
+	readonly transactionIndex?: Uint;
+	readonly transactionHash?: HexString32Bytes;
+	readonly blockHash?: HexString32Bytes;
+	readonly blockNumber?: Uint;
```
Spacesai1or:
```suggestion
	readonly logIndex?: Uint | null;
	readonly transactionIndex?: Uint | null;
	readonly transactionHash?: HexString32Bytes | null;
	readonly blockHash?: HexString32Bytes | null;
	readonly blockNumber?: Uint | null;
```

Legacy spec:

image

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+	eth_getLogs: (filter: Filter) => FilterResults;
+
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/mining.json
+	eth_mining: () => boolean;
+	eth_hashrate: () => Uint;
+	eth_getWork: () => [HexString32Bytes, HexString32Bytes, HexString32Bytes];
+	eth_submitWork: (
+		powHash: HexString32Bytes,
+		seedHash: HexString32Bytes,
+		difficulty: HexString32Bytes,
```
Spacesai1or: These might be incorrect, created an [issue](https://github.com/ethereum/execution-apis/issues/119)

Spacesai1or (changes requested):

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+	) => boolean;
+	eth_submitHashrate: (hashRate: HexString32Bytes, id: HexString32Bytes) => boolean;
+
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/sign.json
+	eth_sign: (address: Address, message: HexStringBytes) => HexString256Bytes;
+	eth_signTransaction: (transaction: TransactionWithSender) => HexStringBytes;
+
+	// https://github.com/ethereum/execution-apis/blob/main/src/eth/state.json
+	eth_getBalance: (address: Address, block: BlockNumberOrTag) => Uint;
```
Spacesai1or:
```suggestion
	eth_getBalance: (address: Address, blockNumber: BlockNumberOrTag) => Uint;
```

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+	eth_getStorageAt: (
+		address: Address,
+		storageSlot: Uint256,
+		block: BlockNumberOrTag,
```
Spacesai1or:
```suggestion
		blockNumber: BlockNumberOrTag,
```

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,293 @@
...
+	) => HexStringBytes;
+	eth_getTransactionCount: (address: Address, block: BlockNumberOrTag) => Uint;
+	eth_getCode: (address: Address, block: BlockNumberOrTag) => HexStringBytes;
```
Spacesai1or:
```suggestion
	eth_getTransactionCount: (address: Address, blockNumber: BlockNumberOrTag) => Uint;
	eth_getCode: (address: Address, blockNumber: BlockNumberOrTag) => HexStringBytes;
```

Spacesai1or (approved):

packages/web3-common/src/web3_base_provider.ts
```diff
@@ -23,7 +31,7 @@ const symbol = Symbol.for('web3/base-provider');
 
 // Provider interface compatible with EIP-1193
 // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md
-export abstract class Web3BaseProvider {
+export abstract class Web3BaseProvider<API extends Web3APISpec = EthExecutionAPI> {
```
jdevcs, line 34: EthExecutionAPI must not be enforced at providers level, as in future we might integrate other APIs. 
Filtering transport for specific API functions will make it harder to use for future APIs.

packages/web3-common/src/web3_base_provider.ts
```diff
@@ -44,10 +52,13 @@ export abstract class Web3BaseProvider {
 	abstract supportsSubscriptions(): boolean;
 
 	// https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#request
-	abstract request<T = JsonRpcResponse, T2 = unknown[], T3 = unknown>(
-		payload: JsonRpcPayload<T2>,
-		providerOptions?: T3,
-	): Promise<JsonRpcResponse<T>>;
+	abstract request<
```
jdevcs, line 55: same for request method, lets not make this API specific, I know its extendable to more JSON RPC APIs but it takes away flexibility.

packages/web3-core/src/web3_subscriptions.ts
```diff
@@ -77,13 +77,16 @@ export abstract class Web3Subscription<
 	}
 
 	public async unsubscribe() {
-		await this._requestManager.send<string>({
+		if (!this.id) {
```
jdevcs, line 80: Its not optional `private _id?: HexString;`

so you can adjust this in id def and in this function as well.

packages/web3-providers-http/src/index.ts
```diff
@@ -1,17 +1,20 @@
...
-	JsonRpcResponse,
 	Web3BaseProviderStatus,
-	InvalidClientError,
 } from 'web3-common';
-import fetch from 'cross-fetch';
-
 import { HttpProviderOptions } from './types';
 
-export class HttpProvider extends Web3BaseProvider {
+export class HttpProvider<API extends Web3APISpec = EthExecutionAPI> extends Web3BaseProvider<API> {
```
jdevcs, line 17: same here, change in design request, we shouldn't enforce provider as only Eth Provider.

packages/web3-providers-ws/src/index.ts
```diff
@@ -1,25 +1,32 @@
...
+	Web3BaseProviderStatus,
 } from 'web3-common';
-import { ClientRequestArgs } from 'http';
-import { ClientOptions, WebSocket, MessageEvent, CloseEvent } from 'isomorphic-ws';
-import { ReconnectOptions, WSRequestItem } from './types';
 import { DeferredPromise } from './deferred_promise';
+import { ReconnectOptions, WSRequestItem } from './types';
 
-export default class WebSocketProvider extends Web3BaseProvider {
+export default class WebSocketProvider<
```
jdevcs, line 27: + same as http provider, it is not only eth provider.

jdevcs (changes requested):

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,298 @@
...
+	readonly receiptsRoot: HexString32Bytes;
+	readonly logsBloom: HexString256Bytes | null;
+	readonly difficulty?: Uint;
+	readonly number: Uint | null;
+	readonly gasLimit: Uint;
+	readonly gasUsed: Uint;
+	readonly timestamp: Uint;
+	readonly extraData: HexStringBytes;
+	readonly mixHash: HexString32Bytes;
+	readonly nonce: HexStringBytes | null;
```
Spacesai1or:
```suggestion
	readonly nonce: Uint | null;
```

Spacesai1or (changes requested):

Spacesai1or (approved):

packages/web3-core/src/web3_subscriptions.ts
nazarhussain: It's not optional but its initialized after you subscribed and set to `undefined` when you `unsubscribe`. So this check make sure you don't unsubscribe a non-subscribed or already unsubscribe subscription.

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -64,22 +70,22 @@ export class Web3RequestManager extends Web3EventEmitter<{
...
-		let newProvider!: Web3BaseProvider;
+	public setProvider(provider: SupportedProviders<API> | string, net?: Socket) {
+		let newProvider!: Web3BaseProvider<API>;
 
 		// autodetect provider
 		if (provider && typeof provider === 'string' && this.providers) {
 			// HTTP
 			if (/^http(s)?:\/\//i.test(provider)) {
-				newProvider = new this.providers.HttpProvider(provider);
+				newProvider = new this.providers.HttpProvider<API>(provider);
```
jdevcs, line 80: if user is passing string how `<API>` is used?

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -64,22 +70,22 @@ export class Web3RequestManager extends Web3EventEmitter<{
 		return this._providers;
 	}
 
-	public setProvider(provider: SupportedProviders | string, net?: Socket) {
-		let newProvider!: Web3BaseProvider;
+	public setProvider(provider: SupportedProviders<API> | string, net?: Socket) {
```
jdevcs, line 73: added issue for validating current design choices https://github.com/ChainSafe/web3.js/issues/4543

packages/web3-common/src/deferred_promise.ts
```diff
@@ -48,6 +48,7 @@ export class DeferredPromise<T> implements Promise<T> {
 	}
 
 	public async catch<TResult>(
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
 		onrejected?: (reason: any) => TResult | PromiseLike<TResult>,
```
jdevcs, line 52: instead of any could use `unknown` ?

packages/web3-utils/src/types.ts
```diff
@@ -1,8 +1,24 @@
 export type HexString = string;
 export type ValueTypes = 'address' | 'bool' | 'string' | 'int256' | 'uint256' | 'bytes' | 'bigint';
-export type Address = string;
 export type Bytes = Buffer | Uint8Array | ArrayBuffer | HexString;
 export type Numbers = number | bigint | string | HexString;
+// Hex encoded 32 bytes
+export type HexString32Bytes = HexString;
```
jdevcs, line 6: All these types should be in web3-common, or along with `eth_execution_api.ts ` if eth_execution_api.ts moved to any other package in future.

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,298 @@
...
+	readonly nonce: Uint;
+	readonly gas: Uint;
+	readonly value: Uint;
+	readonly input: HexStringBytes;
+}
+
+export interface Transaction1559Unsigned extends BaseTransaction {
+	readonly maxFeePerGas: Uint;
+	readonly maxPriorityFeePerGas: Uint;
+	readonly accessList: AccessList;
```
jdevcs, line 47: chainId is missing: `"chainId": {
				"title": "chainId",
				"description": "Chain ID that this transaction is valid on.",
				"$ref": "#/components/schemas/uint"
			}`

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,298 @@
...
+	readonly data?: HexString;
+}
+
+export interface BaseTransaction {
+	readonly to?: Address | null;
+	readonly type: HexStringSingleByte;
+	readonly nonce: Uint;
+	readonly gas: Uint;
+	readonly value: Uint;
+	readonly input: HexStringBytes;
```
jdevcs, line 41: chainId is missing: `"chainId": {
				"title": "chainId",
				"description": "Chain ID that this transaction is valid on.",
				"$ref": "#/components/schemas/uint"
			}`

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,298 @@
...
+}
+
+export interface Transaction1559Unsigned extends BaseTransaction {
+	readonly maxFeePerGas: Uint;
```
jdevcs, line 45: good to specify type as` readonly type: '0x2' ;`

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,298 @@
...
+
+export interface Transaction1559Signed extends Transaction1559Unsigned {
+	readonly yParity: Uint;
+	readonly r: Uint;
+	readonly s: Uint;
+}
+
+export interface Transaction2930Unsigned extends BaseTransaction {
+	readonly gasPrice: Uint;
+	readonly accessList: AccessList;
```
jdevcs, line 58: also here specify type `readonly type: '0x1' ;`

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -0,0 +1,298 @@
...
+// https://github.com/ethereum/execution-apis/blob/main/src/eth/fee_market.json#L53
+export interface FeeHistoryResult {
+	readonly oldestBlock: Uint;
+	readonly baseFeePerGas: Uint;
+	readonly reward: number[][];
+}
+
+// https://github.com/ethereum/execution-apis/blob/main/src/schemas/filter.json#L28
+export interface Filter {
+	readonly fromBlock?: BlockNumberOrTag;
```
jdevcs, line 173: seems it only accepts number `unit` instead of both number and Tag https://github.com/ethereum/execution-apis/blob/main/src/schemas/filter.json#L34

packages/web3-core/src/web3_request_manager.ts
nazarhussain: The string refers to the connection string, not provider type. The `API` here refers to the generic value for the newly created type.

packages/web3-core/src/web3_request_manager.ts
nazarhussain: We need integration tests for every thing, not just this part.

packages/web3-common/src/deferred_promise.ts
nazarhussain: It's been enforced by the inherited native Promise api.

packages/web3-utils/src/types.ts
nazarhussain: These are general types or to say particularly type aliases. Not necessary to be used only with Eth API so why these are in `web3-utils`

packages/web3-common/src/eth_execution_api.ts
nazarhussain: This attribute is not available in the old specs which we are trying to be consistent with https://eth.wiki/json-rpc/API

packages/web3-common/src/eth_execution_api.ts
nazarhussain: `type` could be the input parameter for creating transaction but not the part of the spec for the transaction. Neither on the [legacy specs](https://eth.wiki/json-rpc/API) and not on [execution specs](https://github.com/ethereum/execution-apis).

packages/web3-common/src/eth_execution_api.ts
nazarhussain: Same as above.

packages/web3-common/src/eth_execution_api.ts
nazarhussain: Its accepts the number and tag both, the issue been created on the execution-api. https://github.com/ethereum/execution-apis/issues/118

packages/web3-common/src/eth_execution_api.ts
jdevcs: I think lets follow consistent specs https://github.com/ethereum/execution-apis and make chainId optional

packages/web3-common/src/eth_execution_api.ts
jdevcs: Its specified in EIPs 1559 and 2930

packages/web3-core/src/web3_request_manager.ts
jdevcs: thats exactly my question :) if user is passing string ( connection string ) , this code will use `API` generic type provider by default? and user will not be able to specify it via param?

packages/web3-utils/src/types.ts
jdevcs: ok, if its not only used in Eth API, still web3-common will be good candidate for holding common types. I think this will make dependency less complicated.

nazarhussain:
@jdevcs  In reference to all feedback related to Execution API spec. 

- EthExecutionAPI is not enforced anywhere, rather used as default value of the generic.
- The consumer of the providers can pass any APISpec for generic to use any other API.
- The API Spec design also not enforce Eth pattern, rather JSON-RPC standard. As soon one consume any JSON-RPC compatible API, one can use same provider without any issue with same type safety.
- As already discussed we will move EthExecutionAPI to the web3-eth package in up-coming PRs. That will also resolve the separation of concern.

jdevcs (approved):

PR #4514, opened 2021-11-01
web3 utils validation

### LINKED ISSUE #4351: Migrate validation utils functions from 1.x
### Description

Migrate the following functions form `web3-utils` 1.x branch. 

- isHex
- isHexStrict
- isBloom
- isUserEthereumAddressInBloom
- isContractAddressInBloom
- isTopic
- isTopicInBloom
- isInBloom
- compareBlockNumbers
- checkAddressChecksum

### Acceptance Criteria

- All the functions mentioned above should migrated 
- All the functions must be covered with their unit tests 

### Additional Information

- Don't use `BigNumber` where needed instead use native `BigInt`

luu-alex:
## Description
- isHex
- isHexStrict
- isBloom
- isUserEthereumAddressInBloom
- isContractAddressInBloom
- isTopic
- isTopicInBloom
- isInBloom
- compareBlockNumbers
- checkAddressChecksum
- isAddress

## New functons
- validateHexStringInput
- validateBytesInput
- validateNumbersInput
- validateStringInput
- isBigInt
## Removed functions
- isBN

## Relevant links:
https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-utils/src/utils.js
https://github.com/joshstevens19/ethereum-bloom-filters/blob/master/src/index.ts

# Acceptance Criteria

- All the functions mentioned above should migrated
- All the functions must be covered with their unit tests


nazarhussain (approved):

packages/web3-utils/src/types.ts
```diff
@@ -4,11 +4,11 @@ export type Address = string;
 export type Bytes = Buffer | Uint8Array | ArrayBuffer | HexString;
 export type Numbers = number | bigint | string | HexString;
 export type EncodingTypes = Numbers | boolean | Numbers[] | boolean[];
-export type typedObject = {
+export type TypedObject = {
 	type: string;
 	value: EncodingTypes;
 };
-export type typedObject2 = {
+export type TypedObject2 = {
```
Spacesai1or:
```suggestion
export type TypedObjectAbbreviated = {
```

We should rename this to be more descriptive, doesn't have to be what I have above

packages/web3-utils/src/validation.ts
```diff
@@ -7,39 +9,37 @@ import {
...
-
 export const isHexStrict = (hex: string) =>
 	typeof hex === 'string' && /^(-)?0x[0-9a-f]*$/i.test(hex);
 
+/**
+ * returns true if input is a hexstring, number or bigint
+ */
+export const isHex = (hex: Numbers): boolean =>
+	typeof hex === 'number' ||
+	typeof hex === 'bigint' ||
```
Spacesai1or, line 28: I think these checks are misleading as integers not intended to be interpreted as hex and floats will return `true`.
We could force users to prefix their literal hex numbers with `0x` as mentioned [here](https://stackoverflow.com/a/41333753)

Regarding `bigint`s, I'm not sure what to do besides return `false` as `bigint`s will never be in hex format (they can be converted quite easily, but that's beside the point for a `isHex` function)

packages/web3-utils/test/fixtures/validation.ts
```diff
@@ -0,0 +1,275 @@
+import { Numbers } from '../../src/types';
+
+export const hexStrict: [any, boolean][] = [
+	['0x48', true],
+	['0x123c', true],
+	['0x0dec0518fa672a70027b04c286582e543ab17319fbdd384fa7bc8f3d5a542c0b', true],
+	['0xd115bffabbdd893a6f7cea402e7338643ced44a6', true],
```
Spacesai1or, line 7:
```suggestion
	['0xd115bffabbdd893a6f7cea402e7338643ced44a6', true],
	['0x2C941171bD2A7aEda7c2767c438DfF36EAaFdaFc', true],
```

Adding a checksummed ETH address

Spacesai1or (changes requested):

packages/web3-utils/src/validation.ts
luu-alex: Yea I agree bigint should return `false`.
From the post it seems like we can't check if the typeof `number` is prefixed with a `0x`. 
It seems like in `1.x` `isHex` returns true if the value is typeof `number` so it'll be a breaking change.
We can change it so it only accepts strings with a `0x` prefix, that'll have the exact same functionality as `isHexStrict` function

Spacesai1or: I misread the post I linked, it would be possible if we did `.toString()`, but I think your point about it being a breaking change is the most important detail

Spacesai1or (approved):

PR #4525, opened 2021-11-09
ETH RPC Wrappers

### LINKED ISSUE #4483: Implement JSON-RPC method wrappers
- Includes parameter validation and in/output formatting
---
- [ ] `web3_clientVersion`
- [ ] `web3_sha3`
---
- [x] `eth_protocolVersion`
- [x] `eth_syncing`
- [x] `eth_coinbase`
- [x] `eth_mining`
- [x] `eth_hashrate`
- [x] `eth_gasPrice`
- [x] `eth_accounts`
- [x] `eth_blockNumber`
- [x] `eth_getBalance`
- [x] `eth_getStorageAt`
- [x] `eth_getTransactionCount`
- [x] `eth_getBlockTransactionCountByHash`
- [x] `eth_getBlockTransactionCountByNumber`
- [x] `eth_getUncleCountByBlockHash`
- [x] `eth_getUncleCountByBlockNumber`
- [x] `eth_getCode`
- [x] `eth_sign`
- [x] `eth_signTransaction`
- [x] `eth_sendTransaction`
- [x] `eth_sendRawTransaction`
- [x] `eth_call`
- [x] `eth_estimateGas`
- [x] `eth_getBlockByHash`
- [x] `eth_getBlockByNumber`
- [x] `eth_getTransactionByHash`
- [x] `eth_getTransactionByBlockHashAndIndex`
- [x] `eth_getTransactionByBlockNumberAndIndex`
- [x] `eth_getTransactionReceipt`
- [x] `eth_getUncleByBlockHashAndIndex`
- [x] `eth_getUncleByBlockNumberAndIndex`
- [x] `eth_getCompilers`
- [x] `eth_compileLLL`
- [x] `eth_compileSolidity`
- [x] `eth_compileSerpent`
- [x] `eth_newFilter`
- [x] `eth_newBlockFilter`
- [x] `eth_newPendingTransactionFilter`
- [x] `eth_uninstallFilter`
- [x] `eth_getFilterChanges`
- [x] `eth_getFilterLogs`
- [x] `eth_getLogs`
- [x] `eth_getWork`
- [x] `eth_submitWork`
- [x] `eth_submitHashrate`

### Not being implemented

- `db_putString`
- `db_getString`
- `db_putHex`
- `db_getHex`
- `shh_post`
- `shh_version`
- `shh_newIdentity`
- `shh_hasIdentity`
- `shh_newGroup`
- `shh_addToGroup`
- `shh_newFilter`
- `shh_uninstallFilter`
- `shh_getFilterChanges`
- `shh_getMessages`

Spacesai1or:
Created an [issue](https://github.com/ChainSafe/web3.js/issues/4608) to track remaining TODOs

closes #4483


Lucid36:
995b9dd0faf900b315cc7c27435d1ca4d7fb5bd6

packages/web3-core/src/web3_context.ts
```diff
@@ -19,6 +19,8 @@ export class Web3Context<API extends Web3APISpec> extends Web3Config {
 		return this.requestManager.provider;
 	}
 
+	// TODO Getting: warning  Unexpected any. Specify a different type
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Spacesai1or: This randomly started throwing, not sure what to do about it @nazarhussain

tools/eslint-config-web3-base/ts.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
 		'no-unused-expressions': ['error'],
 		'no-continue': 'off',
 		'no-underscore-dangle': 'off',
+		'no-nested-ternary': 'off',
```
Spacesai1or:
```suggestion
```

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -20,9 +23,8 @@ export interface AccessListEntry {
 export type AccessList = AccessListEntry[];
 export type TransactionHash = HexString;
 export type Uncles = HexString32Bytes[];
-export type BlockTag = 'earliest' | 'latest' | 'pending';
-export type BlockNumberOrTag = Uint | BlockTag;
 
+// TODO Should probably support EIP-2930 and EIP-1559
```
Spacesai1or, line 27: @nazarhussain Do you disagree? Probably need to create an issue in ETH spec repo

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -20,9 +23,8 @@ export interface AccessListEntry {
 export type AccessList = AccessListEntry[];
 export type TransactionHash = HexString;
 export type Uncles = HexString32Bytes[];
-export type BlockTag = 'earliest' | 'latest' | 'pending';
-export type BlockNumberOrTag = Uint | BlockTag;
```
Spacesai1or, line 24: Renamed and moved to [web3-utils](https://github.com/ChainSafe/web3.js/blob/c7933b378db2e220ca0444f3481e1d0c4cb16d04/packages/web3-utils/src/types.ts#L25)

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -168,14 +167,6 @@ export interface FeeHistoryResult {
...
 }
 
-// https://github.com/ethereum/execution-apis/blob/main/src/schemas/filter.json#L28
-export interface Filter {
-	readonly fromBlock?: BlockNumberOrTag;
-	readonly toBlock?: BlockNumberOrTag;
-	readonly address?: Address | Address[];
-	readonly topics?: (Topic | Topic[] | null)[];
-}
-
```
Spacesai1or, line 178: Moved to [web3-utils](https://github.com/ChainSafe/web3.js/blob/c7933b378db2e220ca0444f3481e1d0c4cb16d04/packages/web3-utils/src/types.ts#L103)

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -123,9 +125,6 @@ export interface Block {
 	readonly hash: HexString32Bytes | null;
 }
 
-// https://github.com/ethereum/execution-apis/blob/main/src/schemas/filter.json#L59
-export type Topic = HexString256Bytes;
-
```
Spacesai1or, line 128: Moved to [web3-utils](https://github.com/ChainSafe/web3.js/blob/c7933b378db2e220ca0444f3481e1d0c4cb16d04/packages/web3-utils/src/types.ts#L23)

packages/web3-common/src/formatters.ts
```diff
@@ -47,14 +48,6 @@ export const outputProofFormatter = (proof: Proof): Proof => ({
...
 export const outputBigIntegerFormatter = (number: Numbers) => toNumber(number);
 
-/**
- * Returns true if the given blockNumber is 'latest', 'pending', or 'earliest.
- */
-export const isPredefinedBlockNumber = (blockNumber: string) =>
-	PredefinedBlockNumbers.LATEST === blockNumber ||
-	PredefinedBlockNumbers.PENDING === blockNumber ||
-	PredefinedBlockNumbers.EARLIEST === blockNumber;
-
```
Spacesai1or, line 60: Renamed and moved to [web3-utils](https://github.com/ChainSafe/web3.js/blob/c7933b378db2e220ca0444f3481e1d0c4cb16d04/packages/web3-utils/src/validation.ts#L419)

packages/web3-common/src/types.ts
```diff
@@ -207,12 +207,6 @@ export interface SyncOutput {
 	readonly pulledStates?: bigint | number;
 }
 
-export enum PredefinedBlockNumbers {
-	LATEST = 'latest',
-	PENDING = 'pending',
-	EARLIEST = 'earliest',
-}
-
```
Spacesai1or, line 220: Renamed and moved to [web3-utils](https://github.com/ChainSafe/web3.js/blob/c7933b378db2e220ca0444f3481e1d0c4cb16d04/packages/web3-utils/src/types.ts#L25)

packages/web3-common/test/unit/formatters.test.ts
```diff
@@ -71,36 +69,17 @@ describe('formatters', () => {
...
-
-		it('should return true for "earliest"', () => {
-			expect(isPredefinedBlockNumber('earliest')).toBeTruthy();
-		});
-
-		it('should return false for "unknown"', () => {
-			expect(isPredefinedBlockNumber('unknown')).toBeFalsy();
-		});
-	});
-
```
Spacesai1or, line 91: Moved to [web3-utils](https://github.com/ChainSafe/web3.js/blob/c7933b378db2e220ca0444f3481e1d0c4cb16d04/packages/web3-utils/test/unit/validation.test.ts#L203)

packages/web3-eth/src/convertible_properties.ts
```diff
@@ -0,0 +1,42 @@
...
+	'blockNumber',
+	'cumulativeGasUsed',
+	'gasUsed',
+	'status',
+];
+
+export const convertibleFeeHistoryResultProperties: (keyof FeeHistoryResult)[] = [
+	'oldestBlock',
+	'baseFeePerGas',
+];
```
Spacesai1or: These are manually defined properties that are convertible from `HexString` to `number | NumberString | bigint` for each relevant RPC method

packages/web3-eth/src/errors.ts
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable max-classes-per-file */
+
+import { Web3Error } from 'web3-utils';
+
+export class InvalidTransactionWithSender extends Web3Error {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	public constructor(value: any) {
+		// TODO Discuss this naive approach to logging object
+		// Does not account for non JSON properties
+		super(JSON.stringify(value), 'invalid transaction with sender');
```
Spacesai1or, line 10: @nazarhussain Any idea how to do this better? The values passed to these error constructors really is `any` as we're throwing an error because we came across an unexpected value. If a `bigint` was passed here this would throw an internal uncaught exception

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+import {
+	convertibleBlockProperties,
+	convertibleFeeHistoryResultProperties,
+	convertibleReceiptInfoProperties,
+	convertibleTransactionInfoProperties,
+} from './convertible_properties';
+
+import * as RpcMethods from './rpc_methods';
+
+export class Web3Eth {
```
Spacesai1or: Mentioned in PR description, but the `returnType` functionality is untested and the formatting of user input is not implemented for below methods

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+		blockNumber: BlockNumberOrTag = this._options.defaultBlock,
+		returnType?: ReturnType,
+	) {
+		const response = await RpcMethods.getBalance(this._requestManager, address, blockNumber);
+
+		return convertToValidType(
+			response,
+			returnType ?? this._options.defaultReturnType,
+		) as ValidReturnTypes[ReturnType];
+	}
```
Spacesai1or: Wanted to get everyone's input on the use of multiple optional parameters. For this method, if a user wanted to default `blockNumber`, but wanted to specify `returnType`, they'd need to call the methods like so:

```typescript
getBalance('0x0...', undefined, ReturnType.BigInt);
```

I'm not sure how else to do this, just wanted to point it out because this is something we'll have to make clear in our documentation

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+		const response = isHexString32Bytes(block)
+			? await RpcMethods.getBlockByHash(this._requestManager, block, hydrated)
+			: await RpcMethods.getBlockByNumber(this._requestManager, block, hydrated);
+
+		return convertObjectPropertiesToValidType(
+			response,
+			convertibleBlockProperties,
+			returnType ?? this._options.defaultReturnType,
+		);
+	}
```
Spacesai1or: Another example of multiple optional parameters

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+	) {
+		const response = isHexString32Bytes(block)
+			? await RpcMethods.getBlockTransactionCountByHash(this._requestManager, block)
+			: await RpcMethods.getBlockTransactionCountByNumber(this._requestManager, block);
+
+		return convertToValidType(
+			response,
+			returnType ?? this._options.defaultReturnType,
+		) as ValidReturnTypes[ReturnType];
+	}
```
Spacesai1or: Another example of multiple optional parameters

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+	) {
+		const response = isHexString32Bytes(block)
+			? await RpcMethods.getUncleCountByBlockHash(this._requestManager, block)
+			: await RpcMethods.getUncleCountByBlockNumber(this._requestManager, block);
+
+		return convertToValidType(
+			response,
+			returnType ?? this._options.defaultReturnType,
+		) as ValidReturnTypes[ReturnType];
+	}
```
Spacesai1or: Another example of multiple optional parameters

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+					// @ts-expect-error
+					convertibleTransactionInfoProperties,
+					returnType ?? this._options.defaultReturnType,
+			  );
+	}
+
+	// TODO Can't find in spec
+	// public async getPendingTransactions() {
+
+	// }
```
Spacesai1or: [Docs for 1.x](https://web3js.readthedocs.io/en/v1.2.11/web3-eth.html#getpendingtransactions)

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+			this._requestManager,
+			address,
+			blockNumber,
+		);
+
+		return convertToValidType(
+			response,
+			returnType ?? this._options.defaultReturnType,
+		) as ValidReturnTypes[ReturnType];
+	}
```
Spacesai1or: Another example of multiple optional parameters

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+		seedHash: HexString32Bytes,
+		difficulty: HexString32Bytes,
+	) {
+		return RpcMethods.submitWork(this._requestManager, nonce, seedHash, difficulty);
+	}
+
+	// TODO
+	// public async requestAccounts() {
+
+	// }
```
Spacesai1or: - [1.x implementation](https://github.com/ChainSafe/web3.js/blob/3a3cb32b6434656131f0bfc4a8622130d19d2338/packages/web3-eth/src/index.js#L569)
- [1.x docs](https://web3js.readthedocs.io/en/v1.2.11/web3-eth.html#requestaccounts)
- Apart of [EIP-1102 ](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1102.md)

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+
+	// TODO
+	// public async getChainId() {
+
+	// }
```
Spacesai1or: - [1.x docs](https://web3js.readthedocs.io/en/v1.2.11/web3-eth.html#getchainid)
- [EIP-695](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-695.md)

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+
+	// TODO
+	// public async getNodeInfo() {
+
+	// }
```
Spacesai1or: - [1.x docs](https://web3js.readthedocs.io/en/v1.2.11/web3-eth.html#getnodeinfo)

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+
+	// TODO
+	// public async getProof() {
+
+	// }
```
Spacesai1or: - [1.x docs](https://web3js.readthedocs.io/en/v1.2.11/web3-eth.html#getproof)
- [EIP-1186](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1186.md)

packages/web3-eth/src/rpc_methods.ts
```diff
@@ -0,0 +1,518 @@
...
+	transaction: Partial<TransactionWithSender>,
+	blockNumber: BlockNumberOrTag,
+) {
+	validateBlockNumberOrTag(blockNumber);
+
+	return requestManager.send<'eth_estimateGas'>({
+		method: 'eth_estimateGas',
+		params: [transaction, blockNumber],
+	});
+}
```
Spacesai1or: @nazarhussain I don't think there's a way to validate `Partial<TransactionWithSender>`?

packages/web3-eth/test/fixtures/rpc_methods_with_params.ts
```diff
@@ -0,0 +1,197 @@
...
+	BlockTags,
+	Filter,
+	HexString32Bytes,
+	HexStringBytes,
+	Uint,
+	Uint256,
+} from 'web3-utils';
+
+// TODO Not sure if this is okay to do
+import { isFilterObjectValidData } from 'web3-utils/test/fixtures/validation';
```
Spacesai1or: @nazarhussain Is this going to cause problems? Can't import it regularly since it's not being exported in `index.ts`

packages/web3-utils/src/converters.ts
```diff
@@ -436,3 +437,56 @@ export const jsonInterfaceMethodToString = (
...
+			return toHex(value);
+		case ValidTypes.NumberString:
+			return hexToNumberString(toHex(value));
+		case ValidTypes.Number:
+			return toNumber(value);
+		case ValidTypes.BigInt:
+			return BigInt(toNumber(value));
+		default:
+			// TODO Replace with proper error
+			throw new Error('Invalid desiredType provided');
```
Spacesai1or: TODO

packages/web3-utils/src/converters.ts
```diff
@@ -436,3 +437,56 @@ export const jsonInterfaceMethodToString = (
...
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-expect-error
+		convertedObject[convertibleProperty] = convertToValidType(
+			object[convertibleProperty],
+			desiredType,
+		);
+	}
+
+	return convertedObject;
+}
```
Spacesai1or: @nazarhussain Helped me out with this, not sure how/when we should try cleaning this up

packages/web3-utils/src/errors.ts
```diff
@@ -155,3 +154,33 @@ export class InvalidTypeAbiInput extends Web3Error {
 		super(value, 'components found but type is not tuple');
 	}
 }
+
+// TODO Type of value for all errors should be any as
+// any value can be passed when an error occurs
```
Spacesai1or, line 164: I believe the type of `value` for all errors should be `any` as we are throwing an error because we got an unexpected value, therefore it could be anything. Thoughts?

packages/web3-utils/src/errors.ts
```diff
@@ -155,3 +154,33 @@ export class InvalidTypeAbiInput extends Web3Error {
...
+		super(value, 'invalid 32 byte hex string given');
+	}
+}
+
+export class InvalidFilter extends Web3Error {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	public constructor(value: any) {
+		// TODO Discuss this naive approach to logging object
+		// Does not account for non JSON properties
+		super(JSON.stringify(value), 'invalid filter given');
```
Spacesai1or: Another error prone handling of possible object `value`s

packages/web3-utils/src/types.ts
```diff
@@ -61,4 +73,36 @@ export type JsonEventInterface = {
 	anonymous: boolean;
 };
 
-// need to create Json Contract interface constructor, receive, and fallback
+// TODO need to create Json Contract interface constructor, receive, and fallback
```
Spacesai1or, line 76: @nazarhussain I believe this was your TODO

packages/web3-eth/src/errors.ts
nazarhussain: In such cases if its just the matter of errors and exceptions then we can stick to one format. And when ever we throw error we should use that. e.g. We can say that what ever value will be user will parse to string and then create error constructor. This way it will be consistent to handle.

packages/web3-eth/test/fixtures/rpc_methods_with_params.ts
nazarhussain: Yes we should not import anything from `test` directory. If that's really a reusable piece of data/code then it must be kept in `src`. One can import from the source directory with a particular path.

packages/web3-eth/src/rpc_methods.ts
nazarhussain: I don't understand what you means by validate? The partial just suggests that value can be undefined but if its not then we can validate if its the valid value or not.

packages/web3-utils/src/converters.ts
nazarhussain: Let's split this out as a separate issue, discuss its scope and then work on it. It's really complex to to handle the nested objects so we should do it in a separate PR.

packages/web3-utils/src/types.ts
nazarhussain: Yes that's seems mine.

packages/web3-utils/src/converters.ts
nazarhussain: This code needs to have extensive tests coverage with variety of input values.

packages/web3-utils/src/errors.ts
```diff
@@ -155,3 +154,33 @@ export class InvalidTypeAbiInput extends Web3Error {
...
+	}
+}
+
+export class InvalidHexString8Bytes extends Web3Error {
+	public constructor(value: string) {
+		super(value, 'invalid 8 byte hex string given');
+	}
+}
+
+export class InvalidHexString32Bytes extends Web3Error {
```
nazarhussain: I would sugges to create one generic constructor for such cases. e.g. `InvalidHexString` would be sufficient to cover all type of bytes data. We can provide more specific information in the error message itself.

packages/web3-utils/src/errors.ts
nazarhussain: Yes `any` is fine for sepcific cases, but if we can use `unknown` instead its much safer as the end user can use `as` keyword with the value.

packages/web3-utils/src/errors.ts
```diff
@@ -155,3 +154,33 @@ export class InvalidTypeAbiInput extends Web3Error {
 		super(value, 'components found but type is not tuple');
 	}
 }
+
+// TODO Type of value for all errors should be any as
+// any value can be passed when an error occurs
+export class InvalidBlockNumberOrTag extends Web3Error {
```
nazarhussain: Always suffix all error constructorts with `Error`, e,g, `InvalidBlockNumberOrTag` -> `InvalidBlockNumberOrTagError`. Same for other created error consstrutors.

packages/web3-utils/src/errors.ts
nazarhussain: That is fine. Unless we have a proper and standard for handling of the errors have a consistent string values are fine. Once we reach to `logging` aspect of the library we will see how to deal it.

packages/web3-utils/src/validation.ts
```diff
@@ -390,3 +409,100 @@ export function isTopicInBloom(bloom: string, topic: string): boolean {
...
+ * Returns true if the given blockNumber is 'latest', 'pending', or 'earliest.
+ */
+export const isBlockTag = (value: string) =>
+	BlockTags.LATEST === value || BlockTags.PENDING === value || BlockTags.EARLIEST === value;
+
+/**
+ * Returns true if given value is valid hex string and not negative, or is a valid BlockTag
+ */
+export const isBlockNumberOrTag = (value: BlockNumberOrTag) =>
+	(isHexStrict(value) && value.substr(0, 1) !== '-') || isBlockTag(value);
```
nazarhussain: You can use `value.startsWtih` instead of `substr` that's more readable.

packages/web3-core/src/web3_context.ts
nazarhussain: Let's proceed with it. Eventually we have to rewise all `eslint-disable-next-line` in the source code after all.

packages/web3-eth/src/rpc_methods.ts
```diff
@@ -0,0 +1,518 @@
...
+
+export async function getBlockByHash(
+	requestManager: Web3RequestManager,
+	blockHash: HexString32Bytes,
+	hydrated: boolean,
+) {
+	validateHexString32Bytes(blockHash);
+	validateBoolean(hydrated);
+
+	return requestManager.send<'eth_getBlockByHash'>({
```
nazarhussain: While you are using the `{method: 'eth_getBlockByHash'}` you don't need to pass the generic value here. It will infer automatically from the parameter.

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+			blockNumber,
+		);
+
+		return convertToValidType(
+			response,
+			returnType ?? this._options.defaultReturnType,
+		) as ValidReturnTypes[ReturnType];
+	}
+
+	public async sendTransaction(transaction: TransactionWithSender) {
```
nazarhussain: As discussed earlier the input parameters for this function must match with the `1.x` API. Remember these parameters may differ from the underlaying JSON-RPC API. We may get some extra or some less input and library handle those paremters and process and send to the underlaying transport layer. e.g. The discussion was going around `chainId`, we need to get this input from user and then use it to encode the transaction before sending.

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+	Filter,
+} from 'web3-utils';
+import {
+	convertibleBlockProperties,
+	convertibleFeeHistoryResultProperties,
+	convertibleReceiptInfoProperties,
+	convertibleTransactionInfoProperties,
+} from './convertible_properties';
+
+import * as RpcMethods from './rpc_methods';
```
nazarhussain: As its not a class and just an object with different functions use `rpcMethods` instead.

nazarhussain (changes requested):

packages/web3-utils/src/converters.ts
Spacesai1or: https://github.com/ChainSafe/web3.js/issues/4582

packages/web3-eth/src/errors.ts
Spacesai1or: https://github.com/ChainSafe/web3.js/issues/4583

packages/web3-utils/src/errors.ts
Spacesai1or: https://github.com/ChainSafe/web3.js/issues/4583

packages/web3-utils/src/errors.ts
Spacesai1or: https://github.com/ChainSafe/web3.js/issues/4583

packages/web3-utils/src/errors.ts
Spacesai1or: I agree, we have a couple other instances of class names missing `Error` suffix in other packages that we'll need to update

packages/web3-eth/src/index.ts
luu-alex: Is that how 1.x deals with multiple optional parameters?

packages/web3-eth/test/fixtures/rpc_methods_with_params.ts
Spacesai1or: If I understand this correctly, doesn't this data need to be included when the `web3-utils` package is built? If it's not, then the `web3-eth` package is no longer standalone (would require pulling the un-built `web3-utils`), because it would import files not included in the published version of `web3-utils`

I don't think we should include test data in the published version of `web3-utils`, so maybe we just duplicate `isFilterObjectValidData` in `web3-eth`?

Spacesai1or:
@nazarhussain @jdevcs 

https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-eth/src/index.js#L455

[1.x docs](https://web3js.readthedocs.io/en/v1.5.2/web3-eth.html?highlight=sendTransaction#getuncle) says `getUncle` takes a `boolean` parameter to provide transaction objects as opposed to just the hashes, however the [1.x code](https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-eth/src/index.js#L455) for the call doesn't reflect that and only has two parameters. Additionally the RPC spec doesn't mention the boolean parameter - this looks like a mistake, yes?

Spacesai1or:
[1.x's call method](https://web3js.readthedocs.io/en/v1.5.2/web3-eth.html?highlight=sendTransaction#call) says it uses the same transaction object as defined by `sendTransaction`:

```typescript
import { AccessList } from 'web3-common';
import { Address, HexString, Numbers } from 'web3-utils';

export enum ChainNames {
    MAINNET = 'mainnet',
    GOERLI = 'goerli',
    KOVAN = 'kovan',
    RINKEBY = 'rinkeby',
    ROPSTEN = 'ropsten'
};

export enum HardForks {
    CHAIN_START = 'chainstart',
    HOMESTEAD = 'homestead',
    DAO = 'dao',
    TANGERINE_WHISTLE = 'tangerineWhistle',
    SPURIOUS_DRAGON = 'spuriousDragon',
    BYZANTIUM = 'byzantium',
    CONSTANTINOPLE = 'constantinople',
    PETERSBURG = 'petersburg',
    ISTANBUL = 'istanbul',
    BERLIN = 'berlin',
    LONDON = 'london'
};

export interface Transaction {
    from?: Address;
    to?: Address;
    value?: Numbers;
    gas?: Numbers;
    gasPrice?: Numbers;
    type?: Numbers;
    maxFeePerGas?: Numbers;
    maxPriorityFeePerGas?: Numbers;
    accessList?: AccessList;
    data?: HexString;
    nonce?: Numbers;
    chain?: HexString;
    hardfork?: HexString;
    common?: {
        customChain: {
            name?: string;
            networkId: Numbers;
            chainId: Numbers;
        };
        baseChain?: ChainNames;
        hardfork?: HardForks;
    }
}
```

In 1.x the [call method](https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-eth/src/index.js#L535) uses the [inputTransactionFormatter](https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-core-helpers/src/formatters.js#L202) which calls [_txInputFormatter](https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-core-helpers/src/formatters.js#L133) which only checks if `transaction.to` is defined and formats it

Currently in 1.x, `transaction.to` is not enforced to be defined, even though the RPC spec says it must be. My proposal is for us to deviate from 1.x for the parameter type of `web3Eth.call` and make `transaciton.to` required

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,374 @@
...
+} from './convertible_properties';
+
+import * as rpcMethods from './rpc_methods';
+
+export class Web3Eth {
+	private readonly _requestManager: Web3RequestManager;
+	private readonly _options: Web3Config;
+
+	public constructor(provider: Web3BaseProvider | string, options: Web3Config) {
+		this._requestManager = new Web3RequestManager(provider);
```
jdevcs: According to architecture discussed, `requestManager` instance must be used from `web3Context`.

jdevcs (changes requested):

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,374 @@
...
+		this._options = options;
+	}
+
+	public async getProtocolVersion() {
```
jdevcs, line 39: I think we must specify return type of this and all following functions.

jdevcs, line 41: we must specify return type of this and all following functions.

packages/web3-eth/src/index.ts
```diff
@@ -0,0 +1,374 @@
...
+import {
+	convertibleBlockProperties,
+	convertibleFeeHistoryResultProperties,
+	convertibleReceiptInfoProperties,
+	convertibleTransactionInfoProperties,
+} from './convertible_properties';
+
+import * as rpcMethods from './rpc_methods';
+
+export class Web3Eth {
```
jdevcs: according to 1.x , web3.eth should also have following functions exposed:
setProvider, given providers, current Provider, ( all these functions should be wrapper to web3Context if you are using instance, else if using inheritance these should be visible to user)

jdevcs: Its also missing exposing ( batchRequest and defaultConfig items to user according to 1.x ( defaultAccount, defaultBlock, ..etc )

jdevcs (changes requested):

jdevcs:
> @nazarhussain @jdevcs
> 
> https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-eth/src/index.js#L455
> 
> [1.x docs](https://web3js.readthedocs.io/en/v1.5.2/web3-eth.html?highlight=sendTransaction#getuncle) says `getUncle` takes a `boolean` parameter to provide transaction objects as opposed to just the hashes, however the [1.x code](https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-eth/src/index.js#L455) for the call doesn't reflect that and only has two parameters. Additionally the RPC spec doesn't mention the boolean parameter - this looks like a mistake, yes?

yes. seems doc is copied from getBlock to getUncle and modified.

jdevcs:
> [1.x's call method](https://web3js.readthedocs.io/en/v1.5.2/web3-eth.html?highlight=sendTransaction#call) says it uses the same transaction object as defined by `sendTransaction`:
> 
> ```ts
> import { AccessList } from 'web3-common';
> import { Address, HexString, Numbers } from 'web3-utils';
> 
> export enum ChainNames {
>     MAINNET = 'mainnet',
>     GOERLI = 'goerli',
>     KOVAN = 'kovan',
>     RINKEBY = 'rinkeby',
>     ROPSTEN = 'ropsten'
> };
> 
> export enum HardForks {
>     CHAIN_START = 'chainstart',
>     HOMESTEAD = 'homestead',
>     DAO = 'dao',
>     TANGERINE_WHISTLE = 'tangerineWhistle',
>     SPURIOUS_DRAGON = 'spuriousDragon',
>     BYZANTIUM = 'byzantium',
>     CONSTANTINOPLE = 'constantinople',
>     PETERSBURG = 'petersburg',
>     ISTANBUL = 'istanbul',
>     BERLIN = 'berlin',
>     LONDON = 'london'
> };
> 
> export interface Transaction {
>     from?: Address;
>     to?: Address;
>     value?: Numbers;
>     gas?: Numbers;
>     gasPrice?: Numbers;
>     type?: Numbers;
>     maxFeePerGas?: Numbers;
>     maxPriorityFeePerGas?: Numbers;
>     accessList?: AccessList;
>     data?: HexString;
>     nonce?: Numbers;
>     chain?: HexString;
>     hardfork?: HexString;
>     common?: {
>         customChain: {
>             name?: string;
>             networkId: Numbers;
>             chainId: Numbers;
>         };
>         baseChain?: ChainNames;
>         hardfork?: HardForks;
>     }
> }
> ```
> 
> In 1.x the [call method](https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-eth/src/index.js#L535) uses the [inputTransactionFormatter](https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-core-helpers/src/formatters.js#L202) which calls [_txInputFormatter](https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-core-helpers/src/formatters.js#L133) which only checks if `transaction.to` is defined and formats it
> 
> Currently in 1.x, `transaction.to` is not enforced to be defined, even though the RPC spec says it must be. My proposal is for us to deviate from 1.x for the parameter type of `web3Eth.call` and make `transaciton.to` required

As 1.x doc is refering to sendTransaction so when we are sending tx for contract creation `to` is optional and it should be set address(0) by lib, but for eth_call rpc specs says its not optional. I agree for making it required for eth_call.

packages/web3-eth/src/index.ts
Spacesai1or: Typescript automatically infers the return type, no need to explicitly declare it

packages/web3-eth/src/index.ts
Spacesai1or: Yes, in the process of updating this one - just a bit confused on how to use `Web3Context` since it's an abstract class

packages/web3-eth/src/index.ts
Spacesai1or: > ...setProvider, given providers, current Provider...

Will be fixed when `Web3Context` is used.

Default config is provided [here](https://github.com/ChainSafe/web3.js/pull/4525/files/f1b956264f03648d780b9c69926b422853c2a99f#diff-3113e3877d2e33382025ff60b3d62ce1eb326e60e9ac21cb4162d9577b7d6784R36) and subsequently used like [so](https://github.com/ChainSafe/web3.js/pull/4525/files/f1b956264f03648d780b9c69926b422853c2a99f#diff-3113e3877d2e33382025ff60b3d62ce1eb326e60e9ac21cb4162d9577b7d6784R95)

I'm not sure about `batchRequest` I think it should be apart of `requestManager`, but I need to look into this

Spacesai1or:
> > [1.x's call method](https://web3js.readthedocs.io/en/v1.5.2/web3-eth.html?highlight=sendTransaction#call) says it uses the same transaction object as defined by `sendTransaction`:
> > ```ts
> > import { AccessList } from 'web3-common';
> > import { Address, HexString, Numbers } from 'web3-utils';
> > 
> > export enum ChainNames {
> >     MAINNET = 'mainnet',
> >     GOERLI = 'goerli',
> >     KOVAN = 'kovan',
> >     RINKEBY = 'rinkeby',
> >     ROPSTEN = 'ropsten'
> > };
> > 
> > export enum HardForks {
> >     CHAIN_START = 'chainstart',
> >     HOMESTEAD = 'homestead',
> >     DAO = 'dao',
> >     TANGERINE_WHISTLE = 'tangerineWhistle',
> >     SPURIOUS_DRAGON = 'spuriousDragon',
> >     BYZANTIUM = 'byzantium',
> >     CONSTANTINOPLE = 'constantinople',
> >     PETERSBURG = 'petersburg',
> >     ISTANBUL = 'istanbul',
> >     BERLIN = 'berlin',
> >     LONDON = 'london'
> > };
> > 
> > export interface Transaction {
> >     from?: Address;
> >     to?: Address;
> >     value?: Numbers;
> >     gas?: Numbers;
> >     gasPrice?: Numbers;
> >     type?: Numbers;
> >     maxFeePerGas?: Numbers;
> >     maxPriorityFeePerGas?: Numbers;
> >     accessList?: AccessList;
> >     data?: HexString;
> >     nonce?: Numbers;
> >     chain?: HexString;
> >     hardfork?: HexString;
> >     common?: {
> >         customChain: {
> >             name?: string;
> >             networkId: Numbers;
> >             chainId: Numbers;
> >         };
> >         baseChain?: ChainNames;
> >         hardfork?: HardForks;
> >     }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In 1.x the [call method](https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-eth/src/index.js#L535) uses the [inputTransactionFormatter](https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-core-helpers/src/formatters.js#L202) which calls [_txInputFormatter](https://github.com/ChainSafe/web3.js/blob/281217277dfdbeacb22c1869a530dcb7b0877d5c/packages/web3-core-helpers/src/formatters.js#L133) which only checks if `transaction.to` is defined and formats it
> > Currently in 1.x, `transaction.to` is not enforced to be defined, even though the RPC spec says it must be. My proposal is for us to deviate from 1.x for the parameter type of `web3Eth.call` and make `transaciton.to` required
> 
> As 1.x doc is refering to sendTransaction so when we are sending tx for contract creation `to` is optional and it should be set address(0) by lib, but for eth_call rpc specs says its not optional. I agree for making it required for eth_call.

`to` should not be set to `address(0)` for contract creation, it **must** be `null`. Setting `to` to `0x0000000000000000000000000000000000000000` would not result in a deployed contract (the gas would just be consumed and nothing would happen)

From [Solidity docs](https://docs.soliditylang.org/en/latest/introduction-to-smart-contracts.html#index-8)
> If the target account is not set (the transaction does not have a recipient or the recipient is set to null), the transaction creates a new contract.

packages/web3-eth/src/index.ts
jdevcs: As discussed in architecture, this class should be inheriting web3Context through web3transaction package. we can discuss this in call today.

packages/web3-eth/src/index.ts
jdevcs: You are right, I have same understanding, `batchRequest` should not be part of Eth package but it should be exposed/exported from eth package. so we maintain 1.x feature parity.

packages/web3-utils/src/validation.ts
```diff
@@ -390,3 +407,100 @@ export function isTopicInBloom(bloom: string, topic: string): boolean {
...
+export const validateHexString8Bytes = (value: HexString8Bytes) => {
+	if (!isHexString8Bytes(value)) throw new InvalidHexStringError(value, 8);
+};
+
+export const isHexString32Bytes = (value: HexString32Bytes) =>
+	isHexStrict(value) && value.length === 66; // 32 bytes + 0x = 66
+
+export const validateHexString32Bytes = (value: HexString32Bytes) => {
+	if (!isHexString32Bytes(value)) throw new InvalidHexStringError(value, 32);
+};
```
Spacesai1or: Should also validate values without `0x` prefix with a `hasPrefix: boolean` argument

packages/web3-eth/src/index.ts
Spacesai1or: > Is that how 1.x deals with multiple optional parameters?

@luu-alex No, `1.x` uses the `...arguements` approach and doesn't specify each parameter in the function signature like we are doing now with TypeScript

packages/web3-utils/src/converters.ts
Spacesai1or: These tests have been written [here](https://github.com/ChainSafe/web3.js/pull/4525/files#diff-f8ab0cde9e2a4da690869725ec95f381288fd4ff11dc3b0df8081e73d1e9d36bR274)

packages/web3-eth/src/index.ts
Spacesai1or: Will be addressed in subsequent PR

packages/web3-eth/src/index.ts
Spacesai1or: Will be addressed in subsequent PR

packages/web3-core/src/types.ts
```diff
@@ -37,3 +38,18 @@ export type Web3BaseProviderConstructor = new <API extends Web3APISpec>(
 	url: string,
 	net?: Socket,
 ) => Web3BaseProvider<API>;
+
+export interface ConfigOptions {
```
nazarhussain: Earlier it was a private type for the `Web3Config` class then the context was fine. As we are exporting now, would be nice to rename it to `Web3ConfigOptions`

packages/web3-core/src/web3_context.ts
```diff
@@ -10,9 +10,20 @@ export class Web3Context<API extends Web3APISpec> extends Web3Config {
...
+	) {
 		super();
 		this.requestManager = new Web3RequestManager<API>(provider);
+
+		if (options !== undefined)
+			for (const key of Object.keys(options)) {
+				// TODO
+				// @ts-expect-error Not sure how to type this
+				// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+				this[key] = options[key];
```
nazarhussain: This loop will not check valid keys for the config options and can cause to overrid any property or method of the context even the private one. Better to add a check for config optiosns before we override.

packages/web3-eth-accounts/package.json
```diff
@@ -28,7 +28,7 @@
 		"@types/jest": "^27.0.3",
 		"@typescript-eslint/eslint-plugin": "^5.4.0",
 		"@typescript-eslint/parser": "^5.4.0",
-		"eslint": "^8.3.0",
+		"eslint": "^7.32.0",
```
nazarhussain: Why this package is downgradded?

packages/web3-eth-contract/package.json
```diff
@@ -34,7 +34,7 @@
 		"@types/jest": "^27.0.3",
 		"@typescript-eslint/eslint-plugin": "^5.4.0",
 		"@typescript-eslint/parser": "^5.4.0",
-		"eslint": "^8.3.0",
+		"eslint": "^7.32.0",
```
nazarhussain: Same here.

packages/web3-utils/src/converters.ts
```diff
@@ -383,3 +400,109 @@ export const toChecksumAddress = (address: Address): string => {
 	}
 	return checksumAddress;
 };
+
+/**
+ *  used to flatten json abi inputs/outputs into an array of type-representing-strings
+ */
+export const flattenTypes = (includeTuple: boolean, puts: Components[]): string[] => {
```
nazarhussain, line 407: Does this code relates to this PR?

nazarhussain (changes requested):

packages/web3-eth-accounts/package.json
Spacesai1or: This addresses this issue:

```bash
✔ Preparing...
⚠ Running tasks...
  ↓ No staged files match *.js [SKIPPED]
  ❯ Running tasks for *.ts
    ✔ prettier --write
    ✖ eslint [FAILED]
  ↓ No staged files match *.{json,md} [SKIPPED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up...

✖ eslint:

Oops! Something went wrong! :(

ESLint: 8.3.0

Error: Rules with suggestions must set the `meta.hasSuggestions` property to `true`. `meta.docs.suggestion` is ignored by ESLint.
```

I forgot I fixed that here, but will break it out into a separate PR

packages/web3-utils/src/converters.ts
Spacesai1or: I didn't add this, it was merged in a [separate PR](https://github.com/ChainSafe/web3.js/pull/4535/files#diff-217301804b75cf6f71d5a808105e97fd258371e0846a38f44841668b5c852148R399), but I'm not sure why it's showing it as an addition here

packages/web3-eth-accounts/package.json
Spacesai1or: [PR](https://github.com/ChainSafe/web3.js/pull/4628) that merges just these changes

packages/web3-eth/src/types.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+interface Transaction1559UnsignedFormatted<
+	DesiredType extends ValidTypes = ValidTypes.HexString,
+	ReturnType = ValidReturnTypes[DesiredType],
+> extends BaseTransactionFormatted<DesiredType> {
+	readonly maxFeePerGas: ReturnType;
+	readonly maxPriorityFeePerGas: ReturnType;
+	readonly accessList: AccessList;
+}
```
nazarhussain, line 81: We can simply all these types with better extended generics. Please create an issue just stating to simply all formatted types. So we address it later.

packages/web3-eth/test/fixtures/rpc_methods_with_params.ts
nazarhussain: For this use case, duplicating code to avoid such dependency will be a good way to go.

nazarhussain (approved):
LGTM. Apart from 2 left over points. 

https://github.com/ChainSafe/web3.js/pull/4525/files#r755062664

https://github.com/ChainSafe/web3.js/pull/4525/files#r768671139

It's also recommend to exclude any `package.json` changes related to `eslint` from this PR.

packages/web3-eth/src/types.ts
```diff
@@ -0,0 +1,194 @@
...
+export enum ChainNames {
+	MAINNET = 'mainnet',
+	GOERLI = 'goerli',
+	KOVAN = 'kovan',
+	RINKEBY = 'rinkeby',
+	ROPSTEN = 'ropsten',
+}
+
+export enum HardForks {
+	CHAIN_START = 'chainstart',
```
jdevcs, line 23: this was called Frontier

packages/web3-eth/src/types.ts
```diff
@@ -0,0 +1,194 @@
...
+	HOMESTEAD = 'homestead',
+	DAO = 'dao',
+	TANGERINE_WHISTLE = 'tangerineWhistle',
+	SPURIOUS_DRAGON = 'spuriousDragon',
+	BYZANTIUM = 'byzantium',
+	CONSTANTINOPLE = 'constantinople',
+	PETERSBURG = 'petersburg',
+	ISTANBUL = 'istanbul',
```
jdevcs, line 31: [Muir Glacier ](https://ethereum.org/en/history/#muir-glacier)

```suggestion
	ISTANBUL = 'istanbul',
	MUIRGLACIER = muirGlacier,
```

packages/web3-eth/src/types.ts
```diff
@@ -0,0 +1,194 @@
...
+	BERLIN = 'berlin',
+	LONDON = 'london',
+}
+
+export interface Transaction {
```
jdevcs, line 36: shouldnt we use same transaction def https://github.com/ChainSafe/web3.js/blob/2ea05a8e9b251f618bdc54f0319f0c45896822af/packages/web3-common/src/eth_execution_api.ts#L78

packages/web3-eth/src/types.ts
```diff
@@ -0,0 +1,194 @@
...
+	readonly cumulativeGasUsed: ReturnType;
+	readonly gasUsed: ReturnType;
+	readonly contractAddress: Address | null;
+	readonly logs: Log[];
+	readonly logsBloom: HexString256Bytes;
+	readonly root: HexString32Bytes;
+	readonly status: ReturnType;
+}
+
+export interface BlockFormatted<
```
jdevcs, line 160: also for block and other types defined here these are also in web3 common, we should reuse types from web3 common https://github.com/ChainSafe/web3.js/blob/2ea05a8e9b251f618bdc54f0319f0c45896822af/packages/web3-common/src/eth_execution_api.ts#L102

packages/web3-eth/src/validation.ts
```diff
@@ -0,0 +1,104 @@
...
+
+	return true;
+}
+
+export function isTransaction1559Unsigned(value: Transaction1559Unsigned): boolean {
+	if (!isBaseTransaction(value)) return false;
+	if (!isHexStrict(value.maxFeePerGas)) return false;
+	if (!isHexStrict(value.maxPriorityFeePerGas)) return false;
+	if (!isAccessList(value.accessList)) return false;
+
```
jdevcs, line 52: should check here its not `isTransactionLegacyUnsigned`

jdevcs:
One minor thing, package.json have` "private": true`, it should not be there.

packages/web3-eth/src/types.ts
Spacesai1or: Created an [issue](https://github.com/ChainSafe/web3.js/issues/4632) for this, and added it to [Web3Eth TODO list](https://github.com/ChainSafe/web3.js/issues/4608)

packages/web3-eth/src/types.ts
Spacesai1or: Created an [issue](https://github.com/ChainSafe/web3.js/issues/4633) for this, and added it to [Web3Eth TODO list](https://github.com/ChainSafe/web3.js/issues/4608)

packages/web3-eth/src/types.ts
Spacesai1or: Created an [issue](https://github.com/ChainSafe/web3.js/issues/4633) for this, and added it to [Web3Eth TODO list](https://github.com/ChainSafe/web3.js/issues/4608)

packages/web3-eth/src/types.ts
Spacesai1or: Created an [issue](https://github.com/ChainSafe/web3.js/issues/4632) for this, and added it to [Web3Eth TODO list](https://github.com/ChainSafe/web3.js/issues/4608)

packages/web3-eth/src/types.ts
Spacesai1or: Created an [issue](https://github.com/ChainSafe/web3.js/issues/4632) for this, and added it to [Web3Eth TODO list](https://github.com/ChainSafe/web3.js/issues/4608)

Spacesai1or:
> One minor thing, package.json have` "private": true`, it should not be there.

@jdevcs This is supposed to be there at the root `package.json`, the root package (mono repo) should never be deployed. Secondly, this was apart of the initial lerna/yarn setup and was generated by that process

jdevcs (approved):

Issue #4472, opened 2021-10-19
Accounts Creation


jdevcs:
Having functionality for following:

- [ ] https://github.com/ChainSafe/web3.js/pull/4496

- [ ] `create`: Generates an account object with private key and public key for optional provided entrypy.

- [ ] `privateKeyToAccount`: Creates an account object from a given private key.

PR #4577, opened 2021-11-24
Accounts Creation - 4.x rewrite

luu-alex:
## Description
Closes: #4472 
Created Account functions:
- create
- privateKeyToAccount

Added missing functions from utils:
- randomHex

fixed toChecksum function

### Additional details:
Breaking changes, create doesn't take in a parameter anymore, due to the switch in libraries (from eth-lib to ethereum-cryptography for security purposes)
former library we used to generate [privateKeys](https://github.com/MaiaVictor/eth-lib/blob/da0971f5b09964d9c8449975fa87933f0c9fef35/src/account.js#L25)


packages/web3-eth-accounts/test/unit/account.test.ts
```diff
@@ -0,0 +1,29 @@
+import { isHexStrict } from 'web3-utils';
+import { create, privateKeyToAccount } from '../../src/account';
+import { validPrivateKeytoAccountData } from '../fixtures/account';
+
+describe('accounts', () => {
+	describe('create', () => {
+		describe('valid cases', () => {
+			it('%s', () => {
+				for (let i = 0; i < 10; i += 1) {
```
luu-alex: Theres probably a better way to test generating random accounts

CHANGELOG.md
```diff
@@ -334,3 +334,13 @@ Released with 1.0.0-beta.37 code base.
 1. The function `outputBigNumberFormatter` in `web3-core-helper` renamed to `outputBigIntFormatter` under `web3-common`
 2. Removed `this.defaultBlock` context from `inputDefaultBlockNumberFormatter` in `web3-core-helper` and converted to additional parameter
 3. Removed `this.defaultBlock` context from `inputTransactionFormatter` in `web3-core-helper` and converted to additional parameter
+
+#### web3-utils
+
+1. The following functions `soliditySha3` `soliditySha3Raw` `encodePacked` now includes type validation and requires type specficiation, instead of guessing the value type
+2. The functions `soliditySha3` `soliditySha3Raw` `encodePacked` does not support BN and now supports `BigInt`
+3. The functions `flattenTypes` and `jsonInterfaceMethodToString` moved to the `web3-eth-abi` package
```
nazarhussain, line 342: 👍

packages/web3-eth-accounts/package.json
```diff
@@ -37,5 +37,9 @@
 		"prettier": "^2.4.1",
 		"ts-jest": "^27.0.7",
 		"typescript": "^4.5.2"
+	},
+	"dependencies": {
+		"ethereum-cryptography": "^0.2.0",
+		"web3-utils": "4.0.0-alpha.0"
```
nazarhussain:
```suggestion
		"web3-utils": "^4.0.0-alpha.0"
```

packages/web3-eth-accounts/src/account.ts
```diff
@@ -0,0 +1,63 @@
...
+export const sign = (): boolean => true;
+
+// Will be added later
+export const signTransaction = (): boolean => true;
+
+/**
+ * Get address from private key
+ */
+export const privateKeyToAccount = (
+	privateKey: string | Uint8Array,
```
nazarhussain: In `4.x` rewrite we prefer to use `Buffer` instead of Typed Array.

packages/web3-eth-accounts/src/types.ts
```diff
@@ -0,0 +1 @@
+export type HexString = string;
```
nazarhussain: You can use this from `web3-utils` package.

packages/web3-eth-accounts/test/unit/account.test.ts
nazarhussain: Why it's required to test this several times? I believe you should just test it once without loop and that's fine.

nazarhussain (changes requested):

packages/web3-eth-accounts/src/account.ts
```diff
@@ -0,0 +1,63 @@
...
+// Will be added later
+export const sign = (): boolean => true;
+
+// Will be added later
+export const signTransaction = (): boolean => true;
+
+/**
+ * Get address from private key
+ */
+export const privateKeyToAccount = (
```
jdevcs, line 24: 1.x doc also have optional `ignoreLengthCheck` boolean but I dnt see any ignoring usage scenario for that, as PK is 64 random hex characters or 32 random bytes, that must be checked.

packages/web3-eth-accounts/src/account.ts
```diff
@@ -0,0 +1,63 @@
...
+	privateKey: string | Uint8Array,
+): {
+	address: string;
+	privateKey: string;
+	signTransaction: () => boolean; // From 1.x
+	sign: () => boolean;
+	encrypt: () => boolean;
+} => {
+	const stringPrivateKey =
```
jdevcs: First validate input privateKey, before processing input: 

- length . (for hex: 64 random hex characters or 32 random bytes)
- null or undefined checks

CHANGELOG.md
```diff
@@ -334,3 +334,13 @@ Released with 1.0.0-beta.37 code base.
...
+
+### web3-eth-accounts
+
+1. `create` function does not take in the optional parameter `entropy`
```
jdevcs, line 347: If we are not including optional `ignoreLength` in `privateKeyToAccount` like 1.x [doc](https://web3js.readthedocs.io/en/v1.5.2/web3-eth-accounts.html#privatekeytoaccount) it should be mentioned here. Its not in 1.x code but in doc

packages/web3-eth-accounts/src/account.ts
luu-alex: Yes It isn't a useful parameter but it is a breaking change, i'll document that.

nazarhussain (approved):

package.json
```diff
@@ -55,5 +55,6 @@
 		"ts-node": "^10.4.0",
 		"typescript": "^4.5.2"
 	},
-	"packageManager": "yarn@1.22.15"
+	"packageManager": "yarn@1.22.15",
+	"dependencies": {}
```
spacesailor24, line 59:
```suggestion
```

Spacesai1or:
```suggestion
	"packageManager": "yarn@1.22.15"
```

packages/web3-eth-accounts/src/account.ts
```diff
@@ -0,0 +1,70 @@
...
+import { InvalidPrivateKeyError, PrivateKeyLengthError } from './errors';
+
+// Will be added later
+export const encrypt = (): boolean => true;
+
+// Will be added later
+export const sign = (): boolean => true;
+
+// Will be added later
+export const signTransaction = (): boolean => true;
```
```suggestion
// TODO Will be added later
export const encrypt = (): boolean => true;

// TODO Will be added later
export const sign = (): boolean => true;

// TODO Will be added later
export const signTransaction = (): boolean => true;
```

Adding `TODO` makes it easier to find using an IDE

packages/web3-eth-accounts/src/account.ts
```diff
@@ -0,0 +1,70 @@
...
+): {
+	address: string;
+	privateKey: string;
+	signTransaction: () => boolean; // From 1.x
+	sign: () => boolean;
+	encrypt: () => boolean;
+} => {
+	if (!privateKey) {
+		throw new InvalidPrivateKeyError(privateKey);
+	}
```
Spacesai1or: The current validation allows false positives, I think we should create `isValidStringInput` and `isBuffer` methods that return a boolean and use them for validation here. After those methods are created we could revise this to:

```typescript
if (!isValidStringInput(privateKey) || !isBuffer(privateKey)) throw new InvalidPrivateKeyError(privateKey);
```

Even though these validation methods are simple, declaring them once and reusing them across the library gives us a single method to update if we discover a bug or want to make the validation stricter. It also reduces the opportunity for accidental less/more strict checks in various parts of the codebase

FYI the new `isBuffer` method (that would be added to `web3-utils/src/validation.ts`) would use: [Buffer.isBuffer](https://www.w3schools.com/nodejs/met_buffer_isbuffer.asp)

packages/web3-eth-accounts/src/account.ts
```diff
@@ -0,0 +1,70 @@
...
+
+	const stringPrivateKey =
+		typeof privateKey === 'object' ? Buffer.from(privateKey).toString('hex') : privateKey;
```
```suggestion
		typeof Buffer.isBuffer(privateKey) ? Buffer.from(privateKey).toString('hex') : privateKey;
```

I'm assuming this is going to work as expected

packages/web3-eth-accounts/src/account.ts
```diff
@@ -0,0 +1,70 @@
...
+
+	const updatedKey = stringPrivateKey.startsWith('0x')
```
Spacesai1or: Nit picky, but I'd rename `updatedKey` to `stringPrivateKeyNoPrefix` for readability's sake

packages/web3-eth-accounts/src/account.ts
```diff
@@ -0,0 +1,70 @@
...
+		? stringPrivateKey.slice(2)
+		: stringPrivateKey;
+
+	// Must be 64 hex characters
```
Spacesai1or, line 46:
```suggestion
    // TODO Replace with isHexString32Bytes function in web3-eth PR:
    // https://github.com/ChainSafe/web3.js/blob/afa6a4bcdfbc122246f35e749d251fd1afb683d0/packages/web3-utils/src/validation.ts#L437
	// Must be 64 hex characters
```

packages/web3-eth-accounts/src/account.ts
```diff
@@ -0,0 +1,70 @@
...
+	if (updatedKey.length !== 64) {
+		throw new PrivateKeyLengthError(updatedKey);
+	}
+
+	const publicKey = getPublicKey(updatedKey);
+
+	const publicKeyString = `0x${publicKey.slice(2)}`;
```
Spacesai1or, line 53: A comment for why `.slice(2)` is being performed would be helpful

packages/web3-eth-accounts/src/account.ts
```diff
@@ -0,0 +1,70 @@
...
+	const publicHash = sha3Raw(publicKeyString);
+	const publicHashHex = bytesToHex(publicHash);
+	const address = toChecksumAddress(publicHashHex.slice(-40));
```
Spacesai1or: A comment for why `.slice(-40)` is being performed would be helpful

packages/web3-eth-accounts/src/errors.ts
```diff
@@ -0,0 +1,27 @@
...
+	public constructor(value: unknown, msg: string) {
+		super(`Invalid value given "${String(value)}". Error: ${msg}.`);
+		this.name = this.constructor.name;
+		Error.captureStackTrace(this, Web3Error);
+	}
+
+	public toJSON() {
+		return { name: this.name, message: this.message };
+	}
+}
```
Spacesai1or: Should be imported from `web3-common` instead of being re-declared (will help us remain consistent if we ever make changes to `Web3Error`)

packages/web3-eth-accounts/src/account.ts
Spacesai1or: Could also use the method [validateHexString32Bytes](https://github.com/ChainSafe/web3.js/pull/4525/files#diff-170228df6bd950804f5f0071be007d1f2acba8a41ef2f9959b815e6ca82a1d97R440) which will throw a `InvalidHexStringError` with the message: `not a valid 32 byte hex string` - Unless you'd prefer the error be more specific with `PrivateKeyLengthError` then you could use [isHexString32Bytes](https://github.com/ChainSafe/web3.js/blob/afa6a4bcdfbc122246f35e749d251fd1afb683d0/packages/web3-utils/src/validation.ts#L437) which returns a boolean

packages/web3-eth-accounts/src/account.ts
Spacesai1or: If above change is added (i.e. `isBuffer` is added) then this would change to:

```typescript
typeof isBuffer(privateKey) ? Buffer.from(privateKey).toString('hex') : privateKey;
```

Spacesai1or (changes requested):

packages/web3-utils/src/converters.ts
```diff
@@ -355,28 +355,30 @@ export const toWei = (number: Numbers, unit: EtherUnits): string => {
 };
 
 export const toChecksumAddress = (address: Address): string => {
-	if (typeof address === 'undefined') return '';
-
-	if (!/^(0x)?[0-9a-f]{40}$/i.test(address)) {
+	if (!/^(0x)?[0-9a-f]{40}$/i.test(address) && isHexStrict(address)) {
```
Spacesai1or: Should use [isAddress](https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-utils/src/validation.ts#L215)

packages/web3-utils/src/random.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Returns a random byte array by the given bytes size
+ *
+ * @param {Number} size
+ * @returns {Buffer}
+ */
+ export const randomBytes = (byteSize: number): Buffer => {
+	const randomValues =
+		typeof window !== 'undefined' && window.crypto && window.crypto.getRandomValues
```
Spacesai1or, line 11:
```suggestion
		window !== undefined && window.crypto && window.crypto.getRandomValues
```

Nit picky, don't necessary have to commit this, just used to checking `undefined` like this

packages/web3-eth-accounts/src/account.ts
Spacesai1or: Or this might not be needed if it's tracked in an issue. I just like the idea of all TODOs being labeled in the code so we can search for anything we might've missed

packages/web3-eth-accounts/src/account.ts
luu-alex: I'm not too sure, ether and our old library did this when creating addresses

jdevcs (approved):

packages/web3-utils/src/converters.ts
luu-alex: We cant use `isAddress` as it expects the address to be checksummed

packages/web3-eth-accounts/src/account.ts
luu-alex: Noted, will be added when web3-eth PR is merged

packages/web3-utils/src/converters.ts
Spacesai1or: Oh yea, then I would suggest we add a second argument to `isAddress` called `checkChecksum` which is default to `true` to remain consistent with `1.x`. In this case we would call `isAddress(address, false)`

Again, this is just to remain consistent across the repository, so we have one place to check how an address is validated

packages/web3-utils/src/random.ts
Spacesai1or: [This](https://stackoverflow.com/a/4725697) is actually a really good point as why we should use `typeof window !== 'undefined'` here

Spacesai1or (approved):

PR #4578, opened 2021-11-24
Create Contract constructor with basic encoding/decoding methods

### LINKED ISSUE #4561: Create Contract constructor with basic encoding/decoding methods
### Description

Need to create `Contract` constructor with existing configurations and available encoding/decoding methods. 

### Acceptance Criteria

- Package must export `Contract` constructor 
- The `Contract` constructor must have a properties matching to existing `1.x` branch
- The contract must have following encoding/decoding functions 
- _encodeEventABI
- _decodeEventABI
- _encodeMethodABI
- _decodeMethodReturn
- Where possible keep the functions as pure functions. 

### Additional Information

- All the function should be covered with unit tests
- The utility functions should not be exported by default

nazarhussain:
## Description

Create basic contract constructor with encoding/decoding functions. 

Fixes #4561

## Type of change

## Checklist:


packages/web3-common/src/formatters.ts
```diff
@@ -260,7 +263,7 @@ export const outputTransactionFormatter = (tx: TransactionInput): TransactionOut
  * @param {Object} log object
  * @returns {Object} log
  */
-export const outputLogFormatter = (log: LogsInput): LogsOutput => {
+export const outputLogFormatter = (log: Partial<LogsInput>): LogsOutput => {
```
Spacesai1or, line 266: Ideally these formatters would be replaced by the same process we're using for formatting `web3-eth` responses. This could probably exist as a cleanup `TODO`

packages/web3-eth-abi/src/utils.ts
```diff
@@ -32,6 +33,13 @@ export const isAbiFunctionFragment = (item: unknown): item is JsonAbiFunctionFra
 	(item as { type: string }).type !== undefined &&
 	(item as { type: string }).type === 'function';
 
+export const isAbiConstructorFragment = (item: unknown): item is JsonAbiConstructorFragment =>
```
Spacesai1or, line 37: Might want to move these `is_` methods to a `validation.ts` file to match `web3-eth` and `web3-utils`

packages/web3-eth-contract/src/contract.ts
```diff
@@ -0,0 +1,122 @@
...
+		name: string,
+		_abi: JsonAbiFunctionFragment,
+	): ContractBoundFunction {
+		return name;
+	}
+
+	// eslint-disable-next-line class-methods-use-this
+	private _createContractEvent(name: string, _abi: JsonAbiEventFragment): ContractBoundEvent {
+		return name;
+	}
```
Spacesai1or, line 130: Could these be `static`?

Spacesai1or:
Can delete unused file `packages/web3-eth-contract/src/errors.ts`

Spacesai1or (approved):
Not a huge fan of doing things mostly 1:1 with `1.x`, but we do need to get the initial version of the rewrite out

nazarhussain:
@spacesailor24 Yes I am also not fan for doing this way, but the encoding/decoding functions are tightly linked to the way `abi` package is written. So have to stay like this. But this code will evolve a lot before reaching to alpha stage.

luu-alex (approved):

packages/web3-eth-contract/src/encoding.ts
```diff
@@ -0,0 +1,194 @@
...
+			topics: data.topics,
+		},
+	};
+};
+
+export const encodeMethodABI = (
+	abi: (JsonAbiFunctionFragment | JsonAbiConstructorFragment) & { signature?: string },
+	args: unknown[],
+	deployData?: HexString,
+) => {
```
luu-alex, line 138: Is it worth anything to specify a return type for decode/encode functions?

nazarhussain: The return type inferred by the the type of variable which is used with `return`.

packages/web3-eth-abi/src/utils.ts
nazarhussain: Will look into doing this in next PR. End of the day all exported as functions from root so does not matter a lot.

packages/web3-common/src/formatters.ts
nazarhussain: Yes. It would be.

PR #4584, opened 2021-11-29
add: custom transaction polling interval

### LINKED ISSUE #2621: Transaction confirmation poll rate as a constructor option
### Description
In the `TransactionObserver` class, the polling rate is set to poll every 1000 ms. While this is fine for the mainnet and testnets, people using private networks or ganache have experienced a significant performance dip.
I personally have fixed this locally by changing the observer to poll every 100ms.
Perhaps this could be added as a configurable option (with a default of 1000), similar to the options added under this new transaction confirmation workflow.
Plus it's nice to remove hardcoded values in the codebase.

#### Expected behavior
Transaction confirmations are processed multiple times per second.

#### Actual behavior
Confirmations take up to 1000ms each, a hardcoded value.

Configure web3 with any private network that has a high transaction throughput and try to run many transactions in a row.

### Versions 
* web3.js: 1.0.0-beta.51

sirpy:
## Description

allow to set via options the interval which polls for TX status
#2621 

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -484,7 +484,8 @@ Released with 1.0.0-beta.37 code base.
 
 ### Added
 - `maxPriorityFeePerGas` and `maxFeePerGas` added to `Transaction` and `TransactionConfig` interfaces (#4232) (#4585)
-
+- `transactionPollingInterval` added to web3, contract and method constructor options. defaults to 1 second. (#4584)
+- 
 ### Fixed
  -  Fix readthedoc's build for web3js documentation (#4425)
  -  Fix response sorting for batch requests (#4250)
--- a/docs/web3-eth-contract.rst
+++ b/docs/web3-eth-contract.rst
@@ -380,6 +380,27 @@ Returns
 
 ------------------------------------------------------------------------------
 
+.. _eth-contract-module-transactionpollinginterval:
+
+transactionPollingInterval
+=====================
+
+.. code-block:: javascript
+
+    web3.eth.Contract.transactionPollingInterval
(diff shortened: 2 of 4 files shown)


nazarhussain (changes requested):
Hi @sirpy Thanks for contribution. Please add relevant docs under `docs/` folder to reflect the changes. 

Also update the `CHANGELOG.md` to specify what new config been added.

nazarhussain:
@sirpy Did you get a chance to look into feedback I posted earlier. I will encourage you to complete it so we can include this PR in upcoming sprint.

sirpy:
@nazarhussain Hey, actually looking at it again, I dont think what I did works.
Not only that, i'm not able to understand and track how all those defaults are passed down to the contract and then to the method sending. Can some enlighten me?

sirpy:
@nazarhussain updated

nazarhussain (approved):

jdevcs:
`Build / e2e_windows` is failing on this

jdevcs (changes requested):
`Build / e2e_windows (pull_request) ` is failing on this PR, needs to be fixed

also could you add test for passing custom `transactionPollingInterval`

sirpy:
@jdevcs I dont see how this PR is related to this error. In anycase I wouldnt have any idea how to fix it.

Spacesai1or:
@sirpy FYI I merged this PR into a branch I just created and opened PR #4672, the Windows test is no longer failing. I don't have an answer as to why this works, but we used this to solve the same issue previously

sirpy:
@spacesailor24 thanks

PR #4604, opened 2021-12-06
Implement static type interface for Contract

### LINKED ISSUE #4562: Implement static type interface for Contract
### Description

Need to create a strong type interface for `Contract` matching the provided ABI.  

### Acceptance Criteria

- The interface must support 1.x API 
- The features using callback should be removed 
- All the interfaces must be strong type 
- 

### Additional Information

- All the function should be covered with unit tests
- Explore the need and use of [typechain](https://www.npmjs.com/package/typechain)

nazarhussain:
## Description

Add contract static type support. 

Fixes #4562

## Type of change

## Checklist:


packages/web3-eth-abi/src/number_map_type.ts
```diff
@@ -0,0 +1,7 @@
+// TODO: Convert it to static file
+type _SolidityIndexRange = 1 | 2 | 3 | 4 | 5;
```
nazarhussain: As the fixed array size is passed through the ABI as string so we need relevant map for integer values. We can set this up to maximum we want to support.

packages/web3-eth-contract/test/fixtures/types/web3-v1-contracts/types.ts
```diff
@@ -0,0 +1,64 @@
+/* Autogenerated file. Do not edit manually. */
```
Spacesai1or: Is this file actually generated? Also curious why we're supporting `bn.js`?

packages/web3-eth-abi/src/number_map_type.ts
Spacesai1or: But isn't the range X >= 0? So it's some very large number that we'd have to support up to?

spacesailor24:
I don't really understand TypeScript enough to give a definitive review on this

packages/web3-eth-contract/test/fixtures/types/web3-v1-contracts/types.ts
nazarhussain: It's accidentally committed, was comparing the interface with v1 interfaces.

packages/web3-eth-abi/src/number_map_type.ts
nazarhussain: Theoretically ECMA-262 supports 2^32 -1 which is very less than what solidity supports. We can add as much number we want, but the larger numbers here will make the intelisence of TS very slow.  So we have to restrict this number to some limited values.

packages/web3-eth-contract/test/fixtures/types/web3-v1-contracts/types.ts
Spacesai1or: Okay, so `packages/web3-eth-contract/test/fixtures/types/web3-v1-contracts/abi.ts` and `packages/web3-eth-contract/test/fixtures/types/web3-v1-contracts/types.ts` (this file) need to be removed?

packages/web3-eth-abi/src/types.ts
```diff
@@ -1,57 +1,187 @@
...
+
+export type FilterAbis<Abis extends ContractAbi, Filter, Abi = Abis[number]> = Abi extends Filter
+	? Abi
+	: never;
+
+type _TypedArray<Type, Size extends string> = Size extends ''
+	? Type[]
+	: FixedSizeArray<Type, ConvertToNumber<Size>>;
+
+export type PrimitiveAddressType<Type extends string> = Type extends `address[${infer Size}]`
```
jdevcs, line 97: As discussed , could you add doc comments or in readme file list of supported ( solidity) types checking.

jdevcs:
Please test type checking support for famous community contracts: 
Open Zeppelin :
- [x] ERC20, 
- [ ] ERC777, 
- [x] ERC721, and 

Gnosis  :
- [ ]  Safe / multisig

packages/web3-eth-abi/src/types.ts
```diff
@@ -1,57 +1,187 @@
...
+	? Numbers
+	: never;
+
+export type PrimitiveBytesType<Type extends string> = Type extends `bytes${string}[${infer Size}]`
+	? _TypedArray<Bytes, Size>
+	: Type extends 'bytes' | `bytes${string}`
+	? Bytes
+	: never;
+
+export type PrimitiveTupleType<
```
jdevcs, line 129: also as discussed could you see if complex types checking like mappings , mappings of structs or mappings of mappings support is applicable for this feature?

packages/web3-eth-abi/src/types.ts
```diff
@@ -1,57 +1,187 @@
...
+	readonly type: 'function';
+	readonly stateMutability: 'nonpayable' | 'payable' | 'pure' | 'view';
+	readonly inputs: ReadonlyArray<AbiParameter>;
+	readonly outputs: ReadonlyArray<AbiParameter>;
 
 	// legacy properties
-	constant?: boolean;	// stateMutability == 'pure' or stateMutability == 'view'
-	payable?: boolean; // stateMutability == 'payable'
-}
+	readonly constant?: boolean; // stateMutability == 'pure' or stateMutability == 'view'
```
jdevcs, line 66: constants and payable are not mentioned in specs, we only have `stateMutability` field for this

packages/web3-eth-abi/src/types.ts
```diff
@@ -1,57 +1,187 @@
...
+	readonly inputs: ReadonlyArray<AbiParameter>;
+	readonly anonymous?: boolean;
+};
 
 // https://docs.soliditylang.org/en/latest/abi-spec.html#json
-export type JsonAbiFragment =
-	| JsonAbiConstructorFragment
-	| JsonAbiFunctionFragment
-	| JsonAbiEventFragment;
+export type AbiFragment = AbiConstructorFragment | AbiFunctionFragment | AbiEventFragment;
```
jdevcs: there is fallback function type as well

packages/web3-eth-contract/src/types.ts
```diff
@@ -20,3 +23,72 @@ export interface ContractInitOptions {
 	readonly gasLimit: Uint;
 	readonly provider: SupportedProviders<EthExecutionAPI> | string;
 }
+
+// TODO: Add correct type
```
jdevcs: for all these TODOs could you open a new issue and mention these in that, will be easier to plan in sprints.

packages/web3-eth-abi/src/types.ts
nazarhussain: All the solidity types which can be used in `ABI` are supported as per my knowledge. So seems no need to mention which types are supported.

packages/web3-eth-abi/src/types.ts
nazarhussain: As per my knowledge mapping and mapping of structures are only be used internally by solidity. Can't be used as input or output of a public method or event. So not needed to cover in this the scope of this PR.

packages/web3-eth-abi/src/types.ts
nazarhussain: Yes its not mentioned, but legacy ABI versions support these. So to avoid nay compatibility issue used it as optional. Similar approach was use in earlier implementation as well.

nazarhussain:
@jdevcs I had added type tests for ERC20 and ERC721. I believe these are good enough to test the type interface for contract. Gnosis safe/multisig is splited among different contracts, if you think think that's necessary to add please share reference to its ABI to use.

packages/web3-eth-abi/src/types.ts
jdevcs: wts the min version of solidity being type validated with current functions? will be useful to mention min version in code as comments and also in doc.

jdevcs:
> @jdevcs I had added type tests for ERC20 and ERC721. I believe these are good enough to test the type interface for contract. Gnosis safe/multisig is splited among different contracts, if you think think that's necessary to add please share reference to its ABI to use.

For now just include ERC777 along with , including multisig in this PR will increase scope/time , add that in tests TODO items.

nazarhussain:
@jdevcs Created an issue for additional type tests https://github.com/ChainSafe/web3.js/issues/4631

packages/web3-eth-abi/src/types.ts
nazarhussain: I updated and mentioned in README file. I tested it personally for the ABIs generated by `v0.4.x` solidity. In future if required we can add more tests for different versions of solidity.

packages/web3-eth-abi/src/number_map_type.ts
nazarhussain: Added the limitation reference in the README file.

packages/web3-eth-abi/package.json
```diff
@@ -41,6 +41,7 @@
 		"jest": "^27.3.1",
 		"jest-extended": "^1.1.0",
 		"jest-when": "^3.4.1",
+    "@humeris/espresso-shot": "^4.0.0",
```
Spacesai1or: I don't think it's such a great idea to use such a young and unused library for our tests. Having something is better than nothing, but I've created [an issue](https://github.com/ChainSafe/web3.js/issues/4635) to revisit this

packages/web3-eth-abi/test/unit/types.test.ts
```diff
@@ -0,0 +1,151 @@
...
+			typecheck('should extend the string type fixed array', () => {
+				const b: MatchPrimitiveType<'string[3]', []> = ['a', 'a', 'a'];
+
+				return expectTypeOf(b).toExtend<[string, string, string]>();
+			});
+		});
+
+		describe('address', () => {
+			typecheck('should extend correct type', () =>
+				expectTypeOf<MatchPrimitiveType<'address', []>>().toExtend<Address>(),
```
Spacesai1or, line 41: Why switch to this style of tests when the above use `const b: MatchPrimitiveType<...> = ...;` format?

packages/web3-eth-contract/README.md
```diff
@@ -30,6 +30,29 @@ yarn add web3-eth-contract
...
+import { Contract } from 'web3-eth-contract';
+
+const abi = [...] as const;
+
+const contract = new Contract(abi);
+```
+
+-   We prefer that you use `web3.eth.Contract` API in normal usage.
+-   The use of `as const` is necessary to have fully type-safe interface for the contract.
+-   As the ABIs does not extensive in size, we suggest to copy those in your TS project and declare them `as const`.
```
Spacesai1or:
```suggestion
-   As the ABIs are not extensive in size, we suggest declaring them `as const` in your TS project.
```

packages/web3-eth-contract/README.md
```diff
@@ -30,6 +30,29 @@ yarn add web3-eth-contract
...
+-   This approach is more flexible and seamless compared to other approaches of off-line compiling ABIs to TS interfaces.
```
Spacesai1or:
```suggestion
-   This approach is more flexible and seamless compared to other approaches of off-line compiling ABIs to TS interfaces (such as [TypeChain](https://github.com/dethcrypto/TypeChain).
```

packages/web3-eth-contract/README.md
```diff
@@ -30,6 +30,29 @@ yarn add web3-eth-contract
...
+
+## Compatibility
+
+We had tested the Typescript interface support for the ABIs compiled with solidity version `v0.4.x` and above. If you face any issue regarding the contract typing, please create an issue to report to us.
```
Spacesai1or:
```suggestion
We have tested the Typescript interface support for the ABIs compiled with solidity version `v0.4.x` and above. If you face any issue regarding the contract typing, please create an issue to report to us.
```

packages/web3-eth-contract/README.md
```diff
@@ -30,6 +30,29 @@ yarn add web3-eth-contract
...
+
+The Typescript support for fixed length array types are supported for limited numbers 30. See more details [here](https://github.com/ChainSafe/web3.js/blob/nh%2F4562-contract-typing/packages/web3-eth-abi/src/number_map_type.ts#L1). This limitation is only to provide more performant developer experience in IDEs. In future we may come up with a workaround to avoid this limitation. If you have any idea feel free to share.
```
Spacesai1or:
```suggestion
The Typescript support for fixed length array types are supported up 30 elements. See more details [here](https://github.com/ChainSafe/web3.js/blob/nh%2F4562-contract-typing/packages/web3-eth-abi/src/number_map_type.ts#L1). This limitation is only to provide more performant developer experience in IDEs. In future we may come up with a workaround to avoid this limitation. If you have any idea feel free to share.
```

Spacesai1or (approved):
Still don't fully understand all the TypeScript going on, but the tests seem to show it works as needed. I think we should find some more complex contracts to throw at this before releasing `4.x`

nazarhussain:
@spacesailor24 Already created an issue to add more complex contract types tests. #4631

packages/web3-eth-abi/package.json
nazarhussain: This is quite a new topic to tests types so why we don't have any library. I had reviewed all the code of that library before using it, so good to go for now.

packages/web3-eth-abi/test/unit/types.test.ts
nazarhussain: Fixed that feedback by using consistent pattern.

luu-alex (approved):

jdevcs (approved):

packages/web3-eth-abi/src/types.ts
```diff
@@ -1,57 +1,207 @@
...
+	? Type[]
+	: FixedSizeArray<Type, ConvertToNumber<Size>>;
+
+export type PrimitiveAddressType<Type extends string> = Type extends `address[${infer Size}]`
+	? _TypedArray<Address, Size>
+	: Type extends 'address'
+	? Address
+	: never;
+
+export type PrimitiveStringType<Type extends string> = Type extends `string${string}[${infer Size}]`
```
dapplion, line 103: I would love to have extensive JSDocs here explaining the rationale of this types. For external contributors this is hard to read and comments would help a lot

dapplion: Are there benchmarks of the compilation cost of extensive use of Template Literal Types? Docs say:

> We generally recommend that people use ahead-of-time generation for large string unions, but this is useful in smaller cases. [ref](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html)

nazarhussain: Never found any such, but these types provides real-time developer feedback in IDE so fast enough to use.

dapplion: Is there any demo project I can try with this feature? Would love to see some usage guide too

dapplion:
To better understand this PR, please correct me if I'm wrong:

- The goal of this PR is to strongly type consumers of contracts, for example DApps and test suites
- This is an alternative to Typechain but instead of doing code generation it derives the types from the JSON ABI directly on each run

dapplion:
Very impressive use of Template Literal Types :+1: technically this is really cool

Issue #4473, opened 2021-10-19
Signing and Recovery


jdevcs:
- `signTransaction` : Signs an Ethereum transaction with a given private key.

- `recoverTransaction` : Recovers the Ethereum address which was used to sign the given RLP encoded transaction.

- `sign` : Signs arbitrary data

- `recover` : Recovers the Ethereum address which was used to sign the given data.

- `hashMessage` : Hashes the given message to be passed web3.eth.accounts.recover() function. The data will be UTF-8 HEX decoded and enveloped as follows: "\x19Ethereum Signed Message:\n" + message.length + message and hashed using keccak256.

PR #4607, opened 2021-12-07
Signing and Recovery 4.x Accts 

jdevcs:
## Description

This PR will address issue #4473

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -341,7 +341,7 @@ Released with 1.0.0-beta.37 code base.
 
 1. `create` function does not take in the optional parameter `entropy`
 2. `ignoreLength` will be removed as an optional parameter for `privateKeyToAccount`
-3. `signTransaction` will not fill any default values and it will only sign and return result. For filling default values use web3.transaction package
+3. `signTransaction` will not fill any default values and it will only sign and return result. For filling default values use `web3-eth-transaction` package
 4. `recover` function's last param is boolean `hashed`, it is used to indicate if data provided is already hashed or not. By default this function will assume data is not hashed.
 5. The `Wallet` no more supports address/number indexing. Have to use `wallet.get` instead.
 6. `Wallet.create` function doesn't accepts `entropy` param
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -347,3 +347,4 @@ Released with 1.0.0-beta.37 code base.
 1. `create` function does not take in the optional parameter `entropy`
 2. `ignoreLength` will be removed as an optional parameter for `privateKeyToAccount`
 3. `signTransaction` will not fill any default values and it will only sign and return result. For filling default values use web3.transaction package
+4. `recover` function's last param is boolean `hashed`, it is used to indicate if data provided is already hashed or not. By default this function will assume data is not hashed.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -346,3 +346,4 @@ Released with 1.0.0-beta.37 code base.
 
 1. `create` function does not take in the optional parameter `entropy`
 2. `ignoreLength` will be removed as an optional parameter for `privateKeyToAccount`
+3. `signTransaction` will not fill any default values and it will only sign and return result. For filling default values use web3.transaction package
--- a/packages/web3-common/src/constants.ts
+++ b/packages/web3-common/src/constants.ts
@@ -29,6 +29,7 @@ export const ERR_TX_NO_CONTRACT_ADDRESS = 403;
 export const ERR_TX_CONTRACT_NOT_STORED = 404;
 export const ERR_TX_REVERT_WITHOUT_REASON = 405;
 export const ERR_TX_OUT_OF_GAS = 406;
+export const ERR_TX_UNDEFINED = 407;
 
 // Connection error codes
 export const ERR_CONN = 500;
@@ -45,7 +46,12 @@ export const ERR_INVALID_PROVIDER = 601;
 export const ERR_INVALID_CLIENT = 602;
 export const ERR_SUBSCRIPTION = 603;
 
+// Privatekey error codes
 export const ERR_PRIVATE_KEY_LENGTH = 701;
 export const ERR_INVALID_PRIVATE_KEY = 702;
+
+// Signature error codes
+export const ERR_SIGNATURE_FAILED = 801;
+export const ERR_INVALID_SIGNATURE = 802;
 
 export const GENESIS_BLOCK_NUMBER = '0x0';
--- a/packages/web3-common/src/constants.ts
+++ b/packages/web3-common/src/constants.ts
@@ -29,7 +29,7 @@ export const ERR_TX_NO_CONTRACT_ADDRESS = 403;
 export const ERR_TX_CONTRACT_NOT_STORED = 404;
 export const ERR_TX_REVERT_WITHOUT_REASON = 405;
 export const ERR_TX_OUT_OF_GAS = 406;
-export const ERR_TX_UNDEFINED = 407;
+export const ERR_RAW_TX_UNDEFINED = 407;
 
 // Connection error codes
 export const ERR_CONN = 500;
--- a/packages/web3-common/src/errors.ts
+++ b/packages/web3-common/src/errors.ts
@@ -37,6 +37,9 @@ import {
 	ERR_ABI_ENCODING,
 	ERR_INVALID_PRIVATE_KEY,
 	ERR_PRIVATE_KEY_LENGTH,
+	ERR_SIGNATURE_FAILED,
+	ERR_INVALID_SIGNATURE,
+	ERR_TX_UNDEFINED,
 } from './constants';
 import { isResponseWithError } from './json_rpc';
 
@@ -206,7 +209,7 @@ export class InvalidResponseError<ErrorType = unknown> extends ResponseError<Err
 export class TransactionError extends Web3Error {
 	public code = ERR_TX;
 
-	public constructor(message: string, public receipt: Receipt) {
+	public constructor(message: string, public receipt?: Receipt) {
 		super(message);
 	}
 
@@ -283,6 +286,13 @@ export class TransactionOutOfGasError extends TransactionError {
 			receipt,
 		);
 		this.code = ERR_TX_OUT_OF_GAS;
+	}
+}
+
+export class UndefinedRAWTransactionError extends TransactionError {
+	public constructor() {
+		super(`Transaction Undefined`);
+		this.code = ERR_TX_UNDEFINED;
 	}
 }
 
@@ -407,5 +417,19 @@ export class InvalidPrivateKeyError extends Web3Error {
 	public code = ERR_INVALID_PRIVATE_KEY;
 	public constructor() {
 		super(`Error: not a valid string or buffer.`);
+	}
+}
+
+export class SignerError extends Web3Error {
+	public code = ERR_SIGNATURE_FAILED;
+	public constructor(errorDetails: string) {
+		super(`Error: Invalid signature. "${errorDetails}"`);
+	}
+}
+
+export class InvalidSignatureError extends Web3Error {
+	public code = ERR_INVALID_SIGNATURE;
+	public constructor(errorDetails: string) {
+		super(`Error: "${errorDetails}"`);
 	}
 }
--- a/packages/web3-common/src/errors.ts
+++ b/packages/web3-common/src/errors.ts
@@ -398,14 +398,14 @@ export class AbiError extends Web3Error {
 
 export class PrivateKeyLengthError extends Web3Error {
 	public code = ERR_PRIVATE_KEY_LENGTH;
-	public constructor(value: string | Buffer) {
-		super(`Invalid value given "${value.toString()}". Error: Private key must be 32 bytes.`);
+	public constructor() {
+		super(`Error: Private key must be 32 bytes.`);
 	}
 }
 
 export class InvalidPrivateKeyError extends Web3Error {
 	public code = ERR_INVALID_PRIVATE_KEY;
-	public constructor(value: string | Buffer) {
-		super(`Invalid value given "${String(value)}". Error: not a valid string or buffer.`);
+	public constructor() {
+		super(`Error: not a valid string or buffer.`);
 	}
 }
--- a/packages/web3-common/src/errors.ts
+++ b/packages/web3-common/src/errors.ts
@@ -39,7 +39,7 @@ import {
 	ERR_PRIVATE_KEY_LENGTH,
 	ERR_SIGNATURE_FAILED,
 	ERR_INVALID_SIGNATURE,
-	ERR_TX_UNDEFINED,
+	ERR_RAW_TX_UNDEFINED,
 } from './constants';
 import { isResponseWithError } from './json_rpc';
 
@@ -289,10 +289,10 @@ export class TransactionOutOfGasError extends TransactionError {
 	}
 }
 
-export class UndefinedRAWTransactionError extends TransactionError {
+export class UndefinedRawTransactionError extends TransactionError {
 	public constructor() {
-		super(`Transaction Undefined`);
-		this.code = ERR_TX_UNDEFINED;
+		super(`Raw transaction undefined`);
+		this.code = ERR_RAW_TX_UNDEFINED;
 	}
 }
 
@@ -409,27 +409,27 @@ export class AbiError extends Web3Error {
 export class PrivateKeyLengthError extends Web3Error {
 	public code = ERR_PRIVATE_KEY_LENGTH;
 	public constructor() {
-		super(`Error: Private key must be 32 bytes.`);
+		super(`Private key must be 32 bytes.`);
 	}
 }
 
 export class InvalidPrivateKeyError extends Web3Error {
 	public code = ERR_INVALID_PRIVATE_KEY;
 	public constructor() {
-		super(`Error: not a valid string or buffer.`);
+		super(`Not a valid string or buffer.`);
 	}
 }
 
 export class SignerError extends Web3Error {
 	public code = ERR_SIGNATURE_FAILED;
 	public constructor(errorDetails: string) {
-		super(`Error: Invalid signature. "${errorDetails}"`);
+		super(`Invalid signature. "${errorDetails}"`);
 	}
 }
 
 export class InvalidSignatureError extends Web3Error {
 	public code = ERR_INVALID_SIGNATURE;
 	public constructor(errorDetails: string) {
-		super(`Error: "${errorDetails}"`);
+		super(`"${errorDetails}"`);
 	}
 }
--- a/packages/web3-eth-accounts/package.json
+++ b/packages/web3-eth-accounts/package.json
@@ -41,7 +41,7 @@
 	},
 	"dependencies": {
 		"@ethereumjs/tx": "^3.4.0",
-		"ethereum-cryptography": "^0.2.0",
+		"ethereum-cryptography": "^0.2.1",
 		"secp256k1": "^4.0.2",
 		"web3-common": "^1.0.0-alpha.0",
 		"web3-utils": "^4.0.0-alpha.0"
--- a/packages/web3-eth-accounts/package.json
+++ b/packages/web3-eth-accounts/package.json
@@ -41,6 +41,7 @@
 	"dependencies": {
 		"@ethereumjs/tx": "^3.4.0",
 		"ethereum-cryptography": "^0.2.0",
+		"secp256k1": "^4.0.2",
 		"web3-common": "^1.0.0-alpha.0",
 		"web3-utils": "^4.0.0-alpha.0"
 	}
--- a/packages/web3-eth-accounts/package.json
+++ b/packages/web3-eth-accounts/package.json
@@ -39,6 +39,7 @@
 		"typescript": "^4.5.2"
 	},
 	"dependencies": {
+		"@ethereumjs/tx": "^3.4.0",
 		"ethereum-cryptography": "^0.2.0",
 		"web3-common": "^1.0.0-alpha.0",
 		"web3-utils": "^4.0.0-alpha.0"
--- a/packages/web3-eth-accounts/src/account.ts
+++ b/packages/web3-eth-accounts/src/account.ts
@@ -13,6 +13,7 @@ import {
 	HexString,
 	isBuffer,
 	isValidString,
+	Address,
 } from 'web3-utils';
 import { InvalidPrivateKeyError, PrivateKeyLengthError } from 'web3-common';
 import { signTransactionFunction, signTransactionResult } from './types';
@@ -60,6 +61,17 @@ export const signTransaction = (
 		rawTransaction: rawTx,
 		transactionHash: `0x${Buffer.from(txHash).toString('hex')}`,
 	};
+};
+
+/**
+ * Recovers the Ethereum address which was used to sign the given RLP encoded transaction.
+ */
+export const recoverTransaction = (rawTransaction: string): Address => {
+	if (rawTransaction === undefined) throw new Error('Invalid rawTransaction');
+
+	const tx = TransactionFactory.fromSerializedData(Buffer.from(rawTransaction.slice(2), 'hex'));
+
+	return toChecksumAddress(tx.getSenderAddress().toString());
 };
 
 /**
--- a/packages/web3-eth-accounts/src/account.ts
+++ b/packages/web3-eth-accounts/src/account.ts
@@ -92,13 +92,13 @@ export const signTransaction = (
 	transaction: TxData | AccessListEIP2930TxData | FeeMarketEIP1559TxData,
 	privateKey: string,
 ): signTransactionResult => {
-	//	TO DO : Send calls to web3.transaction package for :
+	//	TODO : Send calls to web3.transaction package for :
 	//		Transaction Validation checks
 
 	const tx = TransactionFactory.fromTxData(transaction);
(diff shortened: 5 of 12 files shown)


fallingicarus (approved):

packages/web3-eth-accounts/package.json
```diff
@@ -39,7 +39,9 @@
 		"typescript": "^4.5.2"
 	},
 	"dependencies": {
+		"@ethereumjs/tx": "^3.4.0",
```
nazarhussain, line 43: Please create an issue for future reference to avoid using this dependency.

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,166 @@
...
+		rawTransaction: rawTx,
+		transactionHash: `0x${Buffer.from(txHash).toString('hex')}`,
+	};
+};
+
+/**
+ * Recovers the Ethereum address which was used to sign the given RLP encoded transaction.
+ */
+export const recoverTransaction = (rawTransaction: string): Address => {
+	if (rawTransaction === undefined) throw new Error('Invalid rawTransaction');
```
nazarhussain: Not needed here. If want extensive checks then also include other incompatible types.

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,166 @@
...
+		s: `0x${signedTx.s.toString('hex')}`,
+		rawTransaction: rawTx,
+		transactionHash: `0x${Buffer.from(txHash).toString('hex')}`,
+	};
+};
+
+/**
+ * Recovers the Ethereum address which was used to sign the given RLP encoded transaction.
+ */
+export const recoverTransaction = (rawTransaction: string): Address => {
```
nazarhussain:
```suggestion
export const recoverTransaction = (rawTransaction: HexString): Address => {
```

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,166 @@
...
+	if (typeof data === 'object') {
+		const signatureStr = `${data.r}${data.s.slice(2)}${data.v.slice(2)}`;
+		return recover(data.messageHash, signatureStr, true);
+	}
+
+	if (signature === undefined) throw new Error('signature string undefined');
+
+	const hashedMessage = hashed ? data : hashMessage(data);
+
+	const v = signature.substring(130); // 0x + r + s + v
```
nazarhussain: It's preferable to avoid magic numbers in the code. Try creating constants for it then use it here. 

Does it really pass the linting? We had following rule enabled for `src`. 

https://eslint.org/docs/rules/no-magic-numbers

packages/web3-eth-accounts/test/unit/account.test.ts
```diff
@@ -29,4 +41,73 @@ describe('accounts', () => {
...
+
+			const tx = {
+				to: '0x2c7536e3605d9c16a7a3d7b1898e529396a65c23',
+				value: '0x16345785d8a0000',
+				gasLimit: '0x5208',
+				gasPrice: '0x2710',
+				chainId: 1,
+				nonce: '0x11',
+				data: '0x',
+			};
```
nazarhussain: We need more examples at least one for each transaction type for this test.

packages/web3-eth-accounts/test/unit/account.test.ts
```diff
@@ -29,4 +41,73 @@ describe('accounts', () => {
...
+		it('recover transaction', () => {
+			const account = create();
+
+			const rawTx = {
+				to: '0x118C2E5F57FD62C2B5b46a5ae9216F4FF4011a07',
+				from: account.address,
+				value: toHex(toWei('0.1', 'ether')),
+				gasLimit: toHex(21000),
+				gasPrice: toHex('1000000'),
+			};
```
nazarhussain: Same here.

nazarhussain (changes requested):

tsconfig.json
```diff
@@ -16,5 +16,6 @@
 		"sourceMap": true,
 		"declarationMap": true,
 		"strict": true,
-  },
-}
\ No newline at end of file
+		"esModuleInterop": true
```
nazarhussain: Enabling this also enables [allowSyntheticDefaultImports](https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports). We need to set it `esModuleInterop` to `false` which is the default value for the compiler. Importing a module with custom name as default import is a not a good practice, end up having different names for same package on different code files.

nazarhussain (changes requested):

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,172 @@
...
+	);
+
+	const r = Buffer.from(signObj.signature.slice(0, 32));
+	const s = Buffer.from(signObj.signature.slice(32, 64));
+	const v = signObj.recid + 27;
+
+	return {
+		message: data,
+		messageHash: hash,
+		v: `0x${v.toString(16)}`,
```
luu-alex, line 95: Is there any difference using `toString(16)` and `toString("hex")`

jdevcs: v here is number and we can only pass an integer in its [toString(](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toString#parameters)) from 2 to 36 as base.

packages/web3-eth-accounts/src/account.ts
jdevcs: made this change. It was passing as can be seen in PR CI / CD actions.

packages/web3-eth-accounts/package.json
jdevcs: issue created

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,173 @@
...
+
+/**
+ * Recovers the Ethereum address which was used to sign the given data
+ */
+export const recover = (
+	data: string | signatureObject,
+	signature?: string,
+	hashed?: boolean,
+): Address => {
+	if (typeof data === 'object') {
```
luu-alex, line 160: data should be validated, otherwise lgtm

tsconfig.json
jdevcs: I enabled this for allowing default import of:
`../../node_modules/ethereumjs-util/dist/constants.d.ts(2,8): error TS1259: Module '"/web3.js/node_modules/@types/bn.js/index"' can only be default-imported using the 'esModuleInterop' flag`

CHANGELOG.md
```diff
@@ -346,3 +346,5 @@ Released with 1.0.0-beta.37 code base.
 
 1. `create` function does not take in the optional parameter `entropy`
 2. `ignoreLength` will be removed as an optional parameter for `privateKeyToAccount`
+3. `signTransaction` will not fill any default values and it will only sign and return result. For filling default values use web3.transaction package
```
```suggestion
3. `signTransaction` will not fill any default values and it will only sign and return result. For filling default values use `web3-eth-transaction` package
```

packages/web3-common/src/errors.ts
```diff
@@ -286,6 +289,13 @@ export class TransactionOutOfGasError extends TransactionError {
 	}
 }
 
+export class UndefinedRAWTransactionError extends TransactionError {
```
Spacesai1or:
```suggestion
export class UndefinedRawTransactionError extends TransactionError {
```

packages/web3-common/src/errors.ts
```diff
@@ -286,6 +289,13 @@ export class TransactionOutOfGasError extends TransactionError {
...
+	public constructor() {
+		super(`Transaction Undefined`);
```
Spacesai1or:
```suggestion
		super(`Raw transaction undefined`);
```

packages/web3-common/src/constants.ts
```diff
@@ -29,6 +29,7 @@ export const ERR_TX_NO_CONTRACT_ADDRESS = 403;
 export const ERR_TX_CONTRACT_NOT_STORED = 404;
 export const ERR_TX_REVERT_WITHOUT_REASON = 405;
 export const ERR_TX_OUT_OF_GAS = 406;
+export const ERR_TX_UNDEFINED = 407;
```
Spacesai1or:
```suggestion
export const ERR_RAW_TX_UNDEFINED = 407;
```

packages/web3-common/src/errors.ts
```diff
@@ -37,6 +37,9 @@ import {
 	ERR_ABI_ENCODING,
 	ERR_INVALID_PRIVATE_KEY,
 	ERR_PRIVATE_KEY_LENGTH,
+	ERR_SIGNATURE_FAILED,
+	ERR_INVALID_SIGNATURE,
+	ERR_TX_UNDEFINED,
```
Spacesai1or:
```suggestion
	ERR_RAW_TX_UNDEFINED,
```

packages/web3-common/src/errors.ts
```diff
@@ -286,6 +289,13 @@ export class TransactionOutOfGasError extends TransactionError {
 	}
 }
 
+export class UndefinedRAWTransactionError extends TransactionError {
+	public constructor() {
+		super(`Transaction Undefined`);
+		this.code = ERR_TX_UNDEFINED;
```
Spacesai1or:
```suggestion
		this.code = ERR_RAW_TX_UNDEFINED;
```

packages/web3-common/src/constants.ts
Spacesai1or: Because this is only used for `UndefinedRAWTransactionError` I would change the name to be more explicit, since we might have a `ERR_TX_UNDEFINED` used for `web3-eth-tx` package

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,173 @@
...
-// TODO Will be added later
-export const sign = (): boolean => true;
+/**
+ * Hashes the given message. The data will be UTF-8 HEX decoded and enveloped as follows: "\x19Ethereum Signed Message:\n" + message.length + message and hashed using keccak256.
+ */
 
-// TODO Will be added later
-export const signTransaction = (): boolean => true;
+export const hashMessage = (message: string): string => {
+	const messageHex = isHexStrict(message) ? message : utf8ToHex(message);
```
Spacesai1or, line 60: We should use the [validateStringInput](https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-utils/src/validation.ts#L135) method before this for `message`

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,173 @@
...
+
+	const ethMessage = Buffer.concat([Buffer.from(preamble), Buffer.from(messageBytes)]);
+
+	return `0x${Buffer.from(keccak256(ethMessage)).toString('hex')}`;
+};
+
+/**
+ * Signs arbitrary data. The value passed as the data parameter will be UTF-8 HEX decoded and wrapped as follows: "\x19Ethereum Signed Message:\n" + message.length + message
+ */
+export const sign = (data: string, privateKey: string): signResult => {
```
Spacesai1or: Validation for `data` and `privateKey` should be added, additionally I think we should support `buffer`s are input for `privateKey` since we support it in `web3-eth-accounts`

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,173 @@
...
+	// 64 hex characters + hex-prefix
+	if (privateKey.length !== 66) {
```
Spacesai1or: [isHexString32Bytes](https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-utils/src/validation.ts#L447)

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,173 @@
...
+		signature: `0x${Buffer.from(signObj.signature).toString('hex')}${v.toString(16)}`,
+	};
+};
+
+/**
+ *  Signs an Ethereum transaction with a given private key.
+ */
+export const signTransaction = (
+	transaction: TxData | AccessListEIP2930TxData | FeeMarketEIP1559TxData,
+	privateKey: string,
```
Spacesai1or: Should support `buffer`

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,173 @@
...
+): signTransactionResult => {
+	//	TO DO : Send calls to web3.transaction package for :
```
```suggestion
	//	TODO : Send calls to web3.transaction package for :
```

So VSCode will detect the `TODO`

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,173 @@
...
+	//		Transaction Validation checks
+
+	const tx = TransactionFactory.fromTxData(transaction);
+	const signedTx = tx.sign(Buffer.from(privateKey.substring(2), 'hex'));
+	if (signedTx.v === undefined || signedTx.r === undefined || signedTx.s === undefined)
+		throw new SignerError('Signer Error ');
```
```suggestion
		throw new SignerError('Signer Error');
```

packages/web3-common/src/errors.ts
```diff
@@ -398,14 +408,28 @@ export class AbiError extends Web3Error {
 
 export class PrivateKeyLengthError extends Web3Error {
 	public code = ERR_PRIVATE_KEY_LENGTH;
-	public constructor(value: string | Buffer) {
-		super(`Invalid value given "${value.toString()}". Error: Private key must be 32 bytes.`);
+	public constructor() {
+		super(`Error: Private key must be 32 bytes.`);
```
Spacesai1or:
```suggestion
		super(`Private key must be 32 bytes.`);
```

No need to have prefix `Error: ` JavaScript's native Error class will append it to the message

packages/web3-common/src/errors.ts
```diff
@@ -398,14 +408,28 @@ export class AbiError extends Web3Error {
...
 	}
 }
 
 export class InvalidPrivateKeyError extends Web3Error {
 	public code = ERR_INVALID_PRIVATE_KEY;
-	public constructor(value: string | Buffer) {
-		super(`Invalid value given "${String(value)}". Error: not a valid string or buffer.`);
+	public constructor() {
+		super(`Error: not a valid string or buffer.`);
```
Spacesai1or:
```suggestion
		super(`Not a valid string or buffer.`);
```

No need to have prefix `Error: ` JavaScript's native Error class will append it to the message

packages/web3-common/src/errors.ts
```diff
@@ -398,14 +408,28 @@ export class AbiError extends Web3Error {
...
+	}
+}
+
+export class SignerError extends Web3Error {
+	public code = ERR_SIGNATURE_FAILED;
+	public constructor(errorDetails: string) {
+		super(`Error: Invalid signature. "${errorDetails}"`);
```
Spacesai1or:
```suggestion
		super(`Invalid signature. "${errorDetails}"`);
```

No need to have prefix `Error: ` JavaScript's native Error class will append it to the message

packages/web3-common/src/errors.ts
```diff
@@ -398,14 +408,28 @@ export class AbiError extends Web3Error {
...
+	}
+}
+
+export class InvalidSignatureError extends Web3Error {
+	public code = ERR_INVALID_SIGNATURE;
+	public constructor(errorDetails: string) {
+		super(`Error: "${errorDetails}"`);
```
Spacesai1or:
```suggestion
		super(`"${errorDetails}"`);
```

No need to have prefix `Error: ` JavaScript's native Error class will append it to the message

packages/web3-eth-accounts/src/account.ts
Spacesai1or: I think this error message could be more descriptive, perhaps `JSON.stringify` `signedTx`?

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,173 @@
...
+	hashed?: boolean,
+): Address => {
+	if (typeof data === 'object') {
+		const signatureStr = `${data.r}${data.s.slice(2)}${data.v.slice(2)}`;
+		return recover(data.messageHash, signatureStr, true);
+	}
+
+	if (signature === undefined) throw new InvalidSignatureError('signature string undefined');
+
+	const V_INDEX = 130;
```
```suggestion
    // r = first 32 bytes, s = second 32 bytes, v = last byte of signature
	const V_INDEX = 130;
```

packages/web3-eth-accounts/src/account.ts
```diff
@@ -26,12 +177,12 @@ export const privateKeyToAccount = (
 ): {
 	address: string;
 	privateKey: string;
-	signTransaction: () => boolean; // From 1.x
-	sign: () => boolean;
+	signTransaction: signTransactionFunction; // From 1.x
+	sign: signFunction;
 	encrypt: () => boolean;
```
spacesailor24, line 182: You don't need to specify the return object like this, TypeScript will automatically infer it

packages/web3-eth-accounts/src/account.ts
```diff
@@ -26,12 +177,12 @@ export const privateKeyToAccount = (
...
 } => {
```
Spacesai1or: You don't need to specify the return object like this, TypeScript will automatically infer it

packages/web3-eth-accounts/src/account.ts
```diff
@@ -63,17 +214,11 @@ export const privateKeyToAccount = (
 export const create = (): {
 	address: HexString;
 	privateKey: string;
-	signTransaction: () => boolean; // From 1.x
-	sign: () => boolean;
+	signTransaction: signTransactionFunction; // From 1.x
+	sign: signFunction;
 	encrypt: () => boolean;
 } => {
```
Spacesai1or: You don't need to specify the return object like this, TypeScript will automatically infer it

packages/web3-eth-accounts/src/types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type signTransactionResult = signatureObject & {
+	rawTransaction: string;
+	transactionHash: string;
+};
+
+export type signTransactionFunction = (
+	transaction: TxData | AccessListEIP2930TxData | FeeMarketEIP1559TxData,
+	privateKey: string,
+) => signTransactionResult;
```
Spacesai1or, line 19:
```suggestion
```
 I don't think this is needed, TypeScript will infer the method signature

packages/web3-eth-accounts/src/types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type signResult = signatureObject & {
+	message?: string;
+	signature: string;
+};
+
+export type signFunction = (data: string, privateKey: string) => signResult;
```
Spacesai1or, line 26:
```suggestion
```

 I don't think this is needed, TypeScript will infer the method signature

packages/web3-eth-accounts/test/fixtures/account.ts
```diff
@@ -24,8 +24,82 @@ export const validPrivateKeytoAccountData: [string, any][] = [
 ];
 
 export const invalidPrivateKeytoAccountData: [any, string][] = [
```
Spacesai1or: Instead of providing just the error string here, you should provide the actual error expected to be thrown ([example](https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-utils/test/fixtures/validation.ts#L100))

packages/web3-eth-accounts/test/fixtures/account.ts
```diff
@@ -24,8 +24,82 @@ export const validPrivateKeytoAccountData: [string, any][] = [
...
+			privateKey: '0xbe6383dad004f233317e46ddb46ad31b16064d14447a95cc1d8c8d4bc61c3728',
+			data: 'Some data!%$$%&@*',
+			// signature done with personal_sign
+			signature:
+				'0x05252412b097c5d080c994d1ea12abcee6f1cae23feb225517a0b691a66e12866b3f54292f9cfef98f390670b4d010fc4af7fcd46e41d72870602c117b14921c1c',
+		},
+	],
+];
+
+export const transactionsTestData: [any][] = [
```
Spacesai1or: Should not use `any` but the transaction types from `@ethereumjs/tx`

Spacesai1or (changes requested):
Please review above suggestions

packages/web3-eth-accounts/src/account.ts
jdevcs: isHexString() will check if its type is string, else inside utf8ToHex() validateStringInput is already checked. so no need to do it explicitly here,

jdevcs: `isHexString()` will check if its type is `string`, else inside `utf8ToHex()` validateStringInput is already checked. so no need to do it explicitly here

packages/web3-eth-accounts/src/account.ts
jdevcs: we dnt have private key as buffer input in 1.x https://web3js.readthedocs.io/en/v1.5.2/web3-eth-accounts.html#sign

tsconfig.json
nazarhussain: I could not find the source of this using the default import. 

https://github.com/ethereumjs/ethereumjs-monorepo/blob/0c7bdfcdf5866fc31df2b87508156f3294bbae45/packages/util/src/constants.ts#L1-L2

If you are using it somewhere, you can import such with syntax as `import * as name from 'package'` without enabling this option.

nazarhussain (changes requested):

packages/web3-eth-accounts/src/account.ts
jdevcs: data is validated inside `hashMessage()`

packages/web3-eth-accounts/src/account.ts
jdevcs: we only have private key as hex string input in 1.x , its 1.x parity

packages/web3-eth-accounts/test/fixtures/account.ts
jdevcs: this test and is not part of this PR, please add new issue for improvements in existing approved PRs

packages/web3-eth-accounts/test/fixtures/account.ts
jdevcs: not required to validating types in this test

packages/web3-eth-accounts/src/account.ts
jdevcs: ts will infer but good to mention for user's usage or visibility

packages/web3-eth-accounts/src/account.ts
jdevcs: ts will infer but good to mention for user's usage or visibility

packages/web3-eth-accounts/src/types.ts
jdevcs: as we are returning this function in return type of other functions so good to mention type for user's usage or visibility

packages/web3-eth-accounts/src/types.ts
jdevcs: as we are returning this function in return type of other functions so good to mention type for user's usage or visibility

tsconfig.json
jdevcs: its not used in our package but as error is indicating its in node_modules, could you help me and check this branch on your system with esModuleInterop to false and see its reported by lib in node_modules

nazarhussain: I fixed it by changing the `tsconfig.json` only for that package where it cause error. 

Reported and issue to etherumjs  https://github.com/ethereumjs/ethereumjs-monorepo/issues/1615 and also created an issue on our repo for tracking https://github.com/ChainSafe/web3.js/issues/4636

nazarhussain (approved):

luu-alex (approved):

packages/web3-eth-accounts/test/fixtures/account.ts
Spacesai1or: It's not required but it helps the developer understand what the expected data is for the tests. Other packages use this ideology and we should remain consistent

packages/web3-eth-accounts/test/fixtures/account.ts
Spacesai1or: https://github.com/ChainSafe/web3.js/issues/4637

packages/web3-eth-accounts/src/account.ts
Spacesai1or: I understand, but to remain consistent with `web3-eth-accounts`, I'm suggesting to add support for buffers since it's a trivial addition

Aside from this, `privateKey` should be using the type `HexString` to match the rest of library where hex strings are expected and not arbitrary strings

packages/web3-eth-accounts/src/account.ts
Spacesai1or: Should use type `HexString`

packages/web3-eth-accounts/src/account.ts
```diff
@@ -1,22 +1,175 @@
...
+	};
+};
+
+/**
+ *  Signs an Ethereum transaction with a given private key.
+ */
+export const signTransaction = (
+	transaction: TxData | AccessListEIP2930TxData | FeeMarketEIP1559TxData,
+	privateKey: string,
+): signTransactionResult => {
```
spacesailor24, line 94:
```suggestion
) => {
```

Don't need to specify return type, it's inferred

packages/web3-eth-accounts/src/account.ts
jdevcs: `hashMessage` have some data validation and it throws error

Spacesai1or (approved):

PR #4617, opened 2021-12-10
wallet from web3core to wallet/accounts

### LINKED ISSUE #4475: Wallet basic functions
This contains an in memory wallet with multiple accounts. These accounts can be used when using web3.eth.sendTransaction()

`create` : Generates one or more accounts in the wallet. If wallets already exist they will not be overridden. params are number of accounts and optional entropy

`add` : Adds an account using a private key or account object to the wallet. Param is a private key or account object created with web3.eth.accounts.create().

`remove`: Removes an account from the wallet. param is the account address, or index in the wallet.

`clear` : Securely empties the wallet and removes all its accounts.

### LINKED ISSUE #4477: Wallet local storage
`save` : Stores the wallet encrypted and as string in local storage. params are password to encrypt before storage and optional keyname used for local storage position.

`load` : Loads a wallet from local storage and decrypts it. params are password and keyName if used non default during storage.

jdevcs:
## Description

## Type of change

wallet moved to accounts package
dependency updated
renamed local storage wallet it to wallet
removed entropy param for account creation ( breaking change )
updated change log

## Checklist:


packages/web3-common/src/web3_base_wallet.ts
```diff
@@ -22,7 +22,7 @@ export abstract class Web3BaseWallet<T extends Web3BaseWalletAccount> {
 		this._accountProvider = accountProvider;
 	}
 
-	public abstract create(numberOfAccounts: number, entropy: string): this;
```
jdevcs, line 25: Removed entropy as our account creation does not support this, https://github.com/ChainSafe/web3.js/blob/2ea05a8e9b251f618bdc54f0319f0c45896822af/packages/web3-eth-accounts/src/account.ts#L63 its breaking change

jdevcs:
Removed entropy as web3 account creation does not support this, https://github.com/ChainSafe/web3.js/blob/2ea05a8e9b251f618bdc54f0319f0c45896822af/packages/web3-eth-accounts/src/account.ts#L63 its breaking change

CHANGELOG.md
```diff
@@ -346,3 +341,5 @@ Released with 1.0.0-beta.37 code base.
 
 1. `create` function does not take in the optional parameter `entropy`
 2. `ignoreLength` will be removed as an optional parameter for `privateKeyToAccount`
+3. The `Wallet` no more supports address/number indexing. Have to use `wallet.get` instead.
+4. `Wallet.create` function doesn't accepts `entropy` param
```
nazarhussain, line 345: Would suggest to introduce such sensitive breaking change by creating an issue and linking it to PR. Referring back to a comment on PR is not an effective way to trace a breaking change.

packages/web3-eth-accounts/src/local_storage_wallet.ts
```diff
@@ -7,7 +7,7 @@ import {
 
 type BrowserError = { code: number; name: string };
 
-export class LocalStorageWallet<
+export class Wallet<
```
nazarhussain, line 10: Existing implementation is better matching with the name `LocalStorageWallet`. If want to keep old name for the sake of 1.x then export it in the `index.ts` instead with a deprecation warning.

nazarhussain (changes requested):

luu-alex (approved):

CHANGELOG.md
jdevcs: https://github.com/ChainSafe/web3.js/issues/4475#issuecomment-991093194

packages/web3-eth-accounts/src/local_storage_wallet.ts
jdevcs: lets closely follow 1.x

nazarhussain (approved):

PR #4628, opened 2021-12-14
Downgrade eslint from ^8.3.0 to ^7.32.0

### LINKED ISSUE #4593: Linting error with git commit on 4.x branch

### Current Behavior

```
✔ Preparing...
⚠ Running tasks...
  ↓ No staged files match *.js [SKIPPED]
  ❯ Running tasks for *.ts
    ✔ prettier --write
    ✖ eslint [FAILED]
  ↓ No staged files match *.{json,md} [SKIPPED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up...

✖ eslint:

Oops! Something went wrong! :(

ESLint: 8.3.0

Error: Rules with suggestions must set the `meta.hasSuggestions` property to `true`. `meta.docs.suggestion` is ignored by ESLint.
```

### Expected Behavior

It should run the linter without any error. 

### Steps to Reproduce

- Run `yarn` to register `husky` hooks
- Make some changes to any file
- Try to commit with git

### Web3.js Version

4.x

### Environment

- Operating System: Mac
- Browser: N/A
- Node.js Version: 16.x
- NPM Version: 7.x

### Anything Else?

- Running the standard linter with `npx eslint` or `yarn format` works fine.
- The only issue is with while running under git commit hook
- This seems to be introduced after merging this PR https://github.com/ChainSafe/web3.js/pull/4567 though the git commit was working fine on that particular branch.

Spacesai1or:
closes #4593

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
 	},
 	"devDependencies": {
 		"@types/node": "*",
-		"eslint": "^8.3.0",
+		"eslint": "^7.32.0",
 		"husky": "^7.0.4",
 		"jest": "^27.3.1",
 		"lerna": "^4.0.0",
--- a/packages/web3-common/package.json
+++ b/packages/web3-common/package.json
@@ -31,7 +31,7 @@
 		"@types/jest": "^27.0.3",
 		"@typescript-eslint/eslint-plugin": "^5.4.0",
 		"@typescript-eslint/parser": "^5.4.0",
-		"eslint": "^8.3.0",
+		"eslint": "^7.32.0",
 		"eslint-config-prettier": "^8.3.0",
 		"eslint-config-web3-base": "0.1.0",
 		"eslint-plugin-import": "^2.25.3",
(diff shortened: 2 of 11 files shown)


nazarhussain (changes requested):
We need to find the exact reason why we are facing that issue. Downgrading the package does not seems to be the right solution to me. 

The `lint-staged` package already [dependent](https://github.com/okonet/lint-staged/blob/e7b6412fb128f314346e28329c17a676cf691135/package.json#L53) on `eslint@8.x` so it should not be issue.

Spacesai1or:
> We need to find the exact reason why we are facing that issue. Downgrading the package does not seems to be the right solution to me.
> 
> The `lint-staged` package already [dependent](https://github.com/okonet/lint-staged/blob/e7b6412fb128f314346e28329c17a676cf691135/package.json#L53) on `eslint@8.x` so it should not be issue.

@nazarhussain [This](https://stackoverflow.com/a/69578831) is part of the reason why we're facing the error. From what I understand, one our co/dependencies has not updated according to `eslint@^8.0.0`

nazarhussain:
@spacesailor24 I am pretty sure this is not the problem. Try running `npx eslint` on any package and you will see it works fine. We need to inspect the problem further.

PR #4658, opened 2021-12-30
Replace the custom validation logic across packages with `web3-validator`

### LINKED ISSUE #4650: Replace the custom validation logic across packages with `web3-validator`
### Description 

Currently different validation logic is used across different packages. So to make it consistent we have to replace it's usage with `web3-validator` 

### Acceptance Criteria
- There should not custom validation error reporting 
- There should not be a validation logic in any package which is supported by `web3-validator` 
- The unused code should be removed.  

### Additional Information 
Need to look over each package for the validation usages. Some tests may be required to update/fix.

nazarhussain:
## Description

Replace all usages of validation with `web3-validator`.

Fixes #4650

## Type of change

## Checklist:


packages/web3-eth/src/rpc_methods.ts
```diff
@@ -97,9 +88,11 @@ export async function getStorageAt(
 	storageSlot: Uint256,
 	blockNumber: BlockNumberOrTag,
 ) {
-	validateAddress(address);
-	validateHexStringInput(storageSlot);
-	validateBlockNumberOrTag(blockNumber);
+	validator.validate(
+		['address', 'hex', 'blockNumberOrTag'],
+		[address, storageSlot, blockNumber],
+		{ silent: false },
```
Spacesai1or:
```suggestion
```

This doesn't need to be included, right? `silent: false` is the default [here](https://github.com/ChainSafe/web3.js/blob/e21d55bd6cdc8c181289e8faaf9afa691508d2e1/packages/web3-validator/src/web3_validator.ts#L57):

image

packages/web3-eth/src/rpc_methods.ts
```diff
@@ -505,10 +486,12 @@ export async function getFeeHistory(
...
 	rewardPercentiles: number[],
 ) {
-	validateHexStringInput(blockCount);
-	validateBlockNumberOrTag(newestBlock);
+	validator.validate(['hex', 'blockNumberOrTag'], [blockCount, newestBlock]);
+
 	for (const rewardPercentile of rewardPercentiles) {
-		validateNumbersInput(rewardPercentile, { onlyIntegers: false });
+		validator.validate(['number'], [rewardPercentile], {
+			silent: false,
```
Spacesai1or: Is the `{ silent: false }` necessary here?

packages/web3-utils/src/converters.ts
```diff
@@ -72,7 +64,7 @@ export type EtherUnits = keyof typeof ethUnitMap;
 
 /** @internal */
 const bytesToBuffer = (data: Bytes): Buffer | never => {
-	validateBytesInput(data);
+	validator.validate(['bytes'], [data], { silent: false });
```
Spacesai1or: More `{ silent: false }`s in this file

packages/web3-utils/src/converters.ts
```diff
@@ -106,7 +98,11 @@ export const hexToBytes = (bytes: HexString): Buffer => bytesToBuffer(bytes);
  * Converts value to it's number representation
  */
 export const hexToNumber = (value: HexString): bigint | number => {
-	validateHexStringInput(value);
+	validator.validateJSONSchema(
+		{ type: 'array', items: [{ type: 'string', format: 'hexStrict' }] },
+		[value],
+		{ silent: false },
+	);
```
Spacesai1or: Is there a reason this isn't:

```typescript
validator.validate(['hex'], [value]);
```

packages/web3-utils/src/converters.ts
```diff
@@ -127,7 +123,7 @@ export const toDecimal = hexToNumber;
  * Converts value to it's hex representation
  */
 export const numberToHex = (value: Numbers): HexString => {
-	validateNumbersInput(value, { onlyIntegers: true });
+	validator.validate(['int'], [value], { silent: false });
```
Spacesai1or:
```suggestion
	validator.validate(['number'], [value]);
```

packages/web3-utils/src/converters.ts
```diff
@@ -332,7 +329,14 @@ export const fromWei = (number: Numbers, unit: EtherUnits): string => {
...
 export const toWei = (number: Numbers, unit: EtherUnits): string => {
-	validateNumbersInput(number, { onlyIntegers: false });
+	validator.validateJSONSchema(
+		{
+			type: 'array',
+			items: { anyOf: [{ type: 'string', format: 'number' }, { type: 'number' }] },
+		},
+		[number],
+		{ silent: false },
+	);
```
Spacesai1or:
```suggestion
	validator.validate(['number'], [number]);
```

packages/web3-utils/src/string_manipulation.ts
```diff
@@ -11,7 +11,7 @@ export const padLeft = (value: Numbers, characterAmount: number, sign = '0'): st
 		return value.padStart(characterAmount, sign);
 	}
 
-	validateNumbersInput(value, { onlyIntegers: true });
+	validator.validate(['int'], [value], { silent: false });
```
Spacesai1or: More `{ silent: false }`s in this file

packages/web3-utils/src/validation.ts
```diff
@@ -1,132 +1,89 @@
...
+	isTopicInBloom as isTopicInBloomValidator,
+	isUserEthereumAddressInBloom as isUserEthereumAddressInBloomValidator,
+} from 'web3-validator';
+import { InvalidBlockError } from './errors';
+import { Numbers } from './types';
 
-export const isHexStrict = (hex: unknown) =>
-	typeof hex === 'string' && /^(-)?0x[0-9a-f]*$/i.test(hex);
+/**
+ * @deprecated Will be removed in next release. Please use `web3-validator` package instead.
```
Spacesai1or, line 17: I don't think there's a plan to deprecate the methods

packages/web3-utils/test/fixtures/string_manipulation.ts
```diff
@@ -43,56 +41,44 @@ export const padRightData: [[Numbers, number, string], HexString][] = [
...
 	[[BigInt(9), 1], -7],
 ];
 
 export const toTwosComplementInvalidData: [[Numbers, number], string][] = [
-	[['ab', 32], 'Invalid value given "ab". Error: not a valid integer.'],
-	[['-ab', 3], 'Invalid value given "-ab". Error: not a valid integer.'],
-	[['ab0x', 2], 'Invalid value given "ab0x". Error: not a valid integer.'],
-	[[25.5, 32], 'Invalid value given "25.5". Error: not a valid integer.'],
-	[['-120.0', 4], 'Invalid value given "-120.0". Error: not a valid integer.'],
+	// solidity only store 32 bytes numbersw
```
Spacesai1or:
```suggestion
	// solidity only store 32 bytes numbers
```

nazarhussain:
@spacesailor24 I tend to use `{silent: false}` to avoid breaking changes in future to dependent packages if we change default behavior of the validator. I had fixed all the places now.

packages/web3-utils/src/validation.ts
nazarhussain: Considering a particular package just for validation, we should not expose these functions from the `web3-utils`. For now I kept it for the API compatibility. But in future we must remove those. I didn't mentioned the future version in which we deprecate. So when we plan we can put the notice for a specific version.

Spacesai1or (approved):

packages/web3-utils/src/converters.ts
```diff
@@ -106,7 +98,10 @@ export const hexToBytes = (bytes: HexString): Buffer => bytesToBuffer(bytes);
  * Converts value to it's number representation
  */
 export const hexToNumber = (value: HexString): bigint | number => {
-	validateHexStringInput(value);
+	validator.validateJSONSchema(
+		{ type: 'array', items: [{ type: 'string', format: 'hexStrict' }] },
```
luu-alex: why do we use `validateJSONSchema` instead of `validate(['hexstring', [value]])`?

Spacesai1or: We should be using `validate(['hexstring', [value]])` (that's what a bulk of my comments were about above, looks like I missed this one, thank you!)

luu-alex (approved):

PR #4666, opened 2022-01-03
Web3 IPC Provider

### LINKED ISSUE #4643: Create web3-providers-ipc package structure for the 4.x
### Description
We need a package skeleton to be available to start rewriting the `web3-providers-ipc` package.

### Acceptance Criteria
- Package must have same package information as 1.x branch
- Package must have jest unit test configurations

### Additional Information
Unit tests should have test coverage support as well.

### LINKED ISSUE #4644: Create Web3IPCProvider class and initialize IPC connection library
### Description 
Create a skeleton for the class `Web3IPCProvider` which is extended by the `Web3BaseProvider` class. 

### Acceptance Criteria
- All the abstract functions must be filled by place holders 
- The class should initialize the underlying library to be used for the handling of IPC connection. 
- Unit tests must cover the initialization of the library and the class

### Additional Information 
In 1.x a library called `oboe`​​ is used to parse and process data streams. Such a stream manipulation should be handled by the low level libraries not by us. See if there is an alternative library available to avoid such stream handling.

### LINKED ISSUE #4645: Implement the connection related methods for  `Web3IPCProvider`
### Description 
We need to implement the abstract methods `connect`, `reconnect` and `reset` for the `Web3IPCProvider`. 

### Acceptance Criteria

- The `connect` method should initialize the wire connection not before. 
- The connection statuses must be properly managed and accessible by `getStatus` 

### Additional Information 
In case some of these methods are found irrelevant to implement, fill those in with empty implementation.

### LINKED ISSUE #4646: Implement the `request` method for `Web3IPCProvider`
### Description 

We need to implement the abstract method `request` for the `Web3IPCProvider`. The method should follow the similar approach used in other `Web3HTTPProvider` and `Web3WSProvider`. 

### Acceptance Criteria

- The `request` should use the underlying connection to send the data 
- The `request` method must be asynchronous and return the result based on generic type.

### Additional Information 

The `request` method contains an `options` parameter, which is open to the implementation. See what kind of options would be useful per request. The request flow is asynchronous so it must return a deferred promise.

### LINKED ISSUE #4647: Implement the events related behaviour for  `Web3IPCProvider`
### Description 
We need to implement the abstract methods, `on`, `once`, `removeListener` and others related to event handling  for the `Web3IPCProvider`. The method should follow the similar approach used in other `Web3HTTPProvider` and `Web3WSProvider`. 

### Acceptance Criteria
- The events does not support subscription then it must be reported by `supportsSubscriptions`
- The `disconnect` should remove all existing event listeners. 
- The `reconnect` should keep the existing event listeners. 

### Additional Information 
The total of `message`, `connect`, `disconnect` , `chainChanged` and `accountsChanged` events are required to be implemented. If some of the events can’t be implemented over the IPC then we can skip those.

nazarhussain:
## Description

IPC Provider for web3.

Fixes #4643 , #4644 , #4645 , #4646 , #4647 

## Type of change

## Checklist:


packages/web3-providers-http/package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "web3-providers-http",
 	"version": "4.0.0-alpha.1",
-	"description": "HTTP provider for Web3 4.x.x",
+	"description": "Module to handle web3 RPC connections over IPC sockets.",
```
jdevcs: this package json is in http package
```suggestion
	"description": "HTTP provider for Web3 4.x.x",
```

packages/web3-providers-ipc/package.json
```diff
@@ -0,0 +1,44 @@
+{
+	"name": "web3-providers-ipc",
+	"version": "4.0.0-alpha.1",
```
jdevcs:
```suggestion
	"version": "4.0.0-alpha.0",
```

packages/web3-providers-ipc/package.json
```diff
@@ -0,0 +1,44 @@
...
+	"description": "Package template for Web3 4.x.x",
```
jdevcs:
```suggestion
	"description": "Module to handle web3 RPC connections over IPC sockets",
```

jdevcs (dismissed):

packages/web3-providers-ipc/src/index.ts
```diff
@@ -0,0 +1,193 @@
...
+		this._connectionStatus = 'disconnected';
+		this._emitter.emit('disconnect');
+	}
+
+	private _onClose(event: CloseEvent): void {
+		this._clearQueues(event);
+		this._removeSocketListeners();
+	}
+
+	private _addSocketListeners(): void {
```
Spacesai1or, line 180: [1.x](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-providers-ipc/src/index.js#L86) had listeners `error` and `timeout`, should we add them?

packages/web3-providers-ipc/src/index.ts
```diff
@@ -0,0 +1,193 @@
...
+		this._emitter.on(type, callback);
+	}
+
+	public once<T = JsonRpcResult>(type: string, callback: Web3BaseProviderCallback<T>): void {
+		this._emitter.once(type, callback);
+	}
+
+	public removeListener(type: string, callback: Web3BaseProviderCallback): void {
+		this._emitter.removeListener(type, callback);
+	}
```
Spacesai1or, line 70: Should we be validating `callback` as done in [1.x](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-providers-ipc/src/index.js#L225)?

packages/web3-providers-ipc/src/index.ts
```diff
@@ -0,0 +1,193 @@
...
+		if (request.id === undefined) throw new Error('Request Id not defined');
+
+		if (this.getStatus() !== 'connected') {
+			throw new ConnectionNotOpenError();
+		}
+
+		try {
+			this._socket.write(JSON.stringify(request));
+			const defPromise = new DeferredPromise<JsonRpcResponse<ResponseType>>();
+			this._requestQueue.set(request.id, defPromise);
```
Spacesai1or:
```suggestion
			const defPromise = new DeferredPromise<JsonRpcResponse<ResponseType>>();
			this._requestQueue.set(request.id, defPromise);
			this._socket.write(JSON.stringify(request));
```

Should it be written this way to avoid possible race conditions?

packages/web3-providers-ipc/src/index.ts
```diff
@@ -0,0 +1,193 @@
...
+
+	public once<T = JsonRpcResult>(type: string, callback: Web3BaseProviderCallback<T>): void {
+		this._emitter.once(type, callback);
+	}
+
+	public removeListener(type: string, callback: Web3BaseProviderCallback): void {
+		this._emitter.removeListener(type, callback);
+	}
+
+	public connect(): void {
```
spacesailor24, line 72: Should update `_connectionStatus`

packages/web3-providers-ipc/src/index.ts
```diff
@@ -0,0 +1,193 @@
...
+
+		try {
+			this._addSocketListeners();
+			this._socket.connect({ path: this._socketPath });
+		} catch (e) {
+			throw new InvalidConnectionError(this._socketPath);
+		}
+	}
+
+	public disconnect(): void {
```
spacesailor24, line 85: Should update `_connectionStatus`

spacesailor24, line 85: Should this call `_onDisconnect`?

Spacesai1or (approved):
Couple of minor questions

packages/web3-providers-ipc/src/index.ts
nazarhussain: As per the docs the `timeout` is really just an notification of inactivity of idle socket. So we should not be worried about it. 

> Emitted if the socket times out from inactivity. This is only to notify that the socket has been idle. The user must manually close the connection.


The `error` event is unnecessary in case we have `close` event, as `close` will be triggered directly after. 

> Emitted when an error occurs. The 'close' event will be called directly following this event.

packages/web3-providers-ipc/src/index.ts
nazarhussain: We are just passing it to the internal `this._emitter.removeListener`, if any validation required to remove it will be done `this._emitter.removeListener`

luu-alex:
LGTM, it looks like some IPC methods need tests though and can be done in a seperate PR/issue?

luu-alex (approved):

PR #4675, opened 2022-01-07
web3-eth-accounts: Update to latest version of uuid package

LucianBuzzo:
This change updates to the latest version of uuid, to fix an issue with Math.random() (displayed in a warning on install).
Older versions may use Math.random() in certain circumstances, which is known to be problematic.
See https://v8.dev/blog/math-random for details.

## Type of change

## Checklist:


LucianBuzzo:
Because the `web3-eth-accounts` module uses `uuid` with the `random` option, it appears to be unaffected by the issue described in my commit message. However it doesn't hurt to update the dependency and prevent a potentially misleading warning message from appearing.

jdevcs:
@LucianBuzzo Thanks for pointing this. 

Please could you remove all changes of your PR in  

- dist/web3.min.js 
- dist/web3.min.js.map and 
- packages/web3-eth-accounts/package-lock.json .

jdevcs (changes requested):
@LucianBuzzo Thanks for pointing this.

Please could you remove all changes of your PR in

- dist/web3.min.js
- dist/web3.min.js.map and
- packages/web3-eth-accounts/package-lock.json .

packages/web3-eth-accounts/package.json
```diff
@@ -20,7 +20,7 @@
         "eth-lib": "0.2.8",
         "ethereumjs-util": "^7.0.10",
         "scrypt-js": "^3.0.1",
-        "uuid": "3.3.2",
+        "uuid": "8.3.2",
```
nazarhussain, line 23: That's a drastic change. Let's hope nothing breaks 🤞

LucianBuzzo: @nazarhussain Are there specific things you are concerned about that we could add test cases for? What do you think the biggest risk factors are?

LucianBuzzo:
@jdevcs Done, btw, should the dist files be generated in an `npm prepack` stage instead of committed to version control?

jdevcs:
> @jdevcs Done, btw, should the dist files be generated in an `npm prepack` stage instead of committed to version control?

Thanks @LucianBuzzo , We dnt commit dist files in feature, bug fix or lib-update PRs. 

@spacesailor24 @luu-alex Review/approval for uuid lib update.

luu-alex:
Hey thank you for this PR :) from what it seems in the codebase we only use `uuid` [once](https://github.com/ChainSafe/web3.js/blob/2279a67e0702343764db5cae2dffc04048083952/packages/web3-eth-accounts/src/index.js#L598). It does look like the method is affected by the version upgrade but it is not breaking. 
Perhaps adding a unit testcase using `uuid.v4()` with the attribute random as an arg wouldn't hurt. Thoughts?

Also updating the [changelog](https://github.com/ChainSafe/web3.js/blob/1.x/CHANGELOG.md)  and detailing the change

jdevcs:
@LucianBuzzo please merge down 1.x branch into your branch and fix merge conflicts. Thanks

avkos (approved):

jdevcs (commented):
This change is addressed in https://github.com/web3/web3.js/pull/5529 so I am closing this PR.

PR #4678, opened 2022-01-10
Implement `Contract` public interfaces 

### LINKED ISSUE #4564: Implement `Contract` public interfaces
### Description

We need to implement a contract public interfaces which is not part of the contract ABI. e.g. `clone` 

### Acceptance Criteria

- The functions must follow the same API as `1.x` branch 
- The callback support must be dropped 
- All the public interfaces from `1.x` must be implemented 

### Additional Information

- The code must be covered with unit tests

nazarhussain:
## Description

Add all public interfaces for the Contract package. 

Fixes #4564

## Type of change

## Checklist:


packages/web3-core/src/web3_subscription_manager.ts
```diff
@@ -1,16 +1,20 @@
...
+	private readonly _subscriptions: Map<
+		string,
+		InstanceType<RegisteredSubs[keyof RegisteredSubs]>
+	> = new Map();
 
 	public constructor(
-		public readonly requestManager: Web3RequestManager,
-		public readonly registeredSubscriptions: ST,
+		public readonly requestManager: Web3RequestManager<API>,
+		public readonly registeredSubscriptions: RegisteredSubs,
```
Spacesai1or, line 17: We should stay consistent and not abbreviate if not needed

packages/web3-eth-abi/src/types.ts
```diff
@@ -33,21 +33,23 @@ export type AbiParameter = {
 type FragmentTypes = 'constructor' | 'event' | 'function' | 'fallback';
 
 export type AbiBaseFragment = {
-	readonly type: FragmentTypes;
+	readonly type: string | FragmentTypes;
```
Spacesai1or, line 30: Why must we add arbitrary `string`?

packages/web3-eth-contract/src/contract.ts
```diff
@@ -132,17 +172,132 @@ export class Contract<Abi extends ContractAbi>
...
+	private _createContractMethod(abi: AbiFunctionFragment): ContractBoundMethod {
+		return (...params) => {
+			validator.validate(abi.inputs, params);
+
+			if (abi.stateMutability === 'payable' || abi.stateMutability === 'pure') {
+				return {
+					arguments: params,
+					call: async (options?: PayableCallOptions, block?: BlockNumberOrTag) =>
+						this._contractMethodCall(abi, params, options, block),
+					// TODO: Use `web3-eth-tx` package to return `PromiEvent` instead.
```
Spacesai1or, line 235: Perhaps we should be creating [web3-core-promievent](https://github.com/ChainSafe/web3.js/tree/1.x/packages/web3-core-promievent)

packages/web3-eth-contract/src/contract.ts
```diff
@@ -132,17 +172,132 @@ export class Contract<Abi extends ContractAbi>
...
+			}),
+		);
+	}
+
+	private async _contractMethodEstimateGas<
+		Options extends PayableCallOptions | NonPayableCallOptions,
+	>(abi: AbiFunctionFragment, params: unknown[], options?: Options, block?: BlockNumberOrTag) {
+		return hexToNumber(
+			await this.requestManager.send({
+				method: 'eth_estimateGas',
```
Spacesai1or, line 306: Should be using `web3Eth.estimateGas`

packages/web3-eth-contract/src/contract.ts
```diff
@@ -132,17 +172,132 @@ export class Contract<Abi extends ContractAbi>
...
+					getSendTxParams({
+						abi,
+						params,
+						options,
+						contractOptions: this.options,
+					}),
+				],
+			}),
+		);
+	}
```
Spacesai1or: https://github.com/ChainSafe/web3.js/issues/4685

packages/web3-eth-contract/src/log_subscription.ts
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	protected _buildSubscriptionParams() {
+		return ['logs', { address: this.address, topics: this.topics }] as [
+			'logs',
+			{ address?: string; topics?: string[] },
+		];
+	}
+
+	// eslint-disable-next-line class-methods-use-this
```
Spacesai1or: Is this necessary? Method uses `this.abi` and `this.emit`

packages/web3-eth-contract/src/types.ts
```diff
@@ -34,40 +42,40 @@ export interface ContractOptions {
...
 
-export interface NonPayableTx {
+export interface NonPayableCallOptions {
 	nonce?: Numbers;
 	chainId?: Numbers;
 	from?: Address;
 	to?: Address;
 	data?: HexString;
-	gas?: Numbers;
+	gas?: string;
```
Spacesai1or, line 51: Why not `Numbers`?

packages/web3-eth-contract/src/types.ts
```diff
@@ -34,40 +42,40 @@ export interface ContractOptions {
...
 	maxPriorityFeePerGas?: Numbers;
 	maxFeePerGas?: Numbers;
-	gasPrice?: Numbers;
+	gasPrice?: string;
```
Spacesai1or, line 54: Why not `Numbers`?

packages/web3-eth-contract/src/types.ts
```diff
@@ -34,40 +42,40 @@ export interface ContractOptions {
...
 }
 
-export interface PayableTx extends NonPayableTx {
-	value?: Numbers;
+export interface PayableCallOptions extends NonPayableCallOptions {
+	value?: string;
```
Spacesai1or, line 58: Why not `Numbers`?

packages/web3-utils/src/objects.ts
```diff
@@ -0,0 +1,29 @@
+const isIterable = (item: unknown): item is Record<string, unknown> =>
+	typeof item === 'object' && item !== null && !Array.isArray(item) && !Buffer.isBuffer(item);
+
+export const mergeDeep = (
```
Spacesai1or, line 6: Should have extensive tests before merging

packages/web3-core/src/web3_subscription_manager.ts
nazarhussain: Yes, it was old code. in refactored I removed the abbreviation.

packages/web3-eth-abi/src/types.ts
nazarhussain: If one uses Typescript but the ABI is declaring  without `as const` then the types for these attributes are not matched with the union values, so why we have to use a generic type for that cases.

packages/web3-eth-contract/src/contract.ts
nazarhussain: The `PromiEvent` is already created in the `web3-common`.

packages/web3-eth-contract/src/contract.ts
nazarhussain: Yes and no both. I want to avoid the circular dependency. 

If `web3-eth` is using `web3-eth-contract` package then we can't use the other way around. To provide `web3.eth.Contract` interface to users that dependency is required. 

Suggest if you have any other workaround?

packages/web3-eth-contract/src/contract.ts
nazarhussain: It's related to this comment as well. https://github.com/ChainSafe/web3.js/pull/4678#discussion_r782113877

packages/web3-eth-contract/src/log_subscription.ts
nazarhussain: Not necessary, will remove it.

packages/web3-eth-contract/src/types.ts
nazarhussain: It was not compatible with the type we used in the Eth specification. Either we have to change that specification type or here, or we parse convert the type in between.

packages/web3-eth-contract/src/types.ts
nazarhussain: Same as https://github.com/ChainSafe/web3.js/pull/4678#discussion_r782115733

packages/web3-eth-contract/src/types.ts
nazarhussain: Same as https://github.com/ChainSafe/web3.js/pull/4678#discussion_r782115733

packages/web3-common/src/json_rpc.ts
```diff
@@ -33,13 +34,22 @@ export const isResponseWithError = <Error = unknown, Result = unknown>(
 	(typeof response.id === 'number' || typeof response.id === 'string');
 
 export const isResponseWithNotification = <Result>(
-	response: JsonRpcNotification<Result>,
+	response: JsonRpcNotification<Result> | JsonRpcSubscriptionResult,
 ): response is JsonRpcNotification<Result> =>
 	!Array.isArray(response) &&
 	!!response &&
 	response.jsonrpc === '2.0' &&
 	response.params !== undefined;
```
jdevcs:
```suggestion
	response.params !== undefined &&
	response.method !== undefined;
```

packages/web3-common/src/json_rpc.ts
```diff
@@ -33,13 +34,22 @@ export const isResponseWithError = <Error = unknown, Result = unknown>(
...
 
+export const isSubscriptionResult = <Result>(
+	response: JsonRpcNotification<Result> | JsonRpcSubscriptionResult,
```
jdevcs, line 46: we should only allow passing `JsonRpcSubscriptionResult` type here.

packages/web3-common/src/json_rpc.ts
```diff
@@ -33,13 +34,22 @@ export const isResponseWithError = <Error = unknown, Result = unknown>(
 	(typeof response.id === 'number' || typeof response.id === 'string');
 
 export const isResponseWithNotification = <Result>(
-	response: JsonRpcNotification<Result>,
+	response: JsonRpcNotification<Result> | JsonRpcSubscriptionResult,
```
jdevcs, line 37: we should only allow passing `JsonRpcNotification` type here.

packages/web3-common/src/types.ts
```diff
@@ -38,6 +38,15 @@ export interface JsonRpcNotification<T = JsonRpcResult> {
 	readonly jsonrpc: JsonRpcIdentifier;
 	readonly method: string; // for subscription
 	readonly params: SubscriptionParams<T>; // for subscription results
+	readonly result: never;
+}
+
+export interface JsonRpcSubscriptionResult {
```
jdevcs, line 44: this is for getting subscription id in result field?

packages/web3-common/src/web3_base_provider.ts
```diff
@@ -6,18 +6,15 @@ import {
...
+	JsonRpcSubscriptionResult,
 } from './types';
 
-export interface ProviderMessage<T = JsonRpcResult> {
-	type: string;
-	data: T;
-}
-
 // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#connectivity
 export type Web3BaseProviderStatus = 'connecting' | 'connected' | 'disconnected';
```
jdevcs: These types should be in types.ts file:
```
Web3BaseProviderStatus
Web3BaseProviderCallback
```

jdevcs (changes requested):

packages/web3-core/src/web3_context.ts
```diff
@@ -2,21 +2,37 @@ import { Web3APISpec } from 'web3-common';
...
 	public static givenProvider?: SupportedProviders<never>;
 	public readonly providers = Web3RequestManager.providers;
 
 	public readonly requestManager: Web3RequestManager<API>;
+	public readonly subscriptionManager?: Web3SubscriptionManager<API, RegisteredSubs>;
 
 	public constructor(
 		provider: SupportedProviders<API> | string,
 		options?: Partial<Web3ConfigOptions>,
+		registeredSubscriptions?: RegisteredSubs,
```
jdevcs, line 24: Is this for continuation of already subscribed subs?

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -44,11 +44,11 @@ export class Web3RequestManager<
 	public constructor(provider?: SupportedProviders<API> | string | string, net?: Socket) {
 		super();
 
+		this._providers = Web3RequestManager.providers;
```
jdevcs:
```suggestion
		
```

Also remove :
`private readonly _providers: { [key: string]: Web3BaseProviderConstructor };`

and 

```
	public static get providers() {
		return {
			HttpProvider: HttpProvider as Web3BaseProviderConstructor,
			WebsocketProvider: WSProvider as Web3BaseProviderConstructor,
			IpcProvider: IpcProvider as Web3BaseProviderConstructor,
		};
	}
```
and directly return supported provider constructors inside get providers functions:

```
	public get providers() {
	return {
			HttpProvider: HttpProvider as Web3BaseProviderConstructor,
			WebsocketProvider: WSProvider as Web3BaseProviderConstructor,
			IpcProvider: IpcProvider as Web3BaseProviderConstructor,
		};
}
```

packages/web3-core/src/web3_subscription_manager.ts
```diff
@@ -52,7 +52,11 @@ export class Web3SubscriptionManager<
 		return this._subscriptions;
 	}
 
-	public async addSubscription(sub: Web3Subscription) {
+	public async addSubscription(sub: InstanceType<RegisteredSubs[keyof RegisteredSubs]>) {
+		if (!this.supportsSubscriptions()) {
+			throw new SubscriptionError('The current provider do not support subscriptions');
```
jdevcs:
```suggestion
			throw new SubscriptionError('The current provider does not support subscriptions');
```

jdevcs (changes requested):

packages/web3-common/src/json_rpc.ts
nazarhussain: If it's a single type, then there is no need of a type guard. Why we have this type guard is because the response we get is mix of these two types.

packages/web3-common/src/json_rpc.ts
nazarhussain: The response we get is a mix of these two types. Already tested it with local ganache.

packages/web3-common/src/types.ts
nazarhussain: Subscription Id is not **in** the result field rather the result field itself. 

```
>> {"id": 1, "method": "eth_subscribe", "params": ["newHeads", {}]}
<< {"jsonrpc":"2.0","id":1,"result":"0xcd0c3e8af590364c09d0fa6a1210faf5"}
```

packages/web3-core/src/web3_context.ts
nazarhussain: Not subscribed subscriptions, but available subscriptions to be subscribed from subscription manager.

packages/web3-eth-contract/src/contract.ts
```diff
@@ -37,28 +46,22 @@ export class Contract<Abi extends ContractAbi>
...
 
 		this._parseAndSetAddress(address);
 		this._parseAndSetJsonInterface(jsonInterface);
 
 		this.options = {
-			address: null,
+			address: undefined,
 			jsonInterface: [],
 			gas: options?.gas ?? options?.gasLimit ?? null,
 			gasPrice: options?.gasPrice ?? null,
```
jdevcs, line 67: These options doesn't have EIP1559 params support, please add support or create TODO issue.

jdevcs, line 106: These options doesn't have EIP1559 tx params support, please add support or create TODO issue

jdevcs (changes requested):

packages/web3-eth-contract/src/types.ts
```diff
@@ -34,40 +42,40 @@ export interface ContractOptions {
 	readonly from?: Address;
 	readonly data?: Bytes;
 	jsonInterface: ContractAbi;
-	address?: Address | null;
+	address?: Address;
```
jdevcs:
```suggestion
	address?: Address; //All transactions generated by web3.js from this contract will contain this address as the "to".
```

jdevcs (changes requested):

packages/web3-eth-contract/src/types.ts
```diff
@@ -34,40 +42,40 @@ export interface ContractOptions {
...
 }
 
 export interface ContractInitOptions {
-	readonly gas: Uint | null;
-	readonly gasPrice: Uint | null;
+	readonly gas?: Uint | null;
+	readonly gasPrice?: Uint | null;
```
jdevcs, line 36: Should also support EIP1559 type params, or create TODO new issue

packages/web3-eth-contract/src/types.ts
```diff
@@ -34,40 +42,40 @@ export interface ContractOptions {
 	readonly from?: Address;
```
jdevcs, line 28: Should also support EIP1559 type params, or create TODO new issue

packages/web3-eth-contract/src/utils.ts
```diff
@@ -0,0 +1,117 @@
...
+
+	if (!options?.from && !contractOptions.from) {
+		throw new Error('Contract "from" address not specified');
+	}
+
+	let txParams = mergeDeep(
+		{
+			to: contractOptions.address,
+			gas: contractOptions.gas,
+			gasPrice: contractOptions.gasPrice,
```
jdevcs, line 31: Should also support EIP1559 type params, or create TODO new issue

packages/web3-eth-contract/src/utils.ts
```diff
@@ -0,0 +1,117 @@
...
+		txParams = {
+			...txParams,
+			data: encodeMethodABI(abi, params, txParams.data),
+		};
+	}
+
+	return txParams;
+};
+
+export const getTxCallParams = ({
```
jdevcs:
```suggestion
export const getEthCallParams = ({
```

also change its usage in _contractMethodCall function inside contract.ts, more readability

packages/web3-eth-contract/src/utils.ts
```diff
@@ -0,0 +1,117 @@
...
+}): Partial<TransactionWithSender> => {
+	if (!options?.to && !contractOptions.address) {
+		throw new Error('Contract address not specified');
+	}
+
+	let txParams = mergeDeep(
+		{
+			to: contractOptions.address,
+			gas: contractOptions.gas,
+			gasPrice: contractOptions.gasPrice,
```
jdevcs, line 103: Should also support EIP1559 type params, or create TODO new issu

packages/web3-eth-contract/src/utils.ts
```diff
@@ -0,0 +1,117 @@
...
+	options,
+	contractOptions,
+}: {
+	abi: AbiFunctionFragment;
+	params: unknown[];
+	options?: PayableCallOptions | NonPayableCallOptions;
+	contractOptions: ContractOptions;
+}): TransactionWithSender & { data: string } => {
+	if (!options?.to && !contractOptions.address) {
+		throw new Error('Contract address not specified');
```
jdevcs: good to use web3 error constructor

packages/web3-eth-contract/src/utils.ts
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	if (!options?.from && !contractOptions.from) {
+		throw new Error('Contract "from" address not specified');
```
jdevcs: good to use web3 error constructor

packages/web3-eth-contract/src/utils.ts
```diff
@@ -0,0 +1,117 @@
...
+	options,
+	contractOptions,
+}: {
+	abi: AbiFunctionFragment;
+	params: unknown[];
+	options?: PayableCallOptions | NonPayableCallOptions;
+	contractOptions: ContractOptions;
+}): TransactionCall => {
+	if (!options?.to && !contractOptions.address) {
+		throw new Error('Contract address not specified');
```
jdevcs: good to use web3 error constructor , and same for rest of PR code

jdevcs:
Could you check CI Build / unit is also failing

nazarhussain:
@jdevcs Yes as mentioned this PR is WIP, so had not looked into tests yet. Related to your comments about EIP-1559  transactions, I used the [existing implementation](https://web3js.readthedocs.io/en/v1.5.0/web3-eth-contract.html#methods-mymethod-call) as reference. I wonder why it's already not supporting EIP-1559, and if not then should we include it in 4.x or not.

packages/web3-core/src/web3_context.ts
jdevcs: in that case why we need to pass available subscriptions via new context to new instance of subscription manager?

jdevcs:
> @jdevcs Yes as mentioned this PR is WIP, so had not looked into tests yet. Related to your comments about EIP-1559 transactions, I used the [existing implementation](https://web3js.readthedocs.io/en/v1.5.0/web3-eth-contract.html#methods-mymethod-call) as reference. I wonder why it's already not supporting EIP-1559, and if not then should we include it in 4.x or not.

Its imp to support eip1559 in 4.x for contract Txs. If its increasing scope of current PR , will be ok to open new issue.

packages/web3-common/src/web3_base_provider.ts
nazarhussain: Yes I agree to this change, will include in this PR. Would be nice to narrow down such points in the relevant PRs.

packages/web3-core/src/web3_context.ts
nazarhussain: Every instance of subscription manager needs to know which type of subscriptions is supported, so it can provide valid type safe interface for it.

packages/web3-core/src/web3_request_manager.ts
nazarhussain: Updated. Would be nice to review such points in the relevant PRs.

packages/web3-eth-contract/src/contract.ts
nazarhussain: https://github.com/ChainSafe/web3.js/issues/4685

packages/web3-eth-contract/src/types.ts
nazarhussain: https://github.com/ChainSafe/web3.js/issues/4685

packages/web3-eth-contract/src/types.ts
nazarhussain: https://github.com/ChainSafe/web3.js/issues/4685

packages/web3-eth-contract/src/utils.ts
nazarhussain: https://github.com/ChainSafe/web3.js/issues/4685

packages/web3-eth-contract/src/utils.ts
nazarhussain: https://github.com/ChainSafe/web3.js/issues/4685

nazarhussain:
@jdevcs Yes already added it to existing issue https://github.com/ChainSafe/web3.js/issues/4685

jdevcs (approved):

luu-alex (approved):

packages/web3-eth-abi/src/types.ts
```diff
@@ -33,21 +26,23 @@ export type AbiParameter = {
...
```
Spacesai1or, line 30:
```suggestion
        // type will default to string if passed ABI is declared without "as const"
	readonly type: string | FragmentTypes;
```

Spacesai1or: In response to:

> If one uses Typescript but the ABI is declaring without `as const` then the types for these attributes are not matched with the union values, so why we have to use a generic type for that cases.

packages/web3-eth-contract/src/contract.ts
Spacesai1or: I'd like to avoid having multiple implementations of the RPC methods, so we don't repeat the mistakes of `1.x`. I guess that means splitting the rpc wrappers into a separate package (`web3-eth-rpc-methods`)?

nazarhussain: I think `web3-eth-tx` may solve this issue by having common code that can avoid the circular dependency between `web3-eth` and `web3-eth-contract` package.

Spacesai1or: `web3-eth-tx` doesn't exist anymor. The tx util functions also needed access to `Web3Eth` methods, so it was made apart of `Web3Eth` for the same circular dependency issues

nazarhussain: In that case, I don't see any other way to have that small duplicate code to avoid circular dependency. Maybe as things shape up we may find some other solution. If you think of any other way around, please share.

Issue #4601, opened 2021-12-06
Migrate registry ABI from 1.x for 4.x rewrite


luu-alex:
Migrate the following ABI from web3-eth-ens 1.x branch to 4.x rewrite
Testing to deploy on private chain and testing the contract code.
Registry Abi:

owner
resolver
ttl
setOwner
setResolver
setTTL
setSubnodeOwner
setRecord
setSubnodeRecord
setApprovalForAll
isApprovedForAll
recordExists

### Related Resources

- https://docs.ens.domains/deploying-ens-on-a-private-chain
- [Github repo](https://github.com/ensdomains/ens) (may be deprecated)

mconnelly8:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/ChainSafe/web3.js/4601?planning-poker) with ZenHub @jdevcs @luu-alex @nazarhussain @spacesailor24

PR #4700, opened 2022-01-12
ens registry abi - 4.x

luu-alex:
This PR only includes the registry ABI for ENS #4601 

Migrate the following ABI from web3-eth-ens 1.x branch to 4.x rewrite
Testing to deploy on private chain and testing the contract code.
Registry Abi:

owner
resolver
ttl
setOwner
setResolver
setTTL
setSubnodeOwner
setRecord
setSubnodeRecord
setApprovalForAll
isApprovedForAll
recordExists

### DIFF
--- a/packages/web3-eth-ens/src/ABI/Registry.js
+++ b/packages/web3-eth-ens/src/ABI/Registry.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var REGISTRY = [
+const REGISTRY = [
 	{
 		anonymous: false,
 		inputs: [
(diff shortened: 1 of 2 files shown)


jdevcs:
Hi Alex, Could you check and fix, **Build / lint (pull_request)**

luu-alex:
@jdevcs  updated

nazarhussain (changes requested):
Why we have to declare it in JS file and not TS? 

I would suggest to declare it as TS and `as const`

nazarhussain (approved):

Issue #4600, opened 2021-12-06
ENS add new ABI contract functions for feature parity with ENS in 4.x 


luu-alex:
Feature parity with ENS for the ABI contract functions for 4.x rewrite
Includes:

ReverseRegistrar Abi:
- claim
- claimWithResolver
- setName
- node
- defaultResolver

TestRegistrar Abi:
- register
- expiryTimes

mconnelly8:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/ChainSafe/web3.js/4600?planning-poker) with ZenHub @jdevcs @luu-alex @nazarhussain @spacesailor24

PR #4714, opened 2022-01-18
ENS ABI contract ReverseRegistrar and TestRegistrar 

luu-alex:
closes:#4600
Feature parity with ENS for the ABI contract functions for 4.x rewrite
includes the registry ABI for ENS #4601 
Includes:

https://github.com/ensdomains/ens/blob/master/contracts/ReverseRegistrar.sol
ReverseRegistrar Abi:
- claim
- claimWithResolver
- setName
- node
- defaultResolver

https://github.com/ensdomains/ens/blob/master/contracts/TestRegistrar.sol
TestRegistrar Abi:
- register
- expiryTimes

Migrate the following ABI from web3-eth-ens 1.x branch to 4.x rewrite
Registry Abi:

owner
resolver
ttl
setOwner
setResolver
setTTL
setSubnodeOwner
setRecord
setSubnodeRecord
setApprovalForAll
isApprovedForAll
recordExists


packages/web3-eth-ens/src/ABI/ReverseRegistrar.ts
```diff
@@ -0,0 +1,140 @@
+const REVERSEREGISTRAR = [
```
nazarhussain: Why such naming convention? Why not `REVERSE_REGISTRAR` or simply `rerverseRegistrar`?

nazarhussain (changes requested):

packages/web3-eth-ens/src/ABI/ReverseRegistrar.ts
luu-alex: I was following for what was done in 1.x but I'm not too sure if we are following a convention. It seems like upper case letters is widely used but either sounds fine to me, ill go with `REVERSE_REGISTRAR `

nazarhussain (approved):

jdevcs (commented):
lgtm, 
could you also put links referring to solidity contract in comments?

jdevcs (approved):

Issue #4608, opened 2021-12-07
Remaining TODOs for `web3-eth`


Spacesai1or:
#### What's Missing

- ~`web3-eth` methods that use a `transaction` type for input need to be re-implemented to have parity for `1.x` (e.g. [sendTransaction](https://github.com/ChainSafe/web3.js/blob/0ff0e95ed1835f7e75b8866dfefeefc825a83c5a/packages/web3-eth/src/index.ts#L303))~ Going to be included in `web3-eth-tx`
- [x] Input formatting for `web3-eth` methods
- Various discussions needs to be had around added `TODO`s
    - [x] https://github.com/ChainSafe/web3.js/pull/4525#discussion_r755062664
- [getPendingTransactions](https://github.com/ChainSafe/web3.js/pull/4525#discussion_r755058180)
    - [x] Implemented
    - [x] Tested
- [requestAccounts](https://github.com/ChainSafe/web3.js/pull/4525#discussion_r755059602)
    - [x] Implemented
    - [x] Tested
- [getChainId](https://github.com/ChainSafe/web3.js/pull/4525#discussion_r755060082)
    - [x] Implemented
    - [x] Tested
- [getNodeInfo](https://github.com/ChainSafe/web3.js/pull/4525#discussion_r755060562)
    - [x] Implemented
    - [x] Tested
- [getProof](https://github.com/ChainSafe/web3.js/pull/4525#discussion_r755060976)
    - [x] Implemented
    - [x] Tested
- [x] [Replace](https://github.com/ChainSafe/web3.js/pull/4525#discussion_r755063447) with error constructor
- [x] Inherit `Web3Context` instead of [instantiating it](https://github.com/ChainSafe/web3.js/blob/0ff0e95ed1835f7e75b8866dfefeefc825a83c5a/packages/web3-eth/src/index.ts#L38)
- ~#4632~
- ~#4633~

PR #4718, opened 2022-01-19
ETH Tx Utilities (#4652)

### LINKED ISSUE #4652: `WebEthTx` class
- Should create an [@ethereumjs/tx](https://www.npmjs.com/package/@ethereumjs/tx) transaction instance using provided transaction object
- Static methods
    - The below methods should support a user defined method being passed as an argument to override default way of determination
        - `detectTransactionType` method
        - `validateTransaction` method
        - `fillTransaction` method
- Should have setting in `Web3Context` for default transaction type when `detectTransactionType` is inconclusive

Spacesai1or:
### What This PR Does

- Adds Ethereum transaction related util methods to format, populate, validate, and prepare transaction object to be signed/used
- Moves RPC wrapper methods from `packages/web3-eth/src/index.ts` to `packages/web3-eth/src/rpc_method_wrappers.ts` so that each wrapper is importable individually
- Adds transaction types that can be pass a valid [Numbers](https://github.com/ChainSafe/web3.js/blob/a808d0dc9af2af3a35db6cc5482d872d2cfb8e7b/packages/web3-utils/src/types.ts#L4) to specify type of numbers related properties

### Significant Changes Made

- `packages/web3-eth/src/eth_tx.ts`
    - Adds the following methods:
        - `formatTransaction` - Used to format a transaction object with `Numbers` property to a desired type
        - `detectTransactionType` - Returns a hex string of detected transaction type, or `undefined`
        - `validateTransactionForSigning` - Validates transaction object has needed properties to sign transaction
        - `populateTransaction` - Populates transaction object with `Web3Context` defaults or fetched values from provider
        - `prepareTransactionForSigning` - Calls `populateTransaction`, then `validateTransactionForSigning`, and returns an `@ethereumjs/tx` object that can be used for signing
- `packages/web3-eth/src/rpc_method_wrappers.ts`
    - Moved wrapped RPC methods from `Web3Eth` class to here as individually exported methods
- `packages/web3-eth/src/index.ts`
    - Class methods now use imported methods from`packages/web3-eth/src/rpc_method_wrappers.ts`
- `packages/web3-eth/src/types.ts`
    - Created [Transaction](https://github.com/ChainSafe/web3.js/blob/6ea21eb9063eb032904c12b7676024df6ea24de8/packages/web3-eth/src/types.ts#L44) type (uses `Numbers` for number properties)
    - Created [PopulatedUnsignedTransaction](https://github.com/ChainSafe/web3.js/blob/6ea21eb9063eb032904c12b7676024df6ea24de8/packages/web3-eth/src/types.ts#L68) type which represents a `Transaction` object with all necessary properties for signing
- `packages/web3-eth/src/validation.ts`
    - Helper validation methods for `packages/web3-eth/src/eth_tx.ts`
        - `validateCustomChainInfo`
        - `validateChainInfo`
        - `validateLegacyGas`
        - `validateFeeMarketGas`
        - `validateGas`
- `packages/web3-utils/src/converters.ts`
    - Add `InvalidConvertibleValueError` error to `convertToValidType`

closes #4652


packages/web3-eth/src/errors.ts
```diff
@@ -1,21 +1,159 @@
...
-		// Does not account for non JSON properties
+	public constructor(value: unknown) {
 		super(JSON.stringify(value), 'invalid transaction call');
 	}
 }
+
+export class MissingCustomChainError extends Web3Error {
+	public constructor() {
+		// TODO - Discuss what to pass as value
+		super('N/A', 'If tx.common is provided it must have tx.common.customChain');
```
Spacesai1or: `Web3Error` requires a string value to be passed to be included in the error message. For this error it doesn't make sense to pass a value in and I wasn't sure what to pass to `Web3Error`'s constructor

packages/web3-eth/src/errors.ts
```diff
@@ -1,21 +1,159 @@
...
+	}
+}
+
+export class MissingCustomChainIdError extends Web3Error {
+	public constructor() {
+		super(
+			// TODO - Discuss what to pass as value
+			'N/A',
+			'If tx.common is provided it must have tx.common.customChain and tx.common.customChain.chainId',
```
Spacesai1or: `Web3Error` requires a string value to be passed to be included in the error message. For this error it doesn't make sense to pass a value in and I wasn't sure what to pass to `Web3Error`'s constructor

packages/web3-eth/src/errors.ts
```diff
@@ -1,21 +1,159 @@
...
+	}
+}
+
+export class CommonOrChainAndHardforkError extends Web3Error {
+	public constructor() {
+		// TODO - Discuss what to pass as value, would it make sense to pass common, chain, and hardfork for error message here?
+		super(
+			'N/A',
+			'Please provide the @ethereumjs/common object or the chain and hardfork property but not all together.',
+		);
```
Spacesai1or: `Web3Error` requires a string value to be passed to be included in the error message. For this error it doesn't make sense to pass a value in and I wasn't sure what to pass to `Web3Error`'s constructor

packages/web3-eth/src/errors.ts
```diff
@@ -1,21 +1,159 @@
...
+
+export class MissingChainOrHardforkError extends Web3Error {
+	public constructor(value: { chain: string | undefined; hardfork: string | undefined }) {
+		// TODO - Discuss what to pass as value
+		super(
+			'N/A',
+			`When specifying chain and hardfork, both values must be defined. Received "chain": ${
+				value.chain ?? 'undefined'
+			}, "hardfork": ${value.hardfork ?? 'undefined'}`,
+		);
```
Spacesai1or: `Web3Error` requires a string value to be passed to be included in the error message. For this error it doesn't make sense to pass a value in and I wasn't sure what to pass to `Web3Error`'s constructor

packages/web3-eth/src/errors.ts
```diff
@@ -1,21 +1,159 @@
...
+			`nonce: ${value.nonce ?? 'undefined'}, chainId: ${value.chainId ?? 'undefined'}`,
+			'Nonce or chainId is lower than 0',
+		);
+	}
+}
+
+export class UnableToPopulateNonceError extends Web3Error {
+	public constructor() {
+		// TODO - Discuss what to pass as value
+		super('N/A', 'unable to populate nonce, no from address available');
```
Spacesai1or: `Web3Error` requires a string value to be passed to be included in the error message. For this error it doesn't make sense to pass a value in and I wasn't sure what to pass to `Web3Error`'s constructor

packages/web3-eth/src/errors.ts
```diff
@@ -1,21 +1,159 @@
...
+	}
+}
+
+export class Eip1559NotSupportedError extends Web3Error {
+	public constructor() {
+		// TODO - Discuss what to pass as value
+		super('N/A', "Network doesn't support eip-1559");
```
Spacesai1or: `Web3Error` requires a string value to be passed to be included in the error message. For this error it doesn't make sense to pass a value in and I wasn't sure what to pass to `Web3Error`'s constructor

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,326 @@
...
+					transaction.common?.customChain?.networkId,
+					desiredType,
+				),
+				chainId: convertToValidType(transaction.common?.customChain?.chainId, desiredType),
+			},
+		},
+	};
+	// TODO - TSC is complaining that ReturnType could be instantiated with an
+	// arbitrary type which could be unrelated to 'string | number | bigint | undefined'
+	return formattedTransaction as unknown as Transaction<ReturnType>;
```
Spacesai1or: @nazarhussain Do you have any idea on how to fix this? I thought `DesiredType extends ValidTypes` would fix this, but maybe I'm missing something?

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,319 @@
...
+	// TODO - Add default to Web3Context
+	if (populatedTransaction.chain === undefined) populatedTransaction.chain = 'mainnet';
+	// TODO - Add default to Web3Context
+	// TODO - Update default to berlin? (It's london in 1.x)
+	if (populatedTransaction.hardfork === undefined) populatedTransaction.hardfork = 'london';
+
+	if (populatedTransaction.chainId === undefined) {
+		if (populatedTransaction.common?.customChain.chainId === undefined) {
+			// TODO - web3Eth.getChainId not implemented
+			// populatedTransaction.chainId = await web3Eth.getChainId();
```
Spacesai1or: Is apart of #4608

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+	const block = await web3Eth.getBlock();
+	const hexTxType = toHex(populatedTransaction.type);
+
+	if (hexTxType < '0x0' || hexTxType > '0x2')
+		throw new UnsupportedTransactionTypeError(populatedTransaction.type);
+
+	if (hexTxType === '0x0' || hexTxType === '0x1') {
+		// transaction.type not supported before Berlin hardfork
+		// TODO - Maybe add check for populatedTransaction.hardfork >= Berlin before deleting
+		// if (hexTxType === '0x0') populatedTransaction.type = undefined;
```
Spacesai1or: So if `hardfork` is before `berlin`, we cannot include `type` as part of the transaction object. However, `type` should be included even if `type === '0x0'` if `hardfork` is >= `berlin`. So instead of just doing the following arbitrarily:

```typescript
 if (hexTxType === '0x0') populatedTransaction.type = undefined;
```

Should we first check if the `hardfork` is < `berlin`?

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+					BigInt(block.baseFeePerGas) * BigInt(2) +
+					BigInt(populatedTransaction.maxPriorityFeePerGas);
+		}
+	}
+
+	// TODO - TSC returns that as only PopulatedUnsignedBaseTransaction
+	return formatTransaction(
+		populatedTransaction,
+		desiredType,
+	) as PopulatedUnsignedTransaction<ReturnType>;
```
Spacesai1or: @nazarhussain Any ideas on what to do about this? The return type should be:

```typescript
export type PopulatedUnsignedTransaction<NumberType = Numbers> =
	| PopulatedUnsignedBaseTransaction<NumberType>
	| PopulatedUnsignedEip2930Transaction<NumberType>
	| PopulatedUnsignedEip1559Transaction<NumberType>;
```

but in my testing the IDE only picks up `PopulatedUnsignedBaseTransaction<NumberType>` and you have to do something like:

```typescript
const result = await populateTransaction(input, web3Context, ValidTypes.HexString);
(result as PopulatedUnsignedEip1559Transaction).maxPriorityFeePerGas
```

for TSC/ide not to complain about accessing properties that don't exist on `PopulatedUnsignedBaseTransaction`

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+}
+
+// TODO - Replace use of Web3Context with Web3Eth
```
Spacesai1or: I'm not sure if it makes sense to pass `Web3Context` to `populateTransaction` to make use of the `provider` and default values, of if passing `Web3Eth` makes more sense?

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+	if (transaction.maxFeePerGas !== undefined || transaction.maxPriorityFeePerGas !== undefined)
+		return '0x2';
+	if (transaction.hardfork === 'london') return '0x2';
+	if (transaction.common?.hardfork === 'london') return '0x2';
+
+	if (transaction.accessList !== undefined) return '0x1';
+	if (transaction.hardfork === 'berlin') return '0x1';
+	if (transaction.common?.hardfork === 'berlin') return '0x1';
+
+	return undefined;
```
jdevcs: If tx is legacy transaction without an envelope, it should be type 0.

packages/web3-eth/src/errors.ts
jdevcs: instead of 'N/A' may be good to pass empty string

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+};
+
+const validateCustomChainInfo = (transaction: Transaction) => {
```
jdevcs: I think it should return boolean, instead of user relying on error throw for detection of validation

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+			transaction.chainId !== transaction.common.customChain.chainId
+		)
+			throw new ChainIdMismatchError({
+				txChainId: transaction.chainId,
+				customChainId: transaction.common.customChain.chainId,
+			});
+	}
+};
+
+const validateChainInfo = (transaction: Transaction) => {
```
jdevcs: same here it should return boolean for easy validation checks.

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+	if (
+		transaction.common !== undefined &&
+		transaction.chain !== undefined &&
+		transaction.hardfork !== undefined
+	)
+		throw new CommonOrChainAndHardforkError();
```
jdevcs: I think user can define all these three together, or is there any reference I am missing.

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+	if (
+		(transaction.chain !== undefined && transaction.hardfork === undefined) ||
+		(transaction.hardfork !== undefined && transaction.chain === undefined)
+	)
+		throw new MissingChainOrHardforkError({
+			chain: transaction.chain,
+			hardfork: transaction.hardfork,
+		});
+};
```
jdevcs: We should also add check if `hardfork` is defined in `tx.hardfork` and inside `tx.common.hardfork` thn it should match

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+
+const validateGas = (transaction: Transaction<HexString>) => {
+	if (
+		transaction.gas === undefined &&
```
jdevcs: In legacy transactions we have `gasPrice` and `gasLimit`, what you are referring here in `transaction.gas` ?

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+		(transaction.chain !== undefined && transaction.hardfork === undefined) ||
+		(transaction.hardfork !== undefined && transaction.chain === undefined)
+	)
+		throw new MissingChainOrHardforkError({
+			chain: transaction.chain,
+			hardfork: transaction.hardfork,
+		});
+};
+
+const validateGas = (transaction: Transaction<HexString>) => {
```
jdevcs: Should also return boolean if validGas params in tx

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+	privateKey?: HexString,
+	overrideMethod?: (
+		transaction: Transaction,
+		web3Context: Web3Context<EthExecutionAPI>,
+	) => PopulatedUnsignedTransaction<ReturnType>,
+): Promise<PopulatedUnsignedTransaction<ReturnType>> {
+	if (overrideMethod !== undefined) return overrideMethod(transaction, web3Context);
+
+	const populatedTransaction = { ...transaction };
+	const web3Eth = new Web3Eth(web3Context.currentProvider);
```
jdevcs, line 223: I think for scalability requirement, its more good not to create new `Web3Eth` object on every call inside this function.

jdevcs: I think for scalability requirement, its more good not to create new `Web3Eth` object on every call inside this function.

jdevcs (changes requested):

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	populatedTransaction.type = detectTransactionType(populatedTransaction);
+	// TODO - After web3Context.defaultTxType is implemented
+	if (populatedTransaction.type === undefined) populatedTransaction.type = '0x0'; // web3Context.defaultTxType;
+
+	const block = await web3Eth.getBlock();
+	const hexTxType = toHex(populatedTransaction.type);
+
+	if (hexTxType < '0x0' || hexTxType > '0x2')
```
jdevcs:
```suggestion
	if (hexTxType < '0x0' || hexTxType > '0x7f') //https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2718.md#transactions
```

jdevcs (changes requested):

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+	if (populatedTransaction.gasLimit === undefined) {
+		if (populatedTransaction.gas !== undefined)
+			populatedTransaction.gasLimit = populatedTransaction.gas;
+	}
+
+	populatedTransaction.type = detectTransactionType(populatedTransaction);
+	// TODO - After web3Context.defaultTxType is implemented
+	if (populatedTransaction.type === undefined) populatedTransaction.type = '0x0'; // web3Context.defaultTxType;
+
+	const block = await web3Eth.getBlock();
```
jdevcs: `const block = await web3Eth.getBlock();` should be moved after line#288 : `if (hexTxType === '0x2') {`

as `block.baseFeePerGas` is only used if tx is 0x2, little performance imp.

packages/web3-eth/src/errors.ts
nazarhussain: Just pass in the name of the error constructor itself. `MissingCustomChainError`

packages/web3-eth/src/errors.ts
nazarhussain: Would suggest to pass name of the error constructor `MissingCustomChainIdError`. Or if it's a redundant scenario, create a error e.g. `Web3EthError` constructor which passes a generic message  like `eth error`  then create all other errors from it without any message.

packages/web3-eth/src/errors.ts
```diff
@@ -1,21 +1,159 @@
...
+}
+
+export class InvalidNonceOrChainIdError extends Web3Error {
+	public constructor(value: { nonce: Numbers | undefined; chainId: Numbers | undefined }) {
+		super(
+			`nonce: ${value.nonce ?? 'undefined'}, chainId: ${value.chainId ?? 'undefined'}`,
+			'Nonce or chainId is lower than 0',
+		);
+	}
+}
```
nazarhussain: In all above cases of different errors, I would suggest to assign different codes to each error as well. e.g. https://github.com/ChainSafe/web3.js/blob/5a00ab5b4098654280811bf098030e7db485441c/packages/web3-common/src/constants.ts#L24-L32

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+	MissingCustomChainError,
+	MissingCustomChainIdError,
+	MissingGasError,
+	UnableToPopulateNonceError,
+	UnsupportedFeeMarketError,
+	UnsupportedTransactionTypeError,
+} from './errors';
+import { PopulatedUnsignedTransaction, Transaction } from './types';
+
+export function formatTransaction<
```
nazarhussain, line 35: For all these formatters, if we can define valid JSON-Schema (which I believe we can) then we can write a generic formatter which takes objects+schema+format and convert it. That can be more extensible approach.

packages/web3-eth/src/eth_tx.ts
nazarhussain: Assign the variable earlier with a proper type then you don't need to override here. This will also help to identify which attributes its complaining about. 

```ts
const formattedTransaction: Transaction<ReturnType> = {
```

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+	};
+	// TODO - TSC is complaining that ReturnType could be instantiated with an
+	// arbitrary type which could be unrelated to 'string | number | bigint | undefined'
+	return formattedTransaction as unknown as Transaction<ReturnType>;
+}
+
+export const detectTransactionType = (
+	transaction: Transaction,
+	overrideMethod?: (transaction: Transaction) => Numbers | undefined,
+): Numbers | undefined => {
```
nazarhussain: That override method should be passed from the package level. And `detectTransactionType` should be default value for it. Means `web3.eth` would have a configuration `transactionTypeIdentifier` which will have default value `detectTransactionType`, so user can pass it's own implementation if want.

packages/web3-eth/src/eth_tx.ts
nazarhussain: The default behaviour of validation in the whole library is to throw error, so that's seems fine. 

In case decide to return boolean value then would better to rename it to `isValidCustomChainInfo`

packages/web3-eth/src/eth_tx.ts
nazarhussain: Same as above, if decided to return boolean rename it to `isValidChainInfo`. Make it a consistent pattern across all functions.

packages/web3-eth/src/eth_tx.ts
```diff
@@ -0,0 +1,315 @@
...
+			hardfork: transaction.hardfork,
+		});
+};
+
+const validateGas = (transaction: Transaction<HexString>) => {
+	if (
+		transaction.gas === undefined &&
+		transaction.gasLimit === undefined &&
+		transaction.maxPriorityFeePerGas === undefined &&
+		transaction.maxFeePerGas === undefined
```
nazarhussain: I prefer to destruct all such params into in condition into boolean identifiers expressions and then user them to construct complex conditions. That make it easy to read and help simplify conditions. 

```ts
const isGasPresent = transaction.gas === undefined && transaction.gasLimit === undefined;
const isMaxPriorityFeePresent = transaction.maxPriorityFeePerGas === undefined && transaction.maxFeePerGas === undefined;


if (isGasPresent && isMaxPriorityFeePresent) {
   hrow new MissingGasError({
}
```

packages/web3-eth/src/eth_tx.ts
nazarhussain: The return type `PopulatedUnsignedTransaction` is a union. If a function return type is set that, then return value could be any of union types. Users has to use `as` to cast to a concrete type before using it. 

Ideally the function `populateTransaction` should accept a generic which can be inferred from the parameter as well then return one concrete type. 

```ts
onst result = await populateTransaction<PopulatedUnsignedEip1559Transaction>(input, web3Context, ValidTypes.HexString);

// Now use can use it without type cast. 
result.maxPriorityFeePerGas
```

If we can somehow extract this generic type from any of the parameters `input`, `web3Context` then user don't need to pass the generic, but generic still be defined with the function.

packages/web3-eth/src/types.ts
```diff
@@ -8,135 +8,89 @@ import {
...
-	gasPrice?: Numbers;
-	type?: Numbers;
-	maxFeePerGas?: Numbers;
-	maxPriorityFeePerGas?: Numbers;
+	value?: NumberType;
+	gas?: NumberType;
+	gasPrice?: NumberType;
+	type?: NumberType;
+	maxFeePerGas?: NumberType;
+	maxPriorityFeePerGas?: NumberType;
```
nazarhussain: I am not fan for this approach. This approach becomes very open cause problems while using types. e.g. a user can specify `Error` as `NumberType` that obviously can't be valid. 

Ideal solution would be that `Numbers` type should be self aware what types it can be represented and then you pass generic to it with that given constraints.

nazarhussain (changes requested):

packages/web3-eth/src/errors.ts
Spacesai1or: That's a good idea, thank you! :rocket:

packages/web3-eth/src/errors.ts
Spacesai1or: I think including the error constructor is the better idea. I like being able to use `instanceOf` for each specific error

packages/web3-eth/src/eth_tx.ts
Spacesai1or: I would like to keep the `return undefined` so that it's explicit when this method fails to detect type (throwing an error is also an option, users would then just wrap in a `try/catch`). However, we could add detection for `0x0` transaction that most input to this method will probably satisfy before erroring/returning `undefined`

packages/web3-eth/src/eth_tx.ts
Spacesai1or: This is an internal method that's used only for `validateTransactionForSigning`, so I think it would make since to error out. We could add `isValidCustomChainInfo` in the future if it's needed, but as @nazarhussain said, it's the library default to error for `validate_` methods

Throwing errors and returning `void` allows for the developer to just call `validate_(input)` and carry on instead of having to always add a check e.g. `if(!validate_(input))`

packages/web3-eth/src/eth_tx.ts
Spacesai1or: This is [1.x behavior](https://github.com/ChainSafe/web3.js/blob/5df6818ee4a0a30fb64dd3e94d97964826489b7f/packages/web3-eth-accounts/src/index.js#L316)

packages/web3-eth/src/eth_tx.ts
Spacesai1or: This isn't a requirement in [1.x](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-eth-accounts/src/index.js#L175), and would be considered a breaking change

packages/web3-eth/src/eth_tx.ts
Spacesai1or: I think it's just an alias defined in 1.x ([usage](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-eth-accounts/src/index.js#L194) and [type](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-core/types/index.d.ts#L147), but it's weird the type doesn't include `gasLimit`)

packages/web3-eth/src/eth_tx.ts
Spacesai1or: Yea I agree, this was a quick and dirty way to get it done. Some alternatives I can think of:

1. Import the needed methods from [rpc_methods.ts](https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-eth/src/rpc_methods.ts) - this way the user only needs to pass [web3Context](https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-core/src/web3_context.ts), and we pass the `web3Context.requestManager` to the `rpc_method`. However, we do loose the benefit of input validation and automatic output formatting since those are only defined in `Web3Eth` (the reason I went initing `Web3Eth`)
2. Export each `Web3Eth` method (basically just already exported `rpc_methods.ts` with input validation and output formatting, plus some alias methods like [getBlockTransactionCount](https://github.com/ChainSafe/web3.js/blob/5a00ab5b4098654280811bf098030e7db485441c/packages/web3-eth/src/index.ts#L151) that use provided input to figure out what `rpc_method` to call), and then have `Web3Eth` include them as class methods - I think this makes more sense than 1., users/devs still have access to use the bare bones implementation (`rpc_methods.ts`) or the validated and formatted implementation

packages/web3-eth/src/eth_tx.ts
Spacesai1or: My reasoning was to be explicit in which types the library supports, but I see now that it doesn't make sense to error given the broad use of our library

packages/web3-eth/src/eth_tx.ts
jdevcs: Good to check `hardfork is >= berlin` for including tx type.

packages/web3-eth/src/eth_tx.ts
jdevcs: I think passing instance of web3Eth is more good, both for : 

- interacting with Eth without creating its instance in other function 
- getting context variables

packages/web3-eth/src/eth_tx.ts
jdevcs: Its recommended to never use try/catch for **control logic**. 
Currently in our case: if not valid throw error else donot throw error
Reason: performance, good pattern,

packages/web3-eth/src/eth_tx.ts
jdevcs: Is there any limitation of not passing object of Web3Eth to this function?

packages/web3-eth/src/eth_tx.ts
Spacesai1or: I think this method could be written a little more clearer:

```typescript
export const detectTransactionType = (
	transaction: Transaction,
	overrideMethod?: (transaction: Transaction) => HexString | undefined,
): HexString | undefined => {
	// TODO - Refactor overrideMethod
	if (overrideMethod !== undefined) return overrideMethod(transaction);
	if (transaction.type !== undefined)
		return convertToValidType(transaction.type, ValidTypes.HexString);

	if (
		transaction.maxFeePerGas !== undefined ||
		transaction.maxPriorityFeePerGas !== undefined ||
		transaction.hardfork === 'london' ||
		transaction.common?.hardfork === 'london'
	) return '0x2';

	if (
		transaction.accessList !== undefined ||
		transaction.hardfork === 'berlin' ||
		transaction.common?.hardfork === 'berlin'
	) return '0x1';

	return undefined;
};
```

But the [1.x implementation](https://github.com/ChainSafe/web3.js/blob/8783f4d64e424456bdc53b34ef1142d0a7cee4d7/packages/web3-eth-accounts/src/index.js#L353):

```typescript
function _handleTxType(tx) {
    // Taken from https://github.com/ethers-io/ethers.js/blob/2a7ce0e72a1e0c9469e10392b0329e75e341cf18/packages/abstract-signer/src.ts/index.ts#L215
    const hasEip1559 = (tx.maxFeePerGas !== undefined || tx.maxPriorityFeePerGas !== undefined);

    let txType;

    if (tx.type !== undefined) {
        txType = utils.toHex(tx.type)
    } else if (tx.type === undefined && hasEip1559) {
        txType = '0x2'
    }

    if (tx.gasPrice !== undefined && (txType === '0x2' || hasEip1559))
        throw Error("eip-1559 transactions don't support gasPrice");
    if ((txType === '0x1' || txType === '0x0') && hasEip1559)
        throw Error("pre-eip-1559 transaction don't support maxFeePerGas/maxPriorityFeePerGas");

    if (
        hasEip1559 ||
        (
            (tx.common && tx.common.hardfork && tx.common.hardfork.toLowerCase() === HardForks.London) ||
            (tx.hardfork && tx.hardfork.toLowerCase() === HardForks.London)
        )
    ) {
        txType = '0x2';
    } else if (
        tx.accessList ||
        (
            (tx.common && tx.common.hardfork && tx.common.hardfork.toLowerCase() === HardForks.Berlin) ||
            (tx.hardfork && tx.hardfork.toLowerCase() === HardForks.Berlin)
        )
    ) {
        txType = '0x1';
    }

    return txType
}
```

only returns: `utils.toHex(tx.type)`, `'0x2'`, `'0x1'`, and `undefined`. So the only difference between `detectTransactionType` and `1.x` is `detectTransactionType` will return `transaction.type` (what's passed by the user) whenever it's not `undefined`, while `1.x` would still run through the type checks and possibly overwrite what the user had set

packages/web3-eth/src/eth_tx.ts
Spacesai1or: For now, I don't think so. The `populateTransaction` method just needs access to `Web3Context`, so might as well only require that instead of an entire `Web3Eth` class instance that may not be available when this method is imported

packages/web3-eth/src/eth_tx.ts
jdevcs: This clean implementation looks more good as compared to 1.x, In 4.x we should not mutate obj / not update tx type inside this function and should return tx type. 

The changes in tx object should be done inside one `filling defaults function` ( unless user explicitly specified not to via boolean )

packages/web3-eth/src/types.ts
Spacesai1or: I believe this solves the issue?

```typescript
export interface Transaction<NumberType extends Numbers = Numbers> { ... }
```

Spacesai1or:
### TODOs

#### Unimplemented Feedback

- [Replacing formatters with JSON schemas](https://github.com/ChainSafe/web3.js/pull/4718/files#r789942037)

#### High Priority

- Migrated test from `1.x` doesn't have same signed transaction values
    - [r value doesn't match](https://github.com/ChainSafe/web3.js/blob/49171a15d58b3f541337cd3d95178e49f7d18ed7/packages/web3-eth/test/fixtures/prepare_transaction_for_signing.ts#L245)
    - [RLP encoded value doesn't match](https://github.com/ChainSafe/web3.js/blob/015d54624d512aef32c190732af804ed2222386f/packages/web3-eth/test/fixtures/prepare_transaction_for_signing.ts#L820)

#### Medium Priority

- [Add code](https://github.com/ChainSafe/web3.js/blob/49171a15d58b3f541337cd3d95178e49f7d18ed7/packages/web3-eth/src/eth_tx.ts#L175) to populate default account from `web3.eth.accounts.wallet` in `populateTransaction`
- [Add code](https://github.com/ChainSafe/web3.js/blob/49171a15d58b3f541337cd3d95178e49f7d18ed7/packages/web3-eth/src/eth_tx.ts#L214) to fetch `chainId` when `web3Eth.getChainId` is added
- [Add code](https://github.com/ChainSafe/web3.js/blob/57f6403de006c12d32d39f8fb5a0b06e06bcd6e2/packages/web3-eth/src/eth_tx.ts#L220) to fetch `networkId` when `web3Net.getNetworkId` is added
- Refactor override method functionality for transaction methods
    - [detectTransactionType](https://github.com/ChainSafe/web3.js/blob/49171a15d58b3f541337cd3d95178e49f7d18ed7/packages/web3-eth/src/eth_tx.ts#L105)
    - [prepareTransactionForSigning](https://github.com/ChainSafe/web3.js/blob/49171a15d58b3f541337cd3d95178e49f7d18ed7/packages/web3-eth/src/eth_tx.ts#L326)
- [Refactor](https://github.com/ChainSafe/web3.js/blob/49171a15d58b3f541337cd3d95178e49f7d18ed7/packages/web3-eth/src/eth_tx.ts#L276) typing to make TSC happy

#### Low Priority

- [Weird behavior](https://github.com/ChainSafe/web3.js/blob/49171a15d58b3f541337cd3d95178e49f7d18ed7/packages/web3-eth/src/eth_tx.ts#L39) when using `Object.assign` in `formatTransaction` method

packages/web3-eth/src/validation.ts
```diff
@@ -103,3 +119,122 @@ export function isTransactionCall(value: TransactionCall): boolean {
...
+			transaction.chainId !== undefined &&
+			transaction.chainId !== transaction.common.customChain.chainId
+		)
+			throw new ChainIdMismatchError({
+				txChainId: transaction.chainId,
+				customChainId: transaction.common.customChain.chainId,
+			});
+	}
+
+	// TODO - Should throw error?
```
Spacesai1or:
```suggestion
```

templates/tsconfig.json.tmpl
```diff
@@ -1,7 +1,8 @@
 {
 	"extends": "../../tsconfig",
 	"compilerOptions": {
-		"outDir": "dist"
+		"outDir": "dist",
+		"esModuleInterop": true
```
Spacesai1or: Needed after adding `@ethereumjs/` packages

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -0,0 +1,388 @@
...
+	rpcMethods.getProtocolVersion(web3Context.requestManager);
+
+export const isSyncing = async (web3Context: Web3Context<EthExecutionAPI>) =>
+	rpcMethods.getSyncing(web3Context.requestManager);
+
+export const getCoinbase = async (web3Context: Web3Context<EthExecutionAPI>) =>
+	rpcMethods.getCoinbase(web3Context.requestManager);
+
+export const isMining = async (web3Context: Web3Context<EthExecutionAPI>) =>
+	rpcMethods.getMining(web3Context.requestManager);
```
nazarhussain: Why we need these wrappers? Why not just pass the `requestManager` where where we are calling these methods?

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -0,0 +1,388 @@
...
+	web3Context: Web3Context<EthExecutionAPI>,
+	returnType?: ReturnType,
+) {
+	const response = await rpcMethods.getHashRate(web3Context.requestManager);
+
+	return convertToValidType(
+		response,
+		returnType ?? web3Context.defaultReturnType,
+	) as ValidReturnTypes[ReturnType];
+}
```
nazarhussain, line 37: If it's all about modifying return type from the RPC method, then why not make it part of the `rpcMethods`?

packages/web3-eth/src/validation.ts
```diff
@@ -103,3 +119,121 @@ export function isTransactionCall(value: TransactionCall): boolean {
...
+	if (
+		// This check is verifying gas and gasPrice aren't less than 0.
+		// transaction's number properties have been converted to HexStrings.
+		// JavaScript doesn't handle negative hex strings e.g. -0x1, but our
+		// numberToHex method does. -0x1 < 0 would result in false, so we must check if
+		// hex string is negative via the inclusion of -
+		transaction.gas === undefined ||
+		transaction.gasPrice === undefined ||
+		transaction.gas.startsWith('-') ||
+		transaction.gasPrice.startsWith('-')
```
nazarhussain, line 168: We can validate this with `uint` type using `web3-validator`

packages/web3-eth/src/validation.ts
```diff
@@ -103,3 +119,121 @@ export function isTransactionCall(value: TransactionCall): boolean {
...
+	)
+		throw new InvalidGasOrGasPrice({
+			gas: transaction.gas,
+			gasPrice: transaction.gasPrice,
+		});
+	if (transaction.maxFeePerGas !== undefined || transaction.maxPriorityFeePerGas !== undefined)
```
nazarhussain, line 174: Same here.

packages/web3-eth/src/validation.ts
```diff
@@ -103,3 +119,121 @@ export function isTransactionCall(value: TransactionCall): boolean {
...
+			maxPriorityFeePerGas: transaction.maxPriorityFeePerGas,
+			maxFeePerGas: transaction.maxFeePerGas,
+		});
+};
+
+/**
+ * This method checks if all required gas properties are present for either
+ * legacy gas (type 0x0 and 0x1) OR fee market transactions (0x2)
+ */
+export const validateGas = (transaction: Transaction<HexString>) => {
```
nazarhussain, line 209: In all these validator functions, where ever possible, use `web3-validator` instead of structural conditions.

packages/web3-eth/test/unit/format_transaction.test.ts
```diff
@@ -0,0 +1,96 @@
...
+	numberStringTransaction,
+	numberTransaction,
+} from '../fixtures/format_transaction';
+
+describe('formatTransaction', () => {
+	it.skip('should call override method', () => {
+		// const overrideFunction = jest.fn();
+		// formatTransaction(hexStringTransaction, ValidTypes.Number, overrideFunction);
+		// expect(overrideFunction).toHaveBeenCalledWith(hexStringTransaction);
+	});
```
nazarhussain, line 15: If this test is not necessary, then we can remove it.

templates/tsconfig.json.tmpl
nazarhussain: Please make this only for the package where necessary required, not for templates. This change applies to all packages.

nazarhussain (changes requested):

packages/web3-core/src/types.ts
```diff
@@ -48,8 +48,11 @@ export interface Web3ConfigOptions {
 	transactionPollingTimeout: number;
 	blockHeaderTimeout: number;
 	maxListenersWarningThreshold: number;
```
jdevcs, line 50: like 1.x If provider does not support event subscription, we should use `transactionPollingInterval` configurable via config options here?

```suggestion
    maxListenersWarningThreshold: number;
	transactionPollingInterval: number;
```

packages/web3-eth/src/rpc_method_wrappers.ts
jdevcs: we can remove this additional layer , by passing requestManager to these functions

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -0,0 +1,388 @@
...
+	rpcMethods.getWork(web3Context.requestManager);
+
+export const submitWork = async (
+	web3Context: Web3Context<EthExecutionAPI>,
+	nonce: HexString8Bytes,
+	seedHash: HexString32Bytes,
+	difficulty: HexString32Bytes,
+) => rpcMethods.submitWork(web3Context.requestManager, nonce, seedHash, difficulty);
+
+// TODO
```
jdevcs, line 288: this PR is already growing large, so good to create new issue with all TODOs in that, for tracking and planning in next sprint.

packages/web3-eth/src/validation.ts
```diff
@@ -103,3 +119,121 @@ export function isTransactionCall(value: TransactionCall): boolean {
 export function validateTransactionCall(value: TransactionCall) {
 	if (!isTransactionCall(value)) throw new InvalidTransactionCall(value);
 }
+
+export const validateCustomChainInfo = (transaction: Transaction) => {
+	if (transaction.common !== undefined) {
+		if (transaction.common.customChain === undefined) throw new MissingCustomChainError();
+		if (transaction.common.customChain.chainId === undefined)
```
jdevcs, line 126: according to 1.x doc `networkId` is also not optional , could you verify this as well

templates/tsconfig.json.tmpl
Spacesai1or: There appears to be a symbolic link between this file in the template and all packages - assumed this was intentional, but maybe it wasn't?

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: Because the rpc methods are the minimal amount of code to perform the request and adhere to the ETH spec, so should always input/output hex strings

packages/web3-eth/test/unit/format_transaction.test.ts
Spacesai1or: The override function is necessary, just not implemented. There are `TODO`s for this already

packages/web3-core/src/types.ts
Spacesai1or: Sure, not relevant to this PR though - created an [issue](https://github.com/ChainSafe/web3.js/issues/4764)

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: Specified [here](https://github.com/ChainSafe/web3.js/pull/4718#issuecomment-1032134238), if this PR is okay to be merged with these outstanding `TODO`s, than I will move them into an issue

packages/web3-eth/src/validation.ts
Spacesai1or: If not specified by the user, `networkId` was fetch via RPC call in [1.x](https://github.com/ChainSafe/web3.js/blob/8783f4d64e424456bdc53b34ef1142d0a7cee4d7/packages/web3-eth-accounts/src/index.js#L296). However, we don't have an RPC method for `net_version` yet

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: Addressed in [this](https://github.com/ChainSafe/web3.js/pull/4718/commits/380002cbae93d3bc3c6e5ba3efd7b74f20b6f0be) commit

packages/web3-eth/src/validation.ts
Spacesai1or: Shouldn't [these](https://github.com/ChainSafe/web3.js/blob/ddecdde7a2f11b8e4f969ec985e5c2c6fefed632/packages/web3-validator/test/fixtures/validation.ts#L68) tests fail for being signed (negative) ints?

packages/web3-eth/src/rpc_method_wrappers.ts
jdevcs: we should create issue for these [todo](https://github.com/ChainSafe/web3.js/pull/4718#issuecomment-1032134238) s,

jdevcs (approved):

luu-alex (approved):

templates/tsconfig.json.tmpl
nazarhussain: Yes symlink is intentional. But if we have to change some config for one package then you can remove that link and create a file. We did same for accounts package as well and there is an issue to revert that change in future https://github.com/ChainSafe/web3.js/issues/4636

packages/web3-eth/src/validation.ts
nazarhussain: Yes you are right it must fail. Created an issue for it. https://github.com/ChainSafe/web3.js/issues/4777

nazarhussain (approved):
@spacesailor24 Apart from `tsconfig` changes, rest of the PR seems good.

Issue #4690, opened 2022-01-11
IBAN Helper functions


jdevcs:
### Description:

- iso13616Prepare: Prepare an IBAN for mod 97 computation
- mod9710: Calculates the MOD 97 10 of the passed IBAN as specified in ISO7064
- mod9710: Calculates the MOD 97 10 of the passed IBAN as specified in ISO7064

Issue #4688, opened 2022-01-11
IBAN Validation and Direct/Indirect checkFunctions


jdevcs:
### Description:

- createIndirect:  Creates an indirect IBAN address from an institution and identifier.
- isValid: Checks if an IBAN address is valid.
- isDirect: Checks if the IBAN instance is direct.
- isInDirect: Checks if the IBAN instance is indirect.

mconnelly8:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/ChainSafe/web3.js/4688?planning-poker) with ZenHub @jdevcs @luu-alex @nazarhussain @spacesailor24

jdevcs:
These function are already covered in https://github.com/ChainSafe/web3.js/pull/4723/

PR #4723, opened 2022-01-20
4687/iban basic functions

### LINKED ISSUE #4687: IBAN Basic Functions
### Description:

- toAddress: Converts a direct IBAN address into an Ethereum address.
- toIban: Converts an Ethereum address to a direct IBAN address.
- From address: Converts an Ethereum address to a direct IBAN instance.
- from Bban: Converts BBAN address to a direct IBAN instance.

luu-alex:
closes: #4687 #4688 #4690
## Description:
- toAddress: Converts a direct IBAN address into an Ethereum address.
- toIban: Converts an Ethereum address to a direct IBAN address.
- From address: Converts an Ethereum address to a direct IBAN instance.
- from Bban: Converts BBAN address to a direct IBAN instance.
- createIndirect: Creates an indirect IBAN address from an institution and identifier.
- isValid: Checks if an IBAN address is valid.
- isDirect: Checks if the IBAN instance is direct.
- isInDirect: Checks if the IBAN instance is indirect.

Helper functions:
- iso13616Prepare: Prepare an IBAN for mod 97 computation
- mod9710: Calculates the MOD 97 10 of the passed IBAN as specified in ISO7064

##Acceptance Criteria 
- All web3-eth-iban `1.x` functions from the issue mentioned above are migrated and unit tested


nazarhussain (approved):

packages/web3-eth-iban/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * Prepare an IBAN for mod 97 computation by moving the first 4 chars to the end and transforming the letters to
+ * numbers (A = 10, B = 11, ..., Z = 35), as specified in ISO13616.
+ */
+const _iso13616Prepare = (iban: string): string => {
+	const A = 'A'.charCodeAt(0);
+	const Z = 'Z'.charCodeAt(0);
+
+	const upperIban = iban.toUpperCase();
+	const modifiedIban = `${upperIban.substr(4)}${upperIban.substr(0, 4)}`;
```
jdevcs, line 14: I think more good to use  `slice()` instead of `substr()` , as [substr](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is not recommended

jdevcs: I think more good to use  `slice()` instead of `substr()` , as [substr](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is not recommended

packages/web3-eth-iban/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+		return this._iban.length === 34 || this._iban.length === 35;
+	}
+
+	/**
+	 * Get the clients direct address from iban
+	 */
+	public toAddress = (): string => {
+		if (this.isDirect()) {
+			// check if Iban can be converted to an address
+			const base36 = this._iban.substr(4);
```
jdevcs: I think more good to use  `slice()` instead of `substr()`

jdevcs (changes requested):

packages/web3-eth-iban/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+
+const _isValid = (iban: string): boolean =>
+	/^XE[0-9]{2}(ETH[0-9A-Z]{13}|[0-9A-Z]{30,31})$/.test(iban) &&
+	_mod9710(_iso13616Prepare(iban)) === 1;
+
+export class Iban {
+	private readonly _iban: string;
+
+	public constructor(iban: string) {
+		if (_isValid(iban)) {
```
jdevcs: shouldnt it be like 
`if (!_isValid(iban)) {` 
in that case throw error

packages/web3-eth-iban/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+			return toChecksumAddress(paddedBigInt);
+		}
+		throw new IbanLengthError();
+	};
+	/**
+	 * This method should be used to create an ethereum address from a direct iban address
+	 */
+	public static toAddress = (iban: string): string => {
+		const ibanObject = new Iban(iban);
+		if (!ibanObject.isDirect()) {
```
jdevcs: This `.isDirect()` is already done inside non static toAddress() , so i think its duplicate and we can remove it here.

packages/web3-eth-iban/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+		const remainder = _mod9710(_iso13616Prepare(`${countryCode}00${bban}`));
+		const checkDigit = `0${(98 - remainder).toString()}`.slice(-2);
+
+		return new Iban(`${countryCode}${checkDigit}${bban}`);
+	}
+
+	/**
+	 * This method should be used to create iban object from an ethereum address
+	 */
+	public static fromAddress(address: string): Iban {
```
jdevcs: I think more good to use [Address](https://github.com/ChainSafe/web3.js/blob/5a00ab5b4098654280811bf098030e7db485441c/packages/web3-utils/src/types.ts#L22) type instead of string

jdevcs: same applies to other `address` types in this PR.

packages/web3-eth-iban/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+		if (_isValid(iban)) {
+			// error
+		}
+		this._iban = iban;
+	}
+
+	/**
+	 * check if iban number is direct
+	 */
+	private isDirect() {
```
jdevcs: as per 1.x:
```suggestion
	public isDirect() {
```

packages/web3-eth-iban/src/index.ts
luu-alex: yes, i forgot i added this. There is actually no error check though and this would be considered a breaking change if we validated. so i'll probably take this out

packages/web3-eth-iban/src/index.ts
```diff
@@ -0,0 +1,157 @@
...
+		block = remainder.slice(0, 9);
+		remainder = `${(parseInt(block, 10) % 97).toString()}${remainder.slice(block.length)}`;
+	}
+
+	return parseInt(remainder, 10) % 97;
+};
+
+const _isValid = (iban: string): boolean =>
+	/^XE[0-9]{2}(ETH[0-9A-Z]{13}|[0-9A-Z]{30,31})$/.test(iban) &&
+	_mod9710(_iso13616Prepare(iban)) === 1;
```
Spacesai1or: Regarding this method and the above un-exported `const`s, should they be private/static methods on the `Iban` class? They're only used in this file for the `Iban` class, so I think it makes sense to have them be at least private methods since that's how they're being used

jdevcs (approved):

Issue #4573, opened 2021-11-23
eip-155 chainId in Transaction


jdevcs:
`chainId` param is missing in [Transaction](https://github.com/ChainSafe/web3.js/blob/16abe310f2af3b98750bdd7e1ec7c792f49c9e61/packages/web3-common/src/eth_execution_api.ts#L35)   EIP155

mconnelly8:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/ChainSafe/web3.js/4573?planning-poker) with ZenHub @jdevcs @luu-alex @nazarhussain @spacesailor24

PR #4749, opened 2022-01-31
adding chainid to transaction

luu-alex:
closes: #4573 eip-155 adding chainid into transaction type


packages/web3-common/src/eth_execution_api.ts
```diff
@@ -41,6 +41,7 @@ export interface BaseTransaction {
 	readonly gas: Uint;
 	readonly value: Uint;
 	readonly input: HexStringBytes;
+	readonly chainId: Uint;
```
Spacesai1or: `chainId` isn't apart of the ETH execution spec for transactions sent via RPC, so I don't think it makes sense to add it here

romanzac: Perhaps you would find interesting how ethers.js team dealt with this problem: 

https://github.com/ethers-io/ethers.js/blob/73a46efea32c3f9a4833ed77896a216e3d3752a0/packages/transactions/src.ts/index.ts

Spacesai1or: I think this is correct, I keep forgetting [this](https://eth.wiki/json-rpc/API) spec is outdated/not all inclusive

nazarhussain: @spacesailor24 From the conversation with the EF team we had on Discord, the impression was that this Wiki is the most accurate specs till now as every client is following it. 

I remember you were doing further research on this topic to conclude it, can you share some reference to what you found? 

@luu-alex Even if we include it into the execution spec, I don't think it will be a required attribute.

nazarhussain (changes requested):

luu-alex:
@nazarhussain on the ethereum [website](https://ethereum.org/en/developers/docs/apis/json-rpc/) i believe they link the most up to date rpc which directs us to this [site](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/eth1.0-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=true&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false). For the rpc method, eth_signtransaction chainid is required

nazarhussain:
@luu-alex Yes all those specs are derived from https://github.com/ethereum/execution-apis. We earlier had discussion on Discord about that repo and concluded specs on the old wiki is up-to-date as most clients are following it for implementation. It's a bit confusing which one to follow. If we include it, don't make it required. 

Let's see what were the findings of @spacesailor24

packages/web3-common/src/eth_execution_api.ts
Spacesai1or: So [our docs](https://web3js.readthedocs.io/en/v1.5.2/web3-eth.html#sendtransaction) say (for `web3.eth.sendTransaction`) that `chainId` is optional from the user's perspective. It does [get defaulted](https://github.com/ChainSafe/web3.js/blob/8783f4d64e424456bdc53b34ef1142d0a7cee4d7/packages/web3-eth-accounts/src/index.js#L293) when the user uses `web3.accounts.signTransaction` though

Judging mainly by how 1.x was written, I think we add `chainId` to our spec as optional, as @nazarhussain has suggested. One thing to keep in mind is a _unsigned_ transaction sent to a provider doesn't _have_ to have all properties defined, especially when the signing of that transaction is going to be handled outside of Web3 (e.g. the provider receives the non-exhaustive transaction, fills in missing properties, then signs (think MetaMask))

p.s. It does [look like](https://github.com/ChainSafe/web3.js/blob/ed825d2b260a249b01c3a55c822a9ca2588e961f/packages/web3-eth/src/index.js#L159) a `defaultChain` can be set for the rpc methods, but doesn't seem to have a default value. The [inputTransactionFormatter](https://github.com/ChainSafe/web3.js/blob/ed825d2b260a249b01c3a55c822a9ca2588e961f/packages/web3-core-helpers/src/formatters.js#L202) (which `web3.eth.sendTransaction` uses) also doesn't handle `chainId`, so it looks like a transaction object can be sent to the provider without a `chainId` specified - I think we should confirm this though in 1.x as the final deciding factor for this

Old spec says it's not required, new spec does, but we should be concerned about how 1.x handles it if none of Web3's users seem to have an issue with how it's handling things

luu-alex:
I have made chainId optional and added the testcase

nazarhussain (approved):

jdevcs (approved):

PR #4755, opened 2022-02-03
`web3-validator` TODOs

### LINKED ISSUE #4667: `web3-validator` TODOs
- [x] [Remove duplicate `web3-validator` code](https://github.com/ChainSafe/web3.js/tree/e21d55bd6cdc8c181289e8faaf9afa691508d2e1/packages/web3-validator/src/validation)
- [x] [Validate nested arrays](https://github.com/ChainSafe/web3.js/blob/e21d55bd6cdc8c181289e8faaf9afa691508d2e1/packages/web3-validator/src/utils.ts#L19)
- [x] [Investiate strippedType error](https://github.com/ChainSafe/web3.js/pull/4625/files?authenticity_token=Jyj1HxIDxE56HhVoTr8sPe_cV9Yu9mtl6oOI1tHXqSDztypQkrLr4Ceiot-NPh1RSlyOmBZjiYuhbahkaEVXIA&file-filters%5B%5D=.lock&file-filters%5B%5D=.ts#diff-e86b1e01f7a57301af06c74bc73538ab099d9ca77afb1ae178aa0faf0acf5c9aR30)

nazarhussain:
## Description

Fix the remaining todo items for `web3-validator` 

Fixes #4667

## Type of change

## Checklist:


luu-alex (approved):

packages/web3-utils/src/converters.ts
```diff
@@ -100,14 +100,9 @@ export const hexToBytes = (bytes: HexString): Buffer => bytesToBuffer(bytes);
 export const hexToNumber = (value: HexString): bigint | number => {
 	validator.validate(['hex'], [value]);
```
Spacesai1or, line 101: Should this be moved to `web3-validator`'s implementation?

packages/web3-utils/src/converters.ts
```diff
@@ -1,4 +1,4 @@
-import { validator, isAddress, isHexStrict } from 'web3-validator';
+import { validator, isAddress, isHexStrict, utils as validatorUtils } from 'web3-validator';
```
Spacesai1or, line 1: Any benefit to doing `utils as validatorUtils` rather than just importing the specific methods used? If I understand correctly, without tree shaking, there is a lot of unused code being imported

nazarhussain: As we have our own `web3-utils` package, so don't want to mix up other utils with it. This way where ever we import utils form `web3-validator` package it will have different unique name.

packages/web3-utils/src/converters.ts
nazarhussain: Yes we should. I will fix it.

nazarhussain: @spacesailor24 Just verified `hexToNumber` is already using `web3-validator` implementation for `hexToNumber`. For the additional input validation, we can't use it in the validator as it will create circular dependency.

Spacesai1or: I was thinking of moving 
`validator.validate(['hex'], [value]);` to `web3-validator`'s implementation of `hexToNumber` so that this wrapper method is as simple as possible

Spacesai1or (approved):

PR #4759, opened 2022-02-07
CCIP-Read + Wildcard resolution

LeonmanRolls:
## Description

The intention of this PR is to add [CCIP-read](https://eips.ethereum.org/EIPS/eip-3668) support to web3.js. 

CCIP-read TLDR: A contract that needs off-chain data can revert with a special error, which will point web3js to a web server where the info can be gathered. The error will also specify which contract function to subsequently call with this off-chain data.

Changes have been made in three main places:

web3-core-method: In order to hook into transaction errors and check for CCIP-read errors. 
web3-ccip-read: Functions that implement the CCIP-read protocol 
web3-http: Module for making simple http requests

This is a draft to check I'm going in the right direction. 

## Type of change

## Checklist:


packages/web3-ccip-read/README.md
```diff
@@ -0,0 +1,46 @@
+# web3-core-method
```
nazarhussain:
```suggestion
# web3-ccip-read
```

packages/web3-ccip-read/README.md
```diff
@@ -0,0 +1,46 @@
...
+All the TypeScript typings are placed in the `types` folder.
+
+[docs]: http://web3js.readthedocs.io/en/1.0/
+[repo]: https://github.com/ethereum/web3.js
+[npm-image]: https://img.shields.io/npm/v/web3-core-method.svg
+[npm-url]: https://npmjs.org/package/web3-core-method
+[deps-image]: https://david-dm.org/ethereum/web3.js/1.x/status.svg?path=packages/web3-core-method
+[deps-url]: https://david-dm.org/ethereum/web3.js/1.x?path=packages/web3-core-method
+[deps-dev-image]: https://david-dm.org/ethereum/web3.js/1.x/dev-status.svg?path=packages/web3-core-method
+[deps-dev-url]: https://david-dm.org/ethereum/web3.js/1.x?type=dev&path=packages/web3-core-method
```
nazarhussain: Please change all references to `web3-ccip-read`

packages/web3-http/README.md
```diff
@@ -0,0 +1,46 @@
+# web3-core-method
+
+[![NPM Package][npm-image]][npm-url] [![Dependency Status][deps-image]][deps-url] [![Dev Dependency Status][deps-dev-image]][deps-dev-url]
```
nazarhussain: Similar feedback as of other package name.

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+const MAX_REDIRECT_COUNT = 3;
+const OFFCHAIN_LOOKUP_PARAMETER_TYPES = ['address', 'string[]', 'bytes', 'bytes4', 'bytes'];
+
+const CCIP_READ_INTERFACE = new Interface([
+    'function callback(bytes memory result, bytes memory extraData)',
+]);
+
+
+var _callGateway = async function (urls, to, callData) {
+    var httpObject = new http.Http();
```
nazarhussain: The specification suggest: 

> The URLs returned by a contract may be of any schema, but this specification only defines how clients should handle HTTPS URLs.

So there should be a pluggable approach to handle further protocols for gateway.

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+const ENCODED_CCIP_READ_FUNCTION_SELECTOR = '556f1830';
+const MAX_REDIRECT_COUNT = 3;
+const OFFCHAIN_LOOKUP_PARAMETER_TYPES = ['address', 'string[]', 'bytes', 'bytes4', 'bytes'];
+
+const CCIP_READ_INTERFACE = new Interface([
+    'function callback(bytes memory result, bytes memory extraData)',
+]);
+
+
+var _callGateway = async function (urls, to, callData) {
```
nazarhussain: Any function which is not exported is private, so don't need to use `_` prefix for such functions.

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+    var httpObject = new http.Http();
+
+    for (const url of urls) {
+        const queryUrl = `${url}${to.toLowerCase()}/${callData.toLowerCase()}.json`;
```
nazarhussain: As per the spec. 

> Given a URL template returned in an OffchainLookup, the URL to query is composed by replacing sender with the lowercase 0x-prefixed hexadecimal formatted sender parameter, and replacing data with the the 0x-prefixed hexadecimal formatted callData parameter.

So here we have to first check if the URL is contains template literals and replace those. Concatenation of such values will not always return the correct query URL.

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+        let response;
+
+        try {
+            response = await httpObject.get(queryUrl);
+            if (response.status >= 200 && response.status <= 299) {
+                return response;
+            }
+        } catch (errorResponse) {
+            if (errorResponse.status >= 400 && errorResponse.status <= 499) {
```
nazarhussain: As per the specs. 

> Clients MUST handle HTTP 4xx and 5xx error responses that have a content type other than application/json appropriately; they MUST NOT attempt to parse the response body as JSON.

So we need to add handling of 5xx error codes to continue with other URL.  Not for every other code.

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+                throw new Error('There was a problem fetching data from the gateway');
+            }
+        }
+
+        console.warn(`Gateway ${url} failed`);
```
nazarhussain: If it's not 5xx code, we should not continue to other URL instead throw error. 

> If the response code from step (5) is in the range 500-599, go back to step (5) and pick a different URL, or stop if there are no further URLs to try.

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+ * @date 2022
+ */
+
+var abi = require('web3-eth-abi');
+var http = require('web3-http');
+var {defaultAbiCoder, Interface} = require('@ethersproject/abi');
+var {hexConcat} = require('@ethersproject/bytes');
+
+const ENCODED_CCIP_READ_FUNCTION_SELECTOR = '556f1830';
+const MAX_REDIRECT_COUNT = 3;
```
nazarhussain: As per spec. 

> Clients MUST implement a limit on the number of lookups they permit for a single contract call, and this limit SHOULD be at least 4.

So we need to increase that limit to 4.

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+ * @author Leon Talbert <leon@ens.domains>
+ * @date 2022
+ */
+
+var abi = require('web3-eth-abi');
+var http = require('web3-http');
+var {defaultAbiCoder, Interface} = require('@ethersproject/abi');
+var {hexConcat} = require('@ethersproject/bytes');
+
+const ENCODED_CCIP_READ_FUNCTION_SELECTOR = '556f1830';
```
nazarhussain: I would prefer to not hardcode this value instead encode using the default encoder. Or at least add a comment explaining how this value is calculated.

nazarhussain (changes requested):

LeonmanRolls:
@nazarhussain Thanks very much for your feedback, will tighten up the code, address your comments and move this out of WIP/Draft shortly.

packages/web3-http/src/index.js
```diff
@@ -0,0 +1,157 @@
...
+
+Http.prototype.post = function(queryUrl, payload= {}) {
+    return new Promise((resolve, reject) => {
+        var _this = this;
+        var request = this._prepareRequest(queryUrl);
+        request.open('POST', queryUrl, true);
+        request.setRequestHeader('Content-Type','application/json');
+        request.timeout = this.timeout;
+
+        request.timeout = this.timeout;
```
luu-alex: remove this line, same as line 115

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,210 @@
...
+
+var gatewayQuery = function (url, to, calldata) {
+    var httpObject = new http.Http();
+
+    const lowerTo = to.toLowerCase();
+    const lowerCalldata = calldata.toLowerCase();
+
+    const senderUrl = url.replace('{sender}', lowerTo);
+
+    if(!url) throw new Error('No gateway url was provided');
```
Arachnid: Shouldn't this come before line 42?

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,210 @@
...
+
+    if(url.includes('{data}')) {
+        return httpObject.get(`${senderUrl.replace('{data}', lowerCalldata)}.json`);
+    }
+
+    return httpObject.post(senderUrl, {sender: lowerTo, data: lowerCalldata});
+};
+
+var parseGatewayError = function (errorResponse) {
```
Arachnid: 'parse' seems like the wrong terminology here.

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,210 @@
...
+            //5xx errors
+            console.warn(parsedError);
+        }
+    }
+
+    throw new Error('All gateways failed');
+};
+
+var hasCcipReadFunctionSelector = function (encodedString) {
+    return encodedString && encodedString.substring(0, 10) === `0x${ENCODED_CCIP_READ_FUNCTION_SELECTOR}`;
```
Arachnid: Why not include the 0x in the constant?

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,210 @@
...
+ * @author Leon Talbert <leon@ens.domains>
+ * @date 2022
+ */
+
+var abi = require('web3-eth-abi');
+var http = require('web3-http');
+var {defaultAbiCoder, Interface} = require('@ethersproject/abi');
+var {hexConcat} = require('@ethersproject/bytes');
+
+const ENCODED_CCIP_READ_FUNCTION_SELECTOR = '556f1830';
```
Arachnid: This is an 'error selector' rather than a 'function selector', really.

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,210 @@
...
+ */
+var isOffChainLookup = function (err, result) {
+    const normalizedResponse = normalizeResponse(err, result);
+    return !!normalizedResponse.data;
+};
+
+/**
+ * Should be used to encode list of params
+ *
+ * @method encodeParameters
```
Arachnid: Wrong method name and parameters.

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,210 @@
...
+ *
+ * @return {String} encoded list of params
+ */
+var ccipReadCall = async function (errorObject, result, payload, send, options) {
+    if (send.ccipReadCalls) {
+        send.ccipReadCalls++;
+    } else {
+        send.ccipReadCalls = 1;
+    }
+    if (send.ccipReadCalls > MAX_REDIRECT_COUNT) {
```
Arachnid: Can this be something users can provide at runtime?

packages/web3-ccip-read/src/index.js
```diff
@@ -0,0 +1,210 @@
...
+    let gatewayResult;
+    if(options.ccipReadGatewayCallback) {
+        try{
+            gatewayResult = await options.ccipReadGatewayCallback(finalUrls, sender, callData);
+        } catch(e) {
+            console.error('ccipReadGatewayCallback error.');
+            throw e;
+        }
+    } else {
+        const result = await callGateway(finalUrls, sender, callData);
```
Arachnid: You should probably export `callGateway` so that implementers of `ccipReadGatewayCallback` can use it.

packages/web3-ccip-read/README.md
```diff
@@ -0,0 +1,46 @@
...
+### Node.js
+
+```bash
+npm install web3-core-method
+```
+
+## Usage
+
+```js
+const Web3Method = require('web3-core-method');
```
jdevcs: could you update usage example for ccip package here.

packages/web3-ccip-read/package.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "web3-ccip-read",
+    "version": "1.0.0",
```
jdevcs:
```suggestion
    "version": "1.0.0.alpha.0",
```

packages/web3-core-method/package.json
```diff
@@ -18,7 +18,8 @@
         "web3-core-helpers": "1.7.0",
         "web3-core-promievent": "1.7.0",
         "web3-core-subscriptions": "1.7.0",
-        "web3-utils": "1.7.0"
+        "web3-utils": "1.7.0",
+        "web3-ccip-read": "1.0.0"
```
jdevcs:
```suggestion
        "web3-ccip-read": "1.0.0.alpha.0"
```

packages/web3-http/package.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "web3-http",
```
jdevcs: I think you should use existing http provider package in web3.js and modify/add new functionality in that instead of creating new package here.

jdevcs:
Hi @LeonmanRolls  Thanks for your contributions in web3.js , could you remove package lock files from this PR:

- package-lock.json
- packages/web3-ccip-read/package-lock.json
- packages/web3-http/package-lock.json

Also currently following are not passing:
```
Build / lint (pull_request)
Build / e2e (unit_and_e2e_clients) (pull_request)
Build / e2e (e2e_browsers) (pull_request) 
Build / e2e (e2e_ganache) (pull_request)
Build / e2e_windows (pull_request) 
```

could you check and fix failing tests on your local system.

jdevcs (changes requested):
Hi @LeonmanRolls  Thanks for your contributions in web3.js , could you remove package lock files from this PR:

- package-lock.json
- packages/web3-ccip-read/package-lock.json
- packages/web3-http/package-lock.json

Also currently following are not passing:
```
Build / lint (pull_request)
Build / e2e (unit_and_e2e_clients) (pull_request)
Build / e2e (e2e_browsers) (pull_request) 
Build / e2e (e2e_ganache) (pull_request)
Build / e2e_windows (pull_request) 
```

could you check and fix failing tests on your local system.

packages/web3-http/src/index.js
LeonmanRolls: done thanks!

packages/web3-http/package.json
LeonmanRolls: Hey are you suggesting modifying/adding functionality to the provider itself or adding a separate class to the provider package? Also I think that a provider as an abstraction is quite different to general http requests, and most of the other packages seem to have a single purpose. Happy to add the functionality to the provider package if you still think it should go there.

jdevcs:
Hi, @LeonmanRolls , will you have time for discussion or addressing feedback of this PR? Thanks

LeonmanRolls:
> Hi, @LeonmanRolls , will you have time for discussion or addressing feedback of this PR? Thanks

Hi yes sure I will address your feedback sorry for the delay, i did also respond to one of your comments but I think you missed it so will post here:

I think you should use existing http provider package in web3.js and modify/add new functionality in that instead of creating new package here.

Hey are you suggesting modifying/adding functionality to the provider itself or adding a separate class to the provider package? Also I think that a provider as an abstraction is quite different to general http requests, and most of the other packages seem to have a single purpose. Happy to add the functionality to the provider package if you still think it should go there.

packages/web3-ccip-read/src/index.js
LeonmanRolls: Thanks you are right, this has been implemented now

packages/web3-ccip-read/src/index.js
LeonmanRolls: If I understand you correctly this is what is happening now, when the error is 5xx the loop continues to the next iteration.

packages/web3-ccip-read/src/index.js
LeonmanRolls: If the error is 4xx an error is thrown

nazarhussain (changes requested):
I would second @jdevcs feedback. We should not introduce new package `web3-http` instead use existing `web3-providers-http` package. You can create a private instance of the provider and use `.send` method to achieve the same.

LeonmanRolls:
> I would second @jdevcs feedback. We should not introduce new package `web3-http` instead use existing `web3-providers-http` package. You can create a private instance of the provider and use `.send` method to achieve the same.

Happy to do it, I have a couple of points I need guidance on:

- The current provider doesn’t support GET requests, should I modify _prepareRequest to take a method option?
- The provider generally hides details about the request itself, which are needed by ccip-read, e.g. response status code and content-type header. I think I need to add another method to the provider that would expose these details.
- One possibility is moving the current web3-http methods over to the provider class. 

Let me know what you think!

LeonmanRolls:
> Hi @LeonmanRolls Thanks for your contributions in web3.js , could you remove package lock files from this PR:
> 
> * package-lock.json
> * packages/web3-ccip-read/package-lock.json
> * packages/web3-http/package-lock.json
> 
> Also currently following are not passing:
> 
> ```
> Build / lint (pull_request)
> Build / e2e (unit_and_e2e_clients) (pull_request)
> Build / e2e (e2e_browsers) (pull_request) 
> Build / e2e (e2e_ganache) (pull_request)
> Build / e2e_windows (pull_request) 
> ```
> 
> could you check and fix failing tests on your local system.

Hey @jdevcs , I'm presuming you wanted me to resolve the merge conflicts instead of removing the package-lock files so have done that (as they are there on the current 1.x branch). 

Running linting and tests on my machine is a little tricky as many are failing on a fresh pull of the 1.x branch too. I do however now have the same number of failing tests on my branch locally as I do for 1.x, with all the new ones that Iv'e added passing. 

Linting is a bit tricky too because it includes the built files (the ones inside the lib folders). I get 20k+ linting errors on a fresh pull of 1.x locally. 

Please advise on how to proceed, may also be easier to talk on discord.

LeonmanRolls:
Hi @jdevcs @nazarhussain 

I have moved the network request code to be part of the provider now. 

I have also added in wildcard resolution as described here: https://docs.ens.domains/ens-improvement-proposals/ensip-10-wildcard-resolution. They work together to facilitate fetching off-chain ENS data so thought it would make sense to merge them as one PR. 

I've gone through the checklist as best I can. A fresh pull of 1.x branch has 19 failing tests for me, so this branch has the same number of failing tests, none of the tests i've added have introduced new failing tests locally. I haven't checked the `test:unit` and `test:cov` boxes for this reason.

Woodpile37 (approved):

Arachnid:
@jdevcs @nazarhussain would it be possible to take another look at this?

makoto:
hi  @nazarhussain Did you have time to review the PR?

GregTheGreek:
~~deleted because my comment didn't make sense~~

GregTheGreek:
@Arachnid @makoto @LeonmanRolls

This PR is looking to add a specific [EIP](https://eips.ethereum.org/EIPS/eip-3668) as a new core package to web3.js. Unless this EIP is being widely used (mass adoption) in the community I do not think this is something we want to be including within web3.js, at least as its own package. If this is something that is crucial for ENS to work, please consider including it directly in the `web3-ens` package so it can be consumed by your users that way.

We do not want to set a precedent that any single EIP can be implemented into web3.js prior to adoption, with caveats for changes such as the wallet provider API.

Arachnid:
I'm sorry to hear that. This change is crucial to ENS, so we're happy to implement it in the ENS package if that's your preference. Because it's a general-purpose standard for fetching offchain data, we figured it would be useful to web3 users to have it available outside ENS; it seems a shame to implement the functionality for ENS only.

Can you clarify what you mean by "prior to adoption"? ethers.js already supports EIP 3668, and it seems like a bit of a catch 22 to require that it be already used before you support it.

LeonmanRolls:
Hi @GregTheGreek , I've moved the logic for CCIP-Read into the web3-core-method package (as a separate file) as that's the only place it's used at the moment, and including it in web3-eth-ens causes circular dependency issues.

Woodpile37 (approved):

ghost:
@LeonmanRolls Could you rebase your branch and make sure it works with the [new fetch API](https://github.com/ChainSafe/web3.js/pull/5085)? We no longer use the buggy `xhr2-cookies` lib to make HTTP requests anymore.

LeonmanRolls:
Will submit a corresponding PR for v4

PR #4743, opened 2022-01-28
Fix jsonrpc payload and response types

MicaiahReid:
## Description
The new Ganache v7 release is not compatible with Web3 as a provider due to some typing issues. We think this is an issue with some of Ganache's and some of Web3's types. This PR fixes the Web3 types.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] Any dependent changes have been merged and published in downstream modules.
- [x] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [x] I ran `npm run test:unit` with success.
- [x] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code.
- [ ] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
- [ ] I have tested my code on the live network.
- [ ] I have checked the Deploy Preview and it looks correct.
- [x] I have updated the `CHANGELOG.md` file in the root folder.

Issue #443, opened 2016-04-28
Invalid contract output parsing


karalabe:
I have the following output ABI definition for a contract call:

`[{"name":"major","type":"uint32"},{"name":"minor","type":"uint32"},{"name":"patch","type":"uint32"},{"name":"commit","type":"bytes20"},{"name":"pass","type":"address[]"},{"name":"fail","type":"address[]"}]`

Calling the specific method returns the following EVM output:

`000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000945e4c7991a3d210c5efc1a2cb8feeabad7cb9ca0000000000000000000000000000000000000000000000000000000000000000`

Web3 (at least the one included in Geth) parses this as:

`[1, 2, 4, "0x0000000000000000000000000000000000000004", ["0x0000000000000000000000000000000000000002"], []]`

Whereas the Go ABI parser parses it (corrently) as:

`{"Major": 1, "Minor": 2, "Patch": 4, "Commit": "0x0000000000000000000000000000000000000004", "Pass": ["0x945e4c7991a3d210c5efc1a2cb8feeabad7cb9ca"], "Fail": []}`

karalabe:
Btw, you can also test this yourself on the **testnet** with [this abi](https://gist.github.com/karalabe/e78e249b57f37716e774af87e7a4bf85) against the contract deployed at [`0x48a117313039b73ab02fb9f73e04a66defe123ec`](http://testnet.etherscan.io/address/0x48a117313039b73ab02fb9f73e04a66defe123ec), by calling the `proposedVersion()` method.

frozeman:
We currently not returning the named return values, but as an array. Though the address doesn't seem to be parsed correctly.

This is something we can look into once we start refactoring the decoding/encoding part

axic:
Not meaning to hijack this, but I was wondering if ethereumjs-abi (which is a from scratch implementation based on the wiki) parses it properly (= the way Go does), and it does:
```
[ <BN: 1>,
  <BN: 2>,
  <BN: 4>,
  <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 04>,
[...]

PR #4761, opened 2022-02-08
Fix jsonrpc payload and response types (#4743)

jdevcs:
e2e_windows were not passing in [this](https://github.com/ChainSafe/web3.js/pull/4743) PR so opened current PR for checks  / fixes .
______________________________________
4743:
* Make JsonRpcPayload's `params` field optional

Currently jsonrpc.js uses `params: params || []` in the
`toPayload` function, so this type update makes the `params` field
optional to match.

* Fix JsonRpcResponse type

Update `id` to accept `string | number` - this now matches the
`isValidResponse` function in `jsonrpc.js`.

Update `error` to accept an object with optional `code`, `data`,
and non-optional `message` fields to more closely match the
[JSON RPC spec](https://www.jsonrpc.org/specification#error_object)
and the `ErrorResponse` function in `errors.js`.

* Remove errant spaces

* Add PR #443 to CHANGELOG

Co-authored-by: jdevcs <86780488+jdevcs@users.noreply.github.com>

## Description

## Type of change

## Checklist:


luu-alex (approved):

nazarhussain (approved):

Spacesai1or (changes requested):
We've refrained from making any changes to 1.x's types because it's considered a breaking change for those using typescript

davidmurdoch:
@spacesailor24 The JSON RPC error property is 100% wrong in every case as is, so I think that should be fixed as no one can rely on that as it is, and if they somehow are, their runtime code is likely broken.

This is _technically_ not a breaking change in Web3's types, as it aligns it with runtime behavior, but for Ganache to have to make the change on our side in v7 it certainly would be breaking (and wrong).

Please reconsider merging this into 1.x.

Spacesai1or:
> @spacesailor24 The JSON RPC error property is 100% wrong in every case as is, so I think that should be fixed as no one can rely on that as it is, and if they somehow are, their runtime code is likely broken.
> 
> This is _technically_ not a breaking change in Web3's types, as it aligns it with runtime behavior, but for Ganache to have to make the change on our side in v7 it certainly would be breaking (and wrong).
> 
> Please reconsider merging this into 1.x.

@davidmurdoch Okay, yea that totally makes sense. We had a blanket policy of not making 1.x type changes, but I should've spent more time understanding this one - weird that the types were even written this way

Spacesai1or (approved):

Issue #4551, opened 2021-11-17
Migrate ens resolver ABI from 1.x for 4.x rewrite  


luu-alex:
### Description

Migrate the following ABI from web3-eth-ens 1.x branch to 4.x rewrite, as well include the new ens functionality

Resolver Abi:
- supportsInterface
- addr (get ethereum address)
- setAddr
- AddrChanged
- addr (get blockchain address)
- setAddr
- name
- setName
- NameChanged
- contenthash
- setContenthash
- ContenthashChanged
- ABI
- setABI
- pubkey
- setPubkey
- text
- setText
- multicall

### Acceptance Criteria
- All of resolver ABI should be migrated
### Additional Information
[ENS Registry abi reference](https://docs.ens.domains/contract-api-reference/ens)
[ENS Reverse Registrar abi reference](https://docs.ens.domains/contract-api-reference/reverseregistrar)
[ENS TestRegistrar abi reference](https://docs.ens.domains/contract-api-reference/testregistrar)
[1.x web3js contracts](https://github.com/ChainSafe/web3.js/tree/1.x/packages/web3-eth-ens/src/resources/ABI)
[ENS resolver abi reference](https://docs.ens.domains/contract-api-reference/publicresolver)
[ENS Registrar ABI reference](https://docs.ens.domains/contract-api-reference/.eth-permanent-registrar)

PR #4762, opened 2022-02-09
Migrate get registry functions from 1.x to 4.x

### LINKED ISSUE #4603: Migrate get registry functions from 1.x to 4.x
### Description

Migrate the following functions from `web3-eth-ens` contract functions for registry 1.x branch to 4.x rewrite.
- registryAddress
- getOwner
- getTTL
- isApprovedForAll
- recordExists
- getResolver

### Deprecated functions from 1.x
- resolver
- owner

### Acceptance Criteria
- All functions above should be migrated
- All functions should be covered with unit tests

luu-alex:
## Description
closes: #4603 #4551
includes registry functions and a partially made ENS object, created constructor tests for registry.
Includes resolver ABI
### Registry functions 
- getOwner
- getResolver
- getTTL
- isApprovedForAll
- recordExists
- setApprovalForAll
- setOwner
- setResolver
- setSubnodeOwner
- setSubnodeRecord
- setTTL

Criteria 
- finish the registry functions
- add constructor tests
 
## TODO
- Finish the rest of the ENS functions when resolver is complete.
- When eth.net is complete, add the this.contract to the registry object

## Changes
- Promievents were used in 1.x but the events were not being [used](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-eth-ens/src/contracts/Registry.js#L81), making them used just as promises. 
so i changed promievents to promises in 4.x.


luu-alex:
Discussion about this function: https://github.com/ChainSafe/web3.js/pull/4762/files#diff-4356939d9d2e5df02cd6beb2a0bd438d5bb0abd2edaeadbb63359df6c26a24d8R6 we use a library called `idna-uts46-hx` to turn ens names toAscii before checking the registry with the name, an [example](https://github.com/ChainSafe/web3.js/pull/4762/files#diff-e0aa222db9e80d24f9f1edbd279ba95bfc39478347f59b0e1a03c0d94f65e970R17) . I attempted to make a toAscii function but looking from the idna package its a bit difficult. if anyone has thoughts about this, to replace the library or not.
Not too sure if we actually need this library, but will look into it

zhanzekun:
@luu-alex  

I met an error when I was trying to use web3.eth.ens.getAddress(). The error was like 
`
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'decode')
    at mapLabel (uts46.js:18:31)
    at process2 (uts46.js:51:22)
    at Object.toUnicode2 [as toUnicode] (uts46.js:125:12)
    at normalize (index.js:26:23)
    at Object.namehash2 [as hash] (index.js:11:10)
    at ResolverMethodHandler.prepareArguments (ResolverMethodHandler.js:186:25)
    at Object.ResolverMethodHandler.call (ResolverMethodHandler.js:68:41)
    at ENS.getAddress (ENS.js:308:64)
`

and I deep into this error and found that the package 「eth-ens-namehash」use a lib called 「idna-utes46-hx」, and this lib has 3 version to compate different node version

image

lib link: https://www.npmjs.com/package/idna-uts46-hx

and I found this error may due to it. my local node version is node 16.x. 
is this PR may help me solve the problem I'm facing ?

luu-alex:
@zhanzekun  hey, thank you for this comment. No this PR wont address it, but can you please make an issue with the details. We can move forward after the issue has been created :)

packages/web3-eth-ens/src/ens.ts
```diff
@@ -0,0 +1,167 @@
...
+import { NonPayableCallOptions, TransactionReceipt } from 'web3-eth-contract';
+import { Registry } from './registry';
+
+export class ENS {
+	private _registryAddress: string;
+	private readonly registry: Registry;
+
+	public constructor() {
+		this.registry = new Registry();
+		this._registryAddress = ''; // TODO change this when eth.net is finished
```
jdevcs: you should also have a [consts](https://github.com/ChainSafe/web3.js/blob/bd6a9d86bbb9dc6e1fbe74216f9d7d44d7ba2b47/packages/web3-eth-ens/src/config.js#L28) file for registry address and other const values , and you should use that to set default registry addr

jdevcs: and should pass that to `new Registry()` constructor to registry contract instance creation

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,211 @@
...
+					reject(RevertInstructionError);
+				});
+		});
+		return promise;
+	}
+
+	public async setOwner(
+		name: string,
+		address: Address,
+		txConfig: NonPayableCallOptions,
```
jdevcs: `NonPayableCallOptions` refers to EthCall ? 

Please could you mark here :
TODO: once 4.x contracts package is updated to use Eth-package for sending transaction, we need to update `txConfig` type here

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,211 @@
...
+					reject(RevertInstructionError);
+				});
+		});
+		return promise;
+	}
+
+	public async getTTL(name: string) {
+		const promise = new Promise((resolve, reject) => {
+			this.contract.methods
+				.getTTL(namehash(name))
```
jdevcs: I think you should directly call ttl as: https://github.com/ensdomains/ens/blob/ff0f41747c05f1598973b0fe7ad0d9e09565dfcd/contracts/ENS.sol#L29

```suggestion
				.ttl(namehash(name))
```

jdevcs (changes requested):

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,211 @@
...
+				.catch(() => {
+					reject(RevertInstructionError);
+				});
+		});
+		return promise;
+	}
+
+	public async setTTL(
+		name: string,
+		ttl: string,
```
jdevcs: ttl is uint64 in registry so good to use compatible type here instead of string

packages/web3-eth-ens/src/registry.ts
jdevcs: same also for all instances of `txConfig` in this file using `NonPayableCallOptions`

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,211 @@
...
+		});
+		return promise;
+	}
+
+	public async setSubnodeRecord(
+		name: string,
+		label: string,
+		owner: Address,
+		resolver: Address,
+		ttl: string,
```
jdevcs: this `ttl` is also uint64 in contract so it shouldnt be string here, 1.x is `String | Number`

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,211 @@
...
+					reject(RevertInstructionError);
+				});
+		});
+		return promise;
+	}
+
+	public async getResolver(name: string) {
+		const resolverContract = new Promise((resolve, reject) => {
+			this.contract.methods
+				.getResolver(namehash(name))
```
jdevcs: I think according to [registry function](https://github.com/ensdomains/ens/blob/ff0f41747c05f1598973b0fe7ad0d9e09565dfcd/contracts/ENS.sol#L28) it should be 

```suggestion
				.resolver(namehash(name))
```

jdevcs (changes requested):

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,211 @@
...
+				.call()
+				.then(address => {
+					const newContract =
+						typeof address === 'string'
+							? new Contract(resolverABI, address)
+							: new Contract(resolverABI);
```
jdevcs: I think we shouldn't create contract instance in case of no or invalid resolver address, and should throw error.

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	public async setResolver(
+		name: string,
+		address: Address,
+		txConfig: NonPayableCallOptions,
+	): Promise<ReceiptInfo | RevertInstructionError> {
+		const promise = new Promise<ReceiptInfo>((resolve, reject) => {
+			this.contract.methods
+				.getResolver(namehash(name), inputAddressFormatter(address))
```
jdevcs:
```suggestion
				.setResolver(namehash(name), inputAddressFormatter(address))
```

jdevcs (changes requested):

packages/web3-eth-ens/src/utils.ts
```diff
@@ -0,0 +1,28 @@
+import { sha3Raw } from 'web3-utils';
+import { toAscii } from 'idna-uts46-hx';
+
+export const normalize = (name: string) =>
+	name
+		? toAscii(name, { useStd3ASCII: true, transitional: false, verifyDnsLength: false })
+		: name;
+
+export const namehash = (inputName: string) => {
```
jdevcs, line 7: good work for a reduced dependency for this function, 

could you also add unit tests for `namehash` , may be have same testing data as its in eth ens namehash lib

packages/web3-eth-ens/src/ens.ts
```diff
@@ -0,0 +1,167 @@
...
+
+	/**
+	 * Sets the owner, resolver and TTL for a subdomain, creating it if necessary.
+	 */
+	public async setSubnodeRecord(
+		name: string,
+		label: string,
+		owner: Address,
+		resolver: Address,
+		ttl: string,
```
jdevcs: ttl type should be number

packages/web3-eth-ens/src/ens.ts
```diff
@@ -0,0 +1,167 @@
...
+	public async getTTL(name: string): Promise<unknown> {
+		return this.registry.getTTL(name);
+	}
+
+	/**
+	 * Returns the address of the owner of an ENS name.
+	 */
+	public async setTTL(
+		name: string,
+		ttl: string,
```
jdevcs: ttl type should be number

packages/web3-eth-ens/src/ABI/Registry.ts
```diff
@@ -1,5 +1,5 @@
 // https://github.com/ensdomains/ens/blob/master/contracts/ENS.sol
-const REGISTRY = [
+export const REGISTRY = [
```
nazarhussain: BTW this change also done in my PR. so you might have conflicts on these files.

packages/web3-eth-ens/src/ens.ts
```diff
@@ -0,0 +1,167 @@
...
+		this._registryAddress = ''; // TODO change this when eth.net is finished
+	}
+
+	public get registryAddress(): string {
+		return this._registryAddress;
+	}
+
+	public set registryAddress(registryAddress: string) {
+		this._registryAddress = registryAddress;
+	}
```
nazarhussain: If we are not processing anything extra in getters or setters, then we can simply have a public attribute `registryAddress`

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,211 @@
...
+				.owner(namehash(name))
+				.call()
+				.then(owner => {
+					resolve(owner);
+				})
+				.catch(() => {
+					reject(RevertInstructionError);
+				});
+		});
+		return promise;
```
nazarhussain:
```suggestion
           try {
              const result = await this.contract.methods.owner(namehash(name)).call();
              
              return result;
           } catch {
               throw new RevertInstructionError();
           }
```

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,211 @@
...
+		address: Address,
+		txConfig: NonPayableCallOptions,
+	): Promise<ReceiptInfo | RevertInstructionError> {
+		const promise = new Promise<ReceiptInfo>((resolve, reject) => {
+			this.contract.methods
+				.setOwner(namehash(name), inputAddressFormatter(address))
+				.send(txConfig)
+				.then(receipt => {
+					resolve(receipt);
+				})
```
nazarhussain: Use `try/catch` within `async` function as I mentioned in above example.

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,211 @@
...
+				.catch(() => {
+					reject(RevertInstructionError);
```
nazarhussain: Always use error instances instead of constructors. `new RevertInstructionError()`

nazarhussain (changes requested):

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,183 @@
...
+		}
+	}
+
+	public async setOwner(
+		name: string,
+		address: Address,
+		txConfig: NonPayableCallOptions, // TODO: web3-eth txconfig should be replaced with sendTransaction type
+	) {
+		try {
+			const receipt = await this.contract.methods
```
jdevcs: this should return [`PromiEvent<TransactionReceipt | TransactionRevertInstructionError>`](https://web3js.readthedocs.io/en/v1.7.0/callbacks-promises-events.html#promievent) and same for all set functions , as in [1.x](https://web3js.readthedocs.io/en/v1.7.0/web3-eth-contract.html#id36) , 

I think we should not await here and directly return promise, and same for all set functions ,

packages/web3-eth-ens/src/ens.ts
```diff
@@ -0,0 +1,160 @@
...
+
+	public constructor(registryAddr?: string) {
+		this.registry = new Registry(registryAddr);
+		this._registryAddress = registryAddr ?? registryAddress; // TODO change this when eth.net is finished
+	}
+
+	/**
+	 * Returns the Resolver by the given address
+	 */
+	public async getResolver(name: string): Promise<unknown> {
```
nazarhussain: Why it's `unknown`, don't we know the data structure for the resolver?

packages/web3-eth-ens/src/ens.ts
```diff
@@ -0,0 +1,160 @@
...
+	public async setSubnodeRecord(
+		name: string,
+		label: string,
+		owner: Address,
+		resolver: Address,
+		ttl: number,
+		txConfig: NonPayableCallOptions,
+	): Promise<TransactionReceipt | RevertInstructionError> {
+		return this.registry.setSubnodeRecord(name, label, owner, resolver, ttl, txConfig);
+	}
```
nazarhussain, line 47: Anything that we are setting, will not return `PromiEvent` instead of `Promise`?

packages/web3-eth-ens/src/config.ts
```diff
@@ -0,0 +1 @@
+export const registryAddress = '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e';
```
nazarhussain:
```suggestion
export const registryContractAddress = '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e';
```

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,183 @@
...
+import { namehash } from './utils';
+
+export class Registry {
+	private readonly contract: Contract<typeof REGISTRY>;
+
+	public constructor(customRegistryAddress?: Address) {
+		// TODO for contract, when eth.net is finished we can check network
+		this.contract = customRegistryAddress
+			? new Contract(REGISTRY, customRegistryAddress)
+			: new Contract(REGISTRY, registryAddress);
```
nazarhussain:
```suggestion
		this.contract = new Contract(REGISTRY, customRegistryAddress ?? registryAddress);
```

nazarhussain (changes requested):

nazarhussain (approved):

packages/web3-eth-ens/src/registry.ts
```diff
@@ -0,0 +1,182 @@
...
+		}
+	}
+
+	public async setResolver(
+		name: string,
+		address: Address,
+		txConfig: NonPayableCallOptions, // TODO: web3-eth txconfig should be replaced with sendTransaction type
+	): Promise<ReceiptInfo> {
+		try {
+			const promise = await this.contract.methods
```
jdevcs:
```suggestion
			return this.contract.methods
```

Good Work Alex, 

Last item: in all set functions, I think we shouldn't wait for promise to resolve in web3.js. and we should return promise and not result, and let user wait for that promise to resolve in his app/lib?

(in that way user will be able to leverage async performance)

jdevcs: a part from that all other things lgtm

luu-alex: Sounds good, I believe same with the other functions. they should return promises so i took away the awaits

jdevcs (approved):

PR #4768, opened 2022-02-13
`Web3Eth` extends `Web3Context` instead of instantiating

Spacesai1or:
~This branch is based off of `wyatt/4.x/4652-eth-tx-refactor`, so needs to be merged/reviewed afterwards~

This closes the `TODO` item from #4608:

> Inherit `Web3Context` instead of [instantiating it](https://github.com/ChainSafe/web3.js/blob/0ff0e95ed1835f7e75b8866dfefeefc825a83c5a/packages/web3-eth/src/index.ts#L38)


nazarhussain (approved):

templates/tsconfig.json.tmpl
```diff
@@ -1,7 +1,8 @@
 {
 	"extends": "../../tsconfig",
 	"compilerOptions": {
-		"outDir": "dist"
+		"outDir": "dist",
+		"esModuleInterop": true
```
jdevcs: I think we should not set `esModuleInterop` true for templates

jdevcs (approved):
lgtm,  suggested change in template tsconfig

templates/tsconfig.json.tmpl
Spacesai1or: I guess these changes overwrote PR #4718. Will remove this again

PR #4769, opened 2022-02-13
#4608 `Web3Eth` missing RPC methods

Spacesai1or:
### What This PR Does

- Adds missing RPC methods from #4608
    - `getPendingTransactions`
    - `requestAccounts`
        - Returned addresses need to be formatted to checksum, but none of the other calls are doing this for in/output, so it will be as apart of an update to all methods
    - `getChainId`
    - `getNodeInfo`
    - `getProof`
         - `address` and `blockNumber` need to be formatted to checksum and hex string (respectively), but none of the other calls are doing this for in/output, so it will be as apart of an update to all methods
- Moves `formatTransaction` from `packages/web3-eth/src/eth_tx.ts` to `packages/web3-eth/src/format_transaction.ts`
    - This was done to avoid a circular dependency between `eth_tx.ts` and `packages/web3-eth/src/rpc_method_wrappers.ts`
- Adds [Web3EthExecutionAPI](https://github.com/ChainSafe/web3.js/blob/5b8f781516c5e4d9ea53614128f4d4fe141c55eb/packages/web3-eth/src/web3_eth_execution_api.ts)

### Notes

- Research needs to be done to determine if [input](https://github.com/ChainSafe/web3.js/blob/5b8f781516c5e4d9ea53614128f4d4fe141c55eb/packages/web3-common/src/eth_execution_api.ts#L44) is really the correct property. `1.x` accepts `input` as a property, but converts it to `data` [here](https://github.com/ChainSafe/web3.js/blob/bd6a9d86bbb9dc6e1fbe74216f9d7d44d7ba2b47/packages/web3-core-helpers/src/formatters.js#L144), but the [spec](https://github.com/ethereum/execution-apis/blob/68cf607f01bd41aecd0aaafd2bfc01928e188cc2/src/schemas/transaction.json#L162) we're using specidies `input` instead of `data`


packages/web3-eth/src/format_transaction.ts
```diff
@@ -0,0 +1,70 @@
...
+			formattedTransaction.common.customChain.chainId,
+			desiredType,
+		);
+
+	if (formattedTransaction.data !== undefined && formattedTransaction.input !== undefined)
+		throw new TransactionDataAndInputError({
+			data: formattedTransaction.data,
+			input: formattedTransaction.input,
+		});
+	else if (formattedTransaction.input !== undefined) {
```
nazarhussain: Can we not put these properties to `formattableProperties` as well to avod duplicate code.

nazarhussain: I still feel if we can define a valid schema for transaction, then we can write a generic formatter function. May be we think of it some time later.

nazarhussain (changes requested):

packages/web3-eth/src/format_transaction.ts
Spacesai1or: The schema based approach would be ideal, but not high priority right now and should be looked at later

I kept these separate since the properties are nested and I didn't want the for loop to get even more complex by handling nested properties. These if statements can get combined though - I'll make this change

nazarhussain (approved):

luu-alex (approved):

PR #4781, opened 2022-02-16
Web3 Eth Personal

### LINKED ISSUE #4720: Web3 Eth Personal
### Description
**Following function are for interacting with wallet located in Eth node, these are not for local wallet.**
Having following functionality in 4.x:

- newAccount: Creates a new account
- sign : The method used for signing
- ecRecover : Recovers the account that signed the data.
- signTransaction: Signs a transaction. This account needs to be unlocked.
- sendTransaction: This method sends a transaction over the management API.
- unlockAccount: Signs data using a specific account.
- lockAccount: locksthe given account.
- getAccounts: Returns a list of accounts the node controls by using the provider and calling the RPC method personal_listAccounts
- importRawKey: Imports the given private key into the key store, encrypting it with the passphrase.

### Acceptance Criteria
This package must also expose following functionality by using existing packages.

- setProvider: Will change the provider for its module.
- providers: Contains the current available providers.
- givenProvider: When using web3.js in an Ethereum compatible browser, it will set with the current native provider by that browser. Will return the given provider by the (browser) environment, otherwise null.
- currentProvider: Will return the current provider, otherwise null.
- BatchRequest: Class to create and execute batch requests.

nazarhussain:
## Description

Add `web3-eth-personal` package with it's implementation. 

Fixes #4720

## Type of change

## Checklist:


packages/web3-eth-personal/src/index.ts
```diff
@@ -0,0 +1,113 @@
...
+
+		const result = await this.requestManager.send({
+			method: 'personal_sign',
+			params: [dataToSign, address, passphrase],
+		});
+
+		return result;
+	}
+
+	public async ecRecover(signedData: HexString, passphrase: string) {
```
jdevcs:
```suggestion
	public async ecRecover(signedData: HexString, signature: HexString) {
```

personal_ecRecover's second param is signature `{"method": "personal_ecRecover", "params": [message, signature]}` , could you make this change in following code lines as well

packages/web3-eth-personal/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+![ES Version](https://img.shields.io/badge/ES-2020-yellow)
+![Node Version](https://img.shields.io/badge/node-14.x-green)
+[![NPM Package][npm-image]][npm-url]
+[![Dependency Status][deps-image]][deps-url]
+[![Dev Dependency Status][deps-dev-image]][deps-dev-url]
+
+This is a sub-package of [web3.js][repo].
+
+`web3-eth-personal` contains the ideal setup for a Web3.js package.
```
jdevcs:
```suggestion
`web3-eth-personal` Web3 module to interact with the Ethereum blockchain accounts stored in the node.
```

packages/web3-eth-personal/package.json
```diff
@@ -0,0 +1,48 @@
+{
+	"name": "web3-eth-personal",
+	"version": "4.0.0-alpha.1",
```
jdevcs:
```suggestion
	"version": "4.0.0-alpha.0",
```

jdevcs (changes requested):

packages/web3-eth-personal/src/personal_api.ts
```diff
@@ -0,0 +1,14 @@
...
+export type EthPersonalAPI = {
+	personal_listAccounts: () => Address[];
+	personal_newAccount: (password: string) => Address;
+	personal_unlockAccount: (address: Address, password: string, unlockDuration: number) => boolean;
+	personal_lockAccount: (address: Address) => boolean;
+	personal_importRawKey: (keyData: HexString, passphrase: string) => boolean;
+	personal_sendTransaction: (tx: Transaction, passphrase: string) => HexString;
+	personal_signTransaction: (tx: Transaction, passphrase: string) => HexString;
+	personal_sign: (data: HexString, address: Address, passphrase: string) => HexString;
+	personal_ecRecover: (signedData: HexString, passphrase: string) => HexString;
```
jdevcs:
```suggestion
	personal_ecRecover: (signedData: HexString, signature: HexString) => Address;
```

packages/web3-eth-personal/test/unit/eth_personal.test.ts
```diff
@@ -0,0 +1,309 @@
...
+		});
+	});
+
+	describe('ecRecover', () => {
+		it('should call the correct method for request manager', async () => {
+			const data = '0x1234';
+
+			jest.spyOn(utils, 'isHexStrict').mockReturnValue(true);
+
+			await ethPersonal.ecRecover(data, 'password');
```
jdevcs:
```suggestion
			await ethPersonal.ecRecover(data, '0x0000..'); //signature in 2nd param
```

packages/web3-eth-personal/test/unit/eth_personal.test.ts
```diff
@@ -0,0 +1,309 @@
...
+
+			expect(sendSpy).toHaveBeenCalledWith({
+				method: 'personal_ecRecover',
+				params: [data, 'password'],
```
jdevcs:
```suggestion
				params: [data, '0x0000..'], //plz use some valid signature in 2nd param
```

packages/web3-eth-personal/src/index.ts
```diff
@@ -0,0 +1,113 @@
...
+	utf8ToHex,
+	ValidTypes,
+} from 'web3-utils';
+import { Transaction, formatTransaction } from 'web3-eth';
+import { validator } from 'web3-validator';
+import { EthPersonalAPI } from './personal_api';
+
+export class EthPersonal extends Web3Context<EthPersonalAPI> {
+	public async getAccounts() {
+		const result = await this.requestManager.send({
```
jdevcs: I think web3.js shouldn't block these functions using await and it should return promise , and let user wait for promise resolution . so user code is able to leverage async based on his requirements. Same suggestion for all following functions.

jdevcs (dismissed):

packages/web3-eth-personal/src/index.ts
nazarhussain: User of the library will be interacting with `getAccounts` function that will be returning the promise in any case. So either we return Promise chain with `then` block to process result or use `await` both have exactly same user experience. 

Use will be able to leverage `async` behavior, no matter we use `await` in the function or not, because the function user is interacting with is returning the Promise. That's the fundamental of adoption of async/await pattern.

Spacesai1or (changes requested):
Regarding `packages/web3-eth-personal/src/index.ts`, while I understand the efficiencies of a single file that includes the validation and formatting of parameters and the sending of the JSON RPC payload, this goes against the pattern that was established for the `eth_` RPC methods

I still think a more versatile approach would be splitting the methods into the separate files:

- `rpc_methods.ts`: Bare minimum RPC methods that don't include any validation or formatting
- `rpc_method_wrappers.ts`: For methods that make use of validation and/or formatting
- `index.ts`: imports and minimally wraps methods from above two files

Spacesai1or (approved):

luu-alex (approved):

PR #4792, opened 2022-02-24
Wyatt/4.x/4789 transaction rpc methods

### LINKED ISSUE #4789: Missing Transaction related RPC Methods
### Missing RPC Methods

- `sendTransaction`
- `signTransaction`
- `call`

---

These methods need to include [promiEvent](https://web3js.readthedocs.io/en/v1.7.0/callbacks-promises-events.html) support

- `sendTransaction`
- `sendSignedTransaction` (a.k.a `sendRawTransaction`)

---

[Nice to have](https://github.com/ChainSafe/web3.js/issues/3204)

Spacesai1or:
### What This PR Does

- Adds missing RPC methods:
    - `sendTransaction`
    - `signTransaction`
    - `call`
- Adds `PromiEvent` support to:
    - `sendTransaction`
    - `sendSignedTransaction`

### TODOs

- [x] Decide whether to throw an error when [polling times out](https://github.com/ChainSafe/web3.js/blob/96026ec5990cd01a6ccfe33daa68b67de50bd1ad/packages/web3-eth/src/rpc_method_wrappers.ts#L299)
- Replace inline errors with constructors
    - [x] [One](https://github.com/ChainSafe/web3.js/blob/96026ec5990cd01a6ccfe33daa68b67de50bd1ad/packages/web3-eth/src/rpc_method_wrappers.ts#L328)
    - [x] [Two](https://github.com/ChainSafe/web3.js/blob/96026ec5990cd01a6ccfe33daa68b67de50bd1ad/packages/web3-eth/src/rpc_method_wrappers.ts#L332)
- [x] Decided to implement `transaction.to` as a required property for `call` (previously mentioned [here](https://github.com/ChainSafe/web3.js/pull/4525#issuecomment-982330076))

---

In `1.x` there was a config option called `transactionPollingInterval` that used both to configure the interval in which to poll for transaction receipts and confirmations after submission via `sendTransaction`. The PR introduces a couple breaking changes:
    - Removal of `transactionPollingInterval`
    - Addition of `transactionReceiptPollingInterval` and `transactionConfirmationPollingInterval`
This point is being discussed [below](https://github.com/ChainSafe/web3.js/pull/4792#discussion_r814529853), but I am now thinking that we should reintroduce `transactionPollingInterval` and have it update both `transactionReceiptPollingInterval` and `transactionConfirmationPollingInterval` if changed, but also keep this separate config options to give the user more control if needed

closes #4789


packages/web3-core/src/types.ts
```diff
@@ -46,6 +46,8 @@ export interface Web3ConfigOptions {
 	transactionBlockTimeout: number;
 	transactionConfirmationBlocks: number;
 	transactionPollingTimeout: number;
+	transactionReceiptPollingInterval: number;
```
Spacesai1or: This was called `transactionPollingInterval` in `1.x`, but I believe this is a better name as it's more specific to what it affects - this is a breaking change though

packages/web3-core/src/web3_config.ts
```diff
@@ -22,6 +22,8 @@ export abstract class Web3Config
 		transactionBlockTimeout: 50,
 		transactionConfirmationBlocks: 24,
 		transactionPollingTimeout: 750,
+		transactionReceiptPollingInterval: 1000,
```
spacesailor24, line 25: I don't understand how the polling interval can be greater than the timeout, this would result in a timeout every time the polling strategy was taken

packages/web3-core/src/web3_config.ts
```diff
@@ -22,6 +22,8 @@ export abstract class Web3Config
...
+		transactionConfirmationPollingInterval: 13000,
```
spacesailor24, line 26: This doesn't exist in `1.x` and instead uses `transactionPollingInterval` (which I renamed to `transactionReceiptPollingInterval` in this PR),  but I believe it would be better to split these configurations as `1.x` was requesting a block every 1 second (by default) to check for a new block, while the [average block time](https://ycharts.com/indicators/ethereum_average_block_time) is around 13 seconds

packages/web3-core/src/types.ts
nazarhussain: Shouldn't we keep the original one as well and deprecate it.

nazarhussain (changes requested):

packages/web3-core/src/types.ts
Spacesai1or: @nazarhussain  Yea we probably should, we'd just need the getter/setter of one call the other so they never become out of sync

Spacesai1or: For clarification, I reintroduced `transactionPollingInterval` and have it update both `transactionReceiptPollingInterval` and `transactionConfirmationPollingInterval` if changed

jjlitt1 (commented):
Thanks

nazarhussain (approved):

luu-alex (approved):

PR #4794, opened 2022-02-24
4x ENS functionality

jdevcs:
## Description

https://github.com/ChainSafe/web3.js/issues/4552
https://github.com/ChainSafe/web3.js/issues/4602
https://github.com/ChainSafe/web3.js/issues/4553

## Type of change

## Checklist:


packages/web3-eth-ens/src/resolver.ts
```diff
@@ -0,0 +1,125 @@
...
+		resolverContract: Contract<typeof RESOLVER>,
+		methodName: string,
+	) {
+		if (interfaceIds[methodName] === undefined) throw new Error(); //  ResolverMethodMissingError(resolver address, methodName);
+
+		let supported = false;
+		try {
+			supported = (await resolverContract.methods
+				.supportsInterface(interfaceIds[methodName])
+				.call()) as Awaited<Promise<boolean>>;
```
nazarhussain: If contract typing is correct, then we don't need to use `as` here. Please check what type error you were getting here.

packages/web3-eth-ens/src/resolver.ts
```diff
@@ -0,0 +1,125 @@
...
+		} catch (err) {
+			//  TODO throw new Error
+		}
+
+		if (!supported) throw new Error(); // new ResolverMethodMissingError(resolver address, methodName);
```
nazarhussain:
```suggestion
		if (!supported) throw new Error(); // new ResolverMethodMissingError(resolver address, methodName);
		
		return supported; 
```

nazarhussain (changes requested):

packages/web3-eth-ens/src/resolver.ts
jdevcs: `error  Unsafe assignment of an any value  @typescript-eslint/no-unsafe-assignment`

packages/web3-eth-ens/src/resolver.ts
jdevcs: I wasnt using boolean as return result

packages/web3-eth-ens/src/config.ts
```diff
@@ -1 +1,22 @@
...
+
+// https://docs.ens.domains/contract-developer-guide/writing-a-resolver
+// resolver interface Ids
+export const interfaceIds: { [T: string]: string } = {
+	addr: '0x3b3b57de',
+	name: '0x691f3431',
+	abi: '0x2203ab56',
+	pubkey: '0xc8690233',
+	text: '0x59d1d43c',
+	contenthash: '0xbc1c58d1',
```
luu-alex, line 11: in the docs it suggests to support `supportsInterface` with the interface id of `0x01ffc9a7`

packages/web3-eth-ens/src/ens.ts
```diff
@@ -116,46 +119,79 @@ export class ENS {
...
+		return this._resolver.getContenthash(ENSName);
+	}
 
 	// TODO after eth.net.getNetworkType is complete
 	// public checkNetwork (): boolean {
 	// return true;
 	//  };
-
-	// TODO finish in resolver
-	// public supportsInterface(){
```
luu-alex, line 158: ENS should also include the supportsinterface method from resolver

luu-alex:
i think unit tests for the resolver class would be good, other then that LGTM

packages/web3-eth-ens/src/resolver.ts
nazarhussain: @jdevcs  That's not right. Please create an issue to address this later.

nazarhussain (approved):

packages/web3-eth-ens/src/config.ts
jdevcs: Internally we are not sending any eth_call to `supportsInterface` for checking if resolver contract have `supportsInterface` . For external calls, If user want he can directly check it via [this](https://github.com/ChainSafe/web3.js/blob/ad2ee10234ba1f0c96d1dca48bb1c0d9bb4769cf/packages/web3-eth-ens/src/resolver.ts#L75)

jdevcs:
> i think unit tests for the resolver class would be good, other then that LGTM

@luu-alex Resolver class is only sending transactions and sending eth_calls for reads so I think we can test it in integration tests.

packages/web3-eth-ens/src/resolver.ts
jdevcs: I was also wondering with typed contract feature it should have correct type.

luu-alex (approved):

packages/web3-eth-ens/src/resolver.ts
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	private async getResolverContractAdapter(ENSName: string) {
+		//  TODO : (Future 4.1.0 TDB) cache resolver contract if frequently queried same ENS name, refresh cache based on TTL and usage, also limit cache size, optional cache with a flag
+		return this.registry.getResolver(ENSName);
+	}
+
+	//  https://eips.ethereum.org/EIPS/eip-165
+	// eslint-disable-next-line class-methods-use-this
+	public async checkInterfaceSupport(
```
Spacesai1or, line 26: This could be a static method?

Spacesai1or (commented):
You could do this in a subsequent PR, but I think `packages/web3-eth-ens/src/ens.ts` should have sanity check unit tests that test `this._resolver` is receiving the expected arguments from `ENS` class - this ensures the user facing API never changes accidentally

packages/web3-eth-ens/src/resolver.ts
luu-alex: I dont think so, it calls the resolverContract on line 38-40

luu-alex: I dont think this matters too much, resolver is only used internally by us and is only exposed through the ENS class

luu-alex:
@spacesailor24 i included add sanity check for resolver in the ens epic, so we can merge this one in

PR #4824, opened 2022-03-04
ETH Tx Utilities TODOs [4/5] - Function overrides

### LINKED ISSUE #4801: ETH Tx Utilities TODOs [4/5] - Function overrides
#4778 is being split into multiple issues, this issue being tasks needed to be addressed before `v4.0.0-alpha`:

- Refactor override method functionality for transaction methods
    - [detectTransactionType](https://github.com/ChainSafe/web3.js/blob/49171a15d58b3f541337cd3d95178e49f7d18ed7/packages/web3-eth/src/eth_tx.ts#L105)
    - [prepareTransactionForSigning](https://github.com/ChainSafe/web3.js/blob/49171a15d58b3f541337cd3d95178e49f7d18ed7/packages/web3-eth/src/eth_tx.ts#L326)

nazarhussain:
## Description

Add tx functions to configuration and override support.

Fixes #4801

## Type of change

## Checklist:


nazarhussain:
@spacesailor24  Please review the changes, I will fix the tests afterwards.

packages/web3-core/src/types.ts
```diff
@@ -40,6 +43,16 @@ export type Web3BaseProviderConstructor = new <API extends Web3APISpec>(
 	net?: Socket,
 ) => Web3BaseProvider<API>;
 
+// TODO: When we have `web3-types` package we can share TransactionType
+export type TransactionTypeParser = (transaction: Record<string, unknown>) => HexString | undefined;
+
+// TODO: When we have `web3-types` package we can share TransactionType
+export type TransactionBuilder = (options: {
+	transaction: Record<string, unknown>;
```
nazarhussain: We don't have a shared type so why can't use the `Transaction` type here.

packages/web3-eth/src/eth_tx.ts
```diff
@@ -279,17 +263,23 @@ export const prepareTransactionForSigning = async (
 	web3Context: Web3Context<EthExecutionAPI>,
 	privateKey?: HexString | Buffer,
 ) => {
-	const populatedTransaction = await populateTransaction(
-		transaction,
+	const transactionBuilder = web3Context.transactionBuilder ?? defaultTransactionBuilder;
```
nazarhussain: I think top level function `prepareTransactionForSigning` should not be overriden instead inner parts. So why I kept `prepareTransactionForSigning` as it is and made `transactionBuilder` as override.

packages/web3-eth/src/eth_tx.ts
```diff
@@ -279,17 +263,23 @@ export const prepareTransactionForSigning = async (
...
 		web3Context,
-		ValidTypes.HexString,
 		privateKey,
-	);
+	});
+
+	const formattedTransaction = formatTransaction(
+		populatedTransaction,
+		ValidTypes.HexString,
+	) as PopulatedUnsignedTransaction<HexString>;
```
nazarhussain: No matter who build the transaciton, we have to format it properly before signing, so why I extracted formatting part to `prepareTransactionForSigning`

packages/web3-eth/src/eth_tx.ts
```diff
@@ -28,38 +22,33 @@ import {
...
+	if (tx.accessList !== undefined || tx.hardfork === 'berlin' || tx.common?.hardfork === 'berlin')
 		return '0x1';
 
 	return undefined;
 };
 
+export const detectTransactionType = (
+	transaction: Transaction,
+	transactionTypeParser: TransactionTypeParser = defaultTransactionTypeParser,
+) => transactionTypeParser(transaction as unknown as Record<string, unknown>);
```
Spacesai1or:
```suggestion
) =>
    // TODO: When we have `web3-types` package we can share TransactionType
    transactionTypeParser(transaction as unknown as Record<string, unknown>);
```

Just so this doesn't get overlooked

packages/web3-eth/src/eth_tx.ts
```diff
@@ -209,12 +197,8 @@ export async function populateTransaction<
 		}
 	}
 
-	// TODO - Types of property 'gasPrice' are incompatible
-	return formatTransaction(
-		populatedTransaction,
-		desiredType,
-	) as unknown as PopulatedUnsignedTransaction<NumberType>;
-}
+	return populatedTransaction as Record<string, unknown>;
```
Spacesai1or:
```suggestion
    // TODO: When we have `web3-types` package we can share TransactionType
	return populatedTransaction as Record<string, unknown>;
```

packages/web3-eth/src/eth_tx.ts
Spacesai1or: We should remove the `TODO` on line `260`

Spacesai1or (dismissed):
Awesome job Nazar, thank you for implementing this so quickly

Spacesai1or (commented):
Awesome job Nazar, thank you for implementing this so quickly

packages/web3-core/src/types.ts
```diff
@@ -40,6 +43,16 @@ export type Web3BaseProviderConstructor = new <API extends Web3APISpec>(
 	net?: Socket,
 ) => Web3BaseProvider<API>;
 
+// TODO: When we have `web3-types` package we can share TransactionType
+export type TransactionTypeParser = (transaction: Record<string, unknown>) => HexString | undefined;
+
+// TODO: When we have `web3-types` package we can share TransactionType
+export type TransactionBuilder = (options: {
```
spacesailor24, line 50: I'm working on breaking all the methods in `eth_tx.ts` into separate files, and was wanting to create a `transactionBuilder` method that uses `defaultTransactionBuilder` if not overridden like you have setup for `transactionTypeParser`. In doing so, I'm wondering why you chose to pass a object to `TransactionBuilder` instead of just passing the parameters themselves?

packages/web3-core/src/types.ts
```diff
@@ -40,6 +43,16 @@ export type Web3BaseProviderConstructor = new <API extends Web3APISpec>(
...
+	transaction: Record<string, unknown>;
+	web3Context: Web3Context<EthExecutionAPI>;
```
Spacesai1or: Also `TransactionBuilder` should support different APIs being passed e.g. `Web3EthApi`

jjlitt1: Ok

packages/web3-core/src/types.ts
spacesailor24: For more  context, I'd like to create `transactionBuilder` like so:

```typescript
export const transactionBuilder = async (
    transaction: Transaction,
	web3Context: Web3Context<EthExecutionAPI>,
	privateKey?: HexString | Buffer,
    transactionBuilder: TransactionBuilder = defaultTransactionBuilder,
) => transactionBuilder({
    transaction: transaction as unknown as Record<string, unknown>,
    web3Context,
    privateKey
});
```

spacesailor24: I'm realizing you already have similar functionality [here](https://github.com/ChainSafe/web3.js/blob/f932da4f8544ba3f9f07c6981abdb3b1def2d993/packages/web3-eth/src/eth_tx.ts#L266), but I think the `defaultTransactionBuilder` is an implementation detail that a dev using `transactionBuilder` shouldn't have to worry about

Spacesai1or: Any reason to pass parameters as an object here?

Spacesai1or:
@nazarhussain I made PR #4827 that's based off this branch (`nh/4801-override-functions`), so please provide feedback and/or merge if you agree with these changes

packages/web3-core/src/types.ts
nazarhussain: Object parameters are easy to extend in future without making breaking changes. It's also easy to read when used.

Spacesai1or (approved):

luu-alex (approved):

PR #4827, opened 2022-03-07
ETH Transaction Utils Refactors

Spacesai1or:
This PR:

1. Breaks up util functions from `packages/web3-eth/src/eth_tx.ts` into separate files under `packages/web3-eth/src/utils/` as `eth_tx.ts` was getting too big
2. Refactor how `detectTransactionType` and `transactionBuilder` handle defaulting to their respective implementations. These methods now handle the decisions themselves (using provided `Web3Context`) instead of the dev having to specify which implementation to use every time. This is the difference in code:

Before:

```typescript
populatedTransaction.type = detectTransactionType(
		populatedTransaction,
		web3Context.transactionTypeParser ?? defaultTransactionTypeParser,
	);
```

After:

```typescript
populatedTransaction.type = detectTransactionType(populatedTransaction, web3Context);
```

Before:

```typescript
const transactionBuilder = web3Context.transactionBuilder ?? defaultTransactionBuilder;

	const populatedTransaction = await transactionBuilder({
		transaction: transaction as unknown as Record<string, unknown>,
		web3Context,
		privateKey,
	});
```

After:

```typescript
const populatedTransaction = await transactionBuilder(transaction, web3Context, privateKey);
```


nazarhussain (approved):

nazarhussain:
@spacesailor24 LGTM. You can resolve the conflicts and merge the PR.

luu-alex (approved):

PR #4829, opened 2022-03-08
web3-eth-ens 4.x -  ens object

### LINKED ISSUE #4752: create the ENS class - 4.x
## Description
The ENS class needs to be created for the ENS library .

## Criteria
Should be up to parity with 1.x. 
All methods and the ENS object should be tested. 
Resources: https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-eth-ens/src/ENS.js

### LINKED ISSUE #4820: 4.x web3-eth-ens - constructor tests resolver
Include constructor unit tests for resolver in the ENS package

luu-alex:
#4752 Completing the ENS object and constructor unit tests
#4820 Creates constructor unit tests for resolver
In ENS Includes:
-  checkNetwork function


luu-alex:
Question: When creating the ENS object, should the registry address default to main registry contract address(rather than null, or any testnets)?

packages/web3/src/index.ts
```diff
@@ -29,7 +30,7 @@ import {
 import { Address } from 'web3-utils';
 import { ContractError } from './errors';
 
-export class Web3 extends Web3Context<EthExecutionAPI> {
+export class Web3 extends Web3Context<EthExecutionAPI | Web3NetAPI> {
```
nazarhussain: We should not change the API here. It should just be used in `web3-eth-net` package. 

```suggestion
export class Web3 extends Web3Context<EthExecutionAPI> {
```

packages/web3/src/index.ts
```diff
@@ -61,7 +62,7 @@ export class Web3 extends Web3Context<EthExecutionAPI> {
 		};
 	};
 
-	public constructor(provider: SupportedProviders<EthExecutionAPI>) {
+	public constructor(provider: SupportedProviders<EthExecutionAPI | Web3NetAPI>) {
```
nazarhussain:
```suggestion
	public constructor(provider: SupportedProviders<EthExecutionAPI>) {
```

nazarhussain (changes requested):

packages/web3-eth-ens/src/ens.ts
```diff
@@ -195,15 +203,42 @@ export class ENS extends Web3Context<EthExecutionAPI> {
...
+				throw new Error(`Network not synced; last block was ${headAge} seconds ago`);
+			}
+
+			this._lastSyncCheck = now;
+		}
+
+		if (this._detectedAddress) {
+			return this._detectedAddress;
+		}
+		const net = new Web3Net(this);
```
jdevcs: Instead of creating new object here, good to explore: directly using rpc [wrapper](https://github.com/ChainSafe/web3.js/blob/2c557d6a8075e6f95f2fbda2668bd35191c81857/packages/web3-net/src/rpc_method_wrappers.ts#L7) internally , and passing context to that.

packages/web3-eth-ens/src/ens.ts
```diff
@@ -195,15 +203,42 @@ export class ENS extends Web3Context<EthExecutionAPI> {
...
+		const networkType = await net.getId(); // get the network from provider
+		const addr = registryAddresses[networkType];
+
+		if (typeof addr === 'undefined') {
+			throw new Error(`ENS is not supported on network ${networkType}`);
```
jdevcs:
```suggestion
            // TODO use web3 error constructor
			throw new Error(`ENS is not supported on network ${networkType}`);
```

packages/web3-eth-ens/src/resolver.ts
```diff
@@ -12,8 +12,8 @@ import { namehash } from './utils';
 export class Resolver {
 	private readonly registry: Registry;
 
-	public constructor(registry: Registry) {
-		this.registry = registry;
+	public constructor(registry?: Registry) {
+		this.registry = registry ?? new Registry();
```
jdevcs: We should always reuse existing registry instance for multiple resolver. For efficient memory usage. 
or is there any scenario where its unavoidable?

```suggestion
	public constructor(registry: Registry) {
		this.registry = registry;
```

jdevcs (changes requested):

packages/web3/src/index.ts
luu-alex: Understandable 👍  sounds good

packages/web3-eth-ens/src/resolver.ts
luu-alex: I believe we should be able to reuse registry, the only case i could think of is if a user wanted to use a different registry addr

nazarhussain (approved):

jdevcs (approved):

PR #4833, opened 2022-03-08
Make disconnect params optional in types

wbt:
## Description

Though the `disconnect` function [isn't documented](https://web3js.readthedocs.io/en/v1.7.1/search.html?q=disconnect&check_keywords=yes&area=default), in practice it does not seem to require two params, and the main [example given](https://github.com/ChainSafe/web3.js/blob/d805cf9f7d382f19739018810c30d94a6f2f749e/scripts/js/basic_usage.js#L62) omits them.  The second is absent in [multiple](https://github.com/ChainSafe/web3.js/blob/d805cf9f7d382f19739018810c30d94a6f2f749e/test/websocket.ganache.js#L92) [tests](https://github.com/ChainSafe/web3.js/blob/d805cf9f7d382f19739018810c30d94a6f2f749e/test/websocket.ganache.js#L136).  Both are [absent](https://github.com/ChainSafe/web3.js/blob/d805cf9f7d382f19739018810c30d94a6f2f749e/test/websocket.ganache.js#L169) [in](https://github.com/ChainSafe/web3.js/blob/d805cf9f7d382f19739018810c30d94a6f2f749e/test/websocket.ganache.js#L189) [other](https://github.com/ChainSafe/web3.js/blob/d805cf9f7d382f19739018810c30d94a6f2f749e/test/websocket.ganache.js#L400) [tests](https://github.com/ChainSafe/web3.js/blob/d805cf9f7d382f19739018810c30d94a6f2f749e/test/websocket.ganache.js#L437). 

The implementation which does seem to want it can be found [here](https://github.com/ChainSafe/web3.js/blob/d805cf9f7d382f19739018810c30d94a6f2f749e/packages/web3-providers-ws/src/index.js#L366) but the first parameter has an effective default value of `1000` and the second is only passed on to `this.connection.close()` which is implemented [here](https://github.com/theturtle32/WebSocket-Node/blob/cce6d468986dd356a52af5630fd8ed5726ba5b7a/lib/W3CWebSocket.js#L136).  That implementation only passes it on to `this._connection.close()` and as seen by parameterless calls to the same function two lines below each two-parameter call, those are optional. 

## Type of change

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas. [This does not seem necessary for this simple 2-character change.]
- [ ] I have made corresponding changes to the documentation. [The function isn't documented so no changes are needed; this brings the typing more in line with documentation that already does exist.]
- [ ] Any dependent changes have been merged and published in downstream modules. [There aren't any.]
- [x] I ran `npm run test:unit` with success. [Noting that the tests are unreliable on the base branch.]
- [ ] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code. [This command's use of semicolons instead of ampersands mean this superset of the line above has been broken for years. Fixing it here seems unnecessary: Existing test cases cover the change, or would if they were TypeScript, but changing that is beyond the scope of this PR.]
- [ ] I have tested `dist/web3.min.js` in a browser. [This doesn't make changes in compiled code.]
- [ ] I have tested my code on the live network. [This doesn't make changes in compiled code.]


wbt:
The failure is 'Failed to connect to Infura over websockets after 10 tries' which I don't think is a consequence of this very minor PR; it seems more likely to be an issue at Infura related to some connectivity issues they've been having lately.  Also, I'm not sure if the IP address being used for the test is in the range of newly censored IP addresses intentionally blocked from connecting to Infura, or one of those adjacent to it in some way that leads it to get caught up in the block.   

The post about test code coverage changing is also inaccurate.  This PR shouldn't change test coverage, and doesn't even touch the files cited as having new missed lines.

wbt:
Thought it is an improvement worth merging in, the extra change was intended just to kick off the CI again because I don't have permission to do that directly, and note that the tests are unreliable when run locally.  It produced the same issue, which I think is an Infura or testing infrastructure issue rather than a bug newly added in the branch proposed for merging.

nazarhussain (approved):

luu-alex (approved):

jdevcs (approved):

PR #4844, opened 2022-03-14
4555/encode decode parameters

### LINKED ISSUE #4555: Consistent encoding/decoding of the parameters

### Feature Request

Currently the `web3-eth-abi` package have to functions `encodeParameters` and `decodeParameters`. Both of these works fine but don't have a consistent output. 

1. We can't use the output of one function to pass to other 
2. Decoding result in different format of data used to encode

### Use Cases

The following code snippet should work fine.

```js
const abi = ['uint256', 'string'];
const inputParams = ['2345675643', 'Hello!%'];

const encodedOutput = encodeParameters(abi, inputParams);
const decodedParams = decodeParameters(abi, encodedOutput);

assert(inputParams === decodedParams);
```

### Implementation Ideas

The reason behind the difference is the implementation of the output format. Which is not based on array index rather object index. Same behaviour exists in 1.x and 4.x

_No response_

- [ ] Yes

luu-alex:
#4555

an input that is encoded should equal the result of the bytes decoded for consistency
This PR includes testcases to test an input that is encoded will be formatted when decoded

```
const abi = ['uint256', 'string'];
const inputParams = ['2345675643', 'Hello!%'];

const encodedOutput = encodeParameters(abi, inputParams);
const decodedParams = decodeParameters(abi, encodedOutput);

assert(inputParams === decodedParams);
```


nazarhussain (approved):

jdevcs (approved):

jdevcs:
@luu-alex also fix linting error
`web3-eth-abi:   111:85  error  Unsafe argument of type `any` assigned to a parameter of type `{ [key: string]: unknown; }`  @typescript-eslint/no-unsafe-argument`
 , all other changes LGTM

PR #4845, opened 2022-03-14
Drop web3-eth circular dependence on web3-eth-ens

wbt:
## Description
This PR attempts to remove a warning about a circular dependency which looks like it might be causing an issue blocking CI of the code typings under test which blocks substantive review of PR #4841.

## Type of change

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas. [This PR only removes code; there's nothing new to comment and nothing left behind without explanation.]
- [ ] I have made corresponding changes to the documentation. [This change doesn't affect what you would read in the documentation.]
- [ ] Any dependent changes have been merged and published in downstream modules. [There are none.]
- [ ] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code. [This PR doesn't add code and the command is broken.]
- [ ] I have tested `dist/web3.min.js` in a browser.


jdevcs:
@wbt  Thanks for your contributions, Tests are failing in CI , could you check and fix it for this PR. 
```
unit 14 
e2e unit and e2e clients
e2e browsers
```

wbt:
I think there is actually an issue in this one and am not sure that it's ready for merging, but getting it closer to that point now looks like a bigger architecture issue of circular dependency requirements, with the ENS functionality being bound too tightly into the ETH package. I'm not sure I can actually fix that myself.

packages/web3-eth/src/index.js
```diff
@@ -382,9 +377,6 @@ var Eth = function Eth() {
     // add ABI
     this.abi = abi;
 
-    // add ENS
-    this.ens = new ENS(this);
-
```
nazarhussain, line 387: This will introduce a [breaking change](https://web3js.readthedocs.io/en/v1.5.0/web3-eth-ens.html) in `1.x`. I am not sure if we are intended to make such changes while maintaining `1.x`. @jdevcs @spacesailor24 @luu-alex Share your thoughts.

nazarhussain (changes requested):

packages/web3-eth/src/index.js
luu-alex: Removing ENS  from the web3 package might be a little out of scope for maintaining `1.x`. This issue is fixed in `4.x` and might be too big of a change in `1.x`.

luu-alex:
Hey there @wbt as you said there might be a bigger issue at hand with the testcases failing, and this being a breaking change to the web3 package I'll be closing this PR. Feel free to open this again or create a new PR if you do attempt at this issue.

PR #4847, opened 2022-03-14
errorhandling 4583

jdevcs:
## Description

For: https://github.com/ChainSafe/web3.js/issues/4583

## Type of change

## Checklist:


packages/web3-eth/src/errors.ts
```diff
@@ -29,15 +29,27 @@ export class InvalidTransactionWithSender extends Web3Error {
 	public code = ERR_TX_INVALID_SENDER;
 
 	public constructor(value: unknown) {
-		super(JSON.stringify(value), 'invalid transaction with sender');
+		super(
+			JSON.stringify(
+				value,
+				(_, v) => (typeof v === 'bigint' ? v.toString() : v) as unknown,
+			).replace(/['"]+/g, ''),
```
nazarhussain: I would suggest to move this logic of parsing data to string into `web3-utils` package and then use here.

nazarhussain (changes requested):

luu-alex (approved):

Spacesai1or (dismissed):
This `jsonStringify` shouldn't be applied to specific errors, but rather the [Web3Error](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-common/src/errors.ts#L57) class which our errors extend. In this way, the developer doesn't have to think about it every time they introduce a new error (possibly forgetting to do this, causing a runtime error)

I also wouldn't name this functionality `jsonStringify`, since `JSON.stringify` exists and this isn't just handling JSON. Perhaps we can just make a `toString` method that can handle BigInts, objects, Buffers, etc

nazarhussain (approved):

nazarhussain:
Apart from @spacesailor24  comment LGTM.

avkos:
> This `jsonStringify` shouldn't be applied to specific errors, but rather the [Web3Error](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-common/src/errors.ts#L57) class which our errors extend. In this way, the developer doesn't have to think about it every time they introduce a new error (possibly forgetting to do this, causing a runtime error)
> 
> I also wouldn't name this functionality `jsonStringify`, since `JSON.stringify` exists and this isn't just handling JSON. Perhaps we can just make a `toString` method that can handle BigInts, objects, Buffers, etc

yes. I also think we need to implement it inside the Web3Error class.

avkos (changes requested):
yes. I also think we need to implement it inside the Web3Error class.

jdevcs:
> This `jsonStringify` shouldn't be applied to specific errors, but rather the [Web3Error](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-common/src/errors.ts#L57) class which our errors extend. In this way, the developer doesn't have to think about it every time they introduce a new error (possibly forgetting to do this, causing a runtime error)
> 
> I also wouldn't name this functionality `jsonStringify`, since `JSON.stringify` exists and this isn't just handling JSON. Perhaps we can just make a `toString` method that can handle BigInts, objects, Buffers, etc

[Web3Error](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-common/src/errors.ts#L57)  is already getting [string](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-common/src/errors.ts#L61) instead of `unknown` type and all of its sub classes are passing string, so no need of parsing that.

More good candidate of parsing `unknown` is [here](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-utils/src/errors.ts#L6), 

 I think this another [web3error](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-utils/src/errors.ts#L2) is created in utils package for avoiding cyclic dependency bw utils and common, but @spacesailor24 Why you used it in web3-eth package instead of web3-error from [common](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-common/src/errors.ts#L57), I think it should be changed in Eth.

avkos (approved):

jdevcs:
@spacesailor24

Spacesai1or:
> > This `jsonStringify` shouldn't be applied to specific errors, but rather the [Web3Error](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-common/src/errors.ts#L57) class which our errors extend. In this way, the developer doesn't have to think about it every time they introduce a new error (possibly forgetting to do this, causing a runtime error)
> > I also wouldn't name this functionality `jsonStringify`, since `JSON.stringify` exists and this isn't just handling JSON. Perhaps we can just make a `toString` method that can handle BigInts, objects, Buffers, etc
> 
> [Web3Error](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-common/src/errors.ts#L57) is already getting [string](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-common/src/errors.ts#L61) instead of `unknown` type and all of its sub classes are passing string, so no need of parsing that.
> 
> More good candidate of parsing `unknown` is [here](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-utils/src/errors.ts#L6),
> 
> I think this another [web3error](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-utils/src/errors.ts#L2) is created in utils package for avoiding cyclic dependency bw utils and common, but @spacesailor24 Why you used it in web3-eth package instead of web3-error from [common](https://github.com/ChainSafe/web3.js/blob/f71d2fe37cf2848d1368f902b376d5fa6cc96714/packages/web3-common/src/errors.ts#L57), I think it should be changed in Eth.

@jdevcs I didn't realize we had two different implementations of `Web3Error`, dependency cycle aside, this does not seem like the best approach and we should investigate only having one class for errors

I think you're right, `Web3Eth` should be using `Web3Error` from `web3-common` and just pass a formatted string. To do this though, we still require formatting potential `unknown` type values into a `string` to be used for the error message

nikoulai (approved):

jdevcs:
https://github.com/ChainSafe/web3.js/issues/4889

PR #4859, opened 2022-03-16
Refactor formatters to JSON schemas and TypeScript band-aid solution

### LINKED ISSUE #4778: ETH Tx Utilities TODOs [1/5] - Refactor formatters to JSON schemas and TypeScript band-aid solution
Originally mentioned [here](https://github.com/ChainSafe/web3.js/pull/4718#issuecomment-1032134238)

#4778 is being split into multiple issues, this issue being tasks not needed to be addressed before `v4.0.0-alpha`:

- [Replacing formatters with JSON schemas](https://github.com/ChainSafe/web3.js/pull/4718/files#r789942037)
- [Refactor](https://github.com/ChainSafe/web3.js/blob/49171a15d58b3f541337cd3d95178e49f7d18ed7/packages/web3-eth/src/eth_tx.ts#L276) typing to make TSC happy

nazarhussain:
## Description

Fixes #4778 

## Type of change

## Checklist:


packages/web3-common/src/formatter.ts
```diff
@@ -0,0 +1,299 @@
...
+	[FMT_BYTES.BUFFER]: Buffer;
+	[FMT_BYTES.UINT8ARRAY]: Uint8Array;
+};
+
+export type DataFormat = {
+	readonly number: FMT_NUMBER;
+	readonly bytes: FMT_BYTES;
+};
+
+export const DEFAULT_RETURN_FORMAT = { number: FMT_NUMBER.NUMBER, bytes: FMT_BYTES.HEX } as const;
```
Spacesai1or:
```suggestion
export const DEFAULT_RETURN_FORMAT = { number: FMT_NUMBER.HEX, bytes: FMT_BYTES.HEX } as const;
```

packages/web3-common/src/formatter.ts
```diff
@@ -0,0 +1,299 @@
...
+		} else if (result.items && Array.isArray(result.items)) {
+			result = (result.items as JsonSchema[])[parseInt(dataPart, 10)];
+		}
+	}
+
+	return result;
+};
+
+// To get one format value so we can convert to other format
+export const getNumberValue = (value: unknown): bigint => {
```
Spacesai1or: We could rename this to `toBigInt` and move it to `web3-utils`

packages/web3-common/src/formatter.ts
```diff
@@ -0,0 +1,299 @@
...
+
+	if (typeof value === 'string' && isHexStrict(value)) {
+		return BigInt(hexToNumber(value));
+	}
+
+	throw new Error('Invalid type');
+};
+
+// To get one format value so we can convert to other format
+export const getBytesValue = (value: unknown): Buffer => {
```
Spacesai1or: We could rename this to `toBuffer` and move it to `web3-utils`

packages/web3-common/src/formatter.ts
```diff
@@ -0,0 +1,253 @@
...
+		if (schemaProp === undefined) {
+			delete object[key];
+			dataPath.pop();
+
+			continue;
+		}
+
+		// If value is an object, recurse into it
+		if (isObject(value)) {
+			// dataPath.push(key);
```
Spacesai1or: Is this needed?

packages/web3-net/src/rpc_method_wrappers.ts
```diff
@@ -1,31 +1,25 @@
...
+	returnFormat: ReturnFormat,
 ) {
 	const response = await rpcMethods.getPeerCount(web3Context.requestManager);
 
-	return convertToValidType(
-		response,
-		returnType ?? web3Context.defaultReturnType,
-	) as ValidReturnTypes[ReturnType];
+	// Data returned is number in hex format
+	return format({ eth: 'uint' }, response as unknown as number, returnFormat);
```
Spacesai1or, line 22:
```suggestion
	return format({ eth: 'uint' }, response as unknown as HexString, returnFormat);
```

If we want this change, we need to add the import for `HexString`

packages/web3-net/src/rpc_method_wrappers.ts
```diff
@@ -1,31 +1,25 @@
...
-	returnType?: ReturnType,
+	returnFormat: ReturnFormat,
 ) {
 	const response = await rpcMethods.getId(web3Context.requestManager);
 
-	return convertToValidType(
-		response,
-		returnType ?? web3Context.defaultReturnType,
-	) as ValidReturnTypes[ReturnType];
+	return format({ eth: 'uint' }, response as unknown as number, returnFormat);
```
Spacesai1or, line 12:
```suggestion
	return format({ eth: 'uint' }, response as unknown as string, returnFormat);
```

The version (id) is returned as a number string

packages/web3-common/src/formatter.ts
```diff
@@ -0,0 +1,253 @@
...
+		dataToParse = [...data];
+	} else {
+		dataToParse = data;
+	}
+
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+	const jsonSchema: JsonSchema = isObject(schema) ? schema : utils.ethAbiToJsonSchema(schema);
+
+	if (!jsonSchema.properties && !jsonSchema.items && !jsonSchema.eth) {
+		throw new Error('Nazar');
```
Spacesai1or: Nazar, you're a great developer, not an error :slightly_smiling_face:

Spacesai1or (commented):
I don't quite understand all that's going on in `packages/web3-common/src/formatter.ts`, but the use of the formatter seems better than what we had. Thank you Nazar

Only other feedback would be to pull any off the `throw new Error('')` into the respective `errors.ts` files

packages/web3-net/src/rpc_method_wrappers.ts
nazarhussain: The type is casted based on data type and values are casted based on schema. For the utility type to cast the data type properly we have to use either `number` or `bigint`.

packages/web3-net/src/rpc_method_wrappers.ts
nazarhussain: Same as above, the value is identified and casted by schema, type here is used to properly format the output type.

packages/web3-common/src/formatter.ts
nazarhussain: There is a lot of other refactoring coming through, will refactor these error constructors as well.

nazarhussain:
@spacesailor24 We had two challenges to solve in formatters with schema. 

1. To properly format the values
2. To properly transform the output type

The value transformation is done through schema provided. The type formation is done through the source type of the data you pass. I tried to format the types based on schema, but that was very complex and didn't cover a lot edge cases, so we had to transform type from the source type. e.g. 

Say we have data: 

```ts
const data = {
     prop: 2
};

const schema = {type: 'object', properties: {prop: {eth: 'uint'}}};
```

In above the schema tells that `prop` is an unsigned integer value, and formatter will format this value into any format based on `FMT_NUMBER` enum you pass. But second problem to solve here is the type of the object `data`. Currently type of `object` is `{prop: number}`, if we format it to `FMT_NUMBER.HEX` then we have to convert the type to `{prop: string}` as well. 

So in this formatting technique the source data type is very important. Any type attribute which is declared as `number | bigint` will be type formatted as `FMT_NUMBER` and any type attribute which is declared as `Buffer | Uint8Array` will be formatted baed on `FMT_BYTES`. 

So to make it work properly, even if you know an attribute is hex string bytes, don't declare the type as `string` or `HexString`, declare it as `Buffer`, the rest will be handled by `FMT_BYTES` enum you pass, similarly for `FMT_NUMBER`.

packages/web3-common/src/formatter.ts
```diff
@@ -0,0 +1,254 @@
...
+			result = result.items;
+		} else if (result.items && Array.isArray(result.items)) {
+			result = (result.items as JsonSchema[])[parseInt(dataPart, 10)];
+		}
+	}
+
+	return result;
+};
+
+export const convertScalarValue = (value: unknown, ethType: string, format: DataFormat) => {
```
Spacesai1or, line 88: Do you think this should throw an error if both `if`s are falsy (i.e. `ethType` is not `int`, `uint`, or `bytes`)? Something like a `UnsupportedEthTypeError`?

packages/web3-common/src/formatter.ts
```diff
@@ -0,0 +1,254 @@
...
+
+export const convert = (
+	data: Record<string, unknown> | unknown[] | unknown,
+	schema: JsonSchema,
+	dataPath: string[],
+	format: DataFormat,
+) => {
+	// If it's a scalar value
+	if (!isObject(data) && !Array.isArray(data)) {
+		return convertScalarValue(data, schema?.eth as string, format);
```
spacesailor24, line 135: Since it's possible for `schema.eth` to be undefined, would this cause an error to be thrown in `convertScalarValue`?

packages/web3-eth-personal/test/unit/eth_personal.test.ts
```diff
@@ -174,7 +174,10 @@ describe('EthPersonal', () => {
 			await ethPersonal.sendTransaction(tx, 'password');
 
 			expect(eth.formatTransaction).toHaveBeenCalledTimes(1);
-			expect(eth.formatTransaction).toHaveBeenCalledWith(tx, 'HexString');
+			expect(eth.formatTransaction).toHaveBeenCalledWith(tx, {
```
Spacesai1or, line 177: Perhaps we can replace `{ bytes: 'BYTES_HEX', number: 'NUMBER_HEX' }` with `DEFAULT_RETURN_FORMAT`?

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1,47 +1,57 @@
 // Disabling because returnTypes must be last param to match 1.x params
 /* eslint-disable default-param-last */
```
spacesailor24, line 2:
```suggestion
```

packages/web3-eth/src/schemas.ts
```diff
@@ -0,0 +1,386 @@
...
+		totalDifficulty: {
+			eth: 'uint',
+		},
+		baseFeePerGas: {
+			eth: 'uint',
+		},
+		size: {
+			eth: 'uint',
+		},
+		// TODO: This attribute can be array of hashes or transaction info
```
Spacesai1or, line 216: Do the schemas not currently support multiple types for properties like this? What's the difficulty with figuring this out?

packages/web3-eth/src/utils/format_transaction.ts
```diff
@@ -16,56 +21,17 @@ export function formatTransaction<
...
-				formattedTransaction.common.customChain.networkId,
-				desiredType,
-			);
-		if (formattedTransaction.common.customChain.chainId !== undefined)
-			formattedTransaction.common.customChain.chainId = convertToValidType(
-				formattedTransaction.common.customChain.chainId,
-				desiredType,
-			);
-	}
+	const result = format(transactionSchema, formattedTransaction, returnFormat);
```
Spacesai1or: This is SUCH an improvement, fantastic job Nazar

packages/web3-eth/src/utils/format_transaction.ts
```diff
@@ -1,11 +1,16 @@
...
-	DesiredType extends ValidTypes,
-	NumberType extends ValidReturnTypes[DesiredType] = ValidReturnTypes[DesiredType],
->(transaction: Transaction, desiredType: DesiredType): Transaction<NumberType> {
+	ReturnFormat extends DataFormat = typeof DEFAULT_RETURN_FORMAT,
+	TransactionType extends Transaction = Transaction,
+>(
+	transaction: TransactionType,
+	returnFormat: ReturnFormat = DEFAULT_RETURN_FORMAT as ReturnFormat,
+): FormatType<TransactionType, ReturnFormat> {
 	const formattedTransaction = mergeDeep(
```
Spacesai1or:
```suggestion
	let formattedTransaction = mergeDeep(
```

packages/web3-eth/src/utils/format_transaction.ts
Spacesai1or:
```suggestion
	formattedTransaction = format(transactionSchema, formattedTransaction, returnFormat);
```

packages/web3-eth/src/utils/format_transaction.ts
```diff
@@ -16,56 +21,17 @@ export function formatTransaction<
...
+			data: bytesToBuffer(formattedTransaction.data).toString('hex'),
+			input: bytesToBuffer(formattedTransaction.input).toString('hex'),
 		});
 	else if (formattedTransaction.input !== undefined) {
 		formattedTransaction.data = formattedTransaction.input;
 		delete formattedTransaction.input;
 	}
 
-	return formattedTransaction as Transaction<NumberType>;
+	return result as FormatType<TransactionType, ReturnFormat>;
```
Spacesai1or:
```suggestion
	return formattedTransaction as FormatType<TransactionType, ReturnFormat>;
```

packages/web3-eth/src/utils/transaction_builder.ts
```diff
@@ -60,25 +62,30 @@ export async function defaultTransactionBuilder<ReturnType = Record<string, unkn
 	web3Context: Web3Context<EthExecutionAPI & Web3NetAPI>;
 	privateKey?: HexString | Buffer;
 }): Promise<ReturnType> {
-	let populatedTransaction = { ...options.transaction } as unknown as Transaction;
+	let populatedTransaction = { ...options.transaction } as unknown as InternalTransaction;
```
Spacesai1or, line 65: I don't think it's safe to assume that the transaction is `InternalTransaction` (a transaction with hex string numbers and bytes properties) if the parameter `transaction` has type `Record<string, unknown>`

Unless TypeScript has a fit about it, I think we should have `transaction` be of type [Transaction](https://github.com/ChainSafe/web3.js/blob/32b898477c455c8773e094e5b32a3d328a328ea4/packages/web3-eth/src/types.ts#L71) and we should format `populatedTransaction` to `DEFAULT_RETURN_FORMAT`

packages/web3-eth/test/fixtures/detect_transaction_type.ts
```diff
@@ -1,7 +1,7 @@
-import { ValidReturnTypes, ValidTypes } from 'web3-utils';
+import { DEFAULT_RETURN_FORMAT, FormatType } from 'web3-common';
 import { Transaction } from '../../src/types';
 
-export const transactionType0x0: Transaction<ValidReturnTypes[ValidTypes.HexString]>[] = [
+export const transactionType0x0: FormatType<Transaction, typeof DEFAULT_RETURN_FORMAT>[] = [
```
Spacesai1or, line 4: Could these use `InternalTransaction` instead?

packages/web3-utils/src/converters.ts
```diff
@@ -255,6 +252,29 @@ export const toNumber = (value: Numbers): number | bigint => {
...
+		return value;
+	}
+
+	if (typeof value === 'string' && !isHexStrict(value)) {
+		return BigInt(value);
+	}
+
+	if (typeof value === 'string' && isHexStrict(value)) {
+		return BigInt(hexToNumber(value));
+	}
```
Spacesai1or, line 273: Instead of using `isHexStrict` here, could we use `isNumber` and combine both `if` statements?

Also `return BigInt(hexToNumber(value));` could result in a over/underflow, so it would be nice to convert number and hex string straight to BigInt without converting to `number`

packages/web3-common/src/formatter.ts
nazarhussain: We have to return original value e.g. If someone uses `eth: 'string'`, then we dont' need to format and we can simply return the original value.

packages/web3-eth-personal/test/unit/eth_personal.test.ts
nazarhussain: Yes then we have to use `web3-common` as dependency for this package. I thought to avoid adding a dependency just for 2 test cases.

packages/web3-eth/src/schemas.ts
nazarhussain: In JSON-Schema, we use `anyOf` keyword for adding such conditional schema. The formatter does not support it yet, but we can add this support in future.

packages/web3-eth/src/utils/format_transaction.ts
nazarhussain: We didn't reassigned the `formattedTransaction`, so `const` is good to use.

packages/web3-utils/src/converters.ts
nazarhussain: `isNumber` does not work with hex strings.

packages/web3-eth/src/utils/transaction_builder.ts
nazarhussain: Created an issue to address this https://github.com/ChainSafe/web3.js/issues/4892

luu-alex (approved):

packages/web3-utils/src/converters.ts
Spacesai1or: `hexToNumber` actually checks if:

```typescript
if (num > Number.MAX_SAFE_INTEGER) {
		return negative ? -num : num;
	}
```

so if we update it to also check for `Number.MIN_SAFE_INTEGER`, than my concern should be covered

Spacesai1or (approved):

packages/web3-utils/src/converters.ts
nazarhussain: @spacesailor24 Create an issue for it https://github.com/ChainSafe/web3.js/issues/4900

PR #4848, opened 2022-03-14
Fix build command

wbt:
## Description
This partially addresses long-broken parts of the npm testing scripts this PR template requires that contributors complete.
It doesn't fully address all the issues (e.g. that docker is an undocumented dependency required to run the geth instamine script which is triggered by one of the commands below, but things get much farther along than they used to. 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules. [There are none]
- [x] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [x] I ran `npm run test:unit` with success.
- [ ] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code. [This works better now, but still not 100%]
- [ ] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
- [ ] I have tested my code on the live network.
- [ ] I have checked the Deploy Preview and it looks correct. [I'm not sure what this refers to or how to do this; further instructions could be helpful.]
- [x] I have updated the `CHANGELOG.md` file in the root folder.

PR #4861, opened 2022-03-17
Fix Build Command (supersedes #4848) 

Spacesai1or:
Original PR: #4848, this fixed some merge conflicts with `package-lock.json`

Something to note is a ran `npm i` which apparently fixed the version number of all packages, changing them from `1.7.1-rc.0` to `1.7.1`


nazarhussain (approved):

CHANGELOG.md
```diff
@@ -522,6 +522,7 @@ Released with 1.0.0-beta.37 code base.
 
 
 ## [Unreleased]
+- Improve npm script commands for development purposes (#4848)
 
 ## [1.7.2]
```
jdevcs:
```suggestion
## [Unreleased]

## [1.7.2]

### Changed

- Improve npm script commands for development purposes (#4848)


```

jdevcs (changes requested):

jdevcs (changes requested):
We should not make changes in package-lock.json in this PR, as scope of issue is for improvement in build commands

wbt:
@jdevcs this PR introduces a new dependency on `rimraf` for the `rm -rf` functionality, so package-lock.json does need to be on the changed files list.

package.json
```diff
@@ -127,6 +127,7 @@
         "mocha": "^6.2.3",
         "nyc": "^14.1.1",
         "pify": "^4.0.1",
+        "rimraf": "^3.0.2",
```
jdevcs, line 130:
```suggestion
     
```

package.json
```diff
@@ -23,27 +23,27 @@
...
+        "test:e2e:min": "bash ./scripts/e2e.min.sh",
+        "test:e2e:cdn": "bash ./scripts/e2e.cdn.sh",
+        "test:e2e:browsers": "npm run build && npm run test:e2e:chrome && npm run test:e2e:firefox",
+        "test:e2e:publish": "bash ./scripts/e2e.npm.publish.sh",
+        "test:e2e:truffle": "bash ./scripts/e2e.truffle.sh",
+        "test:e2e:mosaic": "bash ./scripts/e2e.mosaic.sh",
+        "test:e2e:ganache:core": "bash ./scripts/e2e.ganache.core.sh",
+        "test:e2e:gnosis:dex": "bash ./scripts/e2e.gnosis.dex.sh",
+        "ci": "bash ./scripts/ci.sh",
+        "cov:clean": "rimraf .nyc_output && rimraf coverage",
```
jdevcs, line 46:
```suggestion
        "cov:clean": "rm -rf .nyc_output; rm -rf coverage",
```

jdevcs (changes requested):

jdevcs:
> @jdevcs this PR introduces a new dependency on `rimraf` for the `rm -rf` functionality, so package-lock.json does need to be on the changed files list.

Hi @wbt ,we initiate changes in package lock files internally. Thanks for your contributions.

jdevcs (changes requested):
> @jdevcs this PR introduces a new dependency on `rimraf` for the `rm -rf` functionality, so package-lock.json does need to be on the changed files list.

@spacesailor24 we need to remove changes from package lock files in this PR. thanks

package.json
wbt: The suggested change actively reduces cross-platform compatibility.  Can you explain why it's important to avoid any changes to package-lock?

jdevcs: @wbt We include changes in package-lock files via dedicated and internally initialized PRs only.

wbt: This is now formally an internally initialized PR.  Going out of your way to reduce cross-platform compatibility and break the command on some platforms just doesn't seem like the right direction.

avkos: `rimraf` works exactly the same as `rm-rf` but with cross-platform support. Why do we need to change `rimraf` to `rm -rf`? What is it solve?

avkos (changes requested):
`rimraf` works exactly the same as `rm-rf` but with cross-platform support. Why do we need to change `rimraf` to `rm -rf`? What is it solve?

package.json
jdevcs: > @wbt We include changes in package-lock files via dedicated and internally initialized PRs only.

Internal and also **dedicated** PRs for package lock changes. Thanks

wbt: That rule means never adding or removing any dependencies that affect package functionality in any way.  Otherwise, the PR would change package.json or code that interacts with the dependency, and that's not allowed if package-lock can only be modified by PRs dedicated specifically to that file.  Am I missing something there?

Spacesai1or:
> `rimraf` works exactly the same as `rm-rf` but with cross-platform support. Why do we need to change `rimraf` to `rm -rf`? What is it solve?

@avkos  Web3.js supports Windows, OSX, and Linux environments, hence the cross platform support needed for removing files

Spacesai1or:
Okay so I deleted all the `package-lock.json` files and regenerated them with `npm i`, and now when running `npm run build` we're getting the errors:

```bash
../../node_modules/@ethereumjs/common/dist/types.d.ts(38,32): error TS1005: ',' expected.
../../node_modules/@ethereumjs/common/dist/types.d.ts(38,58): error TS1005: ',' expected.
../../node_modules/@ethereumjs/common/dist/types.d.ts(40,12): error TS1005: ',' expected.
../../node_modules/@ethereumjs/common/dist/types.d.ts(41,9): error TS1005: ',' expected.
../../node_modules/@ethereumjs/common/dist/types.d.ts(42,12): error TS1005: ',' expected.
```

avkos:
> > `rimraf` works exactly the same as `rm-rf` but with cross-platform support. Why do we need to change `rimraf` to `rm -rf`? What is it solve?
> 
> @avkos Web3.js supports Windows, OSX, and Linux environments, hence the cross platform support needed for removing files

I mean `rimraf` is better than `rm -rf`. We need to use it

package.json
wbt: I should state agreement that I don't think all of the proposed changes to package-lock files, especially those that alter the root package version number, should be included in this PR, but I also disagree with the very broadly stated rule that package-lock **cannot change at all** in any PR because **some** package-lock changes accompany any updates to the sets of active dependencies.

wbt: However, the version of this PR which does not change all the package-lock files this one does was rejected at #4848.  In theory, I believe that one could be reopened and merged, but @spacesailor24 seems to prefer the version with all the additional package-lock changes while @jdevcs strongly disagrees, and meanwhile the command nominally required to be run for every PR remains broken.

avkos (approved):

package.json
jdevcs: @wbt , we discussed this internally that @spacesailor24 will re generate package lock files in this PR, but there are errors when he tried. https://github.com/ChainSafe/web3.js/pull/4861#issuecomment-1078599638 

Our team is occupied in [4.x](https://github.com/ChainSafe/web3.js/tree/4.x) so could you investigate and share it.

wbt: @jdevcs can you just reopen #4848?

jdevcs: Its already merged so couldn't reopen.

wbt: Replaced with #4939.

PR #4876, opened 2022-03-21
Fix confirmation check after transaction getting stuck in a loop in some situations

Darkhogg:
## Description

Sometimes, the checkConfirmation login became completely stuck and unable to exit properly, wasting a lot of CPU cycles and API calls on nothing.

I've seen this behaviour in situations where the generations of new blocks was slow (in which case polling kicks in and both polling and subscription remained working, but only one of them would ever finish) and when using `.once('confirmation')` on the result of a `sendSignedTransaction` call.

This PR fixes the afforementioned cases in the most simple way I came up with that doesn't break other stuff.

## Type of change

## Checklist:

    - Honestly, I have no idea how to test this, it doesn't produce any internally visible anything
- [X] I ran `npm run test:unit` with success.
- [X] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
    - On rinkeby, where I actually had the problem (and the problem is gone)

### DIFF
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -306,7 +306,7 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
                     }
 
                     // check if confirmation listener exists
-                    if (defer.eventEmitter.listeners('confirmation').length > 0) {
+                    if (!canUnsubscribe || defer.eventEmitter.listeners('confirmation').length > 0) {
                         var block;
 
                         // If there was an immediately retrieved receipt, it's already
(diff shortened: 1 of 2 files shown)


Darkhogg:
Ok so apparently this also happens if you just... not subscribe to the `confirmation` event? I might spend some time tomorrow writing a proper set of test cases because I'm going insane right now.

jdevcs:
Thanks @Darkhogg did you get time of adding test and fixing merge conflicts.

avkos (approved):

nazarhussain (approved):

nazarhussain (changes requested):
@Darkhogg Please convert the draft PR to ready and then request the request again.

packages/web3-core-method/src/index.js
```diff
@@ -306,7 +306,7 @@ Method.prototype._confirmTransaction = function (defer, result, payload) {
                     }
 
                     // check if confirmation listener exists
-                    if (defer.eventEmitter.listeners('confirmation').length > 0) {
+                    if (!canUnsubscribe || defer.eventEmitter.listeners('confirmation').length > 0) {
```
jdevcs, line 309: Pending items are ( tests cases and fixing merge conflicts ). There is no activity on this PR from contributor so closing. @Darkhogg feel free to reopen and fix issues in this PR when have time.

PR #4877, opened 2022-03-21
Add `removed` property to `Log` type

samlaf:
Attempt at Fixing https://github.com/ChainSafe/web3.js/issues/4747#issue-1119059382

This is just trying to move forward in fixing this issue. I don't have the time to deep dive into web3js and try to understand how to make sure that this PR is non-breaking, so would like for someone else to help me verify.

I just added the property and ran `tsc --noEmit` in `packages/web3-core` which type checked.
However I am not sure if it breaks some of the other web3 packages, since when running `tsc --noEmit` in the root folder, I get a gazillion type errors from the tests, such as
```
packages/web3-utils/types/tests/string-to-hex-test.ts:40:13 - error TS2345: Argument of type 'boolean' is not assignable to parameter of type 'string'.

40 stringToHex(true);
```
which comes from
```
// $ExpectError
stringToHex(true);
```
Perhaps we could add `// @ts-ignore` above these lines, but I'm not sure if it wouldn't break the `// $ExpectError` commands (maybe only the line before the actual test applies?).


packages/web3-core/types/index.d.ts
```diff
@@ -242,6 +242,7 @@ export interface Log {
     transactionHash: string;
     blockHash: string;
     blockNumber: number;
+    removed: boolean;
```
jdevcs, line 245: @samlaf Thanks for your contribution, 
 Could you update [changelog](https://github.com/samlaf/web3.js/blob/patch-1/CHANGELOG.md) and mention this under 1.7.3

samlaf: Done. Please make sure I did it properly.

avkos (changes requested):
@samlaf Please resolve conflicts

nazarhussain (approved):
Please resolve the conflicts. Rest LGTM.

avkos (approved):

jdevcs:
Merged into 1.x via https://github.com/web3/web3.js/pull/5576

Issue #4556, opened 2021-11-18
Improve `formatParam` to not mutate the input


nazarhussain:
#### Expected behavior

No input parameters should be mutated. 

#### Actual behavior

Currently the `formatParam` function in `web3-eth-abi` package is mutating the input parameters. That can have side effects and may result in introducing bugs. Would be nice fix it. 

https://github.com/ChainSafe/web3.js/blob/d9726d41babe14bc823d5d2ea4e4af15ef5914ac/packages/web3-eth-abi/src/utils.ts#L120

https://github.com/ChainSafe/web3.js/blob/a1c7d71973ec17f9287fbea8939e64a80e589fc6/packages/web3-eth-abi/src/index.js#L249

#### Environment

1.x, 4.x

mconnelly8:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/ChainSafe/web3.js/4556?planning-poker) with ZenHub @jdevcs @luu-alex @nazarhussain @spacesailor24

PR #4882, opened 2022-03-22
4556 - improve formatparameter

luu-alex:
#4556 input does not mutate in formatparameter anymore


avkos (approved):

packages/web3-eth-abi/src/utils.ts
```diff
@@ -123,12 +123,18 @@ export const mapTypes = (
...
+export const formatOddHexstrings = (param: string): string =>
+	param.length % 2 === 1 ? `0x0${param.substring(2)}` : param;
+
 /**
  * Handle some formatting of params for backwards compatibility with Ethers V4
  */
 export const formatParam = (type: string, _param: unknown): unknown => {
 	// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-	let param = _param;
+	const param = _param;
```
jdevcs: More safe approach is to clone `_param` obj, as `param` is also returned in this function.

packages/web3-eth-abi/src/utils.ts
```diff
@@ -169,18 +173,16 @@ export const formatParam = (type: string, _param: unknown): unknown => {
...
+					? rightPad(param as string, size * 2)
+					: hexParam;
+			return formatOddHexstrings(paddedParam as string);
 		}
 
-		// format odd-length bytes to even-length
-		if ((param as string).length % 2 === 1) {
-			param = `0x0${(param as string).substring(2)}`;
-		}
+		return formatOddHexstrings(hexParam as string);
```
jdevcs, line 193: I think good to add 'input does not mutate' unit test.

nazarhussain: If input mutate semantically, the TS and ESLint will complaint.

packages/web3-eth-abi/src/utils.ts
```diff
@@ -123,12 +123,18 @@ export const mapTypes = (
 	return mappedTypes;
 };
 
+/**
+ * format odd-length bytes to even-length
+ */
+export const formatOddHexstrings = (param: string): string =>
+	param.length % 2 === 1 ? `0x0${param.substring(2)}` : param;
+
```
nazarhussain: As it's a utility function exported from package, so would be nice to validate the input it's really a hex string and odd lenght.

nazarhussain (changes requested):
Apart from @jdevcs feedback LGTM.

packages/web3-eth-abi/src/utils.ts
```diff
@@ -123,12 +123,26 @@ export const mapTypes = (
...
 /**
  * Handle some formatting of params for backwards compatibility with Ethers V4
  */
 export const formatParam = (type: string, _param: unknown): unknown => {
 	// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-	let param = _param;
+
+	// clone if _param is an object
+	const param = typeof _param === 'object' && !Array.isArray(_param) ? { ..._param } : _param;
```
jdevcs, line 145: @luu-alex  This PR LGTM, one observation: Even its array we should clone it I think. or is there any reason for not doing that?

nazarhussain (approved):

packages/web3-eth-abi/src/utils.ts
luu-alex: nevermind, i updated to cloning if its an array

luu-alex: https://github.com/ChainSafe/web3.js/blob/e8ce18b293ebb1f848d3d13ac4ed1ac5a1a108e8/packages/web3-eth-abi/src/utils.ts#L158 updated

jdevcs (approved):

PR #4895, opened 2022-03-28
move Subscription functions

avkos:
## Description

[4830](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/chainsafe/web3.js/4830)

## Type of change

## Checklist:


avkos:
change name of subscriptions related to this doc https://geth.ethereum.org/docs/rpc/pubsub :
pendingTransactions => newPendingTransactions
newBlockHeaders => newHeads

packages/web3-eth/src/web3_eth.ts
```diff
@@ -17,6 +17,13 @@ import * as rpcMethodsWrappers from './rpc_method_wrappers';
 import { Transaction, TransactionCall } from './types';
 import { Web3EthExecutionAPI } from './web3_eth_execution_api';
 
+enum SubscriptionNames {
+	logs = 'logs',
+	newPendingTransactions = 'newPendingTransactions',
```
```suggestion
	pendingTransactions = 'pendingTransactions',
```

as per 1.x

packages/web3-eth/src/web3_eth.ts
```diff
@@ -17,6 +17,13 @@ import * as rpcMethodsWrappers from './rpc_method_wrappers';
...
+	newHeads = 'newHeads',
```
```suggestion
	newBlockHeaders = 'newBlockHeaders',
```

jdevcs (changes requested):

packages/web3-eth/src/web3_eth.ts
avkos: I've described this:
change name of subscriptions related to this doc https://geth.ethereum.org/docs/rpc/pubsub :
pendingTransactions => newPendingTransactions
newBlockHeaders => newHeads

And, also in 1.x version, we have comments to rename these things

avkos: we can support both options. what do you think @jdevcs?

packages/web3-core/src/web3_context.ts
```diff
@@ -98,9 +104,23 @@ export class Web3Context<
...
 	}
 
+	private static getDefaultRegisteredSubscriptions(): any {
+		return {
+			logs: LogsSubscription,
+			newPendingTransactions: NewPendingTransactionsSubscription,
+			newHeads: NewHeadsSubscription,
+			syncing: SyncingSubscription,
+		};
+	}
```
nazarhussain: This could be false positive, as the user can pass `registeredSubscriptions` to the context, even user can pass `subscriptionManager`. So the values returned by this function will be false positive.

packages/web3-core/src/web3_context.ts
```diff
@@ -98,9 +104,23 @@ export class Web3Context<
 				this.requestManager,
 				registeredSubscriptions,
 			);
+		} else {
+			this._subscriptionManager = new Web3SubscriptionManager(
+				this.requestManager,
+				Web3Context.getDefaultRegisteredSubscriptions(),
+			);
```
nazarhussain: If you want to set as default behavior, use it as default value of `registeredSubscriptions`

nazarhussain: I would not recommend it though, every context should be responsible to register their own set of subscriptions.

nazarhussain (changes requested):

packages/web3-core/src/web3_context.ts
avkos: Thanks. You are right. I will add a setter for `registeredSubscriptions` and pass them from web3_eth package

packages/web3-eth/src/web3_eth.ts
avkos: I've added both options to support

packages/web3-eth/src/web3_eth.ts
avkos: I've added both options to support

packages/web3-core/src/web3_context.ts
```diff
@@ -135,6 +135,15 @@ export class Web3Context<
...
 	}
 
+	protected setRegisteredSubscriptions(registeredSubscriptions: RegisteredSubs): void {
+		if (!this.subscriptionManager) {
+			this._subscriptionManager = new Web3SubscriptionManager(
+				this.requestManager,
+				registeredSubscriptions,
+			);
+		}
+	}
```
nazarhussain: The type `RegisteredSubs` is generic and needs to be passed in the class definition. So does not seems correct to  specify type (which is subscription classes) at definition and actual values later on. The type `RegisteredSubs` is generic and needs to be passed in the class definition. So does not seems correct to  specify type (which is subscription classes) at definition and actual values later on.

packages/web3-core/src/web3_subscription_manager.ts
```diff
@@ -78,16 +84,17 @@ export class Web3SubscriptionManager<
...
 		}
-
+		const { id } = sub;
 		await sub.unsubscribe();
-		this._subscriptions.delete(sub.id);
+		this._subscriptions.delete(id);
 	}
 
-	public async unsubscribe() {
+	public async unsubscribe(condition?: ShouldUnsubscribeCondition) {
```
nazarhussain, line 92: The function `web3.eth.subscribe` returns a subscription object. If user want to clear that one particular subscription, he should use `unsubscribe` on that object. 

https://web3js.readthedocs.io/en/v1.5.0/web3-eth-subscribe.html#subscribe

On the Subscription Manager level, I would suggest to not add such behavior, it will be duplicate implementation and difficult to maintain in future.

packages/web3-core/src/web3_subscriptions.ts
```diff
@@ -73,28 +72,42 @@ export abstract class Web3Subscription<
...
-	public async unsubscribe() {
-		if (!this.id) {
-			return;
-		}
-
-		await this._requestManager.send({
-			method: 'eth_unsubscribe',
-			params: [this.id] as Web3APIParams<API, 'eth_unsubscribe'>,
-		});
+	public async unsubscribe(cb?: Callback) {
```
nazarhussain: We decided to remove the callback support from 4.x, so why it was not added earlier. Please revert these changes.

packages/web3-eth/src/web3_eth.ts
```diff
@@ -15,8 +20,45 @@ import * as rpcMethods from './rpc_methods';
...
+	public constructor(providerOrContext: SupportedProviders<any> | Web3ContextInitOptions) {
+		super(providerOrContext);
+		this.setRegisteredSubscriptions({
+			logs: LogsSubscription,
+			newPendingTransactions: NewPendingTransactionsSubscription,
+			newHeads: NewHeadsSubscription,
+			syncing: SyncingSubscription,
+			pendingTransactions: NewPendingTransactionsSubscription, // the same as newPendingTransactions. just for support API like in version 1.x
+			newBlockHeaders: NewHeadsSubscription, // the same as newHeads. just for support API like in version 1.x
+		});
```
nazarhussain: I am not sure if this works else where outside constructor considering the type safety. So having this method does not seems right. Would suggest to pass these to the `super` parameters.

packages/web3-eth/src/web3_eth.ts
```diff
@@ -249,4 +291,69 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI> {
...
+		if (typeof cb === 'function') {
+			return { options: args as LogArguments, callBack: cb };
+		}
+		if (typeof args === 'function') {
+			return { options: {} as LogArguments, callBack: args };
+		}
+		return {
+			options: args as LogArguments,
+		};
+	}
```
nazarhussain: Move this as utility function outside class.

packages/web3-eth/src/web3_eth.ts
```diff
@@ -249,4 +291,69 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI> {
...
+
+	public async subscribe(
+		name: keyof typeof SubscriptionNames,
+		args?: LogArguments | Callback | undefined,
+		cb?: Callback,
+	): Promise<SomeSubscription> {
```
nazarhussain: This will not return `SomeSubscription`, instead exactly the one matching with the `name`.

packages/web3-eth/src/web3_eth.ts
```diff
@@ -249,4 +291,69 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI> {
...
+			return { options: {} as LogArguments, callBack: args };
+		}
+		return {
+			options: args as LogArguments,
+		};
+	}
+
+	public async subscribe(
+		name: keyof typeof SubscriptionNames,
+		args?: LogArguments | Callback | undefined,
```
nazarhussain: We decided to remove the support for callbacks from `4.x`, so why its' not implemented in subscription manager and subscription class. Please remove this support from here as well.

nazarhussain: Don't need to specify new type `LogArguments` here, it's not typesafe.  The arguments must match the exact type of the constructor parameters of subscription class matching to the `name`. 

Please see the reference implementation of `subscribe` for the `subscription manager`. In here you just need to proxy that function, no need of any implementation.

packages/web3-eth/src/web3_eth.ts
```diff
@@ -249,4 +291,69 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI> {
...
+		}
+
+		return subscription as SomeSubscription;
+	}
+
+	private static shouldClearSubscription({ sub }: { sub: unknown }): boolean {
+		return !(sub instanceof SyncingSubscription);
+	}
+
+	public clearSubscriptions(notClearSyncing = false): Promise<void[]> | undefined {
```
nazarhussain, line 321: I don't see any reference of this parameter in the docs

https://web3js.readthedocs.io/en/v1.5.0/web3-eth-subscribe.html#clearsubscriptions

nazarhussain (changes requested):

packages/web3-core/src/web3_subscriptions.ts
avkos: I don't know this. And make all things by documentation. But I absolutely agree with you :)

avkos: Did we also decide to remove callback from `subscribe` method? @nazarhussain

packages/web3-eth/src/web3_eth.ts
avkos: please read here https://web3js.readthedocs.io/en/v1.7.1/web3-eth-subscribe.html#clearsubscriptions

packages/web3-eth/src/web3_eth.ts
```diff
@@ -249,4 +283,45 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI> {
...
+					.then(logs => {
+						for (const log of logs) {
+							subscription._processSubscriptionResult(log);
+						}
+					})
+					.catch(e => {
+						subscription._processSubscriptionError(e as Error);
+					});
+			});
+		}
```
nazarhussain, line 313: Can we do this inside `LogsSubscription` implementation? If yes may be a separate issue will be good to refactor it.

nazarhussain (approved):

packages/web3-eth/src/web3_eth.ts
avkos: yes we can move into LogsSubscription, but in this case we need to have duplicate of implementation `getPastLogs` method inside LogsSubscription

tools/isomorphic-ws.js
```diff
@@ -0,0 +1,16 @@
+var ws = null;
+
+if (typeof WebSocket !== 'undefined') {
```
jdevcs, line 3: Isomorphic lib should automatically switch to ws in case of node.js and to global.WebSocket for browsers.

luu-alex (approved):

tools/isomorphic-ws.js
jdevcs, line 3: Isomorphic lib should automatically switch to ws in case of node.js and to global.WebSocket for browsers. Why we need to add this logic explicitly here.

avkos: it is just a copy of isomorphic lib. but I've just added this line `module.exports.WebSocket = ws;`
because isomorphic for NodeJS and for webrowser has different external interface
for browser 
`import WebSocket from 'isomorphic'`
for nodejs
`import {WebSocket} from 'isomorphic'`

package.json
```diff
@@ -37,7 +37,8 @@
 		"ganache": "npx ganache -m 'smart guide what forget tired jungle always expire rescue boring glue champion' --account=\"0x4c3758228f536f7a210f8936182fb5b728046970b8e3215d0b5cb4c4faae8a4e,100000000000000000000\" --account=\"0x34aeb1f338c17e6b440c189655c89fcef148893a24a7f15c0cb666d9cf5eacb3,100000000000000000000\"",
 		"ganache-background": "yarn ganache > /dev/null & client=$! npx wait-port 8545",
 		"geth:build": "docker build ./.github -t gethnode -f .github/Dockerfile",
-		"geth": "docker run -d -p 8545:8545 gethnode --datadir data --networkid 15 --http.addr 0.0.0.0 --http.port 8545 --http",
+		"geth": "docker run -p 8545:8545 gethnode --datadir data --networkid 15 --http --http.addr 0.0.0.0 --http.port 8545 --ws --ws.addr 0.0.0.0 --ws.port 8545",
```
jdevcs:
```suggestion
		"geth": "docker run -p 8545:8545 -p 8546:8546 gethnode --datadir data --networkid 15 --http --http.addr 0.0.0.0 --http.port 8545 --ws --ws.addr 0.0.0.0 --ws.port 8546 --http.api personal,web3,eth,admin,debug,miner,txpool,net --ws.api personal,web3,eth,admin,debug,miner,txpool,net --nodiscover",
```

package.json
```diff
@@ -37,7 +37,8 @@
...
+		"geth-background": "docker run -d -p 8545:8545 gethnode --datadir data --networkid 15 --http --http.addr 0.0.0.0 --http.port 8545 --ws --ws.addr 0.0.0.0 --ws.port 8545",
```
jdevcs:
```suggestion
		"geth-background": "docker run -d -p 8545:8545 -p 8546:8546 gethnode --datadir data --networkid 15 --http --http.addr 0.0.0.0 --http.port 8545 --ws --ws.addr 0.0.0.0 --ws.port 8546 --http.api personal,web3,eth,admin,debug,miner,txpool,net --ws.api personal,web3,eth,admin,debug,miner,txpool,net --nodiscover",
```

jdevcs (changes requested):

jdevcs (approved):

PR #4903, opened 2022-03-29
Fixes calling toHexString on undefined gasPrice

### LINKED ISSUE #4902: handleRevert implementation broken when no value is specified for gasPrice on eth_sendRawTransaction

### Current Behavior

`checkConfirmation` in `web3-core-method/src/index.js` on line `458` ([here](https://github.com/ChainSafe/web3.js/blob/d46d922e2d9714b6e8bd72db665f708f12e222ea/packages/web3-core-method/src/index.js#L458)) throws an internal error because it blindly accesses `parsedTx.gasPrice` to call its `.toHexString()`, but this property is undefined for all EIP-1559 transactions. This is then captured by an internal try/catch, swallowed and turned into `TransactionRevertedWithoutReasonError` and the app receives no `reason`.

### Expected Behavior

I expect the logic to be aware of the fact that `gasPrice` will be `undefined` for EIP-1559 transactions and not call `.toHextString()` on it in that case so that it doesn't cause an internal error that stops fetching the `reason` for the failure from the RPC node with `eth_call`.

### Steps to Reproduce

1. Set `web3.eth.handleRevert = true`
2. Send a transaction that should return a revert reason through the `eth_sendRawTransaction`
3. Catch the error
4. Notice it has no `reason` when it should

### Web3.js Version

1.7.1

### Environment

Shouldn't matter, but I'm on Node.js Version 16.14.0.

### Anything Else?

Simply adding the null propagation operator like so is all that is needed to fix:

```JavaScript
gasPrice: parsedTx.gasPrice?.toHexString(),
```

I have tested this by manually editing the version in `lib` locally and everything works as expected then. If trying to be backwards compatible and not use null propagation operator the following fixes as well:

```JavaScript
gasPrice: parsedTx.gasPrice ? parsedTx.gasPrice.toHexString() : undefined,
```

drub0y:
## Description

Fixes bug where, when `handleRevert` was enabled and the call was to `eth_sendRawTransaction`, the logic would blindly call `parsedTx.gasPrice.toHexString()`. However, for EIP-1559 calls, `gasPrice` will always be `undefined` and this code will now throw an internal exception which is caught and then rethrow an error that indicates that the reason couldn't be retrieved.

Fixes #4902

## Type of change

## Checklist:

- [ ] ~I have commented my code, particularly in hard-to-understand areas.~ N/A
- [ ] ~I have made corresponding changes to the documentation.~ N/A
- [ ] ~Any dependent changes have been merged and published in downstream modules.~ N/A
- [x] I ran `npm run build` ~and tested `dist/web3.min.js` in a browser.~


nikoulai (approved):

CHANGELOG.md
```diff
@@ -534,3 +534,4 @@ Released with 1.0.0-beta.37 code base.
 ### Fixed
 -  Fix jsonrpc payload and response types (#4743) (#4761)
 -  Allowed more flexibility in typing the overly constrained `provider.disconnect` function (#4833)
+-  Fix bug in handleRevert logic for eth_sendRawTransaction (#4902)
```
jdevcs, line 549: @drub0y could you add this under 1.7.3. Thanks

drub0y: Sure, done!

jdevcs (approved):

avkos (approved):

jdevcs:
This PR is merged into 1.x via https://github.com/ChainSafe/web3.js/pull/4940

PR #4904, opened 2022-03-30
Sign transaction with available wallet

### LINKED ISSUE #4814: Add Use of Local Wallet Support to Non-`web3-eth-account` Packages
Various packages (namely `web3-eth`) need to check whether a local wallet has been instantiated so a `transaction.from` can be defaulted and a private key can be accessed to sign the transaction e.g.:

- [x] [One](https://github.com/ChainSafe/web3.js/blob/96026ec5990cd01a6ccfe33daa68b67de50bd1ad/packages/web3-eth/src/rpc_method_wrappers.ts#L385)
- [x] [Two](https://github.com/ChainSafe/web3.js/blob/96026ec5990cd01a6ccfe33daa68b67de50bd1ad/packages/web3-eth/src/rpc_method_wrappers.ts#L459)

nazarhussain:
## Description

Use wallet support to sign transaction. 

Fixes #4814

## Type of change

## Checklist:


Spacesai1or (commented):
I think this looks good so far, we just need to add support for passing an index for `address` in `sendTransaction`

avkos (approved):

packages/web3-common/src/web3_base_wallet.ts
```diff
@@ -4,15 +4,32 @@ export interface Web3BaseWalletAccount {
...
 	readonly address: string;
 	readonly privateKey: string;
-	readonly signTransaction: (tx: Record<string, unknown>) => string;
-	readonly sign: (data: Record<string, unknown> | string) => string;
-	readonly encrypt: (password: string, options?: Record<string, unknown>) => Web3EncryptedWallet;
+	readonly signTransaction: (tx: Record<string, unknown>) => {
+		readonly messageHash: string;
+		readonly r: string;
+		readonly s: string;
+		readonly v: string;
```
jdevcs, line 11: Good to create type and reuse that in sign function return type as well.

jdevcs: Good to create type and reuse that in sign function return type as well.

jdevcs (dismissed):

packages/web3-common/src/web3_base_wallet.ts
Spacesai1or: Could use [HexString](https://github.com/ChainSafe/web3.js/blob/bfae214735803e38ba6bddabcf43f241d715c5d2/packages/web3-utils/src/types.ts#L1) from `web3-utils` as it's already being imported in this package

Spacesai1or (approved):

PR #4910, opened 2022-04-01
Main web3 package - 4.0.0 Alpha.1 System Tests

### LINKED ISSUE #4851: Main web3 package - 4.0.0 Alpha.1 System Tests
System tests for:

## Main web3 package

- [x] Using http, websocket, IPC provider strings create web3 instance
- [x] Using external connection provider create web3 instance
- [x] Using external wallet provider create web3 instance
- [x] setProvider
- [ ] List providers
- [ ] given Provider
- [ ] Current Provider
- [x] BatchRequests
- [ ] All modules instantiation

Users are able to query via eth package and are able to get Tx signed via Accounts package with web3 instances created with above mentioned providers.

#### Min Platform Requirements for 4.x Alpha:
- Node: v14, v16
- Browsers: Firefox, Brave
- Eth Clients: Geth, Ganache

More details are in [Epic](https://github.com/ChainSafe/web3.js/issues/4711)

nikoulai:
## Description

#4851 

## Type of change

## Checklist:


.github/Dockerfile
```diff
@@ -1,5 +1,8 @@
 # syntax=docker/dockerfile:1
 FROM ethereum/client-go
 RUN echo '{"config":{"chainId":15,"homesteadBlock":0,"eip150Block":0,"eip155Block":0,"eip158Block":0,"byzantiumBlock":0,"constantinopleBlock":0,"petersburgBlock":0,"istanbulBlock":0,"berlinBlock":0,"londonBlock":0},"alloc":{"0xdc6bad79dab7ea733098f66f6c6f9dd008da3258":{"balance":"100000000000000000000"},"0x962f9a9c2a6c092474d24def35eccb3d9363265e":{"balance":"100000000000000000000"}},"coinbase":"0x0000000000000000000000000000000000000000","difficulty":"0x20000","gasLimit":"0x2fefd8","nonce":"0x0000000000000042","mixhash":"0x0000000000000000000000000000000000000000000000000000000000000000","parentHash":"0x0000000000000000000000000000000000000000000000000000000000000000","timestamp":"0x00"}' > /genesis.json
-RUN geth init  --datadir data /genesis.json
+RUN geth init  --datadir data /genesis.json 
 EXPOSE 8545
+EXPOSE 8546
```
avkos: you can just remove all expose ports here. all necessary ports expose in ethereum/client-go

.github/workflows/build.yml
```diff
@@ -60,10 +60,10 @@ jobs:
             -   run: yarn
 
             -   if: (env.TEST_CMD == 'e2e_ganache')
-                run: yarn ganache-background && yarn test:integration
+                run: WSPORT=8545 yarn ganache-background && yarn test:integration:ganache
 
             -   if: (env.TEST_CMD == 'e2e_geth')
-                run: yarn geth:build && yarn geth && yarn test:integration
+                run: WSPORT=8546 yarn geth:install-ubuntu && yarn geth-local && yarn test:integration
```
avkos: use the same port for the websocket. 8545

package.json
```diff
@@ -36,39 +36,51 @@
 		"clean": "lerna run clean",
 		"ganache": "npx ganache -m 'smart guide what forget tired jungle always expire rescue boring glue champion' --account=\"0x4c3758228f536f7a210f8936182fb5b728046970b8e3215d0b5cb4c4faae8a4e,100000000000000000000\" --account=\"0x34aeb1f338c17e6b440c189655c89fcef148893a24a7f15c0cb666d9cf5eacb3,100000000000000000000\"",
 		"ganache-background": "yarn ganache > /dev/null & client=$! npx wait-port 8545",
+		"geth:install-ubuntu": "sudo add-apt-repository -y ppa:ethereum/ethereum && sudo apt-get update && sudo apt-get install ethereum && geth init --datadir .github/geth/data `pwd`/.github/geth/genesis.json && sudo apt install screen",
 		"geth:build": "docker build ./.github -t gethnode -f .github/Dockerfile",
-		"geth": "docker run -d -p 8545:8545 gethnode --datadir data --networkid 15 --http.addr 0.0.0.0 --http.port 8545 --http",
+		"geth-local": "screen -dmS geth geth --datadir .github/geth/data --networkid 15 --http.addr 0.0.0.0 --http.port 8545 --http --ws --ws.addr 0.0.0.0 --ws.port 8546 --ws.origins '*' </dev/null &>/dev/null &",
```
avkos: use the same port for ws and http

avkos (changes requested):

nikoulai:
It seems that there is a problem in docker exposing unix sockets, so I had to add scripts to install geth locally.
https://github.com/docker/for-mac/issues/483
https://github.com/ethereum/go-ethereum/issues/17907
Also, ganache doesn't support ipc, for that I had to add the filter option in jest, to ignore all tests files regarding ipc
https://github.com/trufflesuite/ganache/issues/759

jdevcs (commented):
some of e2e tests needs to be be fixed.

package.json
```diff
@@ -36,39 +36,51 @@
 		"clean": "lerna run clean",
 		"ganache": "npx ganache -m 'smart guide what forget tired jungle always expire rescue boring glue champion' --account=\"0x4c3758228f536f7a210f8936182fb5b728046970b8e3215d0b5cb4c4faae8a4e,100000000000000000000\" --account=\"0x34aeb1f338c17e6b440c189655c89fcef148893a24a7f15c0cb666d9cf5eacb3,100000000000000000000\"",
 		"ganache-background": "yarn ganache > /dev/null & client=$! npx wait-port 8545",
+		"geth:install-ubuntu": "sudo add-apt-repository -y ppa:ethereum/ethereum && sudo apt-get update && sudo apt-get install ethereum && geth init --datadir .github/geth/data `pwd`/.github/geth/genesis.json && sudo apt install screen",
 		"geth:build": "docker build ./.github -t gethnode -f .github/Dockerfile",
-		"geth": "docker run -d -p 8545:8545 gethnode --datadir data --networkid 15 --http.addr 0.0.0.0 --http.port 8545 --http",
+		"geth-local": "screen -DmS geth geth --datadir .github/geth/data --networkid 15 --http.addr 0.0.0.0 --http.port 8545 --http --ws --ws.addr 0.0.0.0 --ws.port 8545 --ws.origins '*' --ipcpath=./.github/geth/data/geth.ipc </dev/null &>/dev/null &",
```
nazarhussain: Why we need to attach/reattach a session? Every CI/CD task should have an isolated scope.

nazarhussain: WS port should not be different than http port?

package.json
```diff
@@ -36,39 +36,51 @@
...
 		"test:unit": "lerna run test:unit --stream --parallel",
 		"test:integration": "lerna run test:integration --stream --parallel",
 		"test:integration:browsers": "./node_modules/.bin/karma start karma.config.js",
+		"test:integration:ganache": "WSPORT=8545 lerna run test:integration",
 		"prepare": "husky install",
 		"postinstall": "yarn build"
 	},
 	"devDependencies": {
 		"@types/node": "*",
+		"assert": "^2.0.0",
```
nazarhussain: Where this package is used?

packages/web3-core/src/web3_context.ts
```diff
@@ -78,6 +79,7 @@ export class Web3Context<
 	private _subscriptionManager?: Web3SubscriptionManager<API, RegisteredSubs>;
 	private _accountProvider?: Web3AccountProvider<Web3BaseWalletAccount>;
 	private _wallet?: Web3BaseWallet<Web3BaseWalletAccount>;
+	private static readonly _batchRequest = Web3BatchRequest;
```
nazarhussain: A static but private? Can we not simply use the `Web3BatchRequest` directly?

nazarhussain (changes requested):
There are a lot of tests cases which are commented. If these are not necessary then please remove these. If you plan to implement those later then use `it.skip` on those and remove the comments.

package.json
nikoulai: I need somehow to start geth on the background and continue with the tests https://github.com/ChainSafe/web3.js/blob/a71c534d57f5cccadd7fe9f949fa0444d6b5b709/.github/workflows/build.yml#L66

nikoulai: both ganache and geth can run ws and HTTP on the same port, just for simplicity. Should I undo the change?
@avkos

package.json
nikoulai: removed

packages/web3-core/src/web3_context.ts
nikoulai: I changed it and exposed Web3BatchRequest directly

package.json
avkos: yes it works on the same port because it is a different protocol

avkos: I've added both starts local and background
https://github.com/ChainSafe/web3.js/pull/4895/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519
```json
"geth": "docker run -p 8545:8545 gethnode --datadir data --networkid 15 --http --http.addr 0.0.0.0 --http.port 8545 --ws --ws.addr 0.0.0.0 --ws.port 8545",
"geth-background": "docker run -d -p 8545:8545 gethnode --datadir data --networkid 15 --http --http.addr 0.0.0.0 --http.port 8545 --ws --ws.addr 0.0.0.0 --ws.port 8545",

Comment (avkos) [reply package.json]: ah.you need all this of ipc provider....

File: packages/web3/package.json [hunk 7]
```diff
@@ -30,16 +30,18 @@
...
 	"devDependencies": {
+		"@truffle/hdwallet-provider": "^2.0.4",
 		"@types/jest": "^27.0.3",
 		"@typescript-eslint/eslint-plugin": "^5.4.0",
 		"@typescript-eslint/parser": "^5.4.0",
 		"eslint": "^8.3.0",
 		"eslint-config-prettier": "^8.3.0",
 		"eslint-config-web3-base": "0.1.0",
 		"eslint-plugin-import": "^2.25.3",
+		"isomorphic-ws": "^4.0.1",
```
Spacesai1or: I'm confused why this is needed for this package, it doesn't seem to be used in these changes (assuming the reason is related to `tools/isomorphic-ws.js`)?

packages/web3/test/integration/external-string-providers/web3_instance_http_string.test.ts
```diff
@@ -0,0 +1,12 @@
+import { externalHttpsStringProvider } from '../../fixtures/config';
+
+import { Web3 } from '../../../src/index';
+
+describe('Web3 instance', () => {
+	describe('Create Web3 class instance with string providers', () => {
+		it('should create instance with string of http provider', async () => {
+			const web3 = new Web3(externalHttpsStringProvider);
+			expect(web3).toBeInstanceOf(Web3);
```
Spacesai1or:
```suggestion
			expect(web3).toBeInstanceOf(Web3);
			expect(web3.provider).toBe(externalHttpsStringProvider);
```

nazarhussain:
@nikoulai Please resolve the conflicts on this PR to review.

karma.config.js
```diff
@@ -76,22 +108,18 @@ module.exports = function (config) {
...
-		// list of files / patterns to load in the browser
-		// Here I'm including all of the the Jest tests which are all under the __tests__ directory.
-		// You may need to tweak this patter to find your test files/
 		files: listOfTests,
-		// preprocess matching files before serving them to the browser
-		// available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
+		exclude: ['**/web3_instance_ipc_string.test.ts'],
 		preprocessors: {
-			// Use webpack to bundle our tests files
+			'./karma-setup.js': ['webpack'],
```
avkos: I think, Should be `'./karma.setup.js': ['webpack', 'browserify'],`

avkos (changes requested):

packages/web3/package.json
nikoulai: You are right, I copied it from another branch and I thought it was required, but, it's not. I'll remove it. @avkos could you please verify it's not needed?

packages/web3/test/integration/external-string-providers/web3_instance_http_string.test.ts
nikoulai: `externalHttpsStringProvider` is just the string of the node (`localhost:8545` in our case`). I believe this won't work or do I miss something? Maybe, I have to change the name of the variable.

packages/web3/test/fixtures/config.ts
```diff
@@ -0,0 +1,41 @@
...
+		address: '0x962f9a9c2a6c092474d24def35eccb3d9363265e',
+		privateKey: '0x34aeb1f338c17e6b440c189655c89fcef148893a24a7f15c0cb666d9cf5eacb3',
+		balance: '100',
+	},
+];
+export const httpStringProvider = 'http://localhost:8545';
+export const wsStringProvider = 'ws://localhost';
+export const ipcStringProvider = '../../../../../.github/geth/data/geth.ipc';
+export const externalHttpsStringProvider =
+	'https://mainnet.infura.io/v3/6120b2e72c304b4eadafe2bf33862ac4';
```
luu-alex: infura keys shouldn't be exposed

packages/web3/test/fixtures/system_test_utils.ts
```diff
@@ -0,0 +1,115 @@
+/*
+This file is part of web3.js.
```
avkos: You do not need to copy this file. Just create a symlink (like packages/web3-eth/test/fixtures/system_test_utils.ts). the fastest way is just to copy symlink from another package and paste it to your package, but you should use Finder for this

packages/web3/test/fixtures/config.ts
avkos: So if it is committed once - it is saved in git forever. So now you have only one way. Remove and create a new one.

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,249 @@
...
+			});
+
+			it('should create instance with string of external ws provider', async () => {
+				web3 = new Web3(externalWssStringProvider);
+				expect(web3).toBeInstanceOf(Web3);
+				await waitForSocketState(web3, 'connected');
+			});
+		},
+	);
+	describeIf(getSystemTestProvider().startsWith('http'))('Web3 providers', () => {
```
avkos: looks like this test is not related to the type of provider. HTTP or WS

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,249 @@
...
+					sign: expect.any(Function),
+					recover: expect.any(Function),
+					encrypt: expect.any(Function),
+					decrypt: expect.any(Function),
+				}),
+			);
+			const erc20Contract = new web3.eth.Contract(erc20Abi);
+			expect(erc20Contract).toBeInstanceOf(Contract);
+		});
+		// it.skip('able to query via eth pacakge', async () => {
```
avkos: Why this test is commented on?

avkos (changes requested):

.github/geth/genesis.json
```diff
@@ -0,0 +1,26 @@
...
+		"0x962f9a9c2a6c092474d24def35eccb3d9363265e": { "balance": "100000000000000000000" }
+	},
+	"coinbase": "0x0000000000000000000000000000000000000000",
+	"difficulty": "0x20000",
+	"gasLimit": "0x2fefd8",
+	"nonce": "0x0000000000000042",
+	"mixhash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+	"parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+	"timestamp": "0x00"
+}
```
nazarhussain: Do we really need this file now?

.github/setup.sh
```diff
@@ -0,0 +1,34 @@
...
+    # Make sure the folder exists.
+    mkdir /var/run/uwsgi
+    # Set owner.
+    chown root:uwsgi /var/run/uwsgi
+    # Set permissions.
+    chmod u=rwX,g=rwX,o=--- /var/run/uwsgi
+}
+
+
+main "$@"
```
nazarhussain: Same for this file? Do we really need these files now?

jest.config.js
```diff
@@ -2,20 +2,19 @@ const path = require('path');
 const { lstatSync, readdirSync } = require('fs');
```
nazarhussain, line 19: Please add license headers to this file.

packages/web3-core/src/web3_context.ts
```diff
@@ -235,6 +236,10 @@ export class Web3Context<
 	public set currentProvider(provider: SupportedProviders<API>) {
 		this.requestManager.setProvider(provider);
 	}
+
+	public get BatchRequest() {
+		return Web3BatchRequest.bind(null, this._requestManager as unknown as Web3RequestManager);
```
nazarhussain, line 250: It should be a strong type `Web3RequestManager<API>`

packages/web3/package.json
```diff
@@ -30,16 +30,18 @@
 		"test:ci": "jest --coverage=true --coverage-reporters=json --verbose",
 		"test:watch": "npm test -- --watch",
 		"test:unit": "jest --config=./test/unit/jest.config.js",
-		"test:integration": "jest --config=./test/integration/jest.config.js --runInBand --passWithNoTests"
+		"test:integration": "jest --config=./test/integration/jest.config.js --runInBand --forceExit"
```
nazarhussain, line 33: Try if you can avoid using `--forceExit` by properly closing the socket connections. This options sometimes hides floating promises.

packages/web3/package.json
nazarhussain: Yes it's not required as dependency in this package.

packages/web3/test/fixtures/config.ts
```diff
@@ -0,0 +1,41 @@
...
+		balance: '100',
+	},
+	{
+		address: '0x962f9a9c2a6c092474d24def35eccb3d9363265e',
+		privateKey: '0x34aeb1f338c17e6b440c189655c89fcef148893a24a7f15c0cb666d9cf5eacb3',
+		balance: '100',
+	},
+];
+export const httpStringProvider = 'http://localhost:8545';
+export const wsStringProvider = 'ws://localhost';
```
nazarhussain: Try not to use these hardcoded values. Instead use the utility function `getSystemTestProvider()`

packages/web3/test/fixtures/config.ts
nazarhussain: If it's in your branch, you can rewrite the commits and force-push to change the history.

packages/web3/test/fixtures/erc20.ts
```diff
@@ -0,0 +1,238 @@
...
+
+web3.js is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+export const erc20Abi = [
```
nazarhussain: The `erc20` and `erc721` abis and other related stuff is defined in shared fixtures. Link the root level `fixtures` directory to this package as `shared_fixtures` and use these values from there.

packages/web3/test/fixtures/system_test_utils.ts
nazarhussain: Just create a sym-link from root level `fixtures` directory or copy the link from some other package.

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,249 @@
...
+			}
+			afterEach(async () => {
+				(web3.currentProvider as Web3BaseProvider).disconnect(1000, 'done');
+				await waitForSocketState(web3, 'disconnected');
+			});
+
+			it('should create instance with string of ws provider', async () => {
+				web3 = new Web3(clientUrl);
+				expect(web3).toBeInstanceOf(Web3);
+				await waitForSocketState(web3, 'connected');
```
nazarhussain: Why do we need to wait for connection in this test case?

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,249 @@
...
+			});
+
+			it('should create instance with string of external ws provider', async () => {
+				web3 = new Web3(externalWssStringProvider);
+				expect(web3).toBeInstanceOf(Web3);
+				await waitForSocketState(web3, 'connected');
```
nazarhussain: Same for here.

tools/filter-tests.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+module.exports = testPaths => {
```
nazarhussain: I think we don't this file.

nazarhussain (changes requested):

packages/web3/test/fixtures/config.ts
nikoulai: it's a node of mine

packages/web3/test/fixtures/erc20.ts
nikoulai: Currently, they are not. They exist only in `web3-eth-contract`, should I move them to `fixtures` and create symlinks in both packages symlinks? Or I miss something?

packages/web3/test/fixtures/config.ts
luu-alex: Yes it is a node of yours, but your node is exposed on github on a huge project and may lead to vulnerabilities if someone decides to use your node. For example if a bad actor maliciously uses your node to continuosly spam requests, they may hit your infura limit and our testcases using your node will fail.
In this case we use github repo [secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets#accessing-your-secrets) so switch out your personal infura node with the github repo secret

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,249 @@
...
+	});
+	describeIf(getSystemTestProvider().startsWith('http'))(
+		'Create Web3 class instance with http string providers',
+		() => {
+			it('should create instance with string provider', async () => {
+				const web3 = new Web3(clientUrl);
+				expect(web3).toBeInstanceOf(Web3);
+			});
+			it('should create instance with string of external http provider', async () => {
+				const web3 = new Web3(externalHttpsStringProvider);
```
jdevcs: we should use `INFURA_HTTP` here as discussed like we are using in [1.x](https://github.com/ChainSafe/web3.js/blob/2c0324af2da467ee1acb72452f20000e916e4306/scripts/js/basic_usage.js#L50)

Following nodes are avalible for your tests:
` INFURA_HTTP` `INFURA_WS`  : mainnet
` INFURA_GOERLI_HTTP` `INFURA_GOERLI_WS` : goerli nw

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,249 @@
...
+				expect(web3).toBeInstanceOf(Web3);
+			});
+			// todo fix ipc test
+			// https://ethereum.stackexchange.com/questions/52574/how-to-connect-to-ethereum-node-geth-via-ipc-from-outside-of-docker-container
```
jdevcs, line 106: we can skip test for instantiation with IPC provider now in alpha release, create a new issue for that.

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,249 @@
...
+					recover: expect.any(Function),
+					encrypt: expect.any(Function),
+					decrypt: expect.any(Function),
+				}),
+			);
+			const erc20Contract = new web3.eth.Contract(erc20Abi);
+			expect(erc20Contract).toBeInstanceOf(Contract);
+		});
+		// it.skip('able to query via eth pacakge', async () => {
+		// 	const web3 = new Web3(httpStringProvider);
```
jdevcs: @nikoulai for remaining tests you may use similar https://github.com/ChainSafe/web3.js/pull/5015/files#diff-865c618d6dea5b91d2bc10ea4ea0491a39278e97532528ec5c33e56bc198f79bR68 via main web3 package instead of specific packages, for checking list providers given provider current providers

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,303 @@
...
+			it('should create instance with string of ws provider', async () => {
+				web3 = new Web3(clientUrl);
+				expect(web3).toBeInstanceOf(Web3);
+			});
+
+			itIf(
+				process.env.INFURA_GOERLI_WS
+					? process.env.INFURA_GOERLI_WS.toString().includes('ws')
+					: false,
+			)('should create instance with string of external http provider', async () => {
```
nazarhussain:
```suggestion
			)('should create instance with string of external ws provider', async () => {
```

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,303 @@
...
+
+			expect(web3.provider).toBe(newProvider);
+		});
+
+		it('providers', async () => {
+			const res = Web3.providers;
+
+			expect(res.HttpProvider).toBeDefined();
+			expect(res.WebsocketProvider).toBeDefined();
+			expect(res.IpcProvider).toBeDefined();
```
nazarhussain: Please match these wit exact classes e.g. `WebsocketProvider`, `HttpProvider`

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,303 @@
...
+					encrypt: expect.any(Function),
+					decrypt: expect.any(Function),
+				}),
+			);
+			const basicContract = new web3.eth.Contract(BasicAbi);
+			expect(basicContract).toBeInstanceOf(Contract);
+		});
+	});
+
+	describeIf(getSystemTestProvider().startsWith('http'))(
```
nazarhussain: Is this necessary to only try with `http`?

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,303 @@
...
+					encodeEventSignature: expect.any(Function),
+					encodeFunctionCall: expect.any(Function),
+					encodeFunctionSignature: expect.any(Function),
+					encodeParameter: expect.any(Function),
+					encodeParameters: expect.any(Function),
+					decodeParameter: expect.any(Function),
+					decodeParameters: expect.any(Function),
+					decodeLog: expect.any(Function),
+				}),
+			);
```
nazarhussain: Such expectations better match with the unit tests. For integration would be better to test the user expectations from these functions.

nazarhussain (changes requested):

avkos (approved):

packages/web3-common/src/deferred_promise.ts
```diff
@@ -15,16 +15,9 @@ You should have received a copy of the GNU Lesser General Public License
 along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+/* eslint-disable max-classes-per-file */
```
nazarhussain: No need for this comment anymore. 

```suggestion
```

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,303 @@
...
+				id: 11,
+				method: 'eth_getBalance',
+				params: [accounts[1], 'latest'],
+			};
+		});
+
+		it('should execute batch requests', async () => {
+			web3 = new Web3(clientUrl);
+
+			// eslint-disable-next-line @typescript-eslint/no-floating-promises, @typescript-eslint/no-unsafe-call
```
nazarhussain: Please don't disable `floating-promises` lint warnings. These are really evil in JS code and can have unseen side effects.

scripts/system_tests_utils.ts
```diff
@@ -45,7 +45,7 @@ export const getSystemTestMnemonic = (): string => getEnvVar('WEB3_SYSTEM_TEST_M
 
 export const getSystemTestBackend = (): string => getEnvVar('WEB3_SYSTEM_TEST_BACKEND') ?? '';
 
-const accounts: {
+export const accounts: {
```
nazarhussain: Don't export this member, rather use `getSystemTestAccounts` helper.

nazarhussain (changes requested):

.github/setup.sh
nazarhussain: I don't see the usage of this file anywhere. Please remove this file.

nazarhussain (changes requested):

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -0,0 +1,301 @@
...
+	});
+
+	describe('Module instantiations', () => {
+		it('should create contract', () => {
+			const basicContract = new web3.eth.Contract(BasicAbi);
+			expect(basicContract).toBeInstanceOf(Contract);
+		});
+	});
+
+	describe('Batch Request', () => {
```
jdevcs, line 203: Batch Request tests are not passing, if there is any bug in batch request, could you take out this from this PR and create new issue for testing Batch Request in new issue + PR.

nazarhussain (approved):

jdevcs (approved):

PR #4915, opened 2022-04-03
:sparkles: Add provider support to contract to set provider 

### LINKED ISSUE #4580: eth.Contract objects are never garbage collected
Looking at the code here:
https://github.com/ChainSafe/web3.js/blob/a1c7d71973ec17f9287fbea8939e64a80e589fc6/packages/web3-eth/src/index.js#L332

It looks like all eth.Contract objects reference each other in a chain of setProvider calls. This means eth.Contract objects can not be garbage collected and will just cause a slow memory leak if you have a code that creates them over time. 

Here is an example from a heapdump that shows how setProvider references nest:
image

Also, I suspect if you create lots of eth.Contract objects (maybe 100k, maybe 1million), setProvider call will probably just cause a stack overflow error.

nazarhussain:
## Description

Add support to set providers for multiple contracts. 

Fixes #4580

## Type of change

## Checklist:


nazarhussain:
Please read this thread while reviewing this PR. https://github.com/ChainSafe/web3.js/issues/4580#issuecomment-1086854311

avkos (approved):

packages/web3/src/index.ts
```diff
@@ -61,6 +61,9 @@ export class Web3 extends Web3Context<EthExecutionAPI> {
 		};
 	};
 
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	private readonly _contracts: Contract<any>[] = [];
```
luu-alex: do we ever need to remove contracts?

luu-alex (approved):

packages/web3/src/index.ts
```diff
@@ -70,6 +73,29 @@ export class Web3 extends Web3Context<EthExecutionAPI> {
...
+			}
+
+			const contract = self.use(Contract, jsonInterface, address, options);
+
+			self._contracts.push(contract);
+
+			return contract;
+		}
+
+		ContractBuilder.setProvider = (_provider: SupportedProviders<EthExecutionAPI>) => {
```
jdevcs: currently contract is using universal provider coming via `Web3Context` > `Web3RequestManager` why do we need to explicitly cascade change after once it set at start.

nazarhussain: To preserve the behaviour of 1.x that once the provider is changed on `Contract` static level, provider must be changed on all instances of any contract ever created in script life cycle.

packages/web3/src/index.ts
nazarhussain: As the instance of the `Web3` object will be garbage collected, these contracts will be removed.

PR #4922, opened 2022-04-05
Web3Error base class

jdevcs:
## Description

Fixes https://github.com/ChainSafe/web3.js/issues/4889

Summary: https://github.com/ChainSafe/web3.js/pull/4922#issuecomment-1108686822 
## Type of change

## Checklist:


jdevcs:
###  Changes Summary:
Now 4.x is using same web3-Error base class.

Earlier there was Error base class in web3-utils and web3-common, so removed it from web3-util, and made web3-utils to use Error base class from web3-common. This caused cyclic dependency. Made following changes for removing cyclic dependency:

#### Moved from web3-utils to web3-common for fixing cyclic dep issue

- bytesToBuffer
- bufferToHexString
- bytesToHex
- hexToBytes
- hexToNumber
- numberToHex
- hexToNumberString
- utf8ToHex
- fromUtf8
- hexToUtf8
- toUtf8
- toNumber
- toBigInt
- toChecksumAddress

- sha3
- sha3Raw

- isIterable
- mergeDeep


There was another cyclic dependency between IBAN and web3-common so:
#### Moved from iban package to web3-common

- isDirectIBAN
- mod9710
- iso13616Prepare
- isValidIBAN
- padLeft
- parseIntValue
- IBANtoAddress

#### Moved from web3-utils to web3-common/src/eth_execution_api  for fixing cyclic dep issue
Topic
Filter

####  Following types are moved from web3-utils to web3-common/src/types.ts for fixing cyclic dep issue

- HexString
- Uint
- Numbers
- Bytes
- Address 
- BlockTags
- BlockTag
- BlockNumberOrTag
- HexString256Bytes
- HexString8Bytes
- HexString32Bytes
- HexStringBytes
- HexStringSingleByte
- Uint256

#### other changes
made unit tests adjustments
fixed imports

avkos (approved):

packages/web3-common/src/common_utils.ts
```diff
@@ -0,0 +1,306 @@
+import { keccak256 } from 'ethereum-cryptography/keccak';
+import { validator, isHexStrict, utils as validatorUtils, isAddress } from 'web3-validator';
+import { InvalidAddressError, InvalidBytesError, InvalidNumberError } from './errors';
+import { Address, Bytes, HexString, Numbers } from './types';
+
+const SHA3_EMPTY_BYTES = '0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470';
+
+export const bytesToBuffer = (data: Bytes): Buffer | never => {
```
nazarhussain, line 8: We have `web3-utils` package particularly for this purpose. The `web3-utils` should be used inside the `web3-common` not the other-way around. 

Think of it it as: 

1. `web3-utils` should be independent library consisting of the high level functions for small utilities. 
2. If we consider `web3` as a framework the `web3-common` is the common and abstract architectural components of the framework. 
3. It does not make sense to split the utility functions and in multiple packages and say `toWei` will be in `web3-utils` but `toBigInt` will be in `web3-common`.

nazarhussain, line 8: This all change seems to be just to reuse the error codes. For which we already discussed that we will be creating a package `web3-errors` when needed. So it seems now it's the need. @spacesailor24 @luu-alex

nazarhussain (changes requested):
I personally don't think the direction of the changes in this PR is correct and splitting up the utility functions among `web3-utils` and `web3-common` would be a nice idea. Or making a `web3-common` a dependency of the `web3-utils` would a right architectural choice. 

Let's hear form others. @ChainSafe/web3-js

packages/web3-common/src/common_utils.ts
jdevcs: > I personally don't think the direction of the changes in this PR is correct and splitting up the utility functions among `web3-utils` and `web3-common` would be a nice idea. Or making a `web3-common` a dependency of the `web3-utils` would a right architectural choice.
> 
> Let's hear form others. @ChainSafe/web3-js

Its done for solving issue of cyclic dependency between utils and common. As there was wrong design of having two web3 Error base classes in existing code. 

web3-common, name suggests that its common functionality used in all or most of other packages, so there should be zero or minimum dependencies in web3-common.

nikoulai (approved):
Changes seem ok, but, regarding structure, I would agree with Nazar.

packages/web3-common/src/common_utils.ts
jdevcs: Its not for reusing error codes but for using same base Error class.

nazarhussain: I don't think there was any cyclic dependency between `web3-utils` and `web3-common` packages. There are not just `Error` but few other duplicate codes and types in other packages as well to avoid the unnecessary dependency and that's not a bad design choice. And to improve such cases we can always create more domain specific packages e.g. in case of this PR `web3-errors` package will the right direction and that was discussed way earlier.

jdevcs: using same base error class from web3 common in whole 4x created cyclic deps, and fix is in this pr.

Spacesai1or (changes requested):
I agree with the points raised by @nazarhussain, and I believe it's perhaps time to create `web3-errors`

jdevcs:
> I agree with the points raised by @nazarhussain, and I believe it's perhaps time to create `web3-errors`

Closing this PR for above suggested direction. 

However, this will make no difference from functional perspective for end user. As discussed in last Retro call good to discuss such items in backlog grooming issues discussion, sprint plannings, and team calls earlier instead of waiting for PR merged or completed, as this and other such items impacts alpha delivery planned goals.

nazarhussain:
@jdevcs It will impact from perspective of end user, the question here is what's the persona of end user you are referring here. 

I also agree that such discussions to be made during grooming, which we usually miss and directly jump to planning. So for our workflow it makes better productive that person who is assigned and working on an issue should initiate the discussion before taking a direction which involves huge refactoring.

jdevcs:
> It will impact from perspective of end user,
It will not from functional perspective.



> which we usually miss and directly jump to planning
Informed earlier: Sprint Plannings are next good candidate for technical discussions as well along with backlog grooming. Ref: https://scrumguides.org/scrum-guide.html#sprint-planning

PR #4937, opened 2022-04-12
chore: fix eth abi typings

chancehudson:
## Description

Copy of #4100 

@avkos @nazarhussain Can we push this through? It's been a year and this is a pretty simple type change.

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-eth-abi/types/index.d.ts
+++ b/packages/web3-eth-abi/types/index.d.ts
@@ -19,7 +19,7 @@
 
 import { AbiInput, AbiItem } from 'web3-utils';
 
-export class AbiCoder {
+declare class AbiCoder {
     encodeFunctionSignature(functionName: string | AbiItem): string;
 
     encodeEventSignature(functionName: string | AbiItem): string;
@@ -40,3 +40,6 @@ export class AbiCoder {
         topics: string[]
     ): { [key: string]: string };
 }
+
+declare const coder: AbiCoder;
+export default coder;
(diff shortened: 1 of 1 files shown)


packages/web3-eth-abi/types/index.d.ts
```diff
@@ -19,7 +19,7 @@
 
 import { AbiInput, AbiItem } from 'web3-utils';
 
-export class AbiCoder {
+declare class AbiCoder {
```
nazarhussain, line 22: Removing the export would be a breaking change, that's why the tests are also failing. Please fix it, add the entry in the CHANGELOG.md file as well.

nazarhussain: Removing the export would be a breaking change, that's why the tests are also failing. Please fix it, and add the entry in the CHANGELOG.md file as well.

nazarhussain (changes requested):

packages/web3-eth-abi/types/index.d.ts
chancehudson: Thanks for the heads up, I think there's no reason to change this, we just need to add a default export.

jdevcs:
Related to https://github.com/ChainSafe/web3.js/pull/3587

avkos (approved):

nazarhussain (approved):

avkos:
create a [copy of this PR](https://github.com/ChainSafe/web3.js/pull/4947) and close

PR #4939, opened 2022-04-12
Fix build command

wbt:
## Description
This replaces #4848 with another copy of the same PR.  Previously, someone changed the base branch and merged it in there automatically, so it can't be reopened and remerged into the original target.  The proposed replacement of #4861 modifies a lot of package-lock files that aren't allowed to be modified, so this version is re-proposed.


jdevcs (commented):
There are conflicts in package-lock.json

wbt:
As a maintainer, you have more ability than I do to revert changes in a particular file and resolve conflicts. 
image
I can't do this because I'm not listed as a maintainer on this package.

Alternatively, a maintainer can see the changes being proposed here for the third time and manually make them again in a new branch, ideally proposing a PR that does not lead to conflict between maintainers about requirements related to package-lock.json.  

This was submitted in the hopes it'd be helpful but the barriers to contribution in this project are in general significantly higher than they should be, meaning a lot of contributions which could add value are not worth making because the cost of the contribution exceeds its value.  Ironically, this PR was intended to reduce that frustration, lowering the friction and barriers to contribution in one of the areas they are unnecessarily high, but the experience only demonstrated how high those barriers remain.

Spacesai1or:
Superseded by #4987

Issue #4853, opened 2022-03-14
Web3 Personal- 4.0.0 Alpha.0 System Tests 


jdevcs:
System tests for:

## Web3.eth.personal

- [ ] newAccount
- [ ] sign
- [ ] ecRecover
- [ ] signTransaction
- [ ] unlockAccount
- [ ] lockAccount
- [ ] getAccounts
- [ ] importRawKey

#### Min Platform Requirements for 4.x Alpha:
- Node: v14, v16
- Browsers: Firefox, Brave
- Eth Clients: Geth, Ganache

More details are in [Epic](https://github.com/ChainSafe/web3.js/issues/4711)

mconnelly8:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/ChainSafe/web3.js/4853?planning-poker) with ZenHub @avkos @jdevcs @luu-alex @nikoulai @spacesailor24

Spacesai1or:
According to [this](https://github.com/trufflesuite/ganache/blob/develop/src/chains/ethereum/ethereum/RPC-METHODS.md), Ganache doesn't appear to support `ecRecover`

luu-alex:
This epic is done, but an investigation has popped from merging the PR for this issue. https://github.com/ChainSafe/web3.js/issues/5046

PR #4984, opened 2022-04-26
4853 - web3 Personal- 4.0.0 Alpha.0 System Test

luu-alex:
#4853 
Geth and Ganache integration Tests for:
 - new Account
 - ecRecover (geth only)
 - lock account
 - unlock account
 - sign (geth only)
 - get accounts
 - import raw key
 - sign Transaction (ganache only)
 - send Transaction


packages/web3-eth-personal/test/integration/personal.test.ts
```diff
@@ -0,0 +1,130 @@
...
+	beforeEach(async () => {
+		accounts = await getSystemTestAccounts();
+	});
+	it('new account', async () => {
+		const newAccount = await ethPersonal.newAccount('!@superpassword');
+		expect(isHexStrict(newAccount)).toBe(true);
+	});
+
+	it('ecRecover', async () => {
+		if (getSystemTestBackend() === 'geth') {
```
nikoulai: Maybe refactor test to use `ifIf(getSystemTestBackend() === 'geth')` for more clear scope and don't run the test on ganache?

packages/web3-eth-personal/test/fixtures/system_test_utils.ts
```diff
@@ -0,0 +1,106 @@
...
+	_accounts = ((await res.json()) as { result: string[] }).result;
+
+	return _accounts;
+};
+
+export const itIf = (condition: (() => boolean) | boolean) =>
+	(typeof condition === 'function' ? condition() : condition) ? test : test.skip;
+
+export const describeIf = (condition: (() => boolean) | boolean) =>
+	(typeof condition === 'function' ? condition() : condition) ? describe : describe.skip;
```
avkos: create a symlink for this file like in the contract package. the fastest way is just to copy symlink from another package and paste it to your package, but you should use Finder for this

avkos (changes requested):

packages/web3-eth-personal/test/config/personal.test.config.ts
```diff
@@ -0,0 +1,29 @@
...
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+type Account = {
+	address: string;
+	privateKey: string;
+	balance: string;
+};
+
+export const importedAccount: Account = {
```
nazarhussain: To avoid any confusion, try not use the `.test` extension in any file other than the test files itself.

packages/web3-eth-personal/test/fixtures/system_test_utils.ts
nazarhussain: Or you can use `ln -s <source> <link>` command for it.

packages/web3-eth-personal/test/integration/personal.test.ts
```diff
@@ -0,0 +1,125 @@
...
+	});
+	beforeEach(async () => {
+		accounts = await getSystemTestAccounts();
+	});
+	it('new account', async () => {
+		const newAccount = await ethPersonal.newAccount('!@superpassword');
+		expect(isHexStrict(newAccount)).toBe(true);
+	});
+
+	itIf(getSystemTestBackend() === 'geth')('ecRecover', async () => {
```
nazarhussain, line 50: 👍

packages/web3-eth-personal/test/integration/personal.test.ts
```diff
@@ -0,0 +1,125 @@
...
+		const signature = await ethPersonal.sign('0x2313', accounts[0], '');
+		const publicKey = await ethPersonal.ecRecover('0x2313', signature); // ecRecover is returning all lowercase
+		// eslint-disable-next-line jest/no-standalone-expect
+		expect(toChecksumAddress(publicKey)).toBe(toChecksumAddress(accounts[0]));
+	});
+
+	it('lock account', async () => {
+		const lockAccount = await ethPersonal.lockAccount(accounts[1]);
+		expect(lockAccount).toBe(true);
+	});
```
nazarhussain: Should we not test if the account is actually locked for it? What if the implementation of `lockAccount` function simply says `return true`, this test will be a false positive in that case.

packages/web3-eth-personal/test/integration/personal.test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+	it('unlock account', async () => {
+		const key = accounts[0];
+		const unlockedAccount = await ethPersonal.unlockAccount(key, '', 100000);
+		expect(unlockedAccount).toBe(true);
+	});
```
nazarhussain: Same here. These tests files will run as `integration` and `e2e`, so the exact behavior of the function should be tested. If not directly in a test, may be through some other tests. 

e.g. In contract deployment tests I didn't tested if the contract was actually deployed or not, because it was indirectly tested in the methods and events case. But the behavior of contract deployment was fully tested end-to-end.

nazarhussain (changes requested):
I make some comments related to the tests. See if it makes sense to you.

packages/web3-eth-personal/package.json
```diff
@@ -22,7 +22,7 @@
 		"test:ci": "jest --coverage=true --coverage-reporters=json --verbose",
 		"test:watch": "npm test -- --watch",
 		"test:unit": "jest --config=./test/unit/jest.config.js",
-		"test:integration": "jest --config=./test/integration/jest.config.js --runInBand --passWithNoTests"
+		"test:integration": "jest --config=./test/integration/jest.config.js --forceExit --runInBand"
```
Spacesai1or: I could totally be wrong, but it seems like the use of `--forceExit` is hiding some bugs and @nazarhussain had previously mentioned

packages/web3-eth-personal/test/integration/personal.test.ts
```diff
@@ -0,0 +1,125 @@
...
+	beforeAll(() => {
+		const clientUrl = getSystemTestProvider();
+		ethPersonal = new EthPersonal(clientUrl);
+	});
+	beforeEach(async () => {
+		accounts = await getSystemTestAccounts();
+	});
+	it('new account', async () => {
+		const newAccount = await ethPersonal.newAccount('!@superpassword');
+		expect(isHexStrict(newAccount)).toBe(true);
```
spacesailor24, line 40:
```suggestion
		expect(isHexStrict(newAccount)).toBeTruthy();
```
The linter should've complained about this :thinking:

packages/web3-eth-personal/test/integration/personal.test.ts
```diff
@@ -0,0 +1,125 @@
...
+	});
+
+	it('unlock account', async () => {
+		const key = accounts[0];
+		const unlockedAccount = await ethPersonal.unlockAccount(key, '', 100000);
+		expect(unlockedAccount).toBe(true);
+	});
+
+	// ganache does not support sign
+	itIf(getSystemTestBackend() === 'geth')('sign', async () => {
```
Spacesai1or, line 100: This test is relying on a correct implementation/output of `ecRecover`. I think it's fine to include the `expect` you already have, but since the signature should remain static for our test environment, I would copy the `signature` that's produced and have:

```
expect(signature).toBe(signatureFromConfig);
```

packages/web3-eth-personal/test/integration/personal.test.ts
```diff
@@ -0,0 +1,125 @@
...
+		expect(accounts[0]).toBe(address);
+	});
+
+	it('getAccounts', async () => {
+		const accountList = await ethPersonal.getAccounts();
+		const accountsLength = accountList.length;
+		// create a new account
+		await ethPersonal.newAccount('cde');
+		const updatedAccountList = await ethPersonal.getAccounts();
+		expect(updatedAccountList).toHaveLength(accountsLength + 1);
```
Spacesai1or: Instead of testing the length of `updatedAccountList`, I think it would make sense to test the expected value with something like:

```typescript
const newAccount = await ethPersonal.newAccount('cde');
expect(updateAccountList).toStrictEqual([newAccount]);
```

You can probably get rid of

```typescript
const accountList = await ethPersonal.getAccounts();
const accountsLength = accountList.length;
```

Since my suggested change would also be testing the length

packages/web3-eth-personal/test/integration/personal.test.ts
luu-alex: From what I understand, there is no direct command to see if an account is locked/unlocked other than sending a transaction and seeing if errors, which is something i can definitely do.

packages/web3-eth-personal/test/integration/personal.test.ts
luu-alex: the signatures are not static on geth, i would have assumed they are as well but it doesnt seem to be the case.

avkos (approved):

packages/web3-eth-personal/test/integration/personal.test.ts
nazarhussain: Yes that's called the [side effect](https://en.wikipedia.org/wiki/Side_effect_(computer_science)). 

> an observable effect besides returning a value (the primary effect) to the invoker of the operation

If there is only way to test the completion of operation through it's side effects then we should test it as part of the integration/system tests.

packages/web3-eth-personal/package.json
nazarhussain: Yes `--forceExit` was hiding the floating promise.

nazarhussain (approved):

packages/web3-eth-personal/test/integration/personal.test.ts
avkos: yes. I think to test if the account is locked we need to send tx for example from it. and receive the error.

jdevcs (commented):
There are some merge conflicts and CI/CD actions not passing, other than that lgtm.

packages/web3-eth-personal/package.json
luu-alex: will update

Abbasagakasiri (commented):
0xeb1ef1c6b1bffa4e118f1769420107a5076ee36b4741ecae29329ae7278d8cb7

packages/web3-eth-personal/test/integration/personal.test.ts
luu-alex: I will create a seperate issue for this, to investigate testing ecRecover so that the signature does not change.

luu-alex:
https://github.com/ChainSafe/web3.js/issues/5046

PR #5008, opened 2022-05-05
4897/websocket provider 4x system tests

### LINKED ISSUE #4897: WebSocket Provider - 4.x System Tests
System tests for:

## WebSocket Provider package

- [x] connect tests
- [ ] subscribe events tests
- [ ] disconnect and reset tests
- [ ] getStatus get and validate all status tests
- [ ] send multiple Requests on same connection with valid payload and receive response tests
- [ ] Support of Basic Auth tests
- [ ] Fault Tolerance tests

#### Min Platform Requirements:
- Node: v14, v16
- Browsers: Firefox, Brave
- Eth Client: Geth

More details are in [Epic](https://github.com/ChainSafe/web3.js/issues/4711)

nikoulai:
## Description

#4897

## Type of change

## Checklist:


packages/web3-providers-ws/package.json
```diff
@@ -22,10 +22,12 @@
 		"test:ci": "jest --coverage=true --coverage-reporters=json --verbose",
 		"test:watch": "npm test -- --watch",
 		"test:unit": "jest --config=./test/unit/jest.config.js",
-		"test:integration": "jest --config=./test/integration/jest.config.js --passWithNoTests"
+		"test:integration": "jest --config=./test/integration/jest.config.js"
```
avkos: Please add -runInBand option to run tests one by one. Prevents nonce problem

packages/web3-providers-ws/test/fixtures/system_test_utils.ts
```diff
@@ -0,0 +1,115 @@
...
+		address: '0xdc6bad79dab7ea733098f66f6c6f9dd008da3258',
+		privateKey: '0x4c3758228f536f7a210f8936182fb5b728046970b8e3215d0b5cb4c4faae8a4e',
+	},
+	{
+		address: '0x962f9a9c2a6c092474d24def35eccb3d9363265e',
+		privateKey: '0x34aeb1f338c17e6b440c189655c89fcef148893a24a7f15c0cb666d9cf5eacb3',
+	},
+];
+
+export const getSystemTestAccounts = async (): Promise<string[]> => {
```
avkos: You do not need to copy this file. Just create a symlink (like `packages/web3-eth/test/fixtures/system_test_utils.ts`)

.github/nginx.conf
```diff
@@ -0,0 +1,77 @@
+user  nginx;
```
avkos: Do you need this file? I see you rewrite auth connection with NodeJS.

avkos (changes requested):

packages/web3-eth/test/integration/jest.config.js
```diff
@@ -1,4 +1,19 @@
-'use strict';
```
luu-alex, line 1: I dont think you want to remove this

nikoulai: you are right, thanks

nikoulai: I tried to restore, but, I get an eslint error `'use strict' is unnecessary inside of modules. 
 `eslint [strict](https://eslint.org/docs/rules/strict). Does this make sense to you? Should I ignore it?

luu-alex: what command do u run to get the eslint error?

luu-alex: i added it, so ill resolve this conversation

avkos (approved):

.github/.htpasswd
```diff
@@ -0,0 +1,2 @@
+# htpasswd -c .htpasswd geth - change password for geth
```
luu-alex: im not familiar with this, do we need this file or is this file being used somewhere

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+	JsonRpcNotification,
+	JsonRpcSubscriptionResult,
+} from 'web3-common';
+import express from 'express';
+import { createProxyMiddleware } from 'http-proxy-middleware';
+
+import { Server } from 'http';
+import { Web3WSProviderError } from 'web3-errors';
+import WebSocketProvider from '../../src/index';
+// import { Web3WSProviderError } from '../../src/errors';
```
luu-alex: remove unnecessary comment

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+				// eslint-disable-next-line jest/expect-expect
+				it('should subscribe on connect', done => {
+					webSocketProvider.on('open', () => {
+						done();
+					});
+				});
+
+				it('should subscribe on close', done => {
+					const code = 1001;
+					// const provider = new WebSocketProvider(clientWsUrl);
```
luu-alex: remove comment

packages/web3-providers-ws/src/index.ts
```diff
@@ -169,7 +169,8 @@ export default class WebSocketProvider<
 	}
 
 	public disconnect(code?: number, reason?: string): void {
-		this._removeSocketListeners();
+		// todo Is next line really needed? onClose will never be called
+		// this._removeSocketListeners();
 		this._webSocketConnection?.close(code, reason);
```
luu-alex, line 174: I tried searching for the answer to my question but couldn't find it, will try it on my end later when i have more time but if someone has the answer:
On `close` does close remove event listeners or is that something we need to do manually?

packages/web3-providers-ws/package.json
```diff
@@ -34,16 +36,19 @@
 		"eslint-config-prettier": "^8.3.0",
 		"eslint-config-web3-base": "0.1.0",
 		"eslint-plugin-import": "^2.25.3",
+		"express": "^4.18.1",
+		"http-proxy-middleware": "^2.0.6",
 		"jest": "^27.3.1",
 		"jest-extended": "^1.1.0",
+		"morgan": "^1.10.0",
```
Spacesai1or: What is thing being used for? Doesn't seem to be used directly in these changes

packages/web3-providers-ws/src/index.ts
```diff
@@ -80,14 +80,14 @@ export default class WebSocketProvider<
 		this._clientUrl = clientUrl;
 		this._wsProviderOptions = wsProviderOptions;
 
-		const DEFAULT_WS_PROVIDER_OPTIONS = {
+		const DEFAULT_RECON_WS_PROVIDER_OPTIONS = {
```
Spacesai1or: I think it's the project preference to not abbreviate things, I'd suggest renaming this to `DEFAULT_PROVIDER_RECONNECTION_OPTIONS`

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+describeIf(getSystemTestProvider().includes('ws'))(
+	'WebSocketProvider - implemented methods',
+	() => {
+		let clientWsUrl: string;
+		let accounts: string[];
+		let webSocketProvider: WebSocketProvider;
+		let jsonRpcPayload: Web3APIPayload<EthExecutionAPI, 'eth_getBalance'>;
+		// helper function
+		let currentAttempt = 0;
+		const waitForOpenConnection = async (provider: WebSocketProvider, status = 'connected') => {
```
Spacesai1or: I would move this into it's own file called `wait_for_open_connection.ts` to declutter this test file, and in case it can be used elsewhere

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+					// eslint-disable-next-line no-plusplus
+					currentAttempt++;
+				}, intervalTime);
+			});
+		};
+		beforeAll(async () => {
+			clientWsUrl = getSystemTestProvider();
+			accounts = await getSystemTestAccounts();
+		});
+		describe('tests', () => {
```
Spacesai1or: I don't think all the `describe`s you have are necessary. If I remember correctly, I think there's even a performance hit the more you nest the `describes`

There should be a single top level `describe` that says something like 

```typescript
describeIf(getSystemTestProvider().includes('ws'))(
	'WebSocketProvider',
	() => { ... }
```

And everything else will be inherently described by the `it('should ...')`

spacesailor24: I don't think you really need the other `describe`s as well (`describe('websocker provider tests'` and `describe('subscribe event tests'`) I suppose it doesn't matter that much if there is no performance hit though

packages/web3-providers-ws/src/index.ts
nikoulai: Maybe @nazarhussain could give us some insight on the topic.
Also, I would like to mention that `this._removeSocketListeners()` is also called on the `onclose` handler when we receive the event from the server.

packages/web3-providers-ws/package.json
nikoulai: removed/ not necessary

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+		let clientWsUrl: string;
+		let accounts: string[];
+		let webSocketProvider: WebSocketProvider;
+		let jsonRpcPayload: Web3APIPayload<EthExecutionAPI, 'eth_getBalance'>;
+		// helper function
+		let currentAttempt = 0;
+		const waitForOpenConnection = async (provider: WebSocketProvider, status = 'connected') => {
+			return new Promise<void>((resolve, reject) => {
+				const maxNumberOfAttempts = 10;
+				const intervalTime = 5000; // ms
```
nazarhussain: This seems unnecessary large waiting time for reconnect.

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+					clientWsUrl,
+					{},
+					{ delay: 1, autoReconnect: false, maxAttempts: 1 },
+				);
+				currentAttempt = 0;
+			});
+			afterEach(async () => {
+				if (webSocketProvider.getStatus() !== 'disconnected') {
+					// make sure we try to close the connection after it is established
+					await waitForOpenConnection(webSocketProvider);
```
nazarhussain: May be we do this only when it's in `connecting` stage.

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+							error: Error | null,
+							result?: JsonRpcSubscriptionResult | JsonRpcNotification<any>,
+						) => {
+							if (error) {
+								done.fail(error);
+							}
+							expect(result?.id).toBe(jsonRpcPayload.id);
+							done();
+						},
+					);
```
nazarhussain: See the usage in contracts test, may be we can convert this callback code to more promise oriented.

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
nikoulai: Would you suggest being moved inside the tests directory or somewhere else?

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+					});
+
+					webSocketProvider['_wsEventEmitter'].emit(
+						'error',
+						new Web3WSProviderError(errorMsg),
+					);
+				});
+
+				// eslint-disable-next-line jest/expect-expect
+				it('should subscribe on connect', done => {
```
nazarhussain: It should subscribe to which event?

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+						done();
+					});
+
+					webSocketProvider['_wsEventEmitter'].emit(
+						'error',
+						new Web3WSProviderError(errorMsg),
+					);
+				});
+
+				// eslint-disable-next-line jest/expect-expect
```
nazarhussain: That's a valid catch by elint. There should not be any test without any expect. Please add what your expectations to the tests and remove this comment.

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+				it('should subscribe on connect', done => {
+					webSocketProvider.on('open', () => {
+						done();
+					});
+				});
+
+				it('should subscribe on close', done => {
```
nazarhussain: Should subscribe `on` or should subscribe `to`

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
nikoulai: Some of them can be removed, but others are needed because I need different initialization in `beforeAll/Each`. Maybe I could separate them completely.

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+
+					webSocketProvider['_sentRequestsQueue'].set(
+						jsonRpcPayload.id as JsonRpcId,
+						reqItem,
+					);
+					expect(webSocketProvider['_sentRequestsQueue'].size).toBe(1);
+
+					webSocketProvider.reset();
+					expect(webSocketProvider['_pendingRequestsQueue'].size).toBe(0);
+					expect(webSocketProvider['_sentRequestsQueue'].size).toBe(0);
```
nazarhussain: In general one should not interact with private data fields or behaviors in the integration tests. If you think something is really necessary to test from user perspective and it does not have a public interface then you should create one.

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+									done.fail(err.message);
+								});
+						})
+						.catch(err => {
+							done.fail(err.message);
+						});
+				});
+			});
+		});
+		describe('Support of Basic Auth', () => {
```
nazarhussain: Will suggest to move this whole block to a separate file, that will make it easy to read and maintain these test files.

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
```diff
@@ -0,0 +1,338 @@
...
+			let server: Server;
+			beforeAll(() => {
+				const app = express();
+				const PORT = 3000;
+				const HOST = 'localhost';
```
nazarhussain: Not necessary to capitalize these, these are standard variables in this context.

scripts/test-runner.sh
```diff
@@ -29,6 +29,8 @@ if [[ ! " ${SUPPORTED_ENGINES[*]} " =~ " ${ENGINE} " ]]; then
 	helpFunction
 fi
 
+echo "$MODE://localhost:$WEB3_SYSTEM_TEST_PORT"
+echo $BACKEND
```
nazarhussain: 👍 

Try to add further descriptions here as well. So when we check on console we know what was echoed.

nazarhussain (changes requested):

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
nikoulai: I took the number from here:

https://github.com/ChainSafe/web3.js/blob/07fa78a1922d362186a87cea3c2d5b3405229e53/packages/web3-providers-ws/src/index.ts#L85

I'll reduce it to 2000

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
nikoulai: Make some changes, now there is only one

packages/web3-providers-ws/test/integration/web_socket_provider_integration.test.ts
nikoulai: These queues are not needed to be exposed to the end-user, but, I want to be sure that they are emptied when `reset` is called. Any ideas on how this could be done without exposing them?

nazarhussain: The thumb rule is, If something is not exposed as public it should be tested in integration tests, instead in unit tests. Integration is tests simply test the user expectations and if something not exposed, means user is not aware of internal implementation. There could be a queue, or there could be a stack or something else as a user I would not be testing these details. 

Just in case you wonder a way, then one way is to convert those `private` to `protected` then while testing we create a dummy class inherited from there and create public methods which use the protected properties and expose them. And then you use that dummy class in tests. In Java and C# there is a concept of Reflections, which can be used to test such scenarios.

packages/web3-providers-ws/src/index.ts
nazarhussain: If we don't remove listeners before closing `_onClose`  will try to reconnect. Because `disconnect` is intentional close so why we remove the listeners first.

nazarhussain (approved):

packages/web3-providers-ws/src/index.ts
luu-alex: Okay, just uncomment `this._removeSocketListeners();` and it lgtm

jdevcs (approved):

PR #5016, opened 2022-05-10
web3: added option to enable/disable autoconnect

### LINKED ISSUE #4957: web3: include an option to enable/disable autoconnect

### Feature Request

Currently web3 supports a feature of `reconnect` by which we can configure the reconnection parameters. A similar feature is needed to enable/disable the autoconnect to the mentioned provider.

Specifically, enveloping the condition of connection only if autoconnect parameter is set as true [here](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-providers-ws/src/index.js#L89) and similar places.

### Use Cases

There are certain scenarios where this feature can come in handy, like when we want to test some codebase containing web3 library, but cannot pass the provider; in such case we can simple turn off autoconnect to avoid connection in the first place.

One of such usages can come in handy while testing [this](https://github.com/hyperledger/cactus/blob/dda3f00c5319e22b03c73fe6dbc645dea41cc58e/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts#L151)

### Implementation Ideas

This can be easily implemented by enveloping [this](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-providers-ws/src/index.js#L89) with a condition which checks if the `autoconnect` parameter is set or not (should be defaulted to `true` to avoid possible code crashes for devs upgrading to the version of web3 including this feature)

_No response_

- [X] Yes

TheJagpreet:
## Description
Added a boolean to enable/disable autoconnect

Fixes #4957

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-providers-ws/src/index.ts
+++ b/packages/web3-providers-ws/src/index.ts
@@ -71,6 +71,7 @@ export default class WebSocketProvider<
 		clientUrl: string,
 		wsProviderOptions?: ClientOptions | ClientRequestArgs,
 		reconnectOptions?: ReconnectOptions,
+		autoConnect?: boolean,
 	) {
 		super();
 		if (!WebSocketProvider._validateProviderUrl(clientUrl))
@@ -98,7 +99,9 @@ export default class WebSocketProvider<
 		this._onCloseHandler = this._onClose.bind(this);
 
 		this._init();
-		this.connect();
+		if(autoConnect!==false){
+			this.connect();
+		}
 	}
 
 	private static _validateProviderUrl(providerUrl: string): boolean {
@@ -185,6 +188,7 @@ export default class WebSocketProvider<
 		Method extends Web3APIMethod<API>,
 		ResponseType = Web3APIReturnType<API, Method>,
 	>(request: Web3APIPayload<API, Method>): Promise<JsonRpcResponse<ResponseType>> {
+		this.connect();
 		if (this._webSocketConnection === undefined)
 			throw new Web3WSProviderError('WebSocket connection is undefined');
 


packages/web3-providers-ws/src/index.ts
```diff
@@ -71,6 +71,7 @@ export default class WebSocketProvider<
 		clientUrl: string,
 		wsProviderOptions?: ClientOptions | ClientRequestArgs,
 		reconnectOptions?: ReconnectOptions,
+		autoConnect?: boolean,
```
jdevcs, line 74: There is already [autoConnect](https://github.com/ChainSafe/web3.js/blob/82574a2587810f9bea26ba65da48fe667162e573/packages/web3-providers-ws/src/types.ts#L21) in `reconnectOptions` 

```suggestion
		
```

jdevcs (changes requested):

packages/web3-providers-ws/src/index.ts
```diff
@@ -98,7 +99,9 @@ export default class WebSocketProvider<
 		this._onCloseHandler = this._onClose.bind(this);
 
 		this._init();
-		this.connect();
+		if(autoConnect!==false){
+			this.connect();
```
jdevcs, line 103: This is for first time connection , not for re connect so what is need of adding this check here?

packages/web3-providers-ws/src/index.ts
```diff
@@ -185,6 +188,7 @@ export default class WebSocketProvider<
 		Method extends Web3APIMethod<API>,
 		ResponseType = Web3APIReturnType<API, Method>,
 	>(request: Web3APIPayload<API, Method>): Promise<JsonRpcResponse<ResponseType>> {
+		this.connect();
```
jdevcs, line 191: Validate will be called before init so why this connect is added here?

packages/web3-providers-ws/src/index.ts
TheJagpreet: Oh, maybe I wasn't using the version of web3 which has this.
Closing the PR as the feature already exists,

PR #5017, opened 2022-05-10
`web3-eth` `rpc_methods` Test Refactors

### LINKED ISSUE #4993: Finish Rewriting `web3-eth` tests
#4920 rewrote the `rpc_method_wrappers.ts` tests, but did not include `web3_eth.ts` and `rpc_methods.ts`. These tests also need to be rewritten to the new testing format for `web3-eth`

Spacesai1or:
### What This PR Does

- Replaces
    - `packages/web3-eth/test/unit/rpc_methods_no_parameters.test.ts`
    - `packages/web3-eth/test/unit/rpc_methods_with_parameters.test.ts`
    With `packages/web3-eth/test/unit/rpc_methods/`

closes #4993


packages/web3-common/test/unit/formatters.test.ts
```diff
@@ -264,7 +264,7 @@ describe('formatters', () => {
 
 			expect(utils.sha3Raw).toHaveBeenCalledWith('blockHashtransactionHashlogIndex');
 
-			expect(result.id).toEqual(`log_${sha3Result.substr(0, 8)}`);
+			expect(result.id).toBe(`log_${sha3Result.substr(0, 8)}`);
```
Spacesai1or: Linter was complaining about this

packages/web3-eth-accounts/src/account.ts
```diff
@@ -244,7 +244,7 @@ export const privateKeyToAddress = (privateKey: string | Buffer): string => {
 
 	const publicKey = getPublicKey(stringPrivateKeyNoPrefix);
 
-	const publicKeyString = `0x${publicKey.slice(2)}`;
+	const publicKeyString = `0x${publicKey.slice(2).toString()}`;
```
Spacesai1or: Linter was complaining about this

avkos: slice always returns a string. so, I think, we don't need to add `toString()` here.

avkos (changes requested):

package.json
```diff
@@ -75,12 +75,14 @@
...
 		"ts-jest": "^27.0.7",
 		"ts-loader": "^9.2.8",
 		"ts-node": "^10.4.0",
 		"typescript": "^4.5.2",
 		"utf-8-validate": "^5.0.9",
 		"webpack": "^5.72.0"
 	},
-	"packageManager": "yarn@1.22.15"
+	"packageManager": "yarn@1.22.15",
+	"dependencies": {}
```
nazarhussain:
```suggestion
	"packageManager": "yarn@1.22.15"
```

packages/web3-eth/test/unit/rpc_methods/estimate_gas.test.ts
```diff
@@ -0,0 +1,69 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+// */
+import { Web3RequestManager } from 'web3-core';
+import { validator } from 'web3-validator';
+
+import { estimateGas } from '../../../src/rpc_methods';
+import { testData } from './fixtures/estimate_gas';
+
+jest.mock('web3-validator');
```
nazarhussain, line 37: I prefer not to only mock the I/O or compute intensive tasks. In this case if we have right input/output then the validator will not be a problem for the tests. What do you think?

nazarhussain (changes requested):

packages/web3-eth/test/unit/rpc_methods/estimate_gas.test.ts
Spacesai1or: I mocked the validator to be able to spy on the `.validate` method to ensure it was being called with the expected data. Since we have explicit tests for `web3-validator`, I don't think we need to be concerned that the validation passes. Instead, I opted for verifying that we're validating the expected data with the expected validation rules (e.g. `blockNumberOrTag`, `string`, `hex`, etc.)

packages/web3-eth-accounts/src/account.ts
jdevcs: It can be operated on array, so not always string @avkos

avkos (approved):

nazarhussain (approved):

jdevcs (approved):

packages/web3-eth-accounts/src/account.ts
```suggestion
	const publicKeyString = `0x${publicKey.slice(2)}`;
```

PR #5034, opened 2022-05-17
Change parameter type and format transaction to default type

### LINKED ISSUE #4892: Improve type for transaction builder
I don't think it's safe to assume that the transaction is `InternalTransaction` (a transaction with hex string numbers and bytes properties) if the parameter `transaction` has type `Record<string, unknown>`

Unless TypeScript has a fit about it, I think we should have `transaction` be of type [Transaction](https://github.com/ChainSafe/web3.js/blob/32b898477c455c8773e094e5b32a3d328a328ea4/packages/web3-eth/src/types.ts#L71) and we should format `populatedTransaction` to `DEFAULT_RETURN_FORMAT`

_Originally posted by @spacesailor24 in https://github.com/ChainSafe/web3.js/pull/4859#discussion_r832874388_

nikoulai:
## Description

#4892 

## Type of change

## Checklist:


packages/web3-eth/src/utils/transaction_builder.ts
```diff
@@ -75,7 +76,7 @@ export const getTransactionType = (
 // Keep in mind that the order the properties of populateTransaction get populated matters
 // as some of the properties are dependent on others
 export async function defaultTransactionBuilder<ReturnType = Record<string, unknown>>(options: {
-	transaction: Record<string, unknown>;
```
Spacesai1or: This line should be reverted, the line that was really in question regarding the use of `InternalTransaction` was [this one](https://github.com/ChainSafe/web3.js/blob/6d87ef347c548e875b7cc58df13708f3c1fae2fa/packages/web3-eth/src/utils/transaction_builder.ts#L83)

The reason why this was deemed an issue is because `InternalTransaction` states that the expected transaction properties would be hex strings (for the appropriate properties e.g. `nonce`, `data`, `chainId`, etc.), but of course the use of `Record<string, unknown>` doesn't guarantee that. I don't think we should change the type of the transaction parameter to type `Transaction`, because the type signature of `defaultTransactionBuilder` needs to remain open to any object structure (since developers can provider their own implementation for a transaction builder), so the type should be reverted back to `Record<string, unknown>`

As mentioned, it's this line:

```typescript
let populatedTransaction = { ...options.transaction } as unknown as InternalTransaction;
```

Where we should be performing:

```typescript
format(transactionSchema, populatedTransaction, DEFAULT_RETURN_FORMAT)
```

That way we can enforce that all the expected properties on `transaction` we are to use in our implementation of `defaultTransactionBuilder` are formatted to `DEFAULT_RETURN_FORMAT` (hex strings) before we use them

avkos (approved):

nazarhussain (approved):

jdevcs (approved):

Issue #4880, opened 2022-03-22
Web3 Eth - 2 - 4.0.0 Alpha.1 System Tests


jdevcs:
System tests for:

## Web3.eth package

- [ ] sendTransaction, 
- [ ] sendSignedTransaction, 
- [ ] sign, 
- [ ] signTransaction, 
- [ ] call 
- [ ] estimateGas,  submitWork
- [x] defaults Being done in #5082
- [x] BatchRequests~ Being done in #5015

#### Min Platform Requirements for 4.x Alpha:
- Node: v14, v16
- Browsers
- Eth Clients: Geth, Ganache

More details are in [Epic](https://github.com/ChainSafe/web3.js/issues/4711)

mconnelly8:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/ChainSafe/web3.js/4880?planning-poker) with ZenHub @avkos @jdevcs @luu-alex @nikoulai @spacesailor24

Spacesai1or:
FYI, the RPC method for `sendSignedTransaction` is actually `sendRawTransaction` (Web3.js just aliases it)

PR #5056, opened 2022-05-20
`web3-eth` Integration Tests (#4880)

Spacesai1or:
### What This PR Does

- Init integration tests:
    - `call`
    - `estimate_gas`
    - `send_signed_transaction`
    - `send_transaction`
    - `sign`
    - `sign_transaction`
    - `submit_work`
- Updates `sendSignedTransaction` (these changes were found because they were causing the tests to hang with unfulfilled promises)
    - Missing a couple of `if (promiEvent.listenerCount('xxxx') > 0)` listener wrappers
    - Missing the resolver override functionality that was added to `sendTransaction`
    - Added mock resolved values to `getTransactionReceipt` in response to this change in `packages/web3-eth/test/unit/rpc_method_wrappers/send_signed_transaction.test.ts`
- Adds `type` to `receiptInfoSchema` and `ReceiptInfo` type (in `packages/web3-eth/src/types.ts`)
- Adds `TransactionWithLocalWalletIndex` type (in `packages/web3-eth/src/types.ts`)
    - This is just the already existing `Transaction` type but with `from` defined as `from?: Numbers`
- Init `LocalWalletNotAvailableError` (in  `packages/web3-eth/src/errors.ts`)
- Update `sendTransaction` (in `packages/web3-eth/src/rpc_method_wrappers.ts`) to better handle locally indexed accounts

#### New Issues

- #5077  (partially addressed in this PR)


packages/web3-eth/test/integration/web3_eth/submit_work.test.ts
```diff
@@ -0,0 +1,44 @@
...
+		}
+	});
+
+	it('should submit work', async () => {
+		const response = await web3Eth.submitWork(
+			'0x0000000000000001',
+			'0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef',
+			'0xD1FE5700000000000000000000000000D1FE5700000000000000000000000000',
+		);
+		expect(response).toBe(false);
```
Spacesai1or, line 48: I'm not sure how we could have a test that submits valid work data so that the response is `true`

jdevcs: With EL and CL merge this function will be discarded as POW will be removed. But I think for other evm POW based NWs if any one want to use web3.js we can test it in some future release:

0. Infrastructure setup ( fully synced local testnet Ethereum client that is enabled for mining )

1. get results of eth_getWork
Returns the hash of the current block, the seedHash, and the boundary condition to be met (“target”).

2. use some external process for calculation of correct nonce , new proposed block hash and mix digest.( May be some external mining module like https://github.com/ethereum-mining/ethminer#ethminer )

3. submit that via submitWork

jdevcs: its low priority , we can skip it

packages/web3-eth/test/integration/web3_eth/call.test.ts
```diff
@@ -0,0 +1,79 @@
...
+	it('should make a call to deployed Greeter contract', async () => {
+		const expectedEncodedGreet =
+			'0x00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000017736f6c79656e7420677265656e2069732070656f706c65000000000000000000';
+		const expectedDecodedGreet = 'solyent green is people';
+		const greetCallData = '0xcfae3217';
+		const transaction: TransactionCall = {
+			to: greeterContractAddress,
+			data: greetCallData,
+		};
+		const response = await web3Eth.call(transaction);
```
jdevcs, line 73: We should also test it without using defaultBlock and passing it explicitly, and also with modified web3.eth.defaultBlock

packages/web3-eth/test/integration/web3_eth/send_signed_transaction.test.ts
```diff
@@ -0,0 +1,82 @@
...
+			type: '0x0',
+			gas: '0x5208',
+		};
+		const gasPricing = await getTransactionGasPricing(
+			transaction,
+			web3Eth.getContextObject() as unknown as Web3Context<EthExecutionAPI>,
+			DEFAULT_RETURN_FORMAT,
+		);
+		const signedTransaction = await web3Eth.signTransaction({ ...transaction, ...gasPricing });
+		const response = await web3Eth.sendSignedTransaction(signedTransaction.raw);
```
jdevcs: In future we should add tests for following promi events:

- sending 
- sent 
- transactionHash 
- receipt 
- confirmation 
- error

packages/web3-eth/test/integration/web3_eth/send_transaction.test.ts
```diff
@@ -0,0 +1,79 @@
...
+		}
+	});
+
+	it('should make a simple value transfer', async () => {
+		const transaction: Transaction = {
+			from: accounts[0],
+			to: '0x0000000000000000000000000000000000000000',
+			value: '0x1',
+		};
+		const response = await web3Eth.sendTransaction(transaction);
```
jdevcs, line 45: we should also test `sendTransaction` by using `web3.eth.defaultAccount `and modified defaultAccount, and using address of account of local wallet.

jdevcs, line 51: we should also have `0x2` type transaction tests for sendTransaction, along with tests of modified defaults ( chain, hardfork )

jdevcs, line 51: we should also test `sendTransaction` by using `web3.eth.defaultAccount `and modified defaultAccount, and using `address` of account of local wallet.

jdevcs (changes requested):

packages/web3-eth/test/integration/web3_eth/submit_work.test.ts
Spacesai1or: Created an [issue](https://github.com/ChainSafe/web3.js/issues/5064) to track this

packages/web3-eth/test/integration/web3_eth/call.test.ts
Spacesai1or: Addressed by [this](https://github.com/ChainSafe/web3.js/pull/5056/commits/01d49e28028a51392582c70b53b08f2a8507cdaa) commit

Spacesai1or: @jdevcs  Regarding modifying the defaults of `Web3Context`, I think that's better handled by explicit tests that @avkos is working on, instead of including these tests for every RPC method in `Web3Eth`

packages/web3-eth/test/integration/web3_eth/send_transaction.test.ts
Spacesai1or: Addressed by [this](https://github.com/ChainSafe/web3.js/pull/5056/commits/1d77c4c141c97999f2ebbb19b4371a0184c07f82) commit

packages/web3-eth/test/integration/web3_eth/send_transaction.test.ts
```diff
@@ -0,0 +1,121 @@
...
+		it('should send a successful type 0x0 transaction', async () => {
+			const transaction: Transaction = {
+				from: accounts[0],
+				to: '0x0000000000000000000000000000000000000000',
+				value: '0x1',
+				type: '0x0',
+			};
+			const response = await web3Eth.sendTransaction(transaction);
+			expect(response.type).toBe('0x0');
+			expect(response.status).toBe('0x1');
```
jdevcs, line 162: Along with tx status, I think we should also check that provided from, to, value and type match with mined tx via `eth_getTransactionReceipt`, for testing any of these are not modified by lib or dependent libs.

jdevcs: same suggested items should be checked for following tests of 0x1 and 0x2 txs

jdevcs (changes requested):

packages/web3-eth/test/integration/web3_eth/send_transaction.test.ts
```diff
@@ -0,0 +1,121 @@
...
+			const transaction: Transaction = {
+				from: accounts[0],
+				data: greeterContractDeploymentData,
+				gas: '0x740b8',
+			};
+			const response = await web3Eth.sendTransaction(transaction);
+			expect(response.status).toBe('0x1');
```
jdevcs, line 119: we should also check in mined tx, provided `from` and `data` match. For verification that its not modified in our lib or any depend libs. Also check that value is 0.

packages/web3-eth/test/integration/web3_eth/send_transaction.test.ts
```diff
@@ -0,0 +1,121 @@
...
+				type: '0x0',
+			};
+			const response = await web3Eth.sendTransaction(transaction);
+			expect(response.type).toBe('0x0');
+			expect(response.status).toBe('0x1');
+		});
+
+		it('should send a successful type 0x1 transaction', async () => {
+			const transaction: Transaction = {
+				from: accounts[0],
```
jdevcs, line 170: accounts[0] will provide address. In this test could you directly use local wallet account index, this will cover one more scenario in e2e, and match with mined tx `from` field in receipt.

packages/web3-common/src/eth_execution_api.ts
```diff
@@ -281,7 +286,7 @@ export type EthExecutionAPI = {
 	eth_sign: (address: Address, message: HexStringBytes) => HexString256Bytes;
 	eth_signTransaction: (
 		transaction: TransactionWithSender | Partial<TransactionWithSender>,
-	) => HexStringBytes;
+	) => HexStringBytes | SignedTransactionInfo;
```
Spacesai1or, line 290: Geth seems to return this `SignedTransactionInfo` object, even though the ETH spec says it should be `HexStringBytes`

packages/web3-eth/package.json
```diff
@@ -40,7 +40,8 @@
 		"prettier": "^2.4.1",
 		"ts-jest": "^27.0.7",
 		"typescript": "^4.5.2",
-		"web3-providers-http": "4.0.0-alpha.1"
+		"web3-providers-http": "4.0.0-alpha.1",
+		"web3-eth-abi": "4.0.0-alpha.0"
```
Spacesai1or: Added `web3-eth-abi` dev dependency to use `decodeParameters` in tests like [eth_call](https://github.com/ChainSafe/web3.js/blob/ad85e4256d2ee6017732c2c48b077d706c3cdde8/packages/web3-eth/test/integration/web3_eth/call.test.ts#L96)

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -395,6 +406,8 @@ export function sendTransaction<
 					) {
 						transactionFormatted = {
 							...transactionFormatted,
+							// TODO gasPrice, maxPriorityFeePerGas, maxFeePerGas
```
Spacesai1or, line 519: This is a point of discussion, `getTransactionGasPricing` is going to return an object:

```typescript
{
    gasPrice: HexString,
    maxPriorityFeePerGas: HexString,
    maxFeePerGas: HexString,
}
```

and either `gasPrice` or `maxPriorityFeePerGas` and `maxFeePerGas` are going to be `undefined`. So unpacking the object (`...{}`) results in the `undefined` property being added to the transaction object that gets submitted to the RPC node. This is generally not going to be a problem for most users, but it personally doesn't sit right with me to add unnecessary undefined properties to a user's object before submission - the library should modify user data the least amount

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -627,9 +669,12 @@ export async function signTransaction<ReturnFormat extends DataFormat>(
 		web3Context.requestManager,
 		formatTransaction(transaction, DEFAULT_RETURN_FORMAT),
 	);
+	const unformattedResponse = isString(response as HexStringBytes)
```
Spacesai1or, line 801: This change is what's handling the `SignedTransacitonInfo`:

```typescript
export interface SignedTransactionInfo {
	raw: HexStringBytes;
	tx: TransactionSigned;
}
```

that Geth returns. Our library method is supposed to return the same `SignTransactionInfo` object, but in this current implementation, there is a difference between the object we return and Geth returns. Geth returns an object such as:

```typescript
{
          type: '0x0',
          nonce: '0x0',
          gasPrice: '0x3b9aca01',
          maxPriorityFeePerGas: null,
          maxFeePerGas: null,
          gas: '0x5208',
          value: '0x1',
          input: '0x',
          v: '0xa95',
          r: '0xdf1960615fd05d894b67c49e03b7bc7e3908ad00b44f25ae85c17a99f47e577a',
          s: '0x327aa2eb8269907dfcca27582e391a764e49c04a8ef79757a06ccf23ddd74987',
          hash: '0x21adb30975cc47c25b28eda40bc940da309990aafec0a831f5627f848e64091d'
}
``` 

While the library returns whatever the user provided to the method (so we'll always be missing properties such as `v`, `r`, `s`, `hash`)

packages/web3-eth/src/rpc_method_wrappers.ts
spacesailor24: This also means that the transaction object that gets emitted for `sending` and `sent` PromiEvents will not match what the user provided as input to the method

packages/web3-eth/test/integration/web3_eth/send_signed_transaction.test.ts
Spacesai1or: Addressed in [this](https://github.com/ChainSafe/web3.js/pull/5056/commits/96a186ebb56441e2ffdae5e185405f0e324ba892) (minus `error` handling)

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: We could add these missing properties to the user's transaction object before returning when only the raw signed transaction is returned from the RPC client

packages/web3-eth/test/integration/web3_eth/send_transaction.test.ts
Spacesai1or: Addressed by [this](https://github.com/ChainSafe/web3.js/pull/5056/commits/d77232fb8c24003dd8d03830ab00bcc97f510495) commit

packages/web3-eth/test/integration/web3_eth/send_transaction.test.ts
Spacesai1or: Addressed by [this](https://github.com/ChainSafe/web3.js/pull/5056/commits/290b8b3a02da93af8f91eb4b7d39b7938b762063) commit

packages/web3-eth/test/integration/web3_eth/send_transaction.test.ts
Spacesai1or: Addressed by [this](https://github.com/ChainSafe/web3.js/pull/5056/commits/2de671fbcab7c82e5de45cf31f039af33bec804f) commit

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -377,16 +394,42 @@ export function sendTransaction<
...
+					}
+
+					let transactionFormatted = formatTransaction(
+						{
+							...transaction,
+							from: transactionFrom,
+						},
+						DEFAULT_RETURN_FORMAT,
+					);
+
```
Spacesai1or: `formatTransaction` doesn't support `transaction.from` being anything other than `HexString | undefined`, and it doesn't make sense for it to be updated to handle a local wallet index, because then it would need access to `Web3Context` or a wallet. So instead, this code is enforcing that `transactionFrom` is either a `string | undefined` where the `string` is sourced from either `transaction.from` (if it's a `HexString` address, we could consider allowing this to be other valid bytes values) or `web3Context.wallet.get().address` - this code was a result of writing a `sendTransaction` test with a index for `transaction.from`

Spacesai1or: FYI this is was abstracted a little bit to make use of the [getTransactionFromAttr](https://github.com/ChainSafe/web3.js/blob/d7e88f645e086ebb82764a8d31835fd7f149fd48/packages/web3-eth/src/utils/transaction_builder.ts#L50) method

avkos (approved):

packages/web3-eth/src/rpc_method_wrappers.ts
nikoulai: Do you believe a check for null and removal is enough? or we are going to handle anything extra (unwanted effects)?

nikoulai (approved):

nazarhussain (approved):

jdevcs (approved):

luu-alex (approved):

PR #5063, opened 2022-05-23
Add typings for contract

Algoritm211:
## Description
The main goal of this PR is to add support for typing events and contract methods. At the moment, almost all TypeScript contract developers generate types for their contracts (using the built-in [hardhat](https://www.npmjs.com/package/hardhat) functionality or libraries like [ethereum-abi-types-generator](https://www.npmjs.com/package/ethereum-abi-types-generator))

In addition, I noticed **_TODO_** on adding types to the contract and decided to do it. I am sure that this innovation will help to interact more conveniently with web3js by passing methods and events of their contracts at the time of creating a web3 instance

This greatly simplifies development, especially considering the auto-complete offered by IDEs such as WebStorm and VSCode.

### Before:
Снимок экрана 2022-05-24 в 00 47 20

### After:
Снимок экрана 2022-05-24 в 00 46 48

As we can see, we have received full typing of our contract methods. I think this is amazing and should be included in a new library update as soon as possible.

## Type of change

## Checklist:


Algoritm211:
@jdevcs Can you please review it? Maybe you have some comments

jdevcs:
@Algoritm211 Thanks for your contribution, In [Web3.js 4.x](https://github.com/ChainSafe/web3.js/tree/4.x) we will be providing full types support. Could you have time for contributions in web3.js 4.x.

packages/web3-eth-contract/types/index.d.ts
```diff
@@ -22,8 +22,8 @@ import {Common, PromiEvent, provider, hardfork, chain, BlockNumber, PastLogsOpti
 import {Accounts} from 'web3-eth-accounts';
 import {AbiItem} from 'web3-utils';
 
-// TODO: Add generic type!
-export class Contract {
+// Here you can provide generated types for methods and events from your contract
+export class Contract<Methods = any, Events = any> {
```
nazarhussain, line 26: What are the constraints on `Methods` and `Events`? What are the params and return types for those methods?  Without constraints it will generate more confusions for the users. 

For `1.x` I will encourage to use [Typechain](https://github.com/dethcrypto/TypeChain/tree/master/packages/target-web3-v1) for exactly this behavior. 

For `4.x` we already had provided strong type-safety for contracts. 

I would like to close this PR, what do you suggest @jdevcs.

Algoritm211: I don't think there should be any restrictions. Everyone will generate types with different libraries, methods will get something like this object:

```typescript
 interface ContractTypes {
  /**
   * Payable: true
   * Constant: false
   * StateMutability: payable
   * Type: function
   */
  enter(): MethodPayableReturnContext;
  /**
   * Payable: false
   * Constant: true
   * StateMutability: view
   * Type: function
   */
  getManager(): MethodConstantReturnContext<string>;
  /**
   * Payable: false
   * Constant: true
   * StateMutability: view
   * Type: function
   */
  getPlayers(): MethodConstantReturnContext<string[]>;
  /**
   * Payable: false
   * Constant: false
   * StateMutability: nonpayable
   * Type: function
   */
  pickWinner(): MethodReturnContext;
  /**
   * Payable: false
   * Constant: true
   * StateMutability: view
   * Type: function
   * @param parameter0 Type: uint256, Indexed: false
   */
  players(parameter0: string): MethodConstantReturnContext<string>;
}
```

It will be clear and comfortable, I can add some examples in the code

nazarhussain: As explained earlier specifying generics without any constraints results into misuse and confusion. e.g. below would be a valid type code but invalid for the usage. 

```
const contract = new web3.eth.Contract<number>();
```

If we go with this direction, then we need to have some constraints on those generics so the usage would stay safe. 

But it depends if we decided to proceed in that direction as we are working hard to release 4.x alpha very soon and it have all these types issues covered.

Algoritm211: Ok, agree, in a common case it will be`Record<string, unknown>` or do you have another idea how to add typings to it?

Algoritm211:
> This PR has been automatically marked as stale because it has not had recent activity. It will be closed in 7 days if no further activity occurs. Thank you for your contributions. If you believe this was a mistake, please comment.

I think it's actual

jdevcs:
Closing this in favor of web3 4.x

Issue #5072, opened 2022-05-25
Update BN.js dependecy to v5.2.1


gregorfelkar:
### This is not a Bug Report, Feature Request, or related to Documentation

- [X] I have searched the existing issues

### Is there an existing issue for this?

- [X] I have searched the existing issues

### What's up?

* Fix serious issue in `.toString(16)`

The hex encoding of some numbers is wrong. Since the string representation
is also used for interoperability between BigInt libraries, this bug is
causing a number modification when a BN.js number is converted into
another BigInt class, for example the BigNumber of ethers.js.

https://github.com/indutny/bn.js/commit/042ab62e70418c15f189b45460709a51faf303cc

Currently it will install:
```
│ └─┬ web3-utils@1.7.3
│   ├── bn.js@4.12.0
```

plusminushalf:
Hey I am new to the repo, will this be a good first issue? If yes I would like to pick this up.

luu-alex:
Thank you for this @gregorfelkar this looks serious and we'll update BN as soon as we can

Issue #5071, opened 2022-05-24
Investigate signtransaction testcases


luu-alex:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

1. For this [testcase](https://github.com/ChainSafe/web3.js/blob/1.x/test/eth.accounts.signTransaction.js#L167) the decoded rawTransaction does not match the return value of r and s from function account signtransaction. 
rawTransaction: `0xf85f800a827c6d94f0109fc8df283027b6285cc889f5aa624eac1f55808025a03cbfff5b8ef4588b930ecbf9b85388795875edf814dfc6c71884f99b6d7555cca057142e729c1c83bfccb2785e629fc32dffb2e613df565e78e119aa4694cb1df9`
Decoded rawTransaction: 
```
{
  "nonce": 0,
  "gasPrice": {
    "_hex": "0x0a"
  },
  "gasLimit": {
    "_hex": "0x7c6d"
  },
  "to": "0xf0109fc8df283027b6285cc889f5aa624eac1f55",
  "value": {
    "_hex": "0x0"
  },
  "data": "0x",
  "v": 37,
  "r": "0x3cbfff5b8ef4588b930ecbf9b85388795875edf814dfc6c71884f99b6d7555cc",
  "s": "0x57142e729c1c83bfccb2785e629fc32dffb2e613df565e78e119aa4694cb1df9"
}
```
While the r value returned is `0x22f17b38af35286ffbb0c6376c86ec91c20ecbad93f84913a0cc15e7580cd9` and s value is `0x83d6e12e82e3544cb4439964d5087da78f74cefeec9a450b16ae179fd8fe20`

### Expected Behavior

decoded rawTransaction from testcase should match r and s value.

### EDIT Part 2 is complete. 

2. For this post EIP1559 [testcase] (https://github.com/ChainSafe/web3.js/blob/1.x/test/eth.accounts.signTransaction.js#L656)
we receive different a `rawTransaction` when maxgas and priorityFee is type bn.
RawTransaction when `maxPriorityFeePerGas` and `maxFeePerGas` is type BN:
`0x02f86e018084c733124884cb72ec20826a4094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca0080c001a08896fb9a5c033e0163b073cf7a951a1db2dca41b26b4188f13a05158eb26fd32a005e8855691199cd0b6dcae88f3325c374e2f0697b9c528a5c10d5bd8dfb6a3e3`
[...]

PR #5074, opened 2022-05-25
5072 - updating bn 1.x

luu-alex:
## Description
#5072

## Type of change

## Checklist:


jdevcs (commented):
We will also need to update https://github.com/ChainSafe/web3.js/blob/2c0324af2da467ee1acb72452f20000e916e4306/packages/web3-utils/src/utils.js#L273 and might need to remove `number-to-bn` dependency. also update ethers to 5.6.8

luu-alex:
@jdevcs  the dependancy `number-to-bn` uses `bigNumber.js` so that should be fine? will update the others

luu-alex:
The failing testcase that is commented is related to this issue #5071. will need to investigate further

test/eth.accounts.signTransaction.js
```diff
@@ -652,29 +652,29 @@ var tests = [
...
-            data: "",
-            common: commonLondon
-        },
-        // signature from eth_signTransaction
-        rawTransaction: "0x02f86e018084c733124884cb72ec20826a4094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca0080c001a08896fb9a5c033e0163b073cf7a951a1db2dca41b26b4188f13a05158eb26fd32a005e8855691199cd0b6dcae88f3325c374e2f0697b9c528a5c10d5bd8dfb6a3e3",
-        oldSignature: "0x02f86e018084c733124884cb72ec20826a4094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca0080c001a08896fb9a5c033e0163b073cf7a951a1db2dca41b26b4188f13a05158eb26fd32a005e8855691199cd0b6dcae88f3325c374e2f0697b9c528a5c10d5bd8dfb6a3e3",
-        transactionHash: "0xd5b7290a477b9c421d39e61d0f566ec33276fb49b9ff85cfd6152a18f1c92dab",
-        messageHash: "0x17e20e530a889ce52057de228b5b97edcad6002468d723346cd0b6b7a9943457"
-    },
+    // {
```
avkos: why the test is commented on?

avkos (changes requested):

test/eth.accounts.signTransaction.js
luu-alex: This testcase is not correct, we were using bn version 4.11.9 and an error was [found](https://github.com/indutny/bn.js/issues/151) parsing hexstirngs incorrectly. it has been incorrectly converting hexstrings to decimals. 
```
> var a = new web3.utils.BN("0x3B9ACA00", 16)
undefined
> a
BN { negative: 0, words: [ 60475904, 526 ], length: 2, red: null }
> 
```
But `0x3B9ACA00` should be `1000000000`

So with the new bn.js update for our dependancies, there was a [breaking change](https://github.com/indutny/bn.js/releases/tag/v5.0.0) and now this testcase is failing and i had to comment it out. Will need to investigate this

nikoulai (approved):

test/eth.accounts.signTransaction.js
nazarhussain: I believe it's a very important test in `signTransaction` so we should not disable it, but rather investigate it and also include its fix in the same release. 

If the upgrade of `bn.js` is causing some misbehavior to transaction signature then we should report the issue to them and don't upgrade `bn.js` right away.

package.json
```diff
@@ -101,9 +101,10 @@
         "@types/bignumber.js": "^4.0.2",
         "@types/bn.js": "^4.11.6",
         "@types/node": "^12.12.68",
+        "@types/prettier": "2.6.0",
         "babel-loader": "^8.1.0",
         "bignumber.js": "^9.0.1",
-        "bn.js": "^4.11.9",
+        "bn.js": "^5.2.1",
```
nazarhussain, line 107: We need to update its type as well. 

https://github.com/ChainSafe/web3.js/blob/2c0324af2da467ee1acb72452f20000e916e4306/package.json#L102

nazarhussain (changes requested):

test/eth.accounts.signTransaction.js
nikoulai: From the tests I have run, the misbehavior seems to be only in the 4.11.9 version, so I just updated the hex values in the test and reenabled it.

I noticed that ethereumjs-utils uses bn:4.11.6, maybe, we should let them know

nikoulai (commented):
Awesome!

In 4.x I saw something like that for the handling negative hex:

```
 if (typeof value == "string" && value.includes("0x")) {
        const [negative, hexValue] = value.toLocaleLowerCase().startsWith('-') ? ["-", value.slice(3)] : ["", value.slice(2)];
        return new BN(negative + hexValue, 16);
    }
    else {
        return new BN(value);
    }
```

What would you say to refactor?

test/eth.accounts.signTransaction.js
luu-alex: test is uncommented and works fine now with @nikoulai's [fix](https://github.com/ChainSafe/web3.js/pull/5074/files#diff-bd97e9c608060469030c1d6a655baa533c0aa0b11939b0670ebdb41aa51d786bR663), the issue is `bn` errors when given strings starting with`0x`, we shouldn't have passed ` bn('0x3B9ACA00')` but instead it should have been `bn('cb72ec20', 16),`

packages/web3-utils/src/utils.js
```diff
@@ -534,8 +534,29 @@ var toNumber = function(value) {
     return typeof value === 'number' ? value : hexToNumber(toHex(value));
 }
 
+// 1.x currently accepts 0x... strings, bn.js after update doesn't. it would be a breaking change
```
luu-alex, line 537: Thinking about this now we should probably remove this, I dont think we should be changing how BN is being created... the [test case](https://github.com/ChainSafe/web3.js/pull/5074/files#diff-bd97e9c608060469030c1d6a655baa533c0aa0b11939b0670ebdb41aa51d786bR663) is working fine and this is adding an extra feature this PR should just be focusing on upgrading and fixing `BN`.

nikoulai: You are right about the scope, but, shouldn't we handle the <change> introduced too? (you mean to do it in another PR? )

luu-alex:
Added an extra [test](https://github.com/ChainSafe/web3.js/pull/5074/files#diff-bd97e9c608060469030c1d6a655baa533c0aa0b11939b0670ebdb41aa51d786bR702), works properly as in [4.x](https://github.com/ChainSafe/web3.js/pull/5070/files#diff-285b5b91489566a7887cb8efb734e5abc35f3b2f1d42d960b077b44d9f48bef6R834) as well.

nazarhussain (approved):

jdevcs (approved):

avkos (approved):

wbt:
FYI, this turned out to be a breaking change on nonbreaking semver, as discussed in more detail in the above-linked comment, and possibly foreshadowed [here](https://github.com/ChainSafe/web3.js/pull/5132).

arijoon:
As @wbt noted, this was a breaking change. When exposing deps directly (such as `BN`), please bump major version on web3 if any of these deps have a major upgrade

PR #4868, opened 2022-03-18
`e2e_windows` Test Failure Fix for #4808

Spacesai1or:
Merge #4808 into a new branch, so that the `e2e_windows` test may have access to the Infura key

PR #5085, opened 2022-05-29
web3-providers-http: Migrate from xhr2-cookies to cross-fetch

ghost:
closes #4868

I have replaced the old `xhr2-cookies` library to `cross-fetch` since it was better, and the `npm i` for `xhr2-cookies` was broken for the latest node LTS version as well.

Working perfectly like before.

### DIFF
--- a/packages/web3-providers-http/src/index.js
+++ b/packages/web3-providers-http/src/index.js
@@ -139,12 +139,6 @@ HttpProvider.prototype.send = function (payload, callback) {
             // Response is a stream data so should be awaited for json response
             result.json().then(function(data) {
                 result = data;
-                if (!isOk) {
-                    _this.connected = false;
-                    error = errors.InvalidResponse(data);
-                    callback(error, result);
-                    return;
-                }
                 _this.connected = true;
                 callback(error, result);
             });
--- a/packages/web3-providers-http/src/index.js
+++ b/packages/web3-providers-http/src/index.js
@@ -132,7 +132,6 @@ HttpProvider.prototype.send = function (payload, callback) {
             clearTimeout(this.timeoutId);
         }
         var result = response;
-        var isOk = result.ok;
         var error = null;
 
         try {
--- a/packages/web3-providers-http/src/index.js
+++ b/packages/web3-providers-http/src/index.js
@@ -161,14 +161,9 @@ HttpProvider.prototype.send = function (payload, callback) {
         callback(errors.InvalidConnection(this.host));
     }
 
-    try {
-        request(payload)
-            .then(success.bind(this))
-            .catch(failed.bind(this));
-    } catch(error) {
-        this.connected = false;
-        callback(errors.InvalidConnection(this.host));
-    }
+    request(payload)
+        .then(success.bind(this))
+        .catch(failed.bind(this));
 };
 
 HttpProvider.prototype.disconnect = function () {
--- a/packages/web3-providers-http/src/index.js
+++ b/packages/web3-providers-http/src/index.js
@@ -103,11 +103,7 @@ HttpProvider.prototype._prepareRequest = function(payload = {}){
         headers['Content-Type'] = 'application/json';
     }
 
-    // https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#sending_a_request_with_credentials_included
-    if (['include', 'same-origin', 'omit'].indexOf(this.withCredentials) !== -1) {
-      options.credentials = this.withCredentials;
-    }
-
+    options.credentials = this.withCredentials;
     options.headers = headers;
 
     if (this.timeout > 0) {
@@ -142,13 +138,16 @@ HttpProvider.prototype.send = function (payload, callback) {
             result.json().then(function(data) {
                 result = data;
                 if (!isOk) {
+                    this.connected = false;
                     error = errors.InvalidResponse(data);
+                    callback(error, result);
+                    return;
                 }
                 this.connected = true;
                 callback(error, result);
             });
         } catch (e) {
-            this.connected = true;
+            this.connected = false;
             callback(errors.InvalidResponse(result));
         }
     };
@@ -163,7 +162,7 @@ HttpProvider.prototype.send = function (payload, callback) {
             callback(errors.ConnectionTimeout(this.timeout));
         }
 
-        this.connected = true;
+        this.connected = false;
         callback(errors.InvalidConnection(this.host));
     }
 
--- a/packages/web3-providers-http/src/index.js
+++ b/packages/web3-providers-http/src/index.js
@@ -56,7 +56,14 @@ var HttpProvider = function HttpProvider(host, options) {
     }
 };
 
-HttpProvider.prototype._prepareRequest = function(payload = {}){
+/**
+ * Should be used to make async request
+ *
+ * @method send
+ * @param {Object} payload
+ * @param {Function} callback triggered on end with (err, result)
+ */
+HttpProvider.prototype.send = function (payload, callback) {
     var options = {
       method: 'POST',
       body: JSON.stringify(payload)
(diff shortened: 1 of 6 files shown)


packages/web3-providers-http/src/index.js
```diff
@@ -19,29 +19,33 @@
...
 
 var errors = require('web3-core-helpers').errors;
-var XHR2 = require('xhr2-cookies').XMLHttpRequest; // jshint ignore: line
+var fetch = require('cross-fetch');
 var http = require('http');
 var https = require('https');
 
+// Apply missing polyfill for IE
+require('es6-promise').polyfill();
+require('abortcontroller-polyfill/dist/polyfill-patch-fetch');
```
ghost, line 33: I have added the polyfills in case of compatibility, but I think CI makes some complain about it

packages/web3-providers-http/src/index.js
```diff
@@ -19,29 +19,33 @@
...
 
 /**
  * HttpProvider should be used to send rpc calls over http
  */
 var HttpProvider = function HttpProvider(host, options) {
     options = options || {};
 
-    this.withCredentials = options.withCredentials || false;
+    this.withCredentials = options.withCredentials;
```
ghost, line 41: This is no longer boolean  https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#sending_a_request_with_credentials_included

packages/web3-providers-http/types/index.d.ts
```diff
@@ -30,7 +30,6 @@ export interface HttpHeader {
 }
 
 export interface HttpProviderAgent {
-    baseUrl?: string;
```
ghost, line 33: Have removed it since this isn't supported by fetch as well as this is not a standard for Http Agent

test/httpprovider.js
```diff
@@ -5,55 +5,78 @@ var https = require('https');
...
+    this.statusText = 'OK';
+  }
+};
+
+function Mock(url, options) {
+  const response = new Response(url, Object.entries(options.headers));
+  return new Promise(function(resolve, reject) {
+    resolve(response);
+  });
+};
```
ghost, line 28: Simple version of fetch-mock function added

test/httpprovider.js
```diff
@@ -67,14 +90,15 @@ describe('web3-providers-http', function () {
...
+        **/
     });
 
     describe('send', function () {
         it('should send basic async request', function (done) {
             var provider = new HttpProvider();
 
             provider.send({}, function (err, result) {
-                assert.equal(typeof result, 'object');
+                assert.equal(typeof result, 'undefined');
```
ghost, line 95: This should not be an object as we didn't had any url to fetch

packages/web3-providers-http/src/index.js
```diff
@@ -52,37 +56,67 @@ var HttpProvider = function HttpProvider(host, options) {
...
-    return request;
+    // Default headers
+    if (!headers['Content-Type']) {
+        headers['Content-Type'] = 'application/json';
+    }
+
+    // https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#sending_a_request_with_credentials_included
+    if (['include', 'same-origin', 'omit'].indexOf(this.withCredentials) !== -1) {
+      options.credentials = this.withCredentials;
+    }
```
nazarhussain: Why only these credentials? Why any value provided from the user?

packages/web3-providers-http/src/index.js
```diff
@@ -93,32 +127,50 @@ HttpProvider.prototype._prepareRequest = function(){
...
+            clearTimeout(this.timeoutId);
+        }
+
+        if (error.name === 'AbortError') {
+            this.connected = false;
+            callback(errors.ConnectionTimeout(this.timeout));
+        }
+
+        this.connected = true;
+        callback(errors.InvalidConnection(this.host));
```
nazarhussain: Seems invalid case, how an invalid connection can be considered connected?

packages/web3-providers-http/types/index.d.ts
```diff
@@ -46,8 +45,9 @@ export interface HttpProviderOptions {
 export class HttpProvider extends HttpProviderBase {
     host: string;
 
-    withCredentials: boolean;
+    withCredentials?: string;
     timeout: number;
+    timeoutId?: NodeJS.Timeout | number;
```
nazarhussain: Do we need to export it to types? It should be a implementation detail not exposed to users.

test/httpprovider.js
```diff
@@ -5,55 +5,78 @@ var https = require('https');
...
-        it('should use the passed custom https agent', function () {
-            agent = new https.Agent();
-            options = {agent: {https: agent}};
-            provider = new HttpProvider('http://localhost:8545', options);
-            result = provider._prepareRequest();
+        it('should use the passed custom https agent', async function () {
+            var agent = new https.Agent();
+            var options = {agent: {https: agent}};
+            var provider = new HttpProvider('http://localhost:8545', options);
+            var result = await provider._prepareRequest();
```
nazarhussain: Should not `_prepareRequest` create only a function which unless invoked should not return a promise?

nazarhussain (changes requested):

packages/web3-providers-http/src/index.js
ghost: @nazarhussain This was the standard credentials documented from Mozilla as I have wrote the link of the documentation to the comments, should we allow this to any type of string?

packages/web3-providers-http/src/index.js
ghost: @nazarhussain You are right, I think I misread the original code

packages/web3-providers-http/types/index.d.ts
ghost: @nazarhussain Right

test/httpprovider.js
ghost: @nazarhussain You mean we should return the function like () => fetch(this.host, options) instead of the promise itself?

ghost:
@nazarhussain Resolved those three comments

nikoulai (commented):
Please check the failing tests (except for e2e_windows)

ghost:
@jdevcs @avkos @luu-alex @spacesailor24 @nazarhussain @nikoulai I have fixed the testcase, would like to have a look at?

avkos (approved):

test/httpprovider.js
nazarhussain: I believe the `_prepareRequest` should not be `async`. It should return the request object synchronously. And later on we should be using `request.send()` to actually send and await. 

If that seems to be a big change, then we can rename the function to `_prepareAndSendRequest` that will make the context right.

nazarhussain (changes requested):
@ayanamidev Thanks for your contributions. Left one small comment. Rest seems good to me. 

There are two other points that need to address. 

1. We don't allow community contributions to change the `package-lock.json` file. So please remove the change from that file in your PR. Some team member will merge the PR and include the changes in `package-lock.json`. 
2. Add an entry to `CHANGELOG.md` describing what you are changing in this PR.

ghost:
@nazarhussain @nikoulai @spacesailor24 Would like to ask your thought about removing `_prepareRequest` function for `web3-providers-http`. This function was `xhr2-cookies` specific and doesn't export it to the wild. So I think it makes sense to remove with the shift to `cross-fetch` since it will only return promise with the current code.

ghost:
@nazarhussain @nikoulai @spacesailor24 Addressed comments

nazarhussain (changes requested):
@ayanamidev LGTM. Please add an entry into `CHANGELOG.md` describing the change.

ghost:
@nazarhussain Oh, sorry fixed.

nazarhussain (approved):

nazarhussain:
@ayanamidev Please resolve the conflicts. Those probably will be in CHANGELOG.md. Then I would be able to merge the PR right away.

ghost:
@nazarhussain Done. Every conflicts resolved without changes from source code

nikoulai (approved):

PR #5086, opened 2022-05-30
Fix odd length hex string conversion

### LINKED ISSUE #5073: `bytesToBuffer` does not work for 0 and 1 integer values

### Current Behavior

```typescript
bytesToBuffer(numberToHex(1))
```

returns empty buffer `<Buffer >`

### Expected Behavior

```typescript
bytesToBuffer(numberToHex(1))
```

should return `<Buffer 01>`

### Steps to Reproduce

```typescript
bytesToBuffer(numberToHex(1))
```

### Web3.js Version

4.x

### Environment

- Operating System: 
- Browser: 
- Node.js Version: 
- NPM Version: 

### Anything Else?

The following code is responsible for the issue. 

https://github.com/ChainSafe/web3.js/blob/82574a2587810f9bea26ba65da48fe667162e573/packages/web3-utils/src/converters.ts#L88-L90

nikoulai:
## Description

#5073 

## Type of change

## Checklist:


packages/web3-utils/src/converters.ts
```diff
@@ -86,7 +86,8 @@ export const bytesToBuffer = (data: Bytes): Buffer | never => {
 	}
 
 	if (typeof data === 'string' && isHexStrict(data)) {
-		return Buffer.from(data.slice(2), 'hex');
+		const finalData = data.length % 2 === 0 ? data : `0${data}`;
```
nazarhussain: This code block executes when `data` is already prefixed with `0x`, so returning `0${data}` will end up wrong hex string. e.g. If `data='0x1` it will return `00x1`.

Also add a few tests for this specific case.

nazarhussain (changes requested):

avkos (approved):

packages/web3-utils/src/converters.ts
avkos:
```suggestion
		const finalData = data.replace(/^(?!0x)/,'0x') ;
```

avkos (changes requested):

packages/web3-utils/src/converters.ts
avkos: It’s just stating if any starting does not start with 0x, just replace the start with 0x.
Also, add a few test cases for this. It’s very tricky but essential code.

nikoulai: I had in mind the input from `numberToHex`, very good point thanks!!
I didn't find any tests for `bytesToBuffer`, I'll add them including this case

nikoulai: I could use this function in preprocessing of `data` variable to ensure it starts with `0x`, but, what I want to do is to ensure that the length of the string is even, in order to `Buffer.from` work as wanted.

packages/web3-utils/src/converters.ts
```diff
@@ -86,7 +86,8 @@ export const bytesToBuffer = (data: Bytes): Buffer | never => {
 	}
 
 	if (typeof data === 'string' && isHexStrict(data)) {
-		return Buffer.from(data.slice(2), 'hex');
+		const finalData = `${/^(..)*$/.test(data) ? '' : '0'}${data.replace(/0(x|X)/, '')}`;
```
nazarhussain: Mixing up the string interpolation with the regular expressions does make it difficult to read. What do you think about this approach? 

```suggestion
        const eventLength = data.length + (data.length % 2);
		const finalData = data.toLowerCase().replace('0x', '').padStart(eventLength, '0')
```

packages/web3-utils/test/fixtures/converters.ts
```diff
@@ -267,3 +267,15 @@ export const toCheckSumValidData: [string, string][] = [
...
 ];
+
+export const bytesToBufferInvalidData: [any, string][] = bytesToHexInvalidData;
+
+export const bytesToBufferValidData: [Bytes, Buffer][] = [
+	[new Uint8Array([72]), Buffer.from('48', 'hex')],
+	[new Uint8Array([72, 12]), Buffer.from('480c', 'hex')],
+	['0x9c12', Buffer.from('9c12', 'hex')],
+	['0X12c6', Buffer.from('12c6', 'hex')],
+	['0X1', Buffer.from('01', 'hex')],
```
nazarhussain, line 280:
```suggestion
	['0X1', Buffer.from('01', 'hex')],
	['0x1', Buffer.from('01', 'hex')],
	['0x0', Buffer.from('00', 'hex')],	
    ['0X0', Buffer.from('00', 'hex')],	
    ['1234', Buffer.from('1233', 'hex')],	
```

nazarhussain: Also add these invalid use cases as well. 

```
    ['1', Buffer.from('01', 'hex')],	    
    ['0', Buffer.from('00', 'hex')],	    

```

nazarhussain (changes requested):

avkos (approved):
Approve, except for comments from Nazar

packages/web3-utils/src/converters.ts
nikoulai: Of course, I'm ok with change! I made it, I just made a modification, because `eventLength` included the `0x` prefix and finally would pad a string with '000`  or  `00`,

packages/web3-utils/test/fixtures/converters.ts
nikoulai: Add some of them, but, made some changes

As valid I added:
['0x1234', Buffer.from('1233', 'hex')] instead of ['1234', Buffer.from('1233', 'hex')]

and in invalids:
['1', Buffer.from('01', 'hex')],	    
['0', Buffer.from('00', 'hex')] 
I changes the `Buffer.from` to the error message.
Are these changes ok?

luu-alex (approved):

jdevcs (approved):

packages/web3-utils/test/fixtures/converters.ts
nazarhussain: Sure

packages/web3-utils/src/converters.ts
nazarhussain: Sure, please make the change accordingly.

nazarhussain (approved):

PR #5095, opened 2022-06-01
:art: Improve the nullish value usage around the codebase

### LINKED ISSUE #5094: Improve the nullish value usage around the codebase

### What's up?

In our codebase we are using mixed of  **nullish** values, sometimes `null` and sometimes `undefined`. Both have different behavior and can result into different side-effects. I suggest to use this consistent approach throughout the codebase. 

1. Use Typescript **optional** parameter feature where ever possible. 
2. Try to avoid `null` type and if needed use `undefined` which is aligned with **optional** attribute behavior 
3. Whenever checking for **nullish** values, always check for both `null` and `undefined` to avoid null pointer reference. 
4. When it's a need of **nullish** value, use `undefined` as value instead of `null`.

### Examples/References

Talk by Tony Hoare:
https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/

Key Takeaways: 

- Null references have historically been a bad idea
- Early compilers provided opt-out switches for run-time checks, at the expense of correctness
- Programming language designers should be responsible for the errors in programs written in that language

nazarhussain:
## Description

Update all the usage of `nullish` values around the codebase as per the guideline in the issue. 

Fixes #5094

## Type of change

- [x] Improvement 

## Checklist:


nazarhussain:
The changes related to doc comments blocks appears because of running `yarn lint:fix`

avkos (approved):

packages/web3-errors/src/web3_error_base.ts
```diff
@@ -36,6 +36,10 @@ export abstract class Web3Error extends Error {
 	public static convertToString(value: unknown, unquotValue = false) {
 		if (value === undefined) return 'undefined';
 
+		// Using "null" value intentionally for validation
+		// eslint-disable-next-line no-null/no-null
+		if (value === null) return 'undefined';
```
jdevcs:
```suggestion
		// Using "null" value intentionally for validation
		// eslint-disable-next-line no-null/no-null
		if (value === null || value === undefined ) return 'undefined';
```

packages/web3-eth/src/types.ts
```diff
@@ -65,7 +65,7 @@ export interface ReceiptInfo {
 	readonly cumulativeGasUsed: Numbers;
 	readonly gasUsed: Numbers;
 	readonly effectiveGasPrice?: Numbers;
-	readonly contractAddress: Address | null;
+	readonly contractAddress: Address;
```
jdevcs:
```suggestion
	readonly contractAddress?: Address;
```

jdevcs (changes requested):
LGTM, only minor changes required.

jdevcs:
and also some tests are failing so needs to be fixed in this PR.

jdevcs (approved):

luu-alex (approved):

PR #5111, opened 2022-06-07
4x webpack build config

jdevcs:
## Description

https://github.com/ChainSafe/web3.js/issues/5068

## Type of change

## Checklist:


.gitignore
```diff
@@ -29,7 +29,7 @@ tmp
 # Distribution directory
 .vscode
 lib/
-dist/
+*/**/dist/
```
nazarhussain: Why we need to keep the build file into Git? It can be generated and published on the runtime.

package.json
```diff
@@ -32,7 +32,7 @@
 		]
 	},
 	"scripts": {
-		"build": "yarn clean && lerna run build --stream",
+		"build": "yarn clean && lerna run build --stream && webpack --config webpackprod.config.js",
```
nazarhussain, line 35: This should be generated as part of the `web3` package not on the workspace level.

webpackprod.config.js
```diff
@@ -0,0 +1,57 @@
+const webpack = require('webpack');
+const path = require('path');
+const { CleanWebpackPlugin } = require('clean-webpack-plugin');
+
+const config = {
+	mode: 'production',
+	entry: path.resolve(__dirname, 'packages/web3/src/index.ts'),
+	output: {
+		path: path.resolve(__dirname, 'dist'),
```
nazarhussain, line 9: The output along with this file should be moved to the `web3` package.

nazarhussain (changes requested):

.gitignore
jdevcs: Like in 1.x so users cloning repo have last build as part of repo. We keep build at multiple places in git repo dist dir, in releases section of repo on github, npm public repository ...

webpackprod.config.js
jdevcs: For last and this comment: I think we should follow same pattern as in 1.x for keeping users familiar,

.gitignore
nazarhussain: I suggests we should not use this approach any further and I don't see any reason or usability for the users to have the last compile build file when they clone the repo.

webpackprod.config.js
nazarhussain: As it's tightly linked to the `web3` package, so the build process must be encapsulated in that package. If we decided to keep the a root level `dist` folder then we can create a symlink for it.

nazarhussain (changes requested):

webpackprod.config.js
```diff
@@ -0,0 +1,68 @@
...
+	resolve: {
+		extensions: ['.ts', '.js'],
+		fallback: {
+			child_process: false,
+			fs: false,
+			net: false,
+			path: false,
+			os: false,
+			util: require.resolve('util'),
+			http: require.resolve('http-browserify'),
```
luu-alex, line 35: why do we need these modules?

.gitignore
jdevcs: Usability is making build available at multiple places, lets wait for team's suggestions on keeping or removing build from dist dir.

webpackprod.config.js
jdevcs: IMO, web3 package is umbrella of all packages as it contains functionality of whole web3 lib, its not specific functionality package, so it makes more sense not to bind it with one package. Thanks for your input, lets wait for other reviewers on it.

.gitignore
avkos: We can store compiled build to some CDN and save link to it in  `README.md`. In this case we don't need to commit compiled build to the repo

webpackprod.config.js
avkos: if we will not commit the dist folder, it is doesn't matter where it will be

webpackprod.config.js
jdevcs: Webpack 5 no longer polyfills Node.js core modules automatically, so for browsers we need to explicitly include these. 
I just assumed need for all these from cypress build. I haven't tested explicitly need of all these except ( fs, net, crypto and stream :  are required).

jdevcs: Most of such polyfills were included automatically in webpack < 5.

luu-alex: Oh I didn't know that, this could be quite troublesome. with the webpack 5 update, polyfills have been a pain point to many of our [users/developers](https://github.com/ChainSafe/web3.js/issues?q=is%3Aissue+polyfill+is%3Aclosed ). these [issues](https://github.com/ChainSafe/web3.js/issues/5059) have not gone away either.

luu-alex: How feasible would it be to replace these modules within our code in the future?

jdevcs: @luu-alex Thanks for highlighting this, Some of these are in deep path so I think if we explicitly include modules for webpack it should solve issue.

nazarhussain (dismissed):
We need to observe these points explicitly. 

1. We are not building whole repository with webpack
2. We are not also not building every package in the mono repo with the webpack
3. We only need `web3` package to be build with the webpack. 

So we have to be very explicit to create the configuration and build steps for particular packages which we need. 

1. Create a symlink of `webpackprod.config.js` under `packages/web3/` to follow the consistent pattern we have for other configuration files. 
2. Create a task `postbuild` under the `packages/web3/package.json` to execute the `webpack`. 
3. Create a directory `packages/web3/dist_web` for the output of webpack. 
5. To keep the compatibility with `1.x` create a symlink `dist` on root level which points to `packages/web3/dist_web` 

I will highly recommend to not commit the build files in the version control. That does not mean we will be breaking anything for users. We can build and keep the `dist` folders when we publish the packages, so it will not be any issue for any user. Users who would be using local clones of the source code they would never bee using prebuild files anyway. 

None of the above points will introduce any breaking changes for users. The way we are managing builds could be different on our end, which is fine as we have differences of many workflows compared to 1.x and 4.x.

avkos:
> We need to observe these points explicitly.
> 
> 1. We are not building whole repository with webpack
> 2. We are not also not building every package in the mono repo with the webpack
> 3. We only need `web3` package to be build with the webpack.
> 
> So we have to be very explicit to create the configuration and build steps for particular packages which we need.
> 
> 1. Create a symlink of `webpackprod.config.js` under `packages/web3/` to follow the consistent pattern we have for other configuration files.
> 2. Create a task `postbuild` under the `packages/web3/package.json` to execute the `webpack`.
> 3. Create a directory `packages/web3/dist_web` for the output of webpack.
> 4. To keep the compatibility with `1.x` create a symlink `dist` on root level which points to `packages/web3/dist_web`
> 
> I will highly recommend to not commit the build files in the version control. That does not mean we will be breaking anything for users. We can build and keep the `dist` folders when we publish the packages, so it will not be any issue for any user. Users who would be using local clones of the source code they would never bee using prebuild files anyway.
> 
> None of the above points will introduce any breaking changes for users. The way we are managing builds could be different on our end, which is fine as we have differences of many workflows compared to 1.x and 4.x.

I also agree with the text above. it makes sense to go this way.

nikoulai (approved):

jdevcs:
> We need to observe these points explicitly.
> 
> 1. We are not building whole repository with webpack
> 2. We are not also not building every package in the mono repo with the webpack
> 3. We only need `web3` package to be build with the webpack.
> 
> So we have to be very explicit to create the configuration and build steps for particular packages which we need.
> 
> 1. Create a symlink of `webpackprod.config.js` under `packages/web3/` to follow the consistent pattern we have for other configuration files.
> 2. Create a task `postbuild` under the `packages/web3/package.json` to execute the `webpack`.
> 3. Create a directory `packages/web3/dist_web` for the output of webpack.
> 4. To keep the compatibility with `1.x` create a symlink `dist` on root level which points to `packages/web3/dist_web`
> 
> I will highly recommend to not commit the build files in the version control. That does not mean we will be breaking anything for users. We can build and keep the `dist` folders when we publish the packages, so it will not be any issue for any user. Users who would be using local clones of the source code they would never bee using prebuild files anyway.
> 
> None of the above points will introduce any breaking changes for users. The way we are managing builds could be different on our end, which is fine as we have differences of many workflows compared to 1.x and 4.x.

Thanks for your feedback. As discussed in call we can make these changes in future if there is any limitation or requirement for keeping webpack script at web3 package level.

nazarhussain:
@jdevcs I was thinking on our discussion lately. During the ETHPrague when I introduced **new schema based eth compatible validator**  to the community. It was one discussion point and I told them they can use these features in their projects independently of the `web3`. With that perspective it's required to build few other packages for the web as well. As it's few lines of change, so we can make it part of the alpha release as well so community can use those. 

Different web based package managers e.g. [bower](https://bower.io/) tend to utilize the [browser](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#browser) field int he `package.json` and we can point this field only from within the package. Without that field user have to always remember and point the distribution files themselves. 

All these points aligned to the direction of building packages independently with webpack (for the packages we tend to think useable).

nazarhussain:
I created a small [PR](https://github.com/ChainSafe/web3.js/pull/5140) to reflect what I am describing here. Let's see what other team members think about it.

avkos (approved):
need to fix one test or maybe just relaunch it

jdevcs:
> As explained in very details already, I don't think the direction of this PR is appropriate to proceed.

@nazarhussain As we discussed in call these are not urgent requirements in alpha for individual builds of other packages but instead some other items are urgent ( doc, breaking changes fixes, and some other PRs ) so when there is requirement we can implement enhancements after going through our sprint process and team discussion.

PR #5124, opened 2022-06-13
library updates

ac12644:
## Description

Updated several outdated npm libraries.

## Type of change

- [ ] Lib Update


nazarhussain:
@ac12644 To protect the larger community base, we don't encourage accepting PR's which makes changes in the `package-lock.json` files, so I have to close this one. If you know any particular packages which are needed to be changed, then make that change through `package.json` file and don't include `package-lock.json` files in that PR.

Issue #4896, opened 2022-03-29
IPC Provider - 4.x System Tests


jdevcs:
System tests for:

## IPC Provider package

- [x] connect tests
- [x] getStatus tests
- [x] Event subscription 'message' | 'connect' | 'disconnect' tests
- [x] removeAllListeners and removeListener tests
- [x] disconnect and reset tests
- [x] send multiple requests and receive response 

#### Min Platform Requirements for 4.x Alpha:
- Node: v14, v16
- Browsers: Firefox, Brave
- Eth Client: Geth

More details are in [Epic](https://github.com/ChainSafe/web3.js/issues/4711)

mconnelly8:
This can be testing for Beta, but if we have time it should be added to Alpha.

PR #5131, opened 2022-06-13
IPC Provider - 4.x System Tests

avkos:
## Description
#4896 

## Type of change

## Checklist:


package.json
```diff
@@ -40,19 +40,23 @@
...
 		"geth:stop": "./scripts/geth.sh stop",
+		"ipc:start:background": "WEB3_SYSTEM_TEST_BACKEND=ipc && ./scripts/ipc.sh start 1",
+		"ipc:stop": "./scripts/ipc.sh stop",
 		"lint": "lerna run lint --stream --parallel",
 		"lint:fix": "lerna run lint:fix --stream --parallel",
 		"format": "lerna run format --stream --parallel",
 		"test": "lerna run test --stream --parallel",
 		"test:unit": "lerna run test:unit --stream --parallel",
 		"test:integration": "lerna run test:integration --stream",
+		"test:integration:ipc": "cd ./packages/web3 && yarn test:integration test/integration/ipc.test.ts",
```
nazarhussain: Would suggest to keep it consistent with other scripts and create that script only in the `web3` package. 


```suggestion
		"test:integration:ipc": "lerna run test:integration:ipc --stream",
```

package.json
```diff
@@ -40,19 +40,23 @@
...
 		"test:e2e:ganache:http": "./scripts/test-runner.sh ganache http",
 		"test:e2e:ganache:ws": "./scripts/test-runner.sh ganache ws",
 		"test:e2e:geth:http": "./scripts/test-runner.sh geth http",
 		"test:e2e:geth:ws": "./scripts/test-runner.sh geth ws",
 		"test:e2e:ganache:ws:electron": "./scripts/test-runner.sh ganache ws electron",
 		"test:e2e:ganache:ws:chrome": "./scripts/test-runner.sh ganache ws chrome",
 		"test:e2e:ganache:ws:firefox": "./scripts/test-runner.sh ganache ws firefox",
+		"test:e2e:ipc": "./scripts/test-runner.sh ipc ipc ipc",
```
nazarhussain: Would be nice to mention the backend, mode and runner. 

```suggestion
		"test:e2e:geth:ipc": "./scripts/test-runner.sh geth ipc electron",
```

scripts/test-runner.sh
```diff
@@ -13,9 +13,9 @@ BACKEND=${ORIGARGS[0]}
 MODE=${ORIGARGS[1]}
 ENGINE=${ORIGARGS[2]}
 
-SUPPORTED_BACKENDS=("geth" "ganache")
-SUPPORTED_MODE=("http" "ws")
-SUPPORTED_ENGINES=("node" "electron" "firefox" "chrome" "")
+SUPPORTED_BACKENDS=("geth" "ganache" "ipc")
+SUPPORTED_MODE=("http" "ws" "ipc")
+SUPPORTED_ENGINES=("node" "electron" "firefox" "chrome" "ipc" "")
```
nazarhussain: We should run the IPC tests with `node` or `electron` 

```suggestion
SUPPORTED_ENGINES=("node" "electron" "firefox" "chrome" "")
```

nazarhussain (changes requested):

package.json
avkos: for IPC it works only with the geth binary backend (not docker). I called it ipc. Regarding runner. may be yes we can run it from browsers and  electron

nikoulai:
Great! Added a comment.
Except for Nazar's comments, LGTM! 
ipc-provider class tests may look similar to ws-provider class, if you want to use anything as input.

scripts/test-runner.sh
avkos: IpcProvider in the electron from cypress can't connect to the IPC file. We've, with @jdevcs, decided to skip an electron to the IPC test

packages/web3/package.json
```diff
@@ -30,7 +30,8 @@
 		"test:ci": "jest --coverage=true --coverage-reporters=json --verbose",
 		"test:watch": "npm test -- --watch",
 		"test:unit": "jest --config=./test/unit/jest.config.js",
-		"test:integration": "jest --config=./test/integration/jest.config.js --runInBand --forceExit"
+		"test:integration": "jest --config=./test/integration/jest.config.js --runInBand --forceExit",
+		"test:e2e:geth:ipc:node": "yarn test:integration test/integration/ipc.test.ts"
```
nazarhussain: I would tend to run all integration tests on the IPC, not just the connection one.

.github/workflows/build.yml
```diff
@@ -65,30 +71,46 @@ jobs:
...
+              node: [ 14, 16 ]
+              engine: ['node']
+      steps:
+          - uses: actions/download-artifact@v3
+            with:
+                name: web3-${{ matrix.node }}.js.tar.gz
+                path: /tmp
+          - run: tar -xf /tmp/web3-${{ matrix.node }}.js.tar.gz -C ./
+          - run: yarn test:e2e:geth:ipc:${{ matrix.engine }}
+            shell: bash
```
nazarhussain: Actually our job `integration` is exactly same as `e2e` tests, it's just that it runs for `node` environment and not for the `browsers`. 

What do you think if we merge the `ipc` tests into the `integration` pipeline, something like this. 

```
  integration:
    name: Integration
    needs: unit
    runs-on: ubuntu-latest
    env:
        INFURA_GOERLI_HTTP: ${{ secrets.INFURA_GOERLI_HTTP }}
        INFURA_GOERLI_WS: ${{ secrets.INFURA_GOERLI_WS }}
    strategy:
      fail-fast: false
      matrix:
        node: [ 14, 16 ]
        backend: ['geth', 'ganache', 'ipc']
        mode: ['http', 'ws', 'ipc']
    steps:
      - uses: actions/checkout@v3
      - uses: actions/setup-node@v3
        with:
          node-version: ${{ matrix.node }}
          cache: yarn
      -   run: yarn
      -   if: ${{ matrix.backend }} != 'ipc' || ( ${{ matrix.backend}} == 'ipc' && ${{ matrix.mode }} == 'ipc' )
      -   run: yarn test:e2e:${{ matrix.backend }}:${{ matrix.mode }}
          shell: bash
```

nazarhussain (changes requested):

.github/workflows/build.yml
avkos: so matrix will create a few empty tests. which only download artifacts. unpack. and do nothing
I mean. combinations:
backend, mode
ganache-ipc
geth-ipc
ipc-http
ipc-ws

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -101,7 +101,8 @@ export class Web3RequestManager<
 			} else if (typeof net === 'object' && typeof net.connect === 'function') {
 				newProvider = new this.providers.IpcProvider<API>(provider, net);
 			} else {
-				throw new ProviderError(`Can't autodetect provider for "${provider}"`);
+				newProvider = new this.providers.IpcProvider<API>(provider);
```
jdevcs, line 111: Why we are always defaulting to IPC provider if IPC provider is created based on condition above.

```
			} else if (typeof net === 'object' && typeof net.connect === 'function') {
				newProvider = new this.providers.IpcProvider<API>(provider, net);
```

packages/web3-core/test/unit/web3_request_manager.test.ts
```diff
@@ -228,7 +228,7 @@ describe('Web3RequestManager', () => {
 			});
 		});
 
-		it('should throw error if can not detect the provider', () => {
+		it.skip('should throw error if can not detect the provider', () => {
```
jdevcs: As mentioned above if no provider is provided it should throw error.

packages/web3-eth-contract/test/integration/setup.js
```diff
@@ -19,6 +19,6 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 require('../config/setup');
 
-const jestTimeout = 15000;
+const jestTimeout = 20000;
```
jdevcs: Instead of `jestTimeout` def in multiple test files, shouldn't we have one used by all provided by cli --testTimeout .

packages/web3-providers-ipc/src/index.ts
```diff
@@ -110,6 +110,23 @@ export default class IpcProvider<
 		this._removeSocketListeners();
 		this._addSocketListeners();
 	}
+	public async waitForConnection(): Promise<void> {
+		return new Promise<void>((resolve, reject) => {
+			const maxNumberOfAttempts = 10;
+			const timeOutTime = 5000; // ms
+			let currentAttempt = 0;
```
jdevcs: Instead of only hard coding these, I think there should be defaults and these should be configurable via setters at IpcProvider class level.

scripts/geth.sh
```diff
@@ -1,35 +1,88 @@
...
+      msys*)    OS="WINDOWS" ;;
+      cygwin*)  OS="ALSO WINDOWS" ;;
+      *)        OS="unknown: $OSTYPE" ;;
+    esac
+}
+getDownloadLink(){
+    case "$OS" in
+      SOLARIS*) LINK="-" ;;
+      OSX*)  LINK="https://gethstore.blob.core.windows.net/builds/geth-darwin-amd64-1.10.19-23bee162.tar.gz" ;;
+      LINUX*)   LINK="https://gethstore.blob.core.windows.net/builds/geth-linux-386-1.10.19-23bee162.tar.gz" ;;
```
jdevcs: This geth bin was not working for me and I used https://gethstore.blob.core.windows.net/builds/geth-linux-amd64-1.10.19-23bee162.tar.gz for local testing.

jdevcs (changes requested):

packages/web3-eth-contract/test/integration/setup.js
avkos: I've reverted all timeouts to 15 sec except web3-eth

packages/web3-core/src/web3_request_manager.ts
avkos: because we `net` is not an object `typeof net === 'object' `
That's why we will never go inside this if

packages/web3-core/test/unit/web3_request_manager.test.ts
avkos: IPC provider is just string. how we can understand is it the correct IPC string or not?

packages/web3-core/src/web3_request_manager.ts
jdevcs: In that case this condition should be adjusted for IPC provider.

packages/web3-core/test/unit/web3_request_manager.test.ts
jdevcs: Named pipe can be any string but for now we can condition here to accept specific format and can use regex for its check. and add doc comment on function for valid IPC provider strings.

jdevcs (approved):
`Build / Integration (14, geth, ws) (pull_request)` needs to be fixed before merging, other thn that lgtm.

packages/web3-eth-contract/cypress.config.js
```diff
@@ -0,0 +1,12 @@
...
+	video: false,
+	e2e: {
+		// We've imported your old cypress plugins here.
+		// You may want to clean this up later by importing these.
+		setupNodeEvents(on, config) {
+			return require('./cypress/plugins/index.js')(on, config);
+		},
+		specPattern: 'test/integration/**/**/*.test.ts',
+	},
+};
```
nazarhussain: Is it possible to move this files to `templates` and then link to packages?

packages/web3-eth/cypress.config.js
```diff
@@ -0,0 +1,12 @@
...
+	video: false,
+	e2e: {
+		// We've imported your old cypress plugins here.
+		// You may want to clean this up later by importing these.
+		setupNodeEvents(on, config) {
+			return require('./cypress/plugins/index.js')(on, config);
+		},
+		specPattern: 'test/integration/**/**/*.test.ts',
+	},
+};
```
nazarhussain: Same for here.

packages/web3-eth/test/integration/batch.test.ts
```diff
@@ -0,0 +1,79 @@
...
+				method: 'eth_getBalance',
+				params: [accounts[0], 'latest'],
+			};
+			const request2 = {
+				id: 11,
+				method: 'eth_getBalance',
+				params: [accounts[1], 'latest'],
+			};
+			batch.add(request1).catch(console.error);
+			batch.add(request2).catch(console.error);
```
nazarhussain: We also need to make sure each request was resolved successfully.

packages/web3-eth/test/integration/helper.ts
```diff
@@ -53,3 +55,72 @@ export const sendFewTxes = async ({
 	}
 	return res;
 };
+
+export const validateTransaction = (tx: TransactionInfo) => {
+	expect(tx.nonce).toBeDefined();
+	expect(tx.hash).toBeDefined();
+	expect(String(tx.hash)?.length).toBe(66);
```
nazarhussain: Would be nice to match it with regular expression instead for valid length and hex.

packages/web3-eth/test/integration/helper.ts
```diff
@@ -53,3 +55,72 @@ export const sendFewTxes = async ({
...
+	expect(Number(tx.type)).toBe(0);
```
nazarhussain: This should be function argument instead. We can validate type2 transaction as well.

packages/web3-eth/test/integration/helper.ts
```diff
@@ -53,3 +55,72 @@ export const sendFewTxes = async ({
...
+	expect(tx.blockHash).toBeDefined();
```
nazarhussain: Same here match with regexp to match the hex and length.

packages/web3-eth/test/integration/helper.ts
```diff
@@ -53,3 +55,72 @@ export const sendFewTxes = async ({
...
+	expect(r.to).toBeDefined();
+	expect(r.cumulativeGasUsed).toBeDefined();
+	expect(r.gasUsed).toBeDefined();
+	expect(r.effectiveGasPrice).toBeDefined();
+	expect(r.logs).toBeDefined();
+	expect(r.logsBloom).toBeDefined();
+	expect(r.status).toBeDefined();
+	expect(String(r.transactionHash)).toHaveLength(66);
+	expect(Number(r.gasUsed)).toBeGreaterThan(0);
+};
```
nazarhussain, line 122: Please apply above suggestions to all expectations.

packages/web3/test/integration/ipc.test.ts
```diff
@@ -0,0 +1,44 @@
...
+	beforeAll(() => {
+		clientUrl = getSystemTestProvider();
+	});
+
+	describeIf(isIpc)('Create Web3 class instance with ipc string providers', () => {
+		// https://ethereum.stackexchange.com/questions/52574/how-to-connect-to-ethereum-node-geth-via-ipc-from-outside-of-docker-container
+		// https://github.com/ethereum/go-ethereum/issues/17907
+		it('should create instance with string of IPC provider', async () => {
+			// eslint-disable-next-line @typescript-eslint/no-unused-vars
+			// eslint-disable-next-line no-new
```
nazarhussain:
```suggestion
			// eslint-disable-next-line no-new @typescript-eslint/no-unused-vars
```

scripts/geth.sh
```diff
@@ -1,23 +1,79 @@
...
+        echo "$FOLDER"
+        mv "$FOLDER/geth" "$TMP_FOLDER/geth"
+        rm -rf "$FOLDER"
+    fi
+}
 
 start() {
 	. scripts/env.sh
-
+#    download
```
nazarhussain: Why we commented this out? If we are still using the docker then we can delete the `download` function.

scripts/system_tests_utils.ts
```diff
@@ -40,8 +40,14 @@ export const getEnvVar = (name: string): string | undefined =>
...
 
+export const DEFAULT_SYSTEM_PROVIDER = 'http://localhost:8545';
+
 export const getSystemTestProvider = (): string =>
-	getEnvVar('WEB3_SYSTEM_TEST_PROVIDER') ?? 'http://localhost:8545';
+	getEnvVar('WEB3_SYSTEM_TEST_PROVIDER') ?? DEFAULT_SYSTEM_PROVIDER;
+
+export const isHttp = getSystemTestProvider().startsWith('http');
+export const isWs = getSystemTestProvider().startsWith('ws');
+export const isIpc = getSystemTestProvider().includes('ipc');
```
nazarhussain, line 50: 👍

nazarhussain (changes requested):

packages/web3-eth-contract/cypress.config.js
avkos: good idea. done

packages/web3-eth/test/integration/batch.test.ts
avkos: I've checked it on next line
```ts 
const [response1, response2] = await batch.execute();
expect(response1.result).toBeDefined()
expect(response2.result).toBeDefined();
```

packages/web3-eth/test/integration/helper.ts
```diff
@@ -53,3 +55,82 @@ export const sendFewTxes = async ({
 	}
 	return res;
 };
+
+const checkHex = (hash: string, length = 64) => {
+	expect(String(hash)).toBeDefined();
+	expect(String(hash)).toHaveLength(length + 2);
+	expect(new RegExp(`0x[abcdef0-9]{${length}}`, 'i').test(hash)).toBe(true);
+};
```
nazarhussain:
```suggestion
const regexHex20 = 0[xX][0-9a-fA-F]{40};
const regexHex32 = 0[xX][0-9a-fA-F]{64};
```

packages/web3-eth/test/integration/helper.ts
```diff
@@ -53,3 +55,82 @@ export const sendFewTxes = async ({
...
+
+type ExpectOptions = {
+	type?: number;
+};
+export const validateTransaction = (
+	tx: TransactionInfo,
+	expectOptions: ExpectOptions = { type: 0 },
+) => {
+	expect(tx.nonce).toBeDefined();
+	checkHex(String(tx.hash));
```
nazarhussain:
```suggestion
	expect(tx.hash).toMatch(regexHex32);
```

packages/web3-eth/test/integration/helper.ts
```diff
@@ -53,3 +55,82 @@ export const sendFewTxes = async ({
...
+	expect(Number(tx.type)).toBe(expectOptions.type);
+	checkHex(String(tx.blockHash));
```
nazarhussain:
```suggestion
	expect(tx.blockHash).toMatch(regexHex32);
```

packages/web3-eth/test/integration/helper.ts
```diff
@@ -53,3 +55,82 @@ export const sendFewTxes = async ({
...
+	expect(Number(tx.blockNumber)).toBeGreaterThan(0);
+	expect(tx.transactionIndex).toBeDefined();
+	checkHex(tx.from, 40);
```
nazarhussain:
```suggestion
     expect(tx.from).toMatch(regexHex20);
```

nazarhussain (changes requested):

nazarhussain (approved):

PR #5132, opened 2022-06-14
updating web3 utils BN

luu-alex:
Was testing [1.7.4-rc.1](https://github.com/ChainSafe/web3.js/pull/5036) and many open zeppelin testcases were breaking due to the change of updating BN, this was due to us adding `BNwrapped` and not adding the prototype and attributes. 

This PR adds the methods/attributes to the BN object and allows the synthetix tests to pass.

### DIFF
--- a/packages/web3-utils/src/utils.js
+++ b/packages/web3-utils/src/utils.js
@@ -545,14 +545,7 @@ var BNwrapped = function (value) {
         return new BN(value);
     } 
 };
-BNwrapped.BN = BNwrapped;
-BNwrapped.wordSize = 26;
-BNwrapped.isBN = BN.isBN;
-BNwrapped.max = BN.max;
-BNwrapped.min = BN.min;
-BNwrapped.mont = BN.mont;
-BNwrapped.red = BN.red;
-BNwrapped._prime = BN._prime;
+Object.setPrototypeOf(BNwrapped, BN);
 Object.setPrototypeOf(BNwrapped.prototype, BN.prototype);
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


packages/web3-utils/src/utils.js
```diff
@@ -545,6 +545,15 @@ var BNwrapped = function (value) {
...
 };
+BNwrapped.BN = BNwrapped;
+BNwrapped.wordSize = 26;
+BNwrapped.isBN = BN.isBN;
+BNwrapped.max = BN.max;
+BNwrapped.min = BN.min;
+BNwrapped.mont = BN.mont;
+BNwrapped.red = BN.red;
+BNwrapped._prime = BN._prime;
+Object.setPrototypeOf(BNwrapped.prototype, BN.prototype);
```
nazarhussain: Setting prototypes on objects and nested prototypes will work for both instance and static members. You don't need to override each property. 

```suggestion
Object.setPrototypeOf(BNwrapped.prototype, BN.prototype);
Object.setPrototypeOf(BNwrapped, BN);
```

nazarhussain (changes requested):
One more thing I observed is that the bundle size for this RC release is increased. Do you suspect any particular reason for it?

jdevcs (approved):

luu-alex:
@nazarhussain  has the bundlesize increased? running `npm run bundlesize` locally for both `1.7.3` and `1.7.4-rc.1` give me the size of 1.39mb

nazarhussain:
@luu-alex  Please check this https://bundlephobia.com/package/web3@1.7.3

nazarhussain (approved):

luu-alex:
@nazarhussain  Ah I see, I believe the bundle size becoming larger is due to upgrading ether packages we were using. To note, our bundle now includes @ethersproject/signing-key@5.6.2

PR #5136, opened 2022-06-15
Init web3_eth_migration_guide

Spacesai1or:
To preview migration guide:

1. `cd web3.js/docs`
2. `yarn start`
3. Navigate to [Web3Eth Migration Guide](http://localhost:3000/docs/guides/web3_eth_migration_guide)
---

Remaining methods to check for breaking changes:

- [x] `sendSignedTransaction`
- [x] `sign`
- [x] `signTransaction`
- [x] `call`
- [x] `estimateGas`
- [x] `getPastLogs`
- [x] `getWork`
- [x] `submitWork`
- [x] `requestAccounts`
- [x] `getChainId`
- [x] `getNodeInfo`
- [x] `getProof`


nikoulai:
BatchRequest is not documented ( I'll copy it from web3 migration guide after it get ok from reviews)

docs/docs/guides/web3_eth_migration_guide.md
```diff
@@ -0,0 +1,1098 @@
+# Web3Eth Migration Guide
+
+## Breaking Changes
+
+### Not Implemented or Exported
+
+-   [extend](https://web3js.readthedocs.io/en/v1.7.3/web3-eth.html#extend) functionality not implemented
+-   [web3.eth.createAccessList](https://web3js.readthedocs.io/en/v1.7.3/web3-eth.html#createaccesslist) not implemented
+-   [web3.eth.personal](https://web3js.readthedocs.io/en/v1.7.3/web3-eth.html#personal) namespace is not exported
+-   [web3.eth.net](https://web3js.readthedocs.io/en/v1.7.3/web3-eth.html#net) namespace is not exported
```
nazarhussain: It's exported

nazarhussain: This complete section seems to be redundant with the other PR.

docs/docs/guides/web3_eth_migration_guide.md
```diff
@@ -0,0 +1,1098 @@
...
+
+### Defaults and Configs
+
+-   `givenProvider` default value is `undefined` instead of `null`
+-   `currentProvider` will never return `null`, provider required upon instantiation as opposed to being optional in 1.x //todo maybe this is not true after after the recent discussion/change
```
nazarhussain: It could return `undefined`

docs/docs/guides/web3_eth_migration_guide.md
```diff
@@ -0,0 +1,1098 @@
...
+-   `web3.eth.defaultAccount` default value is `undefined` instead of `null`
```
nazarhussain: That applies to everywhere. So insted of listing it for individual values, we should highlight it overall. 

All the API level interfaces returning or accepting `null` are replaced with `undefined`

docs/docs/guides/web3_eth_migration_guide.md
```diff
@@ -0,0 +1,1098 @@
...
+-   `web3.eth.defaultHardfork` default is `london` instead of `undefined`
+    -   1.x has `london` documented as default, but in implementation it's `undefined`
+-   `web3.eth.defaultChain` default is `mainnet` instead of `undefined`
+    -   1.x has `mainnet` documented as default, but in implementation it's `undefined`
+
+### Web3Eth Methods
+
+### `web3.eth.getHashrate`
+
+renamed to `getHashRate`
```
nazarhussain: To avoid breaking change, we should create a same function and mark it deprecated.

docs/docs/guides/web3_eth_migration_guide.md
```diff
@@ -0,0 +1,1098 @@
...
+
+### `web3.eth.getGasPrice`
+
+returns a hex string instead of a number string
```
nazarhussain: This should be returning the `bigint` instead. Pleaes double check.

docs/docs/guides/web3_eth_migration_guide.md
```diff
@@ -0,0 +1,1098 @@
...
+// in 1.x
+await web3.eth.getGasPrice(); // '2000000000'
+
+// in 4.x
+await web3.eth.getGasPrice(); // '0x77359400'
+```
+
+### `web3.eth.getFeeHistory`
+
+returns hex strings for `gasUsedRation` array items instead of a numbers
```
nazarhussain: If it's defined as number, then it should be returning `bigint` not the hex string.

docs/docs/guides/web3_eth_migration_guide.md
```diff
@@ -0,0 +1,1098 @@
...
+// {
+//  oldestBlock: '0x0',
+//  baseFeePerGas: [ '0x3b9aca00', '0x342770c0' ],
+//  gasUsedRatio: [ '0x0' ]
+// }
+```
+
+### `web3.eth.getAccounts`
+
+addresses are not returned as checksum
```
nazarhussain: We should rather create and issue and fix this. That's very imporant for the users to have checksum addresses.

docs/docs/guides/web3_eth_migration_guide.md
```diff
@@ -0,0 +1,1098 @@
...
+// ['0xAB089B30f46883C3598B720d85837080b9929D0B']
+
+// in 4.x
+await web3.eth.getAccounts();
+// ['0xab089b30f46883c3598b720d85837080b9929d0b']
+```
+
+### `web3.eth.getBlockNumber`
+
+returns a hex string instead of a number
```
nazarhussain: It should be returing `bigint` instead. 


That's a global change in the API, eveverywhere we have numbers data type, we will be returning `bigint`

docs/docs/guides/web3_eth_migration_guide.md
```diff
@@ -0,0 +1,1098 @@
...
+// in 1.x
+await web3.eth.getBlockNumber(); // 0
+
+// in 4.x
+await web3.eth.getBlockNumber(); // '0x0'
+```
+
+### `web3.eth.getBalance`
+
+returns a hex string instead of a number string
```
nazarhussain: Same here.

docs/docs/guides/web3_eth_migration_guide.md
```diff
@@ -0,0 +1,1098 @@
...
+// in 1.x
+await web3.eth.getBalance('0xAB089B30f46883C3598B720d85837080b9929D0B'); // '115792089237316195423570985008687907853269984665640564039457584007913129639927'
+
+// in 4.x
+await web3.eth.getBalance('0xAB089B30f46883C3598B720d85837080b9929D0B'); // '0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7'
+```
+
+### `web3.eth.getBlock`
+
+-   Returns a hex string instead of a number for the following properties:
```
nazarhussain: Same here.

docs/docs/guides/web3_eth_migration_guide.md
```diff
@@ -0,0 +1,1098 @@
...
+//     totalDifficulty: '0x1',
+//     transactions: [],
+//     transactionsRoot: '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421',
+//     uncles: []
+// }
+```
+
+### `web3.eth.getBlockTransactionCount`
+
+returns a hex string instead of a number
```
nazarhussain: Same applies it should return `bigint`

nazarhussain (changes requested):
I have following observations: 

1. Pleae make the change for the `bigint` everywhere. All the numbers data type are consistenlty should return the `bigint` . If not then it's a bug and we should fix it. 
2. It does not feel like productive to me to document those breaking changes in that way. Will be tedious to review and maintain. Instead we can just document the list of funcitons which now will return `bigint` for all number type attributes.

docs/docs/guides/web3_eth_migration_guide.md
nikoulai: opened an issue: https://github.com/ChainSafe/web3.js/issues/5166

docs/docs/guides/web3_eth_migration_guide.md
nikoulai: Done. Also, added in docs

avkos (approved):
approved. except for Nazar comments

docs/docs/guides/web3_eth_migration_guide.md
nikoulai: is extended implemented too?

nikoulai: Regarding point 2. Sound good to me.
Let's wait for input from the rest of the team for their opinion and to see how they have structured docs.

docs/docs/guides/web3_eth_migration_guide.md
nikoulai: Some values are not returned. I opened an issue for that: https://github.com/ChainSafe/web3.js/issues/5168.
 Except for this,`GasUsedRation` is decimal, I think this isn't supported in bigints.
```gasUsedRatio: [ 0.0018243056410397187 ]```

luu-alex (approved):

nazarhussain (approved):

jdevcs (approved):

PR #5137, opened 2022-06-15
Change default return type for integer values to BigInt

### LINKED ISSUE #5128: Change default return type for integer values to BigInt

### What's up?

Based on the [discussion](https://github.com/ChainSafe/web3.js/discussions/5120). It seems to better to use the `BigInt` as the default return type for the web3. 

### Examples/References

_No response_

nazarhussain:
## Description

Fixes #5128

## Type of change

## Checklist:


nikoulai (approved):

avkos (approved):

avkos (changes requested):
Some of the tests are not passing.  it will be great to fix it

packages/web3-common/src/formatter.ts
```diff
@@ -60,9 +60,9 @@ export type DataFormat = {
 	readonly bytes: FMT_BYTES;
 };
 
-export const DEFAULT_RETURN_FORMAT = { number: FMT_NUMBER.HEX, bytes: FMT_BYTES.HEX } as const;
+export const DEFAULT_RETURN_FORMAT = { number: FMT_NUMBER.BIGINT, bytes: FMT_BYTES.HEX } as const;
+export const ETH_DATA_FORMAT = { number: FMT_NUMBER.HEX, bytes: FMT_BYTES.HEX } as const;
```
jdevcs, line 64:
```suggestion
export const ETH_INPUT_DATA_FORMAT = { number: FMT_NUMBER.HEX, bytes: FMT_BYTES.HEX } as const;
```
I think should rename its all occurrences for better readability.

jdevcs, line 64:
```
export const DEFAULT_RETURN_FORMAT = { number: FMT_NUMBER.BIGINT, bytes: FMT_BYTES.HEX } as const;
export const ETH_DATA_FORMAT = { number: FMT_NUMBER.HEX, bytes: FMT_BYTES.HEX } as co
nst;
```

These should be places in common/types.ts

jdevcs (dismissed):

packages/web3-eth-ens/src/ens.ts
```diff
@@ -289,7 +292,10 @@ export class ENS extends Web3Context<EthExecutionAPI & Web3NetAPI> {
 		if (this._detectedAddress) {
 			return this._detectedAddress;
 		}
-		const networkType = await getId(this, DEFAULT_RETURN_FORMAT); // get the network from provider
+		const networkType = await getId(this, {
+			...DEFAULT_RETURN_FORMAT,
+			number: FMT_NUMBER.HEX,
```
jdevcs, line 297: If we always want to have `number:FMT_NUMBER.HEX` here why we are using default format?

packages/web3-eth/src/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+import { FMT_BYTES, FMT_NUMBER } from 'web3-common';
+
+export const HEX_NUMBER_DATA_FORMAT = { bytes: FMT_BYTES.HEX, number: FMT_NUMBER.HEX } as const;
```
jdevcs: `HEX_NUMBER_DATA_FORMAT ` is only used in send signed transaction tests so should be places in tests dir.

avkos (approved):

packages/web3-eth/test/integration/web3_eth/send_signed_transaction.test.ts
```diff
@@ -59,15 +65,25 @@ describe('Web3Eth.sendSignedTransaction', () => {
 				...transaction,
 				...gasPricing,
 			});
-			const response = await web3Eth.sendSignedTransaction(signedTransaction.raw);
-			expect(response.status).toBe('0x1');
+			const response = await web3Eth.sendSignedTransaction(
+				typeof signedTransaction.raw === 'string'
+					? signedTransaction.raw
+					: signedTransaction.raw.raw,
```
jdevcs, line 78: There shouldn't be nested raw for output of `web3Eth.signTransaction`. I think it has to do with recently discussed geth vs ganache output diff?

nazarhussain: Yes this change didn't appear from this PR. It was done earlier in @spacesailor24 PR and relates to the output difference for ganache and geth.

packages/web3-validator/src/validation/numbers.ts
```diff
@@ -36,6 +36,10 @@ export const isUInt = (
 		return false;
 	}
 
+	if (typeof value === 'string' && value.length === 0) {
```
jdevcs: Just remove line 35,36, and 37 above and can use:
```suggestion
	if ( (!['number', 'string', 'bigint'].includes(typeof value))  ||
	    (typeof value === 'string' && value.length === 0) ) {
```

packages/web3-eth/test/integration/web3_eth/send_signed_transaction.test.ts
jdevcs: ok, https://github.com/ChainSafe/web3.js/issues/5077 with this issue we will need to fix these tests as well.

packages/web3-common/src/formatter.ts
nazarhussain: It's value not the type.

nazarhussain: @jdevcs As per my knowledge, Eth implementation follows the same format for input/output. In that aspect I named this value as `ETH_DATA_FORMAT`. Please suggest if you think `ETH_INPUT_DATA_FORMAT` is better usage?

packages/web3-eth-ens/src/ens.ts
nazarhussain: To not redefine `bytes` format. You can't define just one format while declaring one data format.

nazarhussain:
@jdevcs Need to merge this PR as another PR is based on it. If you think we still need to make further changes, we can do it with a separate issue.

PR #5130, opened 2022-06-13
Improvements to make 4.x compatible with different projects

nazarhussain:
## Description

Fixes #5125

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] Any dependent changes have been merged and published in downstream modules.
- [x] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [x] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code.
- [x] I ran `npm run build` with success.
- [x] I have tested the built `dist/web3.min.js` in a browser.
- [x] I have tested my code on the live network.
- [x] I have checked the Deploy Preview and it looks correct.
- [x] I have updated the `CHANGELOG.md` file in the root folder.

PR #5139, opened 2022-06-16
:sparkles: Make the web3 constructor parameters optional

### LINKED ISSUE #5125: Improvements to make 4.x compatible with different projects
While migrating [truffle](https://github.com/trufflesuite/truffle) came across a few changes which are required to make web3@4x compatible with a lot of projects our there. 

Here is the the list of individual changes: 

- [x] Need to change the TS target to ES2016 for broader compatibility 
- [x] Need to export `formatters` as namespace from `web3-common` package. 
- [x] `web3.eth.net` namespace is not exported 
- [x] `web3.eth.personal` namespace is not exported 
- [x] `web3.utils` namespace is not exported
- [x] `web3.modules` namespace is not exported
- [x] `Web3.utils` static namespace is not exported 
- [x] `Web3.version` static namespace is not exported 
- [x] Parameter for the web3 constructor is required, need to be optional for 1.x compatibility 
- [x] Parameter for web3.setProvider is required, need to be optional for 1.x compatibility. 
- [x] Truffle `HDWalletProvider` type is not assignable to `SupportedProviders` type. 
- [x] Would be nice to rename `PromiEvent` class to `Web3PromiEvent` to make it more compatible with 1.x implementation. 
- [x] Every provider should have have `send` method for backward compatibility.

nazarhussain:
## Description

Fixes #5130

## Type of change

## Checklist:


luu-alex (approved):

avkos (approved):

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -84,6 +84,16 @@ describe('Web3 instance', () => {
 		}
 	});
 
+	it('should be able to create web3 object without provider', () => {
```
jdevcs, line 87: I think we should also add another test with no provider web3 instantiation and using some functionality that is not dependent on provider.

jdevcs, line 90: I think we should also add another test with no provider web3 instantiation and using some functionality that is not dependent on provider.

jdevcs (changes requested):

jdevcs (changes requested):
Also need to fix tests in CI

packages/web3/test/integration/main_web3.test.ts
nazarhussain: Done.

packages/web3/test/integration/main_web3.test.ts
```diff
@@ -75,15 +75,34 @@ describe('Web3 instance', () => {
...
+		expect(() => new Web3()).not.toThrow();
+	});
+
+	it('should be able use functionality with web3 object not dependant on provider', () => {
+		web3 = new Web3();
+
+		expect(web3.utils.hexToNumber('0x5')).toBe(5);
```
jdevcs, line 97: Seems ok, but I was thinking we should have test that goes through multiple scenarios without provider. Like we should add ABI encoding without provider :
```
  const web3 = new Web3();
  const contract = new web3.eth.Contract(ContractABI);
   await contract.methods.Func(12).encodeABI(); // and see it works
```
This will cover more scenarios, web3 instance creation, contract instance creation, encoding etc

jdevcs (changes requested):

packages/web3/test/integration/main_web3.test.ts
nazarhussain: Such behaviour is better candidate for unit tests, not the integration. I had added one test case though in the integration tests.

jdevcs: Only encoding param can be yes, but for scenarios that involve multiple packages, integration tests are more good. like abi encoding for eth_call of specific function of contract without provider or creating tx and offline signing etc.

jdevcs (approved):

packages/web3/test/integration/main_web3.test.ts
nazarhussain: These functions pure functions, have no dependency on any related code that can effect the behavior. So these functions can be better tested in unit tests with all possible input/output. If those functions had any state which are preserved in the class then we would have gone for integration tests.

PR #5157, opened 2022-06-22
hexToNumber: return BigInt if result is bigger than max integer

### LINKED ISSUE #5153: getTransactionReceipt: Number can only safely store up to 53 bits

- [x] I have searched the existing issues

### Current Behavior

```shell
.../node_modules/number-to-bn/node_modules/bn.js/lib/bn.js:6
    if (!val) throw new Error(msg || 'Assertion failed');
                    ^
Error: Number can only safely store up to 53 bits
```

### Expected Behavior

-

### Steps to Reproduce

```ts
// chain ETC (61)

web3.eth.getTransactionReceipt('0x84e8b73c5f9e5f1fdd6e3482022f3c65ceb92d95c4ea92075c04d1a22f6b3152').then(console.log)
```

gasPrice is 10mio GWEI.

representing `effectiveGasPrice` as string would solve the problem (currently typeof number)

https://blockscout.com/etc/mainnet/tx/0x84e8b73c5f9e5f1fdd6e3482022f3c65ceb92d95c4ea92075c04d1a22f6b3152

### Web3.js Version

1.7.3

### Environment

- Operating System: OSX 12.4 (21F79)
- Browser: 
- Node.js Version: 16
- NPM Version: 

### Anything Else?

thanks.

nikoulai:
## Description

#5153

## Type of change

Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


nikoulai:
I changed `hexToNumber` to return BigNumber if the value is greater than `max_sage_in`. The reason is not to introduce a breaking change to the application already using the functionality.

test/utils.toNumber.js
```diff
@@ -1,60 +1,98 @@
...
+    { value: "-0", expected: -0 },
+    { value: "-0x0", expected: -0 },
+    // { value: [1,2,3,{test: 'data'}], expected: '0x5b312c322c332c7b2274657374223a2264617461227d5d', error: true, errorMessage: 'Number can only safely store up to 53 bits'},
+    // { value: {test: 'test'}, expected: '0x7b2274657374223a2274657374227d', error: true, errorMessage: 'Number can only safely store up to 53 bits'},
+    // { value: '{"test": "test"}', expected: '0x7b2274657374223a202274657374227d', error: true, errorMessage: 'Number can only safely store up to 53 bits'},
+    // { value: 'myString', expected: '0x6d79537472696e67', error: true, errorMessage: 'Number can only safely store up to 53 bits'},
+    // { value: 'myString 34534!', expected: '0x6d79537472696e6720333435333421', error: true, errorMessage: 'Number can only safely store up to 53 bits'},
+    { value: new BN(15), expected: 15 },
+    { value: new BigNumber(15), expected: 15 },
+    // { value: 'Heeäööä👅D34ɝɣ24Єͽ-.,äü+#/', expected: '0x486565c3a4c3b6c3b6c3a4f09f9185443334c99dc9a33234d084cdbd2d2e2cc3a4c3bc2b232f', error: true, errorMessage: 'Number can only safely store up to 53 bits'},
```
nikoulai: These tests are no longer errors because `toHex` is called firstly and a valid input is formed. They were failing due to being too big. Are they valid inputs, though?

nazarhussain (changes requested):
Please remove the unnecessary and white space changes so we can focus on what the issue was.

packages/web3-utils/src/utils.js
```diff
@@ -216,24 +234,34 @@ var hexToUtf8 = function(hex) {
...
+    }
+
+    const [negative, hexValue] = String(value).startsWith("-")
+        ? [true, value.slice(1)]
+        : [false, value];
+    const num = toBN(hexValue);
+
+    if (num > Number.MAX_SAFE_INTEGER) {
+        return negative ? -num : num;
     }
```
nazarhussain: Does this basic check not present in the `BN` library?

nazarhussain (changes requested):

packages/web3-utils/src/utils.js
nikoulai: Not sure what you mean, I just want to return `number` if value is less than `MAX_SAFE_INTEGER` and `bigint` otherwise, like in 4.x

packages/web3-utils/src/utils.js
```diff
@@ -232,9 +232,19 @@ var hexToNumber = function (value) {
...
+    const [negative, hexValue] = String(value).startsWith("-")
+        ? [true, value.slice(1)]
+        : [false, value];
+    const num = toBN(hexValue);
+
+    if (num > Number.MAX_SAFE_INTEGER) {
+        return negative ? -num : num;
+    }
+
+    return negative ? -1 * num.toNumber() : num.toNumber();
```
nazarhussain: This seems to be a problem with the `.toNumber()`, so we need to focus on it.

packages/web3-utils/src/utils.js
```diff
@@ -232,9 +232,19 @@ var hexToNumber = function (value) {
...
 
-    return toBN(value).toNumber();
-};
+    const [negative, hexValue] = String(value).startsWith("-")
+        ? [true, value.slice(1)]
+        : [false, value];
+    const num = toBN(hexValue);
+
+    if (num > Number.MAX_SAFE_INTEGER) {
+        return negative ? -num : num;
```
nazarhussain: It will return an integer if negative and a BN object if positive.

packages/web3-utils/src/utils.js
nazarhussain: Try this out, hopefully will solve the problem. 

```suggestion
    return 1 * toBN(value); 
```

nazarhussain (changes requested):

nazarhussain (changes requested):
@nikoulai Did you get a chance to look at the feedback.

nikoulai:
@nazarhussain 
I tried the suggested solution, but, in big value tests ("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff" etc) the result is truncated (1.157920892373162e+77) due to the multiplication.

jdevcs (commented):
It will be breaking change in 1.x as discussed in team, so we should try to avoid that where possible for semver. 
I think more good approach can be to use an optional output formatting param ( hex formatting ) in function and in case user explicitly wants to have hex it should not do formatting of that field to number like https://github.com/ChainSafe/web3.js/blob/aae9d4a9a4ab6d68296f0cc1a2a29540eb3ff433/packages/web3-core-helpers/src/formatters.js#L292 . If user wants to continue output formatted to number we can show a warning. It will be backward compatible in this way, and we will be in compliance with semver.

.prettierignore
```diff
@@ -0,0 +1 @@
+**
```
luu-alex: This file was added by accident?

nikoulai:
Todo: I have to update docs if change is ok

nikoulai:
Todo: I have to update docs if change is ok

nazarhussain (approved):

jdevcs (commented):
LGTM, @nikoulai 
- yes we need to mention this in doc and 
- also could you remove new changes (space etc ) from changelog

jdevcs (approved):

PR #5160, opened 2022-06-22
Improve the rendering of the docs 

### LINKED ISSUE #5159: Improve the rendering of the docs

### What's up?

While writing docs we found some issues: 

1. The performance of loading the docs is a blocker. It could some time take minutes to load. 
2. The `@link` modifier does not work across packages. 
3. The layout of the large number of properties makes them difficult to read 

### Examples/References

_No response_

nazarhussain:
## Description

Fixes #5159

## Type of change

## Checklist:


nazarhussain:
Improve the rendering of the docs in more readable structure. 

image

nazarhussain:
It improves the performance from over a minute to seconds. 

```
Debug: Finished getting entry points in 18856ms
Debug: Finished conversion in 2792ms
Debug: JSON rendering took 668ms
```

nazarhussain:
The only issue it has is the referencing the types across the packages, which is [reported](https://github.com/milesj/docusaurus-plugin-typedoc-api/issues/47) to plugin. I am hopeful that can be resolved and we can upgrade the plugin later.

luu-alex (approved):

docs/docusaurus.config.js
```diff
@@ -44,46 +65,23 @@ const config = {
...
-					'../packages/web3-validator',
-				],
-				entryPointStrategy: 'packages',
-				cleanOutputDir: true,
-				sidebar: {
-					fullNames: true,
+				projectRoot: join(__dirname, '..'),
+				// Monorepo
+				packages: packages.map(p => `packages/${p}`),
+				minimal: false,
```
jdevcs, line 73: Its by default false for rendering minimal layout

docs/docusaurus.config.js
```diff
@@ -44,46 +65,23 @@ const config = {
...
+				debug: true,
+				changelogs: true,
+				readmes: false,
```
jdevcs, line 76: Its also false by default.

docs/docusaurus.config.js
```diff
@@ -17,6 +17,27 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
+const { join } = require('path');
+
+const packages = [
+	'web3',
+	'web3-eth',
+	'web3-eth-contract',
+	'web3-utils',
+	'web3-validator',
+	'web3-common',
+	'web3-core',
```
jdevcs, line 29: What you think should we include documenting web3.js internal packages as well.

docs/docusaurus.config.js
nazarhussain: Yes, but I want to keep it mention explicitly for readability and future safe.

docs/docusaurus.config.js
nazarhussain: Yes, but I want to keep it mention explicitly for readability and future safe.

docs/docusaurus.config.js
```diff
@@ -123,8 +121,7 @@ const config = {
 						position: 'left',
 					},
 					{
-						to: 'docs/api/', // 'api' is the 'out' directory
-						activeBasePath: 'docs',
+						to: 'api', // 'api' is the 'out' directory
 						label: 'API',
 						position: 'left',
 					},
```
jdevcs, line 127: Its not part of this PR but for now ( until we have main 4.0.0 released and as default branch on repo ) lets make:
`href: 'https://github.com/ChainSafe/web3.js',`
to
`href: 'https://github.com/ChainSafe/web3.js/tree/4.x',`

jdevcs (changes requested):

docs/docusaurus.config.js
nazarhussain: These are not just internal packages, these are public packages. So had to be included anyway.

jdevcs: - These are not directly part of eth api
- also considering 1.x parity docs
- We will not test these packages as standalone for system tests

But we can keep them in doc for users ref, and can document on least priority as compared to eth api packages.

nazarhussain: As per the the package definition these are not private packages, so must be published and documented. 

https://github.com/ChainSafe/web3.js/blob/80b07dd2d670749599123e4fdd9617d9ad5d89dd/packages/web3-core/package.json#L2

All of the code in every package is directly inheriting logic form these packages so would not be possible to completely document Eth or other packages without documenting these.

jdevcs: For sure these are **not** private packages, and already mentioned:

> But we can keep them in doc for users ref, and can document on least priority as compared to eth api packages.

( after alpha )

jdevcs:
lgtm, only some tests are failing so may be re trigger.

docs/docusaurus.config.js
nazarhussain: We can't complete the Eth or any other documentation without documenting `web3-core` and `web3-common` packages. e.g. If we need to document property `web3.eth.defaultAccount` that is inherited from the `Web3Context` in `web3-core` package, so it must be documented. Similarly for other components in those shared packages.

nazarhussain:
@jdevcs If the changes looks good to you, then please approve. so I can merge it as soon all tests passes.

avkos (approved):

jdevcs (approved):

docs/docusaurus.config.js
jdevcs: These properties are used and exposed via main API packages like eth , contract etc.

PR #5164, opened 2022-06-24
4969/account wallet docs

luu-alex:
#4969 

## Description
This PR is for the documentation of both wallets and account.
It also includes some bug fixes I discovered when making the documentation:
 - [fixing ](https://github.com/ChainSafe/web3.js/pull/5164/files#diff-92338be6c5cafbe406d5f751665ce1ac5c5c8ac415e6400313d99b50ed4af2d5R61)the return type so that wallet.add returns the wallet instead of a `boolean`
 - [Adding ](https://github.com/ChainSafe/web3.js/pull/5164/files#diff-22a17763d075d0a515469036b2a26046c3f2d1e7138bea6454348517988bbcb1R92)account.wallet to the Web3EthInterface so wallet can be used
 - [Fixing ](https://github.com/ChainSafe/web3.js/pull/5164/files#diff-55c687488c8bc965581ea1915c7fa86698f5c0d6bf74ae9f41fafc6d0f8281f2R51-R54)a bug that the web3 package would produce an `src` file instead of just the compiled files


packages/web3/src/web3.ts
```diff
@@ -91,6 +92,7 @@ export class Web3 extends Web3Context<EthExecutionAPI> {
 			recover: typeof recover;
 			encrypt: typeof encrypt;
 			decrypt: typeof decrypt;
+			wallet: any;
```
nikoulai: Wouldn't be better to have more specific type?

packages/web3-eth-accounts/src/index.ts
```diff
@@ -15,6 +15,585 @@ You should have received a copy of the GNU Lesser General Public License
...
+ * ### Example
+ *
+ * privateKeyToAddress("0x348ce564d427a3311b6536bbcff9390d69395b06ed6c486954e971d960fe8709")
+ * >0xb8CE9ab6943e0eCED004cDe8e3bBed6568B2Fa01
+ *
+ * ## signTransaction
+ *
+ * Signs an Ethereum transaction with a given private key.
+ * ### Parameters
+ * transaction - Takes a transaction object, see {@link TxData} {@link AccessListEIP2930TxData} {@link FeeMarketEIP1559TxData}
```
nikoulai: Linking for `TxData`, `AccessListEIP2930TxData` and `FeeMarketEIP1559TxData` doesn't work for me. Maybe because they are imported from `ethereumjs`?

packages/web3/src/web3.ts
luu-alex: yes, was using this as a work-around to write docs but will fix it 👍  thanks

packages/web3-eth-accounts/src/index.ts
luu-alex: Yes I think tsdocs cannot refer to links that are dependancy types

nazarhussain (approved):

nikoulai (approved):

packages/web3-eth-accounts/src/account.ts
```diff
@@ -81,10 +89,22 @@ export const hashMessage = (message: string): string => {
...
 /**
- * Signs arbitrary data. The value passed as the data parameter will be UTF-8 HEX decoded and wrapped as follows: "\x19Ethereum Signed Message:\n" + message.length + message
- *
- * @param data
- * @param privateKey
+ * Signs arbitrary data.
+ * **_NOTE:_** The value passed as the data parameter will be UTF-8 HEX decoded and wrapped as follows: "\x19Ethereum Signed Message:\n" + message.length + message
+ * @param data - The data to sign
+ * @param privateKey - The 32 byte private key to sign with
+ * @returns The signature Object containing the message, messageHash, r, s, v
```
jdevcs:
```suggestion
 * @returns The signature Object containing the message, signature, messageHash, r, s, v
```

packages/web3-eth-accounts/src/account.ts
```diff
@@ -158,8 +250,12 @@ export const signTransaction = (
 
 /**
  * Recovers the Ethereum address which was used to sign the given RLP encoded transaction.
- *
- * @param rawTransaction
+ * @param rawTransaction - The RLP encoded transaction
```
jdevcs:
```suggestion
 * @param rawTransaction - The hex string having RLP encoded transaction
```

I think with more detail?

jdevcs (changes requested):

packages/web3-eth-accounts/src/account.ts
```diff
@@ -171,10 +267,23 @@ export const recoverTransaction = (rawTransaction: HexString): Address => {
 
 /**
  * Recovers the Ethereum address which was used to sign the given data
- *
- * @param data
- * @param signature
- * @param hashed
+ * @param data - Either a signed message, hash, or the {@link signatureObject}
+ * @param signature - The raw RLP encoded signature
+ * @param hashed - If the message is hashed or not
```
jdevcs: Last param of recover should be `preFixed` Boolean according to 1.x doc and code https://github.com/ChainSafe/web3.js/blob/aae9d4a9a4ab6d68296f0cc1a2a29540eb3ff433/packages/web3-eth-accounts/src/index.js#L487   :

`preFixed - Boolean (optional, default: false): If the last parameter is true, the given message will NOT automatically be prefixed with "\x19Ethereum Signed Message:\n" + message.length + message, and assumed to be already prefixed.`

packages/web3-eth-accounts/src/account.ts
```diff
@@ -263,12 +385,79 @@ export const privateKeyToAddress = (privateKey: string | Buffer): string => {
...
+ *     r: 8,
+ *     p: 1,
+ *     dklen: 32,
+ *     salt: '210d0ec956787d865358ac45716e6dd42e68d48e346d795746509523aeb477dd'
+ *   },
+ *   mac: 'efbf6d3409f37c0084a79d5fdf9a6f5d97d11447517ef1ea8374f51e581b7efd'
+ * }
+ *}
+ *```
+ * Encrypting using pbkdf2 options
```
jdevcs, line 430: Good work for adding both scrypt and pbkdf2 examples !

packages/web3-eth-accounts/src/account.ts
```diff
@@ -410,7 +610,20 @@ export const privateKeyToAccount = (privateKey: string | Buffer): Web3Account =>
 };
 
 /**
- * Returns an acoount
+ *
+ * Generates and returns a Web3Account object that includes the private and public key
```
jdevcs, line 614: As we removed entropy param in 4.x so I think lets make it more visible to users how it works now, may be add some details like following in doc:

`It uses ethereum-cryptography/secp256k1 https://www.npmjs.com/package/ethereum-cryptography#secp256k1-curve for creation of private key that is cryptographically secure random number with certain characteristics. `

packages/web3-eth-accounts/src/wallet.ts
```diff
@@ -95,6 +199,16 @@ export class Wallet<
...
 	}
 
+	/**
+	 * Removes an account from the wallet.
+	 * @param addressOrIndex - The account address, or index in the wallet.
+	 * @returns true if the wallet was removed. false if it couldn’t be found.
+	 * ```ts
+	 * web3.eth.accounts.wallet.add('0xbce9b59981303e76c4878b1a6d7b088ec6b9dd5c966b7d5f54d7a749ff683387');
+	 *
+	 * web3.eth.accounts.wallet.remove('0x85D70633b90e03e0276B98880286D0D055685ed7'); // FIX THIS
```
jdevcs:
```suggestion
	 * web3.eth.accounts.wallet.remove('0x85D70633b90e03e0276B98880286D0D055685ed7'); 
```
Should this be removed? or is there any thing needs to be fixed?

packages/web3-eth-accounts/src/wallet.ts
```diff
@@ -109,6 +223,42 @@ export class Wallet<
 		return false;
 	}
 
+	/**
+	 * Securely empties the wallet and removes all its accounts.
```
jdevcs, line 228:
```suggestion
	 * Securely empties the wallet and removes all its accounts. Use this with *caution as it will remove all accounts stored in local wallet. 
```

packages/web3-eth-accounts/src/wallet.ts
```diff
@@ -117,12 +267,115 @@ export class Wallet<
...
 		return Promise.all(
 			Object.values(this._accounts).map(async account => account.encrypt(password, options)),
 		);
 	}
 
+	/**
+	 * Decrypts keystore v3 objects.
+	 * @param encryptedWallets `string[]` An array of encrypted keystore v3 objects to decrypt
+	 * @param password `String` The password to encrypt with
+	 * @param options decrypt options for the wallets
```
jdevcs, line 281:
```suggestion
	 * @param options (optional) decrypt options for the wallets
```

packages/web3-eth-accounts/src/account.ts
```diff
@@ -171,10 +267,23 @@ export const recoverTransaction = (rawTransaction: HexString): Address => {
 
 /**
  * Recovers the Ethereum address which was used to sign the given data
- *
- * @param data
- * @param signature
- * @param hashed
+ * @param data - Either a signed message, hash, or the {@link signatureObject}
+ * @param signature - The raw RLP encoded signature
```
jdevcs, line 271:
```suggestion
 * @param signature - (optional) The raw RLP encoded signature
```

packages/web3-eth-accounts/src/account.ts
```diff
@@ -263,12 +385,79 @@ export const privateKeyToAddress = (privateKey: string | Buffer): string => {
...
 /**
- * encrypt a private key given a password, returns a V3 JSON Keystore
- * https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition
+ * encrypt a private key with a password, returns a V3 JSON Keystore
+ *
+ * Read more: https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition
+ *
+ * @param privateKey - The private key to encrypt, 32 bytes.
+ * @param password - The password used for encryption.
+ * @param options - Options to configure to encrypt the keystore either scrypt or pbkdf2
```
jdevcs, line 394:
```suggestion
 * @param options - (optional) Options to configure to encrypt the keystore either scrypt or pbkdf2
```

packages/web3-eth-accounts/src/account.ts
```diff
@@ -419,11 +632,42 @@ export const create = (): Web3Account => {
 };
 
 /**
- *  Decrypts a v3 keystore JSON, and creates the account.
+ * Decrypts a v3 keystore JSON, and creates the account.
+ * @param keystore - the encrypted Keystore object or string to decrypt
+ * @param password - The password that was used for encryption
+ * @param nonStrict - if true and given a json string, the keystore will be parsed as lowercase.
```
jdevcs, line 641:
```suggestion
 * @param nonStrict - (optional) if true and given a json string, the keystore will be parsed as lowercase.
```

packages/web3-eth-accounts/src/wallet.ts
```diff
@@ -117,12 +267,115 @@ export class Wallet<
 		return this;
 	}
 
+	/**
+	 * Encrypts all wallet accounts to an array of encrypted keystore v3 objects.
+	 *
+	 * @param password `string` - The password which will be used for encryption
+	 * @param options - encryption options
```
jdevcs, line 259:
```suggestion
	 * @param options - (optional) encryption options
```

jdevcs (changes requested):

jdevcs:
LGTM, except some minor changes.

packages/web3-eth-accounts/src/wallet.ts
luu-alex: This is fixed in nazars [pr](https://github.com/ChainSafe/web3.js/pull/5175#event-6900514583) that was merged. will update now

packages/web3-eth-accounts/src/wallet.ts
luu-alex: Screen Shot 2022-06-29 at 2 03 07 PM
) optional is already included in docs

packages/web3-eth-accounts/src/account.ts
luu-alex: same as picture from above, optional is already included in docs so we dont need to write it

packages/web3-eth-accounts/src/account.ts
luu-alex: same as picture from above, optional is already included in docs so we dont need to write it

packages/web3-eth-accounts/src/account.ts
luu-alex: same as picture from above, optional is already included in docs so we dont need to write it

packages/web3-eth-accounts/src/wallet.ts
luu-alex: same as picture from above, optional is already included in docs so we dont need to write it

jdevcs (approved):

avkos (approved):

luu-alex:
created [issue](https://github.com/ChainSafe/web3.js/issues/5188 ) for function that is not returning expected type.

PR #5175, opened 2022-06-27
Fix bugs for the wallet and PromiEvent 

### LINKED ISSUE #5169: web3.accounts.wallet is not accessible with array index

### Current Behavior

Following always return undefined. 

```
web3.accounts.wallet[0];
```

### Expected Behavior

It should return a wallet object instead. This is a commonly used approach to access the wallet, so we should avoid introducing a breaking change. 

### Steps to Reproduce

```
web3.accounts.wallet[0];
```

### Web3.js Version

4.x

### Environment

_No response_

### Anything Else?

_No response_

### LINKED ISSUE #5171: PromiEvent does not follow the event chaining.

### Current Behavior

Currently the `PromiEvent` first subscription to the event returns the `undefined`. 

### Expected Behavior

It should return the `PromiEvent` object to allow the event chaining. 

### Steps to Reproduce

```ts
await contract.methods.myMethod().send().on('transationHash', console.log).on('receipt', console.log);
```

### Web3.js Version

4.x

### Environment

_No response_

### Anything Else?

_No response_

nazarhussain:
## Description

Fixes #5169, #5171

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-eth-accounts/test/unit/wallet.test.ts
+++ b/packages/web3-eth-accounts/test/unit/wallet.test.ts
@@ -142,6 +142,14 @@ describe('Wallet', () => {
 			expect(wallet.get('my_address')).toEqual(account);
 			expect(wallet.get('my_Address')).toBeUndefined();
 		});
+
+		it('should get account with index', () => {
+			const account = { address: 'my_Address' } as never;
+
+			wallet.add(account);
+
+			expect(wallet[0]).toEqual(account);
+		});
 	});
 
 	describe('remove', () => {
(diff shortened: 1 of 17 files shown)


packages/web3-eth-accounts/test/unit/wallet.test.ts
```diff
@@ -190,6 +198,29 @@ describe('Wallet', () => {
...
+			delete wallet[0];
+
+			// Deleting objects dees not change the length
+			expect(wallet).toHaveLength(1);
+			expect(wallet[0]).toBeUndefined();
+		});
+
+		it('should remove account with array methods', () => {
+			const account = { address: 'my_address' } as never;
+			wallet.add(account);
```
luu-alex, line 216: Can you add a testcase to remove an account that is added using a privatekey？I think it'll be a more complete test if this [part ](https://github.com/ChainSafe/web3.js/blob/477b566ea87bc0aaf862f3af54a83a8cebd14fad/packages/web3-eth-accounts/src/wallet.ts#L67) of the code gets tested too.

nazarhussain: Added the extra test case.

luu-alex (approved):

nikoulai (approved):

PR #5179, opened 2022-06-28
web3-providers-http: Migrate from xhr2-cookies to cross-fetch

nazarhussain:
## Description

Created the PR to merge https://github.com/ChainSafe/web3.js/pull/5085

## Type of change

## Checklist:


nikoulai (approved):

packages/web3-providers-http/types/index.d.ts
```diff
@@ -46,7 +45,7 @@ export interface HttpProviderOptions {
 export class HttpProvider extends HttpProviderBase {
     host: string;
 
-    withCredentials: boolean;
+    withCredentials?: string;
```
jdevcs: This will be breaking change, why we need it to be string?

packages/web3-providers-http/src/index.js
```diff
@@ -19,29 +19,33 @@
...
-    this.withCredentials = options.withCredentials || false;
+    this.withCredentials = options.withCredentials;
     this.timeout = options.timeout || 0;
     this.headers = options.headers;
     this.agent = options.agent;
     this.connected = false;
 
     // keepAlive is true unless explicitly set to false
-    const keepAlive = options.keepAlive !== false;
+    var keepAlive = options.keepAlive !== false;
```
jdevcs:
```suggestion
    const keepAlive = options.keepAlive !== false;
```
User can explicitly set it via options, why we need this change?

jdevcs (changes requested):

nazarhussain:
@jdevcs This PR contains only the `package-lock.json` changes. Rest of these comes from the community PR https://github.com/ChainSafe/web3.js/pull/5085. Opened here to pass the browsers tests.

packages/web3-providers-http/types/index.d.ts
nazarhussain: https://github.com/ChainSafe/web3.js/pull/5085#discussion_r884264478

test/httpprovider.js
```diff
@@ -5,55 +5,72 @@ var https = require('https');
...
+    resolve(response);
+  });
+};
+
 var HttpProvider = SandboxedModule.require('../packages/web3-providers-http', {
     requires: {
-        'xhr2-cookies': require('./helpers/FakeXHR2'),
+        'cross-fetch': Mock
+        // 'xhr2-cookies': require('./helpers/FakeXHR2'),
         // 'xmlhttprequest': require('./helpers/FakeXMLHttpRequest')
```
luu-alex: think we can remove these comments and FakeXMLHttpRequest file

packages/web3-providers-http/types/index.d.ts
jdevcs: We can remain consistent with existing 1.x https://web3js.readthedocs.io/en/v1.7.4/web3.html#configuration and have `withCredentials` as boolean in web3.js options obj, and based on boolean value provided by user we can pass string to fetch ( include, omit ) [here](https://github.com/ChainSafe/web3.js/blob/535c9714549e494d257bd8832dd943f803efd558/packages/web3-providers-http/src/index.js#L113)  and it will go as [fetch wants](https://github.com/ChainSafe/web3.js/blob/535c9714549e494d257bd8832dd943f803efd558/packages/web3-providers-http/src/index.js#L159) plus we will not introduce breaking change.

ghost: @jdevcs @nazarhussain Then it should be something more like `withCredentials?: boolean | string;` and handle boolean value with if block. Great idea.

nazarhussain: @jdevcs  Please re-review.

jdevcs (approved):

PR #5184, opened 2022-06-29
wallet.signTransaction fail with error 

### LINKED ISSUE #5170: wallet.signTransaction fail with error

### Current Behavior

Signing a transaction with wallet object is resulting into following error.

```
Signer Error  gasLimit is too low. given 0, need at least 53572.
```

### Expected Behavior

It should sign the transaction without any issue. 

### Steps to Reproduce

```ts
await web3.eth.accounts.wallet.get(0).signTransaction({
      data:
        "0x1234567000000000000000000000000" +
        account.address.toLowerCase().replace("0x", "") +
        "00000000000000000000000000000000000000000000000000000000000000c8",
      from: account.address.toLowerCase(),
      gas: "0xd658",
      gasPrice: "0xbb8",
      chainId: "0x1",
      nonce: "0x1",
      chain: "mainnet",
      hardfork: "petersburg",
    })
```

### Web3.js Version

4.x

### Environment

_No response_

### Anything Else?

_No response_

nazarhussain:
## Description

Fixes #5170

## Type of change

## Checklist:


packages/web3-eth-accounts/src/account.ts
```diff
@@ -398,7 +390,9 @@ export const privateKeyToAccount = (privateKey: string | Buffer): Web3Account =>
 	return {
 		address: privateKeyToAddress(pKey),
 		privateKey: pKey,
-		signTransaction: (tx: Record<string, unknown>) => signTransaction(tx, pKey),
+		signTransaction: (_tx: Record<string, unknown>) => {
+			throw new SignerError('Don not have network access to sign the transaction');
```
jdevcs:
```suggestion
			throw new SignerError('Do not have network access to sign the transaction');
```

packages/web3/src/web3.ts
```diff
@@ -67,14 +67,56 @@ export class Web3 extends Web3Context<EthExecutionAPI> {
...
+		const createWithState = () => {
+			const account = create();
+
+			return {
+				...account,
+				signTransaction: async (transaction: Transaction) =>
+					signTransactionWithState(transaction, account.privateKey),
+			};
+		};
+
```
jdevcs: These functions are in main web3 package for avoiding eth package dep in accounts pkg, or is there any thing else?

But I think these should be moved to accounts package as: If some one want to import signing functionality directly from accounts package the behavior will be diff. 

Or if you want to keep in current way, move these functions to new file in main web3 package and document that defaults will not be filled if these functions are directly used from accounts package instead of main web3 package.

packages/web3-eth-accounts/src/account.ts
```diff
@@ -142,17 +135,16 @@ export const signTransaction = (
 		throw new SignerError(errorString);
 	}
 
-	const rlpEncoded = signedTx.serialize().toString('hex');
```
jdevcs, line 237: why you are skipping rlp encoding tx?

jdevcs (changes requested):

packages/web3-eth-accounts/src/account.ts
nazarhussain: It's not skipped, just wrapped in `bytesToHex`

nazarhussain: There was a bug in this function, which was causing different transaction hash calculation compared to one if you send raw transaction to network you get different tx hash. I fixed it.

packages/web3/src/web3.ts
nazarhussain: As mentioned earlier in a discussion: 

There are two approaches to do it.

1. Make the signTransaction stateful accepting the context and handle rest 
2. Process state related stuff in the higher level object web3 or web3.eth and keep the signTransaction stateless. 

I used second approach to have minimum changes in this PR.

packages/web3/test/integration/web3.accounts.test.ts
```diff
@@ -0,0 +1,198 @@
...
+					to: nodeAccounts[0],
+					value: web3.utils.toWei('0.1', 'ether'),
+					gas: '0x1',
+					data: '0x1',
+				};
+
+				await expect(account.signTransaction(tx)).rejects.toThrow('gasLimit is too low.');
+			});
+
+			it.skip('should throw error if signed by private key not associated with "from" field', async () => {
```
avkos, line 116: Why is it skipped?

avkos (changes requested):

packages/web3/src/web3.ts
jdevcs: ok, sounds good. Could you do changes as discussed in chat. 

- changelog update (under breaking changes) for these functions exported via web3 main package will have diff behavior as compared to exported in accounts package. ( filling defaults )

- moving these state functions in diff file in same web3 package and continue using second approach

packages/web3/test/integration/web3.accounts.test.ts
nazarhussain: Need further investigation on this tests. This scenario should fail, but its passing but the behavior is same for 1.x. So I kept this test to debug further in future.

packages/web3-eth-accounts/src/account.ts
```diff
@@ -212,15 +211,13 @@ export const sign = (data: string, privateKey: HexString): signResult => {
...
+export const signTransaction = async (
+	transaction: TypedTransaction,
 	privateKey: HexString,
-): signTransactionResult => {
-	//	TODO : Send calls to web3.transaction package for :
-	//		Transaction Validation checks
-
-	const tx = TransactionFactory.fromTxData(transaction);
-	const signedTx = tx.sign(Buffer.from(privateKey.substring(2), 'hex'));
+	// To make it compatible with rest of the API, have to keep it async
```
avkos, line 217: we don't use it very often in our code. Maybe it will be better to not use async if it is just a plain function?

avkos (changes requested):

packages/web3-eth-accounts/src/account.ts
nazarhussain: During the `signTransaction` on the API level `web3.eth.accounts.signTransaction` and `web3.eth.accounts.wallet.signTransaction` we had to connect network to fetch `nonce` and `chainId` and for that we had to keep these functions `async`. 

Currently the functions in `web3-eth-accounts` are not context aware, but in future we will refactor and make this package context aware. So keeping it `async` would be future safe.

nazarhussain: https://github.com/ChainSafe/web3.js/pull/5184/files#diff-aa1ce8fa7725a4b75b0af0817bb747bf937b97057278b70a8f2dcfe1f506bb84R35-R41

avkos: :+1:

avkos (approved):

jdevcs (approved):

PR #5205, opened 2022-07-05
Incompatible return types for 4.x implementations

### LINKED ISSUE #5126: Incompatible return types for 4.x implementations
While migrating [truffle](https://github.com/trufflesuite/truffle) came across a few changes which are required to make web3@4x compatible with a lot of projects our there.

Here is the the list of individual changes:

1. [ ] `web3.eth.getTransaction` should always return transaction or throw. Should not have return type of `undefined`. 
2. [ ] `web3.eth.getTransactionReceipt` should always return receipt or throw. Should not have return type of `undefined`.
3. [ ] `TransactionReceipt` property `address` should not be optional.

nikoulai:
## Description

#5126


packages/web3-eth/src/web3_eth.ts
```diff
@@ -200,7 +200,15 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
...
 		returnFormat: ReturnFormat = DEFAULT_RETURN_FORMAT as ReturnFormat,
 	) {
-		return rpcMethodsWrappers.getTransaction(this, transactionHash, returnFormat);
+		const response = await rpcMethodsWrappers.getTransaction(
+			this,
+			transactionHash,
+			returnFormat,
+		);
+
+		if (!response) throw new Error('Transaction not found!');
```
jdevcs: I think we should use consistent approach for known errors in 4.x, like https://github.com/ChainSafe/web3.js/blob/e6a9aea25d3ce808f9a1cab4a26164366b68d3f4/packages/web3-errors/src/errors/transaction_errors.ts#L122   so could you add some thing like `TransactionNotFound` new error.

packages/web3-eth/src/web3_eth.ts
```diff
@@ -227,7 +235,15 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
...
 		ReturnFormat extends DataFormat = typeof DEFAULT_RETURN_FORMAT,
 	>(transactionHash: Bytes, returnFormat: ReturnFormat = DEFAULT_RETURN_FORMAT as ReturnFormat) {
-		return rpcMethodsWrappers.getTransactionReceipt(this, transactionHash, returnFormat);
+		const response = await rpcMethodsWrappers.getTransactionReceipt(
+			this,
+			transactionHash,
+			returnFormat,
+		);
+
+		if (!response) throw new Error('Transaction not found!');
```
jdevcs: Same here use Web3Errors

jdevcs (changes requested):

nazarhussain (approved):
Apart from @jdevcs comments, looks good to me.

jdevcs (approved):

Issue #5104, opened 2022-06-06
Migration Guide/Breaking Changes - Accounts/Wallet


mconnelly8:
This information relates to the API documentation.

jdevcs:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/ChainSafe/web3.js/5104?planning-poker) with ZenHub @avkos @luu-alex @nikoulai @spacesailor24

PR #5218, opened 2022-07-07
5104 - accounts breaking changes docs 4.x

luu-alex:
#5104
## Description

## Type of change

## Checklist:


docs/docs/guides/web3_migration_guide.md
```diff
@@ -63,3 +63,21 @@ batch.execute();
 // in 4.x
 const responses = await batch.execute();
 ```
+
+### Web3-eth-accounts
+
+#### create and wallet.create
+
```
nikoulai: Maybe we need to move this part to a separate guide? As we will have a separate for eth package.

nikoulai (approved):

docs/docs/guides/web3_migration_guide.md
```diff
@@ -63,3 +63,21 @@ batch.execute();
...
+In 1.x `entropy` is an optional parameter for the create method, 4.x does not have `entropy` as a parameter, create method uses an audited package ethereum-cryptography/secp256k1 that is cryptographically secure random number with certain characteristics.
+
+#### privateKeyToAccount
+
+In 1.x `ignoreLength` is an optional parameter for privateKeyToAccount, in 4.x `ignoreLength` is removed
```
nazarhussain: Don't you think we should fix this to avoid compatibility issues with 1.x?

docs/docs/guides/web3_migration_guide.md
```diff
@@ -63,3 +63,21 @@ batch.execute();
...
+
+#### signTransaction
+
+In 4.x a transaction object must be passed. Must be typed legacy, EIP2930 or EIP 1559 transaction type Read more about [transaction types] (https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/tx#transaction-types)
```
nazarhussain: Yes and no both. The `signTransaction` from the `web3-eth-accounts` package will not be exposed directly to the public API. The relevant methods `web3.eth.signTransaction` would accept the same parameter as 1.x. Similarly for `wallet.signTransaction`

docs/docs/guides/web3_migration_guide.md
```diff
@@ -63,3 +63,21 @@ batch.execute();
...
+
+#### recover
+
+In 4.x, instead of passing in seperate parameters `messageHash`, `r`, `s`, `v`, a signature object is passed containing those parameters
```
nazarhussain: If people are using this method, will be a breaking change for them. And the fix seems very easy, so why shouldn't we fix it?

nazarhussain (changes requested):

packages/web3/src/accounts.ts
```diff
@@ -89,7 +89,6 @@ export const initAccountsForContext = (context: Web3Context<EthExecutionAPI>) =>
 		privateKeyToAccount: privateKeyToAccountWithContext,
 		decrypt: decryptWithContext,
 	});
-
```
avkos: I don't think we need this change. can you revert it, please?

avkos (changes requested):

docs/docs/guides/web3_migration_guide.md
luu-alex: oh, thought migration guide was one page rather than seperate pages. thanks

docs/docs/guides/web3_migration_guide.md
luu-alex: initially we removed this because it was error prone, but I think ur right for compatibility

nazarhussain (approved):

avkos (changes requested):
Good Job! It looks like you have some problems with unit tests.

avkos (approved):

PR #5219, opened 2022-07-07
Update feeHistory Schema

### LINKED ISSUE #5168: `eth.getFeeHistory` doesn't return all values

### Current Behavior

returns 
```
{ oldestBlock: 1n }
```

### Expected Behavior

1.x returns
```
{
  oldestBlock: '0x1',
  baseFeePerGas: [ '0x342770c0', '0x2da89983' ],
  gasUsedRatio: [ 0.0018243056410397187 ]
}
```

### Steps to Reproduce

await web3Eth.getFeeHistory("0x1", "latest", []);

### Web3.js Version

4.x

### Environment

_No response_

### Anything Else?

_No response_

nikoulai:
## Description

#5168


packages/web3-eth/src/schemas.ts
```diff
@@ -408,6 +411,12 @@ export const feeHistorySchema = {
 				},
 			},
 		},
+		gasUsedRatio: {
+			type: 'array',
+			items: {
+				type: 'number',
```
nazarhussain, line 417: Should this not be the `eth: 'uint'` as well?

nazarhussain (changes requested):

nazarhussain (changes requested):
Does adding the schema fixes the return data as well? 

We may add some tests to make sure endpoints returns that data.

avkos (approved):

luu-alex (approved):

packages/web3-eth/test/integration/web3_eth/getFeeHistory.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+		const functionResponse = await web3Eth.getFeeHistory(
+			blockCount,
+			newestBlock,
+			rewardPercentiles,
+		);
+
+		const functionResponseKeys = Object.keys(functionResponse);
+
+		const requestResponse = await requestManager.send({
```
jdevcs: Instead of using and comparing keys with `eth_feeHistory` as we have schema so we should use it for looking keys in result.

jdevcs (changes requested):

packages/web3-eth/test/integration/web3_eth/getFeeHistory.test.ts
nikoulai: It's more to check that the endpoint and our function return the same data, in order to cover the case Nazar has mentioned.

jdevcs: The reason I suggested above change is endpoint ( direct call to 'eth_feeHistory' ) and our function ( `web3Eth.getFeeHistory(..)` ) both are using same rpc call, so we should explicitly look for keys that we are expecting for schema validation.

jdevcs: The reason I suggested above change is endpoint ( direct call to 'eth_feeHistory' ) and our function ( web3Eth.getFeeHistory(..) ) both are using same rpc call, so we should explicitly look for keys that we are expecting for schema validation

nikoulai: The reason I did it this way, was that before the schema update, the API was used to return all the properties but `web3Eth.getFeeHistory()` was missing one of them in its result. 
I changed the test, hope it's ok now.

jdevcs (approved):

nazarhussain (approved):

nazarhussain (changes requested):
@nikoulai Did you looked at my earlier comment on this PR. Does the schema change, reflects changes to types?

jdevcs (approved):

nikoulai:
> @nikoulai Did you looked at my earlier comment on this PR. Does the schema change, reflects changes to types?

@nazarhussain There was an attribute missing, I added it.

nazarhussain (approved):

PR #5221, opened 2022-07-07
ensure backward compatibility for Web3 constructor

Muhammad-Altabba:
## Description
Fixes [#(5196)](https://github.com/ChainSafe/web3.js/issues/5196)

## Type of change

## Checklist:

- [x] I ran `yarn run build` with success.


packages/web3/src/index.ts
```diff
@@ -327,3 +327,8 @@ export * from './web3';
 export * from './types';
 
 export default Web3;
+
+// The following JavaScript command is used instead of the equivalent TypeScript `export = Web3`, 
+//  to allow exporting the types along with Web3 constructor as the default function,
+//  in order to maintain backward compatibility.
+module.exports = Web3;
```
nazarhussain: Need to disable the TS error. 

```suggestion
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-expect-error
module.exports = Web3;
```

nazarhussain (dismissed):
Please make this change to every `default` export in all the packages.

packages/web3/src/index.ts
Muhammad-Altabba: Thanks for the hint. However, I changed the approach :blush:

Muhammad-Altabba:
With the last commit, the `export` and `import` behaves exactly as it was in version 1.x (this is as far as I know. And let me know if this is not true) :smile: 

However, this also means that the `types` inside `web3` package are no longer exported. So, I recommend having a follow-up task to put those types possibly at `web3-core`. This can be done perhaps in `beta` or at a later version. And this could be done along with the refactoring proposed at https://github.com/ChainSafe/web3.js/discussions/5223, or separately.

packages/web3/src/index.ts
```diff
@@ -323,7 +323,4 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
  */
 import { Web3 } from './web3';
 
-export * from './web3';
-export * from './types';
```
nazarhussain, line 327: We can't skip exporting types from this package as well the named export.

nazarhussain (changes requested):
Apart form the mentioned change, we need to apply same changes to all packages which are using `export default`.

packages/web3/src/index.ts
luu-alex: why can't we skip exporting types? using muhammads branch in a test project and looks like it still gets type safety for `Web3EthInterface`

nazarhussain (approved):

packages/web3/src/index.ts
nazarhussain: Unfortunately we have to skip the types export due to a limitation in TS. We can't combine name and default exports in CommonJS.

https://github.com/microsoft/TypeScript/issues/2719#issuecomment-123960556

jdevcs (approved):

jdevcs:
lgtm, plus we should add simplest blackbox tests as well https://github.com/ChainSafe/web3.js/issues/5261 as part of new PR

avkos (approved):

PR #5226, opened 2022-07-11
Return Subscription object without await

### LINKED ISSUE #5203: Subscription object should be returned without await

### Current Behavior

Currently the user have to use `await` to fetch the subscription object before listening to it. 

```ts
(await myContract.events.MyEvent()).on('data', console.log)
```

### Expected Behavior

Logically this can be avoid that will make it consistent with the `1.x`. We can throw `error` event if it fails to subscribe. 

```ts
myContract.events.MyEvent().on('data', console.log).on('error', console.error); 
```

or

```ts
myContract.events.MyEvent().on('data', console.log).catch(console.error); 
```

### Steps to Reproduce

Mentioned in example above. 

### Web3.js Version

4.x

### Environment

- Operating System: 
- Browser: 
- Node.js Version: 
- NPM Version: 

### Anything Else?

To make this work, we have to refactor the subscription manager. It's trivial change but not complex.

nikoulai:
## Description

Fixes #5203


luu-alex (approved):

packages/web3-eth-contract/src/contract.ts
```diff
@@ -1050,7 +1051,9 @@ export class Contract<Abi extends ContractAbi>
 				{ requestManager: this.requestManager },
 			);
 
-			await this.subscriptionManager?.addSubscription(sub);
+			this.subscriptionManager?.addSubscription(sub).catch(() => {
+				throw new SubscriptionError('Failed to subscribe.');
+			});
```
nazarhussain: I would suggest to create an `error` event and `emit` the error on `sub` object in this particular case.

nazarhussain (changes requested):

jdevcs (approved):
apart from Nazar comments LGTM

avkos (approved):

nazarhussain (approved):

PR #5227, opened 2022-07-12
Add Web3BatchProvider

### LINKED ISSUE #5222: Support BatchRequest for getPastEvents
### Is there an existing issue for this? No.

### Feature Request

It would be great if BatchRequest supports getPastEvents ( eth_getLogs ), so that it would reduce calls for fetching events which would fulfill getting event logs from public nodes with rate limits.

### Use Cases

Many public nodes have rate limits as well as limits for block length for a single eth_getLogs call ( 5K block for BSC https://docs.bnbchain.org/docs/rpc ). Without batch requests, it would take a lot of time to fetch historic contract logs due to those limits applied ).

### Implementation Ideas

Support web3.eth.Contract().getPastEvents to call for eth_getLogs, I don't think 4.x branch doesn't support this feature as well ( Although it is technically possible & already available for ethers.js )

_No response_

### Are you willing to implement this feature? No. ( This would create incompatibility for type definitions for either BatchRequest or getPastEvents )

- [ ] Yes

ghost:
Fixes #5222

## Description

Add Web3BatchProvider which aggregates short-term ( 10ms ) RPC requests to the form of a batch request specified from [JSON-RPC 2.0 spec](https://www.jsonrpc.org/specification#batch).

( Inspired by how [BatchProvider from ethers.js](https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/json-rpc-batch-provider.ts#L36) works ) 

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -563,6 +563,9 @@ Released with 1.0.0-beta.37 code base.
 
 ## [Unreleased]
 
+### Added
+-  Added Web3BatchProvider (#5227)
+
 ## [1.7.5]
 
 ### Changed
@@ -584,4 +587,3 @@ Released with 1.0.0-beta.37 code base.
 
 ### Security
 -  Updated `got` lib version and fixed other libs using npm audit fix (#5178) (#5254)
-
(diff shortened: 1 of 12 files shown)


jdevcs:
@ayanamidev could you make contribution by adding this in our [4.x](https://github.com/ChainSafe/web3.js/tree/4.x)

PR #5229, opened 2022-07-12
Document string interface for encodeFunctionCall

TimDaub:
- I was surprised that `abi.encodeFunctionCall` was document as only being compatible with a JSON as it's interface input (https://web3js.readthedocs.io/en/v1.2.11/web3-eth-abi.html#id13). Especially because `abi.encodeFunctionSignature` would allow both submitting a string or a JSON object.
- But I had a look into the source code and found that `encodeFunctionCall` uses `encodeFunctionSignature` internally and so passing a string is fine too: https://github.com/ChainSafe/web3.js/blob/35d8f7f76ca9412b714775cf15a2c1d57001020b/packages/web3-eth-abi/src/index.js#L314


TimDaub:
nevermind, it depends on `jsonInterface.inputs` :(

PR #5251, opened 2022-07-15
Web3Eth Docs

### LINKED ISSUE #4973: Eth Documentation

Spacesai1or:
Adds documentation for `web3_eth.ts` and `rpc_method_wrappers.ts` RPC functions

Non-RPC functions such as:

- `detectTransactionType`
- `formatTransaction`
- `isAccessList`
- `isAccessListEntry`
- `isBaseTransaction`
- `isTransaction1559Unsigned`
- `isTransaction2930Unsigned`
- `isTransactionCall`
- `isTransactionlegacyUnsigned`
- `isTransactionWithSender`
- `prepareTransactionForSigning`
- `transactionBuilder`
- `validateChainInfo`
- `validateCustomChainInfo`
- `validateFeeMarketGas`
- `validateGas`
- `validateLegacyGas`
- `validateTransacitonCall`
- `validateTransactionForSigning`
- `validateTransactionWithSender`

are exported, but not currently manually documented

closes #4973


packages/web3-eth/src/index.ts
```diff
@@ -15,6 +15,25 @@ You should have received a copy of the GNU Lesser General Public License
...
+ * ```ts
+ * import { getBalance, getBlockByNumber } from 'web3-eth';
+ * ```
+ *
+ * To use this package within the `web3` object use:
+ * ```ts
+ * import { Web3 } from 'web3';
+ *
+ * const web3 = new Web3(Web3.givenProvider || 'ws://some.local-or-remote.node:8546');
+ * web3.eth.getBalance('0x0...');
```
Muhammad-Altabba:
```suggestion
 * await web3.eth.getBalance('0x0...');
```

packages/web3-eth/src/web3_eth.ts
```diff
@@ -95,52 +95,169 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
 			registeredSubscriptions,
 		});
 	}
+
+    /**
+     * @returns Returns the ethereum protocol version of the node.
+     *
+     * ```ts
+     * web3.eth.getProtocolVersion().then(console.log);
```
Muhammad-Altabba: I think it is better to use `await` and not use `.then(...)` to make the code snippet more elegant, here and in all other places. Or what do you think?
```suggestion
     * const version = await web3.eth.getProtocolVersion();
     * console.log(version);
```

Muhammad-Altabba (changes requested):
Looks good!
However, I suggest using `await` in the sample coding snippits.

packages/web3-eth/src/web3_eth.ts
Spacesai1or: I thought so as well, but `.then` was used elsewhere by other devs, so was just trying to stay consistent

packages/web3-eth/src/index.ts
Muhammad-Altabba: You may also use `.then(...)` if you feel like staying on the same style of the previous documentation style.

packages/web3-eth/src/index.ts
```diff
@@ -15,6 +15,25 @@ You should have received a copy of the GNU Lesser General Public License
 along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+/**
+ * The `web3-eth` package allows you to interact with an Ethereum blockchain and Ethereum smart contracts.
+ *
+ * To use this package standalone and use its methods use:
+ * ```ts
+ * import { getBalance, getBlockByNumber } from 'web3-eth';
```
jdevcs: I think we should be more descriptive for using such methods, by adding an example here.

packages/web3-eth/src/index.ts
```diff
@@ -15,6 +15,25 @@ You should have received a copy of the GNU Lesser General Public License
...
+ * ```
+ *
+ * To use this package within the `web3` object use:
+ * ```ts
+ * import { Web3 } from 'web3';
+ *
+ * const web3 = new Web3(Web3.givenProvider || 'ws://some.local-or-remote.node:8546');
```
jdevcs, line 41: Do you think we should add a valid public mainnet endpoint in examples? so people copying examples will be facilitated. 
https://github.com/arddluma/awesome-list-rpc-nodes-providers#ethereum

and should add a note that this rpc endpoint not managed by this lib maintainers.

packages/web3-eth/src/web3_eth.ts
```diff
@@ -167,6 +327,68 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
 		return rpcMethodsWrappers.getCode(this, address, blockNumber, returnFormat);
 	}
 
+	/**
+	 * @param block The {@link BlockNumberOrTag} (defaults to {@link Web3Eth.defaultBlock}) or block hash of the desired block.
```
jdevcs, line 333: its desc :

```suggestion
	 * Returns a block matching the block number, block hash or block tag.
	 *
	 * @param block The {@link BlockNumberOrTag} (defaults to {@link Web3Eth.defaultBlock}) or block hash of the desired block.
```

packages/web3-eth/src/web3_eth.ts
```diff
@@ -175,6 +397,22 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
 		return rpcMethodsWrappers.getBlock(this, block, hydrated, returnFormat);
 	}
 
+	/**
+	 * @param block The {@link BlockNumberOrTag} (defaults to {@link Web3Eth.defaultBlock}) or block hash of the desired block.
```
jdevcs, line 403: should add its desc

```suggestion
	 * Returns the number of transaction in a given block.
	 * 
	 * @param block The {@link BlockNumberOrTag} (defaults to {@link Web3Eth.defaultBlock}) or block hash of the desired block.
```

packages/web3-eth/src/web3_eth.ts
```diff
@@ -259,6 +818,108 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
 		return rpcMethodsWrappers.getTransactionCount(this, address, blockNumber, returnFormat);
 	}
 
+	/**
+	 * @param transaction The {@link Transaction} or {@link TransactionWithLocalWalletIndex} to send.
+	 * @param returnFormat ({@link DataFormat} defaults to {@link DEFAULT_RETURN_FORMAT}) Specifies how the return data should be formatted.
+	 * @param options A configuration object used to change the behavior of the `sendTransaction` method.
```
jdevcs, line 826: we should also add an example for usage of `options?: SendTransactionOptions,` options A configuration object used to change the behavior of the `sendTransaction` method.

packages/web3-eth/src/web3_eth.ts
```diff
@@ -282,6 +1040,57 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
 		return rpcMethodsWrappers.sign(this, message, address, returnFormat);
 	}
 
+	/**
```
jdevcs, line 1055: We should add description that node wallet Account is used for signing in this function via `eth_signTransaction` instead of using local wallet.

packages/web3-eth/src/web3_eth.ts
```diff
@@ -291,6 +1100,14 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
 
 	// TODO Decide what to do with transaction.to
 	// https://github.com/ChainSafe/web3.js/pull/4525#issuecomment-982330076
+	/**
+	 * Executes a message call transaction, which is directly executed in the VM of the node, but never mined into the blockchain.
```
jdevcs: Lets be more descriptive for this function
```suggestion
	 * Executes a message call in EVM without creating transaction. It does not publish anything to the blockchain and also does not consume any gas.
```

packages/web3-eth/src/web3_eth.ts
```diff
@@ -299,6 +1116,30 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
 		return rpcMethodsWrappers.call(this, transaction, blockNumber, returnFormat);
 	}
 
+	/**
+	 * Executes a message call or transaction and returns the amount of the gas used.
```
```suggestion
	 *   Generates and returns an estimate of how much gas is necessary to allow the transaction to complete. The transaction will not be added to the blockchain.
```

packages/web3-eth/src/web3_eth.ts
```diff
@@ -307,17 +1148,96 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
...
+	 *     "0xd1ff1c01710000000000000000000000d1ff1c01710000000000000000000000"
+	 * ]
+	 * ```
+	 */
 	public async getWork() {
 		return rpcMethods.getWork(this.requestManager);
 	}
 
+	/**
+	 * Used for submitting a proof-of-work solution.
```
jdevcs, line 1240: Should we mark that its not tested in 4.x

packages/web3-eth/src/web3_eth.ts
```diff
@@ -327,20 +1247,125 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
 	}
 
 	// TODO - Format addresses
+	/**
+	 * This method will request/enable the accounts from the current environment.
```
```suggestion
	 * This method will request/enable the accounts from the current environment and for supporting [EIP 1102](https://eips.ethereum.org/EIPS/eip-1102)
```

jdevcs (changes requested):

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -71,18 +71,80 @@ import { watchTransactionForConfirmations } from './utils/watch_transaction_for_
 import { Web3EthExecutionAPI } from './web3_eth_execution_api';
 import { NUMBER_DATA_FORMAT } from './constants';
 
+/**
+ *
```
jdevcs, line 75: I think we should add some note in documentation like:
All these functions are available in web3Eth class and normal users should use from there. These RPC wrapper functions are used internally in web3Eth class and are also available for advance users of lib if they want.

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -147,6 +275,27 @@ export async function getStorageAt<ReturnFormat extends DataFormat>(
...
+ * Get the code at a specific address.
+ *
+ * @param web3Context ({@link Web3Context}) Web3 configuration object that contains things such as the provider, request manager, wallet, etc.
+ * @param address The address to get the code from.
+ * @param blockNumber ({@link BlockNumberOrTag} defaults to {@link Web3Eth.defaultBlock}) Specifies what block to use as the current state for the code query.
+ * @param returnFormat ({@link DataFormat} defaults to {@link DEFAULT_RETURN_FORMAT}) Specifies how the return data should be formatted.
+ * @returns The [data](https://ethereum.org/en/developers/docs/transactions/#the-data-field) at the provided `address`.
+ *
+ * ```ts
+ * web3.eth.getCode("0x033456732123ffff2342342dd12342434324234234fd234fd23fd4f23d4234").then(console.log);
```
jdevcs, line 288: Mostly functions in Web3Eth class are only returning results of wrappers by passing context, so Could you check if possible, can we use inherit doc feature https://typedoc.org/tags/inheritDoc/ for copying doc from Web3Eth class functions into these wrappers in rendered doc? and add additional @ param  web3Context in these functions. 

This will help us for avoiding duplicate effort for maintaining doc.

jdevcs (changes requested):

packages/web3-eth/src/index.ts
nikoulai: if we do this, I think there are other places we should make the change too.

nikoulai (approved):

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -110,6 +198,23 @@ export async function getBlockNumber<ReturnFormat extends DataFormat>(
...
+ * @param address The address to get the balance of.
+ * @param blockNumber ({@link BlockNumberOrTag} defaults to {@link Web3Eth.defaultBlock}) Specifies what block to use as the current state for the balance query.
+ * @param returnFormat ({@link DataFormat} defaults to {@link DEFAULT_RETURN_FORMAT}) Specifies how the return data should be formatted.
+ * @returns The current balance for the given address in `wei`.
+ *
+ * ```ts
+ * web3.eth.getBalance("0x407d73d8a49eeb85d32cf465507dd71d507100c1").then(console.log);
+ * > 1000000000000n
+ *
+ * web3.eth.getBalance("0x407d73d8a49eeb85d32cf465507dd71d507100c1").then(console.log);
```
luu-alex:
```suggestion
 * web3.eth.getBalance("0x407d73d8a49eeb85d32cf465507dd71d507100c1", { number: FMT_NUMBER.HEX , bytes: FMT_BYTES.HEX }).then(console.log);
```

packages/web3-eth/src/web3_eth.ts
Spacesai1or: Web3 doesn't do anything special with the data the user submits, so it's almost entirely up to the client the user is connected to whether it works or not

packages/web3-eth/src/index.ts
nazarhussain: There is a feature in Docusaurus to provide executable examples. If we want to provide real network endpoints, then we should use that feature to allow users to run these examples.

nazarhussain (approved):

packages/web3-eth/src/index.ts
Spacesai1or: Addressed by [this](https://github.com/ChainSafe/web3.js/pull/5251/commits/f6033c0bd52381c8532fc4afbc36e265d8b8199c) commit

packages/web3-eth/src/index.ts
Spacesai1or: I agree with @nazarhussain, we should investigate executable examples as this was an initial goal of 4.x's documentation. Either way, we run the risk of the endpoint we use being abused/spammed, so we'd want to be careful what endpoint we use as to not cause burden on someone providing a public endpoint

packages/web3-eth/src/web3_eth.ts
Spacesai1or: `block` is the name of the first parameter of this method, referring to the block identifier that's being requested

packages/web3-eth/src/web3_eth.ts
Spacesai1or: The proposed description is the same as the `@returns` description. Additionally, I think the method name, `getBlockTransactionCount` is descriptive enough. The method that have descriptions are same ones that are described in 1.x's docs

packages/web3-eth/src/web3_eth.ts
Spacesai1or: I [added](https://github.com/ChainSafe/web3.js/pull/5251/commits/04881110d73286f137462c136397d76d3e39661a) an exmaple, but I think a better description of the properties of `SendTransactionOptions` should be done in the type's docs

packages/web3-eth/src/web3_eth.ts
Spacesai1or: Addressed by [this](https://github.com/ChainSafe/web3.js/pull/5251/commits/990128fc7f8421a89e6d186e158588189ca37700) commit

packages/web3-eth/src/web3_eth.ts
Spacesai1or: Addressed by [this](https://github.com/ChainSafe/web3.js/pull/5251/commits/c54e3f6d239196516566a7b5c823e344c9727dd2) commit

packages/web3-eth/src/web3_eth.ts
Spacesai1or: Addressed by [this](https://github.com/ChainSafe/web3.js/pull/5251/commits/1414a1a5ed8f1d0f7737b44df6c92656058cee24) commit

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: There's no single place to add that disclaimer, but we did talk about linking to the `Web3Eth` methods for each `rpc_method_wrapper` method

packages/web3-eth/src/index.ts
jdevcs: We will use executable examples in web3 tutorials section, and may be can explore to get provider string from user in that case for example to run.

packages/web3-eth/src/web3_eth.ts
jdevcs: thats correct. I suggested to add description before block param.

packages/web3-eth/src/rpc_method_wrappers.ts
jdevcs: there are multiple places / rpc functions, you are right, but we can add this note at top (may be before first wrapper function only or at top section?)  like one of:
image
https://docusaurus.io/docs/next/markdown-features/admonitions

jdevcs:
Good work Wyatt for lifting Eth package

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: This would definitely be an improvement over the duplicated docs, however, I cannot get `@inheritdoc` to work. Additionally, I don't think the code example would be inherited based off what's listed here:

image

luu-alex (approved):

packages/web3-eth/src/web3_eth.ts
Spacesai1or: Implemented via [this](https://github.com/ChainSafe/web3.js/pull/5251/commits/3ce68c0082d4838db6489bc0384ca04652293184) commit

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: Made an [issue](https://github.com/ChainSafe/web3.js/issues/5295) to track this

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: Made an [issue](https://github.com/ChainSafe/web3.js/issues/5295) to track this

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: Looks like [@remarks](https://typedoc.org/tags/remarks/) should maybe solve the code example being copied

packages/web3-eth/src/index.ts
Spacesai1or: Addressed via [this](https://github.com/ChainSafe/web3.js/pull/5251/commits/0086f4d2b1f32c5992ef1b5678458e5e6a4fadc6) commit

jdevcs (approved):

Muhammad-Altabba (approved):

PR #5262, opened 2022-07-18
Improve the code maintainability by reducing the packages 

### LINKED ISSUE #5243: Improve the code maintainability by reducing the packages

### What's up?

Based on the [discussion](https://github.com/ChainSafe/web3.js/discussions/5223) we need to merge the `web3-common` package into the `web3-core` package. The details and reasoning for the refactor is mentioned in the discussion topic. 

### Examples/References

_No response_

nazarhussain:
## Description

Fixes #5243

To clarify the scope and domain for the each package, renamed the `web3-common` to the `web3-types` package. 

Now `web3-types` will contains 
- Generic types 
- Reusable types from different packages 
- Abstract classes to be used as interfaces in other packages
- Reusable interfaces. 

To keep the consistency with the 1.x the `formatters` are moved to the `web3-core` package along with internally used classes e.g. `Web3EvetEmitter`. 

The formatter logic is for now moved to `web3-utils` package, because the whole functionality is based on a utility function `format`. Similarly for the `jsonRpc` utility functions. 

Now each package have a clear scope and domain, which is more understandable for internal team. The types package will be very handy and useable for different Typescript Projects. 

## Type of changes

## Checklist:


jdevcs (commented):
The scope defined in this PR is different from points mentioned in https://github.com/ChainSafe/web3.js/issues/5243#issuecomment-1184236531 ?

nazarhussain:
The reason is also explained in the comment. To resolve the circular dependency this was the quick and appropriate change.

jdevcs (changes requested):
we are one sprint away from alpha release. and doing last minute changes might impact that, so its not optimal direction as we discussed this isn't functional requirement. 

But If we are able to manage this PR without impacting alpha ( release goal and other PRs in progress ) thn we can select it. Fixing all tests and doc changes required.

nazarhussain:
We are not making last minute change to any behavior, which could have a side effects. It's just the file structure changes and as soon all tests are green there is no point of concern.

avkos (approved):
Great Job! :muscle:

Muhammad-Altabba:
Seams nice work!
However, could we have the documentation update in this MR? Or will it be in another one? It would be nice to review the changes and their documentation, all at once.

packages/web3-errors/src/types.ts
```diff
@@ -15,14 +15,12 @@ You should have received a copy of the GNU Lesser General Public License
 along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-// duplicate types from web3-common for avoiding cyclic dep between web3-errors and web3-common
 export type ConnectionEvent = {
 	code: number;
 	reason: string;
 	wasClean?: boolean; // if WS connection was closed properly
 };
```
Muhammad-Altabba: I think we can now stop duplicating this type here, and import it from web3-types instead.
```suggestion
```

packages/web3-errors/src/types.ts
```diff
@@ -15,14 +15,12 @@ You should have received a copy of the GNU Lesser General Public License
...
 
-// To avoid dependency of `web3-common` have to use fixed types
 export type Log = {
 	readonly removed?: boolean;
 	readonly logIndex?: bigint;
```
Muhammad-Altabba: Can we use `LogAPI` from `web3-types`? As it has the same properties.
```suggestion
```

Also for `TransactionReceipt` bellow. Is n't it the same as `TransactionReceipt` in `web3-types`?

jdevcs (commented):
This PR "**Improve the code maintainability by reducing the packages**" haven't reduced package and increased number of inter dependencies as compared to current 4.x.

nazarhussain:
@jdevcs There were three points mentioned in the discussion. 

image

All of these points are covered in this PR. The reason why not done as per initial recommendation is highlighted in the issue.

jdevcs:
Good work for thinking in this [direction](https://github.com/ChainSafe/web3.js/discussions/5223) for reducing package and improve package dependency. 

> The reason why not done as per initial recommendation is highlighted in the issue.

Considering cost of increasing inter dependencies ( 'average javascript module loading time' ...etc ) this will not aligned for end users.

nazarhussain:
@jdevcs I believed it's a false assumption. I don't see any measurable build time difference for our monorepo and neither I think it will impact the endusers.

Spacesai1or (approved):
Thank you @nazarhussain for putting this together, the addition of `web3-types` was long overdue

I greatly appreciate the cleaning up and cohesion of our shared types via a single package - makes thing a lot easier to maintain and wrap your head around. I'm not entirely sure how this affects the existing documentation that was prepared for the alpha release, but I'm assuming it was minimal being that the documentation for `web3-common` and `web3-core` was only auto generated. I don't know if moving the shared types to a different package affects the type linking for TypeDoc, but I imagine it uses the `import`s to source the types anyways, so it should update automatically

Being that we are about to release the alpha, I could understand waiting to merge this until after the initial alpha release, but given the number of files changed, this PR could grow stale very quickly and become unwieldy to keep updated

My preference would be to merge this now to refrain from having obnoxious merge conflicts, especially because I see the inclusion of a `web3-types` package as inevitable

nazarhussain:
@spacesailor24 Yes this refactoring is not gonna impact documentation as none of the effected code was documented statically (as of my review of the docs). And auto generated API docs will be generated as it was earlier.

CHANGELOG.md
```diff
@@ -329,15 +329,15 @@ Released with 1.0.0-beta.37 code base.
...
-1. The function `outputBigNumberFormatter` in `web3-core-helper` renamed to `outputBigIntFormatter` under `web3-common`
+1. The function `outputBigNumberFormatter` in `web3-core-helper` renamed to `outputBigIntFormatter` under `web3-core`
 2. Removed `this.defaultBlock` context from `inputDefaultBlockNumberFormatter` in `web3-core-helper` and converted to additional parameter
 3. Removed `this.defaultBlock` context from `inputTransactionFormatter` in `web3-core-helper` and converted to additional parameter
 
 #### web3-utils
 
-1. The following functions `soliditySha3` `soliditySha3Raw` `encodePacked` now includes type validation and requires type specficiation, instead of guessing the value type
+1. The following functions `soliditySha3` `soliditySha3Raw` `encodePacked` now includes type validation and requires type specification, instead of guessing the value type
 2. The functions `soliditySha3` `soliditySha3Raw` `encodePacked` does not support BN and now supports `BigInt`
```
Muhammad-Altabba: Did you mean something like this:
```suggestion
2. The functions `soliditySha3`, `soliditySha3Raw` and `encodePacked` did not support BN; But, now supports `BigInt`
```

packages/web3-errors/src/errors/response_errors.ts
```diff
@@ -15,11 +15,23 @@ You should have received a copy of the GNU Lesser General Public License
...
-/* eslint-disable max-classes-per-file */
+// eslint-disable-next-line max-classes-per-file
+import { JsonRpcResponse, JsonRpcResponseWithError } from 'web3-types';
+import { Web3Error } from '../web3_error_base';
+import { ERR_INVALID_RESPONSE, ERR_RESPONSE } from '../error_codes';
 
-import { ERR_INVALID_RESPONSE, ERR_RESPONSE, Web3Error } from 'web3-errors';
-import { isResponseWithError } from './json_rpc';
-import { JsonRpcResponse } from './types';
+// To avoid circular package dependency, copied to code here. If you update this please update same function in `json_rpc.ts`
```
Muhammad-Altabba, line 23: Maybe we can have a future task to merge all `web3-utils`, `web3-errors` and `web3-validator` in one package to simplify dependencies.

jdevcs:
> I believed it's a false assumption. I don't see any measurable build time difference for our monorepo and neither I think it will impact the endusers.

This PR increased 3 inter-dependencies. For end user loading a package with more dependencies it definitely increases loading time if you see js module loading. We are standing at same point or even more complex inter dependencies as compared to 4.x. 

IMO initial attempt "Reduce the packages and [improve package dependency for 4.x](https://github.com/ChainSafe/web3.js/discussions/5223)" of following goals:

- reducing a package
-  improve package dependency

are not achieved in this PR and we will increase inter dependencies  from 97 to 99 so this PR will not make any difference.

jdevcs (dismissed):
If you can reduce inter dependencies to same or less number we can merge this PR. else it will just add additional loading time for end users.

nazarhussain:
@jdevcs There is no reason to avoid dependency of `web3-types` package. I don't see any measurable effect of having this package. @spacesailor24 also agrees to it. If you have some stats showing that `web3-types` package has increased build time please share. 

Where as your other point concerns, it's your misunderstanding about that refactoring. The `web3-common` package is removed which was intended from start. The inclusion of `web3-types` is long awaited and separate thread. As @spacesailor24 explains it's inevitable.

jdevcs:
> If you have some stats showing that web3-types package has increased build time please share.

There isnt any increase in `build time` for our lib, I haven't mentioned that:

 
 But Increasing inter dependencies , if an external user is using lib it increases `loading time ` for him. Please refer to **average javascript module loading time** and impact of increasing inter-dependencies (are increased in this PR) or external dependencies (we haven't increased that in this PR). 
 
In with this PR: 
image

With current 4.x:
image

The complexity is increased and is measurable, you can use `lerna-dependency-graph` text form for numbers.

nazarhussain:
Yes I see this graph way before when doing this refactoring. I am referring to the fact that I don't see any increase in build time. For users point of view they don't need to be worried either as in modern DAAP development ecosystem everything is bundled together and only **one bundle is loaded** in the page.  If user decide to create chunks, is his preference but have no relations to our packages interdependencies. 

If this refactoring would have any reasonable impact on the build time for our project, we would have thought it will impact users as well. Otherwise it's a false assumption that it will impact users.

nazarhussain:
If we assume for a minute that your assumption is correct, even then providing clarity and domain separation is more valuable.

nikoulai (approved):
The PR, seems to me, to achieve its initial purpose of clearing things up, good job! thank you.

Being just a couple of weeks before the alpha release, my initial thought was to wait, but, Wyatt is right that it will stale fast, so I would agree to merge now. 

Regarding the creation of `web3-types`, it introduces one more inner dependency but will be of great use, so I'm ok with this.
What are your thoughts on merging `web3-errors` to `web3-types`?

jdevcs:
> If we assume for a minute that your assumption is correct, even then providing clarity and domain separation is more valuable.

usability vs maintainability. At end of day aim of this lib is facilitating users more as compared to maintainers. so domain separation is less value able for end users as compared to DAPP performance, loading time etc, when there is trade off bw these. 

> If this refactoring would have any reasonable impact on the build time for our project, we would have thought it will impact users as well. Otherwise it's a false assumption that it will impact users.

I mentioned couple of times there is no impact of build time so we shouldn't mix it again and again, and referring loading time or adding more inter-dependencies as false assumption is misleading. Its not assumption Its measurable as I shared numbers. 

If you want to merge this PR we can do that. but as IMO this will add extra cost in params shared earlier.

Muhammad-Altabba (approved):
LGTM.
However, I think that after merging this, we can now draw another dependency diagram and see how to improve further and reduce dependencies.
I see for example, that we can consider merging web3-utils, web3-errors and possibly web3-validator and even with web3-types. Because all those 4 packages are used by most of the other packages (except for web3-validator but that is used by couple of packages but not all most of them. And we may keep it as a separate one).

nazarhussain:
@jdevcs Number of connections is not that important as these does not impact on the load time for the end users. As mentioned earlier, the in modern DAAP development ecosystem, every dependency is bundled together and just that bundle is loaded. So the number of connections does not matter. If something is really impacting the build time then we should be worried. That's why I am stating it's your false assumption that end users will be impacted. 

There are few observations on your discussion: 

1. We created following packages `web3-errors`, `web3-validator`, `web3-core`. Which have a lot more impact, and we never had that discussion. And now for number of connections from `97` to `99` had raise so much suspicion and concerns. The creation of `web3-errors` would have created a lot more connections. If some parameter is really alarming, then it should be in every case. 

2. If that number of connections is that much important then let's merge `web3-core`, `web3-validator`, `web3-errors` and `web3-types` into one **`web3-utils`** package. Everything will work as it is but we will have a lot less connections. That's definitely not what you either wanted, that implies the argument of number of connections in packages is not so important. 

3. There are a lot of very famous projects with a lot of amazing people working on e.g. [hardhat](https://github.com/NomicFoundation/hardhat/tree/master/packages), [truffle](https://github.com/trufflesuite/truffle/tree/develop/packages). If we look closely on their packages we can identify a couple of packages that can be merged together to reduce number of interconnections. But all those projects didn't do that and gave more important to maintainable and coherent codebase with clear separation of concern.

packages/web3-errors/src/errors/response_errors.ts
nazarhussain: For a large code base in JS ecosystem, number of dependencies is not that valuable over a maintainable codebase. That's why the concept of the monorepo became so much popular, as it allows you to create domain specific application components in different packages to have clear understanding. 

Otherwise I would assume over 90% of the users are just concerned to `web3` package, so why not just create one with zero interdependencies.

jdevcs:
> 1.We created following packages web3-errors, web3-validator, web3-core. Which have a lot more impact, and we never had that discussion. And now for number of connections from 97 to 99 had raise so much suspicion and concerns. The creation of web3-errors would have created a lot more connections. If some parameter is really alarming, then it should be in every case.

Ideally It should be in every case. If we didn't consider optimizations in past, ignoring in future is not good choise. Plus at start we were focused more on functional items, now its time for focus on optimizations as well,  as we have most of functionality in place. 

> 2.If that number of connections is that much important then let's merge web3-core, web3-validator, web3-errors and web3-types into one web3-utils package. Everything will work as it is but we will have a lot less connections. That's definitely not what you either wanted, that implies the argument of number of connections in packages is not so important.
Sure we can consider really achieveing initially discussed objectives. 

Based on initial discussion ( reducing packages and reducing inter dependency complexity ) if it can be achieved using above suggestion, it can be considered for sure. If we can have some measurable improvements its useful to discuss. 

> 3.There are a lot of very famous projects with a lot of amazing people working on e.g. [hardhat](https://github.com/NomicFoundation/hardhat/tree/master/packages), [truffle](https://github.com/trufflesuite/truffle/tree/develop/packages). If we look closely on their packages we can identify a couple of packages that can be merged together to reduce number of interconnections. But all those projects didn't do that and gave more important to maintainable and coherent codebase with clear separation of concern

Every project has its own objectives and can be in different development phase. As we are in complete rewrite phase so few things can be different in our case.

PR #5273, opened 2022-07-21
implement handleRevert

### LINKED ISSUE #5069: Implement getRevertReason for 4.x
implement all things related to `handleRevert:true` setting

Documentation:
https://web3js.readthedocs.io/en/v1.7.3/web3-eth.html?highlight=handlerevert#handlerevert
https://web3js.readthedocs.io/en/v1.7.3/web3-eth-contract.html?highlight=handlerevert#handlerevert

Implementation in 1.x
https://github.com/ChainSafe/web3.js/blob/2c0324af2da467ee1acb72452f20000e916e4306/packages/web3-core-method/src/index.js#L917

https://github.com/ChainSafe/web3.js/blob/2c0324af2da467ee1acb72452f20000e916e4306/packages/web3-core-method/src/index.js#L442

Also, add integration test in `web3-eth/test/integration/defaults.test.ts`

nikoulai:
## Description

#5069


packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1071,6 +1071,14 @@ export function sendTransaction<
 							ETH_DATA_FORMAT,
 						);
 
+						if (web3Context.handleRevert) {
+							// eslint-disable-next-line no-use-before-define
+							await getRevertReason(
+								web3Context,
+								transactionFormatted as TransactionCall,
+							);
```
nazarhussain: The return format should also pass to this function.

packages/web3/test/shared_fixtures/build/Basic.json
```diff
@@ -0,0 +1,253 @@
+{
```
nazarhussain: Would be preferred to create a symlink to the `build` folder or particular file.

nazarhussain (changes requested):

packages/web3/test/shared_fixtures/build/Basic.json
avkos: I think, parent folder `shared_fixtures` should be a symlink

avkos (changes requested):

packages/web3/test/shared_fixtures/build/Basic.json
nikoulai: we have `system_tests_utils.ts` in `shared_fixtures`, which is not in the original folder (@avkos )

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1071,6 +1071,15 @@ export function sendTransaction<
 							ETH_DATA_FORMAT,
 						);
 
+						if (web3Context.handleRevert) {
```
jdevcs, line 1074: LGTM, just mention in doc that `handleRevert` is available for `sendTransaction` only for now.

avkos (approved):

nazarhussain (approved):

Issue #5038, opened 2022-05-17
Accounts - 4.x System E2E Tests


mconnelly8:

mconnelly8:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/ChainSafe/web3.js/5038?planning-poker) with ZenHub @jdevcs @luu-alex @nazarhussain @nikoulai @avkos @Muhammad-Altabba

PR #5285, opened 2022-07-27
Accounts - 4.x System E2E Tests

avkos:
## Description
#5038 

## Type of change

## Checklist:


packages/web3-eth-accounts/test/integration/account.test.ts
```diff
@@ -0,0 +1,214 @@
...
+	signatureRecoverData,
+	transactionsTestData,
+	validDecryptData,
+	validEncryptData,
+	validHashMessageData,
+	validPrivateKeytoAccountData,
+	validPrivateKeyToAddressData,
+} from '../fixtures/account';
+
+describe('accounts', () => {
```
jdevcs, line 49: For accounts we discussed in backlog grooming call that these scenarios are covered in unit testing so we dnt need same scenarios in system tests.

avkos: I've decided to run it in browsers/electron to check how it works. and found problem with firefox 
https://github.com/cypress-io/cypress/issues/14600

avkos: and also found problem with cypress .each method and fix it.

scripts/system_tests_utils.ts
```diff
@@ -53,13 +53,22 @@ export const getEnvVar = (name: string): string | undefined =>
...
-export const isHttp = getSystemTestProvider().startsWith('http');
-export const isWs = getSystemTestProvider().startsWith('ws');
-export const isIpc = getSystemTestProvider().includes('ipc');
+export const getSystemTestEngine = (): string =>
+	getEnvVar('WEB3_SYSTEM_TEST_ENGINE') ?? DEFAULT_SYSTEM_ENGINE;
+
+export const isHttp: boolean = getSystemTestProvider().startsWith('http');
+export const isWs: boolean = getSystemTestProvider().startsWith('ws');
+export const isIpc: boolean = getSystemTestProvider().includes('ipc');
+export const isChrome: boolean = getSystemTestEngine() === 'chrome';
```
nazarhussain, line 62: Would be nice to call browser and node as `Client` or `Environment` rather `Engine`.

nazarhussain (changes requested):

jdevcs (commented):
Crypto.subtle is available in [firefox 34](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle#browser_compatibility) , there is error in tests because :  "This feature is available only in secure contexts (HTTPS)" and CI logs shows tests are running without SSL

```
web3-eth-accounts:      TypeError: crypto_1.crypto.web.subtle is undefined
350
web3-eth-accounts:   getBrowserKey@http://localhost:43915/__cypress/tests?p=test/integration/account.test.ts:54252:18
```

packages/web3-eth-accounts/test/integration/wallet.test.ts
```diff
@@ -0,0 +1,304 @@
...
+
+import { Web3AccountProvider } from 'web3-types';
+import { isBrowser, isElectron, itIf } from '../fixtures/system_test_utils';
+import { Wallet } from '../../src';
+import * as accountProvider from '../../src/account';
+import { Web3Account } from '../../dist';
+
+describe('Wallet', () => {
+	let wallet: Wallet;
+	// let clientUrl: string;
```
luu-alex: this can be removed?

avkos: thanks!

luu-alex (approved):

ghost (approved):

packages/web3-eth-accounts/test/integration/account.test.ts
```diff
@@ -0,0 +1,215 @@
...
+
+		describe('recover', () => {
+			it.each(signatureRecoverData)('%s', (data, testObj) => {
+				const address = recover(data, testObj.signatureOrV, testObj.prefixedOrR, testObj.s);
+				expect(address).toEqual(testObj.address);
+			});
+		});
+	});
+
+	describeIf(!isFirefox)('encrypt', () => {
```
Muhammad-Altabba: Why there are tests here that are limited to Firefox? Should not we also test the other behaviors for other browsers?

packages/web3-eth-accounts/src/wallet.ts
```diff
@@ -157,10 +157,12 @@ export class Wallet<
 		if (typeof account === 'string') {
 			return this.add(this._accountProvider.privateKeyToAccount(account));
 		}
-
-		const index = this.length;
+		let index = this.length;
+		const existAccount = this.get(account.address);
+		if (existAccount) {
+			index = this._addressMap.get(account.address.toLowerCase()) ?? index;
```
Muhammad-Altabba, line 164: Just wondering, if the account already exists, why not just return from the function without adding the address again at the same index?

packages/web3-eth-personal/test/integration/personal.test.ts
```diff
@@ -117,45 +111,43 @@ describe('personal integration tests', () => {
 		const account = await ethPersonal.newAccount('cde');
 		const updatedAccountList = await ethPersonal.getAccounts();
 		accountList.push(account);
-		expect(updatedAccountList).toStrictEqual(accountList);
+		expect(updatedAccountList.length).toBeGreaterThan(account.length);
```
Muhammad-Altabba, line 111: Is this a good suggestion?:
```suggestion
		const accountsList = await ethPersonal.getAccounts();
		accountsList.push(account);
		const updatedAccountsList = await ethPersonal.getAccounts();
		expect(updatedAccountsList.length).toStrictEqual(accountsList.length + 1);
```

packages/web3-eth/test/integration/rpc.block.test.ts
```diff
@@ -29,17 +29,17 @@ import {
...
 import { toAllVariants } from '../shared_fixtures/utils';
 import { sendFewTxes, validateBlock, validateTransaction } from './helper';
 
 describe('rpc with block', () => {
 	let web3Eth: Web3Eth;
-	let accounts: string[] = [];
 	let clientUrl: string;
 
-	let contract: Contract<typeof BasicAbi>;
+	let contractInstance: Contract<typeof BasicAbi>;
```
Muhammad-Altabba: Why to rename `contract` here and in other places? It is by convention to use small cases for the instance of any class. 
I mean: Usually, we use PascalCase for class definition and camelCase for instance. Why to use the word `instance` specifically for this variable?

packages/web3-eth/test/integration/subscribe.test.ts
```diff
@@ -23,26 +23,20 @@ import {
...
+	createTempAccount,
 } from '../fixtures/system_test_utils';
 
 describeIf(isWs)('subscribe', () => {
 	let web3Eth: Web3Eth;
 	let provider: WebSocketProvider;
-	let accounts: string[];
 
 	beforeAll(async () => {
-		accounts = await getSystemTestAccounts();
```
Muhammad-Altabba, line 40: I see, throughout this MR, that `getSystemTestAccounts()` was replaced by  `createTempAccount()` this would lead the tests to take a very long time compared to the old approach. Is there a strong reason for that?
What do you think?

packages/web3-eth/test/integration/subscription_logs.test.ts
```diff
@@ -51,43 +51,42 @@ const makeFewTxToContract = async ({
...
 describeIf(isWs)('subscription', () => {
 	let clientUrl: string;
-	let accounts: string[] = [];
 	let web3Eth: Web3Eth;
 	let providerWs: WebSocketProvider;
+	let contractInstance: Contract<typeof BasicAbi>;
 	let contract: Contract<typeof BasicAbi>;
 	let deployOptions: Record<string, unknown>;
 	let sendOptions: Record<string, unknown>;
 	let from: string;
```
Muhammad-Altabba, line 60: since this is only assigned inside on test, I suggest declaring it there.
```suggestion
```

packages/web3-eth/test/integration/subscription_logs.test.ts
```diff
@@ -51,43 +51,42 @@ const makeFewTxToContract = async ({
...
 	});
 	afterAll(() => {
 		providerWs.disconnect();
+		(contractInstance.provider as WebSocketProvider).disconnect();
 	});
 
 	describe('logs', () => {
 		it(`wait for ${checkEventCount} logs`, async () => {
 			web3Eth = new Web3Eth(providerWs as Web3BaseProvider);
+			from = tempAcc.address;
```
Muhammad-Altabba, line 82: Declaring this here instead of above. Because it is assigned here.
```suggestion
			const from = tempAcc.address;
```

packages/web3-eth/test/integration/subscription_new_pending_tx.test.ts
```diff
@@ -35,49 +34,82 @@ describeIf(isWs)('subscription', () => {
...
-				});
-			});
-
-			const receipts = await sendFewTxes({
-				web3Eth,
-				from,
-				to,
-				value,
-				times: checkTxCount,
+				})().catch(console.error);
```
Muhammad-Altabba, line 106: It is dangerous to just log the error without firing it. If there was an error at the start of his promise, then both `txHashes` and `receipts` would have an equal length or they can both be just empty. And this means that the next `for` can falsely not detect any possible issue.

packages/web3-eth/test/integration/watch_transaction.test.ts
```diff
@@ -85,7 +80,7 @@ describeIf(isWs)('watch subscription transaction', () => {
 				// Tx promise is handled separately
 				// eslint-disable-next-line no-void
 				void sentTx.on('confirmation', ({ confirmations }) => {
-					expect(Number(confirmations)).toBe(shouldBe);
+					expect(Number(confirmations)).toBeGreaterThanOrEqual(shouldBe);
```
Muhammad-Altabba, line 83: Why it should be here greater than or equal? Was not the logic here is to check if the confirmations arrive in order? (1 then 2 then 3...) If we will check just the greater-than-or-equal, we do not need other than this: right?
```
if (confirmations >= waitConfirmations) {
    resolve();
}
```

(This applies also to the changes in the next file.)

packages/web3-eth/test/integration/web3_eth/getFeeHistory.test.ts
```diff
@@ -31,12 +31,12 @@ describeIf(getSystemTestBackend().includes('geth'))('Web3Eth.getFeeHistory', ()
 	let web3Eth: Web3Eth;
 	let systemProvider: string;
 
-	beforeAll(async () => {
+	beforeEach(async () => {
```
Muhammad-Altabba: Why change to `beforeEach`, and `afterEach`?

packages/web3-eth/test/integration/web3_eth/send_signed_transaction.test.ts
```diff
@@ -16,45 +16,42 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
 const HEX_NUMBER_DATA_FORMAT = { bytes: FMT_BYTES.HEX, number: FMT_NUMBER.HEX } as const;
 
 describe('Web3Eth.sendSignedTransaction', () => {
 	let web3Eth: Web3Eth;
-	let accounts: Address[];
+	let tempAcc: { address: string; privateKey: string };
 
-	beforeAll(async () => {
+	beforeEach(async () => {
 		web3Eth = new Web3Eth(getSystemTestProvider());
```
Muhammad-Altabba, line 37: It would be unnecessary to initialize `web3Eth` before each test.

I think those `beforeEach` have to be reviewed for such cases.

scripts/ganache.sh
```diff
@@ -21,12 +21,13 @@ start() {
...
 		npx wait-port "$WEB3_SYSTEM_TEST_PORT"
 		echo "Ganache started..."
-	fi 
+	fi
 }
 
 stop() {
 	echo "Stopping ganache ..."
-	# pkill -15 ganache 2>/dev/null
+	processID=`lsof -Fp -i:${WEB3_SYSTEM_TEST_PORT}| grep '^p'`
```
Muhammad-Altabba:
```suggestion
    echo "Stopping ganache ..."
    processID=`lsof -Fp -i:${WEB3_SYSTEM_TEST_PORT}| grep '^p'`
```

scripts/test-runner.sh
```diff
@@ -48,4 +50,4 @@ else
 fi
 
 
-yarn "$BACKEND:start:background" && yarn $TEST_COMMAND && yarn "$BACKEND:stop"
+yarn "$BACKEND:start:background" && yarn generate:accounts && yarn $TEST_COMMAND && yarn "$BACKEND:stop"
```
Muhammad-Altabba, line 54: Just wondering why do we need to `generate:accounts` at the start of the backend?
Thanks,

Muhammad-Altabba (changes requested):
Great job!
However, I have some questions that I need your help answering them...
Thanks,

nazarhussain (approved):
Apart from the existing feedback, LGTM.

scripts/test-runner.sh
avkos: We refill accounts from one main account in tests. So when we run in parallel, we've received the nonce error. That's why we need to add a list of accounts with refilled balances before running tests. And then we can use it one by one in parallel tests

packages/web3-eth/test/integration/watch_transaction.test.ts
avkos: When I run tests in parallel, it mined a lot of transactions in other tests at the same time. so I receive a lot of confirmations at the  same time

packages/web3-eth/test/integration/rpc.block.test.ts
avkos: `contractInstance = new Contract(...)`
so we received an instance of the contract. that's why I called this variable `contractInstance`

packages/web3-eth-personal/test/integration/personal.test.ts
avkos: Also, when I run parallel tests, a few tests add accounts. And it will always be more than  +1.

packages/web3-eth-accounts/src/wallet.ts
avkos: because we need to override whole account (not just an address) if we set it again

avkos, line 164: because we need to override whole account (not just an address) if we set it again

packages/web3-eth-accounts/test/integration/account.test.ts
avkos: fixed thx.

packages/web3-eth/test/integration/web3_eth/send_signed_transaction.test.ts
avkos: fixed

packages/web3-eth/test/integration/web3_eth/getFeeHistory.test.ts
avkos: fixed

packages/web3-eth-accounts/src/wallet.ts
Muhammad-Altabba: Thanks for the clarification.

Muhammad-Altabba: Sorry, the comment was doublicated so I deleted it and here I am resolving it.

packages/web3-eth-accounts/src/wallet.ts
```diff
@@ -157,10 +157,12 @@ export class Wallet<
...
+		}
 		this._addressMap.set(account.address.toLowerCase(), index);
-
 		this[index] = account;
```
Muhammad-Altabba: I think there is here a chance to have a doublication if the passed account is of type `string`.
And here is a suggested draft to overcome that:

```suggestion
		let accountToAdd =
			typeof account === 'string'
				? this._accountProvider.privateKeyToAccount(account)
				: account;
		let index = this.length;
		const existAccount = this.get(accountToAdd.address);
		if (existAccount) {
			index = this._addressMap.get(accountToAdd.address.toLowerCase()) ?? index;
		}
		this._addressMap.set(accountToAdd.address.toLowerCase(), index);
		this[index] = accountToAdd;
```

packages/web3-eth/test/integration/web3_eth/getFeeHistory.test.ts
Muhammad-Altabba: Thanks,
However, the code did not change. Did you miss pushing the code changes?

packages/web3-eth/test/integration/web3_eth/send_signed_transaction.test.ts
Muhammad-Altabba: Thanks,
However, the code did not change. Did you miss pushing the code changes?
(this comment also apply to all code modifications regarding `beforeEach` and `afterEach` in all modified test cases.)

packages/web3-eth/test/integration/rpc.block.test.ts
Muhammad-Altabba: If this is the case, let us keep with the convention of using camleCase for class instances.
Actually, the camleCase indicate an instance. Or else, we need to rename all varaiables like `web3Eth`, `tempAcc`... in this file and in all other files, because they are also all "instances".
Thanks,

packages/web3-eth/test/integration/subscription_logs.test.ts
Muhammad-Altabba: Actually, this variable was used inside both `beforeAll` and in the test case. Now, with the new code modifications, it is used inside only one code block.

packages/web3-eth-accounts/test/integration/account.test.ts
Muhammad-Altabba: Sorry, I could not see where is it fixed. Did you miss to push the code change?
Thanks,

avkos: pushed

packages/web3-eth/test/integration/web3_eth/getFeeHistory.test.ts
avkos: pushed

packages/web3-eth/test/integration/web3_eth/send_signed_transaction.test.ts
avkos: pushed

packages/web3-eth-accounts/src/wallet.ts
avkos: This is not my code. But this code uses recursion if the type is a string. So, your suggestion also can be placed. Logic is the same.

Muhammad-Altabba: Yes, it is not part of the scope of this MR. However, I suggested it to be part of your suggested enhancements as you checked for duplication if the account is an object, but did not when the account is a `string`. So, my suggestion was basically to expand the new `if` that you added. :smile: 
And you are right, it is not necessarily needed to be handled in this MR. So, I created an issue for this bug so handle it later: https://github.com/ChainSafe/web3.js/issues/5388 :smile:

packages/web3-eth/test/integration/rpc.block.test.ts
avkos: renamed. check plz

.github/workflows/build.yml
```diff
@@ -14,14 +14,14 @@ jobs:
   build:
     strategy:
       matrix:
-        node: [14, 16]
```
Spacesai1or: I'm guessing the changes to this file are to cut down the number of actions while you're debugging the failing tests?

packages/web3-eth-accounts/test/integration/account.test.ts
```diff
@@ -0,0 +1,214 @@
...
+			const account = create();
+			const txObj = { ...txData, from: account.address };
+			const signedResult = await signTransaction(
+				TransactionFactory.fromTxData(txObj),
+				account.privateKey,
+			);
+			expect(signedResult).toBeDefined();
+
+			const address: Address = recoverTransaction(signedResult.rawTransaction);
+			expect(address).toBeDefined();
```
Spacesai1or:
```suggestion
```
Don't need this assertion since we're checking `expect(address).toEqual(account.address);`

templates/cypress/plugins/index.js
```diff
@@ -27,6 +27,7 @@ module.exports = (on, config) => {
 	config.env.WEB3_SYSTEM_TEST_MNEMONIC = process.env.WEB3_SYSTEM_TEST_MNEMONIC;
 	config.env.WEB3_SYSTEM_TEST_PORT = process.env.WEB3_SYSTEM_TEST_PORT;
 	config.env.WEB3_SYSTEM_TEST_PROVIDER = process.env.WEB3_SYSTEM_TEST_PROVIDER;
+	config.env.WEB3_SYSTEM_TEST_CLIENT = process.env.WEB3_SYSTEM_TEST_CLIENT;
```
Spacesai1or: Client is a term that generally refers to the node software being run e.g. Ganache, Geth, etc. I would rename this to `WEB3_SYSTEM_TEST_BROWSER` for clarity

templates/cypress.config.js
```diff
@@ -10,3 +10,24 @@ module.exports = {
 		specPattern: 'test/integration/**/**/*.test.ts',
 	},
 };
+
+if (process.env.WEB3_SYSTEM_TEST_CLIENT === 'firefox') {
+	const port = parseInt(String(Math.random() * 10000 + 10000));
```
Spacesai1or: Is there a reason we can't just hardcode a port?

scripts/ganache.sh
```diff
@@ -9,23 +9,28 @@ helpFunction() {
...
 	if [ -z "${ORIGARGS[1]}" ]
 	then
 		npx ganache ethereum -m "$WEB3_SYSTEM_TEST_MNEMONIC" -a 5 -p $WEB3_SYSTEM_TEST_PORT --wallet.passphrase "123"
 	else
 		echo "Starting ganache ..."
-		docker run --detach --publish 8545:8545 trufflesuite/ganache:latest -m "$WEB3_SYSTEM_TEST_MNEMONIC" -a 5 -p $WEB3_SYSTEM_TEST_PORT --wallet.passphrase "123"
+		echo "npx ganache ethereum -m \"$WEB3_SYSTEM_TEST_MNEMONIC\" -a 5 -p $WEB3_SYSTEM_TEST_PORT &"
+        npx ganache ethereum -m "$WEB3_SYSTEM_TEST_MNEMONIC" -a 5 -p $WEB3_SYSTEM_TEST_PORT >/dev/null --wallet.passphrase "123" &
+
+#		docker run --detach --publish 8545:8545 trufflesuite/ganache:latest -m "$WEB3_SYSTEM_TEST_MNEMONIC" -a 5 -p $WEB3_SYSTEM_TEST_PORT --wallet.passphrase "123"
```
Spacesai1or: Once you have the tests passing, we should revert this change and see if it works with docker. I believe making this change back to `npx` will cause the black box tests to fail

packages/web3/package.json
```diff
@@ -33,7 +33,7 @@
 		"test:ci": "jest --coverage=true --coverage-reporters=json --verbose",
 		"test:watch": "npm test -- --watch",
 		"test:unit": "jest --config=./test/unit/jest.config.js",
-		"test:integration": "jest --config=./test/integration/jest.config.js --runInBand --forceExit",
+		"test:integration": "jest --config=./test/integration/jest.config.js --forceExit --runInBand",
```
Spacesai1or:
```suggestion
		"test:integration": "jest --config=./test/integration/jest.config.js --forceExit",
```

Spacesai1or (changes requested):

.github/workflows/build.yml
avkos: yes I am debugging now. and run tests a lot of times. I do not need run all node versions/build doc/unit tests etc.. I need only IPC test and e2e browsers tests

avkos: reverted

packages/web3-eth-accounts/test/integration/account.test.ts
avkos: removed

templates/cypress/plugins/index.js
avkos: but by default, it is `node`
export const DEFAULT_SYSTEM_CLIENT = 'node';
node - is not a browser

scripts/ganache.sh
avkos: reverted

packages/web3/package.json
avkos: done

templates/cypress.config.js
avkos: Yes, it was a reason. I really do not remember why I need a random port. Looks like we need it for a parallel run

scripts/system_tests_utils.ts
```diff
@@ -52,166 +47,29 @@ export const getEnvVar = (name: string): string | undefined =>
 	// eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
 	global.Cypress ? Cypress.env(name) : process.env[name];
 
-export const DEFAULT_SYSTEM_PROVIDER = 'http://localhost:8545';
+export const DEFAULT_SYSTEM_PROVIDER = 'http://127.0.0.1:8545';
```
luu-alex, line 50: was localhost an issue?

scripts/system_tests_utils.ts
```diff
@@ -318,3 +176,110 @@ export const createAccountProvider = (context: Web3Context<EthExecutionAPI>) =>
...
+			to: acc.address,
+			value: '100000000000000000',
+		});
+	}
+
+	return { address: acc.address.toLowerCase(), privateKey: acc.privateKey };
+};
+let tempAccountList: { address: string; privateKey: string }[] = [];
+// let currentIndex = 0;
+// const workersCount = 1;
```
luu-alex: looks like debug code?

scripts/gen_accounts.js
```diff
@@ -0,0 +1,52 @@
...
+	});
+};
+
+const createWallets = () =>
+	new Promise(async resolve => {
+		for (const acc of tempAccountList) {
+			try {
+				await addAccount(acc.address, acc.privateKey);
+			} catch (e) {
+				console.log('error', e);
```
luu-alex, line 46:
```suggestion
				throw e
```

scripts/system_tests_utils.ts
avkos: yes. because I map 127.0.0.1 to SSL certificate

scripts/gen_accounts.js
avkos: no. it is a helper we do not need to throw. it will exit the script in this case

scripts/system_tests_utils.ts
avkos: removed. thx

luu-alex (approved):

nikoulai (approved):

templates/cypress/plugins/index.js
Spacesai1or: Then `WEB_SYSTEM_TEST_ENGINE` could work

scripts/ganache.sh
Spacesai1or: We should probably remove the commented code if we aren't going to use it. Additionally, we should consider replacing the use of `npx` when ganache is not started in `background` mode. Lastly, the `stop` command will only work for Docker, so we need to either remove the use of `npx` or update `stop` to know whether to kill the ganache process or stop the Docker container

templates/cypress/plugins/index.js
jdevcs: we should refer `client` for eth node, in our repo.

packages/web3-eth-accounts/src/wallet.ts
jdevcs, line 164: plus add a console.warn() that `account.address` already exists.

jdevcs (changes requested):
LGTM, Good work. Just need some minor changes.

scripts/gen_accounts.js
```diff
@@ -0,0 +1,52 @@
...
+
+	const web3Personal = new Personal(clientUrl);
+	if (accountList.length === 0) {
+		accountList = await web3Personal.getAccounts();
+		mainAcc = accountList[0];
+	}
+	const web3Eth = new Web3Eth(clientUrl);
+
+	if (!accountList.find(acc => acc.address === address)) {
+		await web3Personal.importRawKey(
```
jdevcs, line 26: For future test cases we will need to use and test local wallet along with most of tests, instead of node wallet. So some future tests scenarios will use local signing functionality instead of loading PK into node wallet, unlocking and sending unsigned transactions. 
This PR is waiting from long time so we can have above mentioned functionality in some next sprint.

avkos: yes. I think, we have a separate task for this feature

scripts/ganache.sh
avkos: fixed. thanks!

templates/cypress/plugins/index.js
avkos: I've set env as WEB3_SYSTEM_TEST_ENGINE. thx

jdevcs (approved):

Spacesai1or (approved):

packages/web3-eth-contract/test/integration/contract_erc20.test.ts
```diff
@@ -32,61 +32,66 @@ describe('contract', () => {
...
+			await expect(
+				contract.deploy(deployOptions).send(sendOptionsLocal),
+			).resolves.toBeDefined();
 		});
 
 		describe('contract instance', () => {
-			beforeAll(async () => {
-				contract = await contract.deploy(deployOptions).send(sendOptions);
+			let contractDeployed: Contract<typeof ERC20TokenAbi>;
+			beforeEach(async () => {
```
Muhammad-Altabba, line 57: Sorry, do we really need here to have those repeated before each?

packages/web3/test/integration/web3.accounts.test.ts
```diff
@@ -18,23 +18,23 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
 const hexRegx = /0[xX][0-9a-fA-F]+/;
 
 describe('web3.accounts', () => {
 	let clientUrl: string;
-	let nodeAccounts: string[];
+	let tempAccount: string;
 	let web3: Web3;
 
-	beforeAll(async () => {
+	beforeEach(async () => {
```
Muhammad-Altabba, line 35: Do all the lines need to be executed `beforeEach`? I mean, It seems the following needed to be executed only once:
```
 		clientUrl = getSystemTestProvider();
		web3 = new Web3(clientUrl);
		await waitForOpenConnection(web3);
```
Right?

templates/cypress.config.js
```diff
@@ -10,3 +10,24 @@ module.exports = {
...
+// 				},
+// 			],
+// 		},
+// 	];
+// 	config.e2e.port = port;
+// 	config.e2e.hosts = {
+// 		'web3.js': '127.0.0.1',
+// 	};
+// 	config.e2e.baseUrl = `https://web3.js:${port}`;
+// }
```
Muhammad-Altabba, line 32: Could you please add a comment that possibly refers to an issue that describes why this code is commented. And when this code is supposed to be uncommented.
Thanks,

packages/web3-eth-contract/test/integration/contract_defaults.test.ts
```diff
@@ -18,61 +18,67 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
-					.deploy(deployOptions)
-					.send({ gas: '1000000' })
-					.on('receipt', receiptHandler);
-
-				// We didn't specify "from" in this call
-				expect(receiptHandler).toHaveBeenCalledWith(
-					expect.objectContaining({ from: accounts[0] }),
-				);
-			});
+			itIf(!isIpc)(
```
Muhammad-Altabba, line 50: Sorry, why this test should not apply to `IPC`? And why it was working when it was applied to all including IPC?

And if there a good reason for that, could we please, have a small comment hereon  why this test should not apply to `IPC`?

Muhammad-Altabba (approved):
Looks great!
However, there are some minor points. Which I hope you modify them in this MR or create follow-up tasks for them.

packages/web3-eth-contract/test/integration/contract_defaults.test.ts
avkos: done

packages/web3-eth-contract/test/integration/contract_erc20.test.ts
avkos: fixed

packages/web3-eth/test/integration/subscription_logs.test.ts
avkos: fixed

packages/web3-eth/test/integration/subscription_logs.test.ts
avkos: done

packages/web3/test/integration/web3.accounts.test.ts
avkos: fixed

templates/cypress.config.js
avkos: done

packages/web3/test/integration/web3.accounts.test.ts
avkos: done

packages/web3-eth-contract/test/integration/contract_erc20.test.ts
avkos: done

packages/web3-eth-contract/test/integration/contract_defaults.test.ts
avkos: fixed

PR #5301, opened 2022-08-01
Adding a wrapper for keccak256 to allow hashing  `string` and `bigint`

Muhammad-Altabba:
Fixes https://github.com/ChainSafe/web3.js/issues/5287 by adding a wrapper for keccak256 to allow hashing  `string` and `bigint`. 
This MR also includes test cases.

## Type of change

## Checklist:


packages/web3-utils/src/hash.ts
```diff
@@ -57,7 +57,25 @@ export const sha3Raw = (data: Bytes): string => {
 	return hash;
 };
 
-export { keccak256 };
+/**
+ * A wrapper for ethereum-cryptography/keccak256 to allow hashing a `string` and a `bigint` in addition to `UInt8Array`
+ */
+export const keccak256Wrapper = (
+	data: Uint8Array | ReadonlyArray<number> | string | bigint,
```
nazarhussain: Prefer to use the union types we have. 


```suggestion
	data: Bytes | Numbers | string,
```

nazarhussain (changes requested):

jdevcs (approved):
apart from Nazar's comments LGTM.

packages/web3-utils/src/hash.ts
Muhammad-Altabba: Sorry, this is incompatible with keccak256. In web3.js v1.x, It throws an exception if we pass a number that is provided from type `Numbers` or if we pass `ArrayBuffer` that is provided from type `Bytes`. 
Additionally, the `readonly numbers[]` is supported as I am using `Buffer.from(...)` and so I included it.

luu-alex (approved):

packages/web3-utils/src/hash.ts
Muhammad-Altabba: As per our discussion, it is nice to support Bytes and Numbers, similar to other utility functions. 
So, I modified the logic to include more types as desired. Kindly re-check now.

nazarhussain (approved):

PR #5305, opened 2022-08-01
Web3RequestManager is not compatible with the Ganache provider

### LINKED ISSUE #5288: Web3RequestManager is not compatible with the Ganache provider

### Current Behavior

There is method `createInterfaceAdapter` in `truffle` which proxies the `web3` adapter to different adapter. 

https://github.com/trufflesuite/truffle/blob/c5de9c238aa07f7a9bed701fa83bab3457011f3b/packages/interface-adapter/lib/adapter/index.ts#L14

Here is a sample code which is creating `proxy` object for the `Ganache` adapter. 

https://github.com/trufflesuite/truffle/blob/c5de9c238aa07f7a9bed701fa83bab3457011f3b/packages/interface-adapter/test/quorum-getBlock.test.ts#L16-L28

Our request manager could not identify the right adapter type for the adapter and hang the request forever. 

### Expected Behavior

Ganache should work fine with the web3 request manager. 

### Steps to Reproduce

Mentioned above. 

### Web3.js Version

4.x

### Environment

- Operating System: N/A
- Browser: N/A
- Node.js Version: 16.x
- NPM Version: N/A

### Anything Else?

The below logic assumes that the provider which have `request` method should accept a callback. But it seems that ganache provider does not accept a callback rather return a promise. 

https://github.com/ChainSafe/web3.js/blob/0b903378be4cee0f2a9dc2c0572ffcbce5e97afd/packages/web3-core/src/web3_request_manager.ts#L176-L178

nazarhussain:
## Description

Fixes #5288

## Type of change

## Checklist:


avkos (approved):

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -219,42 +272,80 @@ export class Web3RequestManager<
...
-			if (response.id && payload.id !== response.id) {
-				throw new InvalidResponseError<ErrorType>(response);
+			const res = {
+				jsonrpc: '2.0',
+				// eslint-disable-next-line no-nested-ternary
+				id: jsonRpc.isBatchRequest(payload)
+					? payload[0].id
+					: 'id' in payload
+					? payload.id
+					: 'NON_EXISTING_ID',
```
jdevcs: If there was an error in detecting the id in the Request object, it MUST be Null. instead of NON_EXISTING_ID 
 https://www.jsonrpc.org/specification

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -219,42 +272,80 @@ export class Web3RequestManager<
 	private _processJsonRpcResponse<ResultType, ErrorType, RequestType>(
 		payload: JsonRpcPayload<RequestType>,
 		response: JsonRpcResponse<ResultType, ErrorType>,
+		{ legacy, error }: { legacy: boolean; error: boolean },
```
jdevcs, line 275: can you create a type for `{ legacy: boolean; error: boolean }` and use that in this PR. It will be more clean.

jdevcs (changes requested):

packages/web3-core/src/web3_request_manager.ts
nazarhussain: https://www.jsonrpc.org/specification#id1
> The use of Null as a value for the id member in a Request object is discouraged, because this specification uses a value of Null for Responses with an unknown id. Also, because JSON-RPC 1.0 uses an id value of Null for Notifications this could cause confusion in handling.


This case is only for the legacy implementations which don't follow fully compliance JSON-RPC specification. If we use `null` the upper level code will fail as it checks the valid response must contains some id.

packages/web3-core/src/web3_request_manager.ts
nazarhussain: This type is created and used only here. So seems fine to just keep inline.

packages/web3-core/src/web3_request_manager.ts
jdevcs: [link you mentioned above ](https://www.jsonrpc.org/specification#id1) is for request object. and object under discussion is response. ( last line ) 

image

If upper level code in our lib always expects some id thn it should always put id in request obj for getting it in response.

nazarhussain:
@jdevcs Please review.

jdevcs (approved):

packages/web3-core/test/unit/web3_request_manager.test.ts
```diff
@@ -641,12 +739,68 @@ describe('Web3RequestManager', () => {
 			});
 		});
 
+		describe('eip15593-provider', () => {
+			beforeEach(() => {
+				jest.spyOn(utils, 'isWeb3Provider').mockReturnValue(false);
+				jest.spyOn(utils, 'isLegacyRequestProvider').mockReturnValue(false);
+				jest.spyOn(utils, 'isLegacySendProvider').mockReturnValue(false);
+				jest.spyOn(utils, 'isLegacySendAsyncProvider').mockReturnValue(false);
```
Muhammad-Altabba: Why sping on all those providers while the test case is just for `eip15593`?

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -169,28 +171,66 @@ export class Web3RequestManager<
 				response = error as JsonRpcResponse<ResponseType>;
 			}
 
-			return this._processJsonRpcResponse(payload, response);
+			return this._processJsonRpcResponse(payload, response, { legacy: false, error: false });
+		}
+
+		if (isEIP1193Provider(provider)) {
```
Muhammad-Altabba, line 177: I think we need later to consider refactoring to put the difference between every provider inside the provider class. So, we would not have to handle the differences in if statements.

Muhammad-Altabba (approved):
Looks good. I just have one discussion point and one question.

packages/web3-core/test/unit/web3_request_manager.test.ts
Spacesai1or:
```suggestion
```
I removed these locally and the tests still pass

packages/web3-core/test/unit/web3_request_manager.test.ts
```diff
@@ -379,6 +380,103 @@ describe('Web3RequestManager', () => {
...
```
Spacesai1or, line 388:
```suggestion
```
I removed these locally and the tests still pass

packages/web3-core/test/unit/web3_request_manager.test.ts
```diff
@@ -641,12 +739,68 @@ describe('Web3RequestManager', () => {
 			});
 		});
 
+		describe('eip15593-provider', () => {
```
Spacesai1or:
```suggestion
		describe('eip1193-provider', () => {
```

packages/web3-core/test/unit/web3_request_manager.test.ts
```diff
@@ -379,6 +380,103 @@ describe('Web3RequestManager', () => {
...
```
Spacesai1or:
```suggestion
		describe('eip1193-provider', () => {
```

packages/web3-core/test/unit/web3_request_manager.test.ts
```diff
@@ -379,6 +380,103 @@ describe('Web3RequestManager', () => {
...
+
+				await expect(manager.send(request)).rejects.toThrow(
+					new InvalidResponseError(errorResponse),
+				);
+				expect(myProvider.request).toHaveBeenCalledTimes(1);
+				expect(myProvider.request).toHaveBeenCalledWith(payload);
+			});
+		});
+
+		describe('eip15593-provider - return non json-rpc compliance response', () => {
```
Spacesai1or:
```suggestion
		describe('eip1193-provider - return non json-rpc compliance response', () => {
```

packages/web3-core/test/unit/web3_request_manager.test.ts
```diff
@@ -379,6 +380,103 @@ describe('Web3RequestManager', () => {
...
+			beforeEach(() => {
+				jest.spyOn(utils, 'isWeb3Provider').mockReturnValue(false);
+				jest.spyOn(utils, 'isLegacyRequestProvider').mockReturnValue(false);
+				jest.spyOn(utils, 'isLegacySendProvider').mockReturnValue(false);
+				jest.spyOn(utils, 'isLegacySendAsyncProvider').mockReturnValue(false);
```
Spacesai1or, line 445:
```suggestion
```
I removed these locally and the tests still pass

packages/web3-core/test/unit/web3_request_manager.test.ts
```diff
@@ -379,6 +380,103 @@ describe('Web3RequestManager', () => {
...
+		describe('eip15593-provider', () => {
+			beforeEach(() => {
+				jest.spyOn(utils, 'isWeb3Provider').mockReturnValue(false);
+				jest.spyOn(utils, 'isLegacyRequestProvider').mockReturnValue(false);
+				jest.spyOn(utils, 'isLegacySendProvider').mockReturnValue(false);
+				jest.spyOn(utils, 'isLegacySendAsyncProvider').mockReturnValue(false);
+				jest.spyOn(utils, 'isEIP1193Provider').mockReturnValue(true);
+			});
+
+			it('should pass request to provider and resolve if provider resolves it', async () => {
```
jdevcs, line 392: We should also have some [system tests](https://github.com/ChainSafe/web3.js/issues/5309), and we can plan that for release after alpha.

packages/web3-core/test/unit/web3_request_manager.test.ts
nazarhussain: To make sure we the only code branch regarding `eip15593` is executed.

packages/web3-core/src/web3_request_manager.ts
nazarhussain: Yes a lot can be refactored but most of that will be deprecated soon anyway.

nazarhussain:
@spacesailor24 I resolved all your feedback. Merging this PR.

PR #5312, opened 2022-08-04
Update `signTransaction` to Consistently Return `SignedTransactionInfoAPI`

### LINKED ISSUE #5077: Geth returns non `HexStringBytes` value for `eth_signTransaction`
Currently [eth_signTransaction](https://github.com/ChainSafe/web3.js/blob/82574a2587810f9bea26ba65da48fe667162e573/packages/web3-common/src/eth_execution_api.ts#L284) is typed to return `HexStringBytes`, but when working on #5056, I discovered Geth returns [an object](https://github.com/ethereum/go-ethereum/blob/c1b69bd12184cc6278f5641c14bc72ca0a42d397/internal/ethapi/api.go#L1735)

[More context](https://github.com/ChainSafe/web3.js/pull/5056/files#r885098419)

To solve the issue mentioned in `More context`, I believe we'd need to implement a util method similar to [ethers.js' parse](https://github.com/ethers-io/ethers.js/blob/master/packages/transactions/src.ts/index.ts#L478)

Spacesai1or:
Before this PR, `web3Eth.signTransaction` was returning whatever the user passed as the `transaction` parameter for `respnose.tx` when the RPC client was returning only the RLP encoded transaction as the RPC response (in the below code example, Ganache was only returning the value labeled as `raw`, while Geth was returning the entire object (i.e. `{ raw: ..., tx: ... }`)). This resulted in varying returns values based on which client the request was submitted to:

If the connected client was Ganache:

```tsc
{
    raw: '0xf86580843b9aca018252089400000000000000000000000000000000000000000180820a95a02bd3ccac498541156149591091ff2af043f0be204a7d723701a9e1366dcc6f5fa02a138f18dcafebcf48d9df2c7b3d7fa5f0741b2921d6f42fdc219f34a7ddd97e',
    tx: {
      from: '0x6e599da0bff7a6598ac1224e4985430bf16458a4',
      nonce: 0n,
      value: 1n,
      gas: 21000n,
      gasPrice: 1000000001n
    }
}
```

If the connected client was Geth:

```tsc
{
   raw: '0xf86580843b9aca018252089400000000000000000000000000000000000000000180820a95a003d7879059608586185d1f1b662a73f2701026403b44ab639b5558b043ca375ca0024054e057f5cc1e1f3d2c7aabf162616caf95bbcfb8215376210ff410d91f32',
   tx: {
      type: 0n,
      nonce: 0n,
      gasPrice: 1000000001n,
      gas: 21000n,
      value: 1n,
      v: 2709n,
      r: '0x03d7879059608586185d1f1b662a73f2701026403b44ab639b5558b043ca375c',
      s: '0x024054e057f5cc1e1f3d2c7aabf162616caf95bbcfb8215376210ff410d91f32',
      data: '0x'
    }
}
```

After this PR:

If the connected client was Ganache:

```tsc
{
    raw: '0xf86580843b9aca018252089400000000000000000000000000000000000000000180820a95a02bd3ccac498541156149591091ff2af043f0be204a7d723701a9e1366dcc6f5fa02a138f18dcafebcf48d9df2c7b3d7fa5f0741b2921d6f42fdc219f34a7ddd97e',
    tx: {
      nonce: 0n,
      gasPrice: 1000000001n,
      gasLimit: 21000n,
      value: 1n,
      data: '0x',
      v: 2709n,
      r: '0x2bd3ccac498541156149591091ff2af043f0be204a7d723701a9e1366dcc6f5f',
      s: '0x2a138f18dcafebcf48d9df2c7b3d7fa5f0741b2921d6f42fdc219f34a7ddd97e',
      type: 0n
    }
}
```

If the connected client was Geth:

```tsc
{
    raw: '0xf86580843b9aca018252089400000000000000000000000000000000000000000180820a95a0e7e229dcf3b6930ec7a090164f3bafe42fe4b343e2ac89f4e7077992e4cae7a9a0368c25b3275b8f8d8c9ea7b4fd7715b474843f4a6f36c9088d1c9e3f0d61dc27',
    tx: {
      type: 0n,
      nonce: 0n,
      gasPrice: 1000000001n,
      gas: 21000n,
      value: 1n,
      v: 2709n,
      r: '0xe7e229dcf3b6930ec7a090164f3bafe42fe4b343e2ac89f4e7077992e4cae7a9',
      s: '0x368c25b3275b8f8d8c9ea7b4fd7715b474843f4a6f36c9088d1c9e3f0d61dc27',
      data: '0x'
    }
}
```

---

One point of discussion that remains: In the last two above code examples, the response from Ganache returns the `gasLimit` property because the `raw` encoded transaction is being parsed by `@ethereumjs/tx`'s [TransactionFactory](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/docs/classes/TransactionFactory.md). However, the response from Geth returns `gasLimit` as `gas` which is what's being done in `1.x`. The decision to be made it whether we will continue to use `gas` to describe `gasLimit` in `4.x`

- Both the old and new ETH RPC spec use `gas` for `eth_sendTransaction`
- Ethers.js uses `gasLimit`

closes #5077


avkos (approved):

nazarhussain (approved):

packages/web3-eth/test/unit/rpc_method_wrappers/fixtures/sign_transaction.ts
```diff
@@ -17,27 +17,20 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 import { Transaction } from 'web3-types';
 
 const transaction: Transaction = {
-	from: '0xEB014f8c8B418Db6b45774c326A0E64C78914dC0',
-	to: '0x3535353535353535353535353535353535353535',
-	value: '0x174876e800',
-	gas: '0x5208',
-	gasPrice: '0x4a817c800',
 	type: '0x0',
```
jdevcs, line 27: We should also add tests for 0x1 and 0x2 transactions

jdevcs (changes requested):

packages/web3-eth/src/utils/decode_signed_transaction.ts
```diff
@@ -0,0 +1,47 @@
...
+import { formatTransaction } from './format_transaction';
+
+/**
+ * Decodes an [RLP](https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/#top) encoded transaction.
+ *
+ * @param encodedSignedTransaction The RLP encoded transaction.
+ * @param returnFormat ({@link DataFormat} Specifies how the return data should be formatted.
+ * @returns {@link SignedTransactionInfoAPI}, an object containing the RLP encoded signed transaction (accessed via the `raw` property) and the signed transaction object (accessed via the `tx` property).
+ */
+export function decodeSignedTransaction<ReturnFormat extends DataFormat>(
```
jdevcs, line 30: does this support EIP-2718 Enveloped transactions decoding? I think not,  Its only for RLP encoded 0x0 tx decoding?

jdevcs:
>  The decision to be made it whether we will continue to use gas to describe gasLimit in 4.x
> Both the old and new ETH RPC spec use gas for eth_sendTransaction
> Ethers.js uses gasLimit

Yellow paper specifies `gasLimit` instead of `gas`, but EL specs is using `gas`. https://github.com/ethereum/execution-apis/issues/283 will see its response. 

IMO we should keep `gas` field for now. and in future should add support of `gasLimit` field as well

Muhammad-Altabba:
Is it good to have both `gasLimit` and `gas`. Where both of them will have the exact same value? This would mean that the library user can read any of them. And this means, for the user, for example, an easier migration from ethers.js.

nikoulai (approved):
Apart from @jdevcs comments, LGTM.

I think we should go with `gasLimit` as users will be more familiar with yellow paper than the execution specs and would, also, make migration from Ethers.js easier. But, would like to hear the rest of the team.

packages/web3-eth/test/unit/rpc_method_wrappers/fixtures/sign_transaction.ts
Spacesai1or: Updated via [this](https://github.com/ChainSafe/web3.js/pull/5312/commits/ec538fdc8b4e3844c391ec3ebe18d388a9bd497f) commit

packages/web3-eth/src/utils/decode_signed_transaction.ts
Spacesai1or: Yes it can handle type transactions as shown by tests still passing after [this](https://github.com/ChainSafe/web3.js/pull/5312/commits/ec538fdc8b4e3844c391ec3ebe18d388a9bd497f) commit. This is because transaction parsing is done via `@ethereumjs/tx`'s `TransactionFactory.fromSerializedData` [here](https://github.com/ChainSafe/web3.js/blob/ec538fdc8b4e3844c391ec3ebe18d388a9bd497f/packages/web3-eth/src/utils/decode_signed_transaction.ts#L38)

Spacesai1or:
I added code to replace `gasLimit` with `gas` when calling [formatTransaction](https://github.com/ChainSafe/web3.js/blob/9a7af91891334e8876615c6d7e99ea8b6e71b7d2/packages/web3-eth/src/utils/format_transaction.ts#L56) thinking this would be a good way to enforce the library only returning `gas` instead of `gasLimit` or `gas`, however this broke a lot of unit tests and wasn't a complete solution as there are other instances where transaction objects are formatted that don't use `formatTransaction` such as [getTransaction](https://github.com/ChainSafe/web3.js/blob/9a7af91891334e8876615c6d7e99ea8b6e71b7d2/packages/web3-eth/src/rpc_method_wrappers.ts#L659). So, I propose we make a separate issue to decide globally how the library should handle `gasLimit` vs `gas` as a lot of code would need to be updated that's not directly related to this PR

---

#5387 was created to track this

jdevcs (approved):

PR #5324, opened 2022-08-07
Handling of the "null" response from provider

### LINKED ISSUE #5319: Handling of the "null" response from provider

### Current Behavior

In some cases the provider returns `null`, which can't be handled by the `request manager` and cause runtime error. 

### Expected Behavior

There should not be an error and the `null` response should be returned from the request manager. 

### Steps to Reproduce

Send a transaction receipt request for a hash which does not exists. 

### Web3.js Version

4.x

### Environment

N/A

### Anything Else?

_No response_

nazarhussain:
## Description

Fixes #5319

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-core/src/web3_request_manager.ts
+++ b/packages/web3-core/src/web3_request_manager.ts
@@ -274,6 +274,16 @@ export class Web3RequestManager<
 		response: JsonRpcResponse<ResultType, ErrorType>,
 		{ legacy, error }: { legacy: boolean; error: boolean },
 	): JsonRpcResponse<ResultType> | never {
+		if (isNullish(response)) {
+			return this._buildResponse(
+				payload,
+				// Some providers uses "null" as valid empty response
+				// eslint-disable-next-line no-null/no-null
+				null as unknown as JsonRpcResponse<ResultType, ErrorType>,
+				error,
+			);
+		}
+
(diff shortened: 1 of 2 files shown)


avkos (approved):

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -274,6 +274,16 @@ export class Web3RequestManager<
 		response: JsonRpcResponse<ResultType, ErrorType>,
 		{ legacy, error }: { legacy: boolean; error: boolean },
 	): JsonRpcResponse<ResultType> | never {
+		if (isNullish(response)) {
```
jdevcs, line 277: is any provider returning null for valid requests (for which some valid result is expected)?

jdevcs, line 277: can you share rpc request and eth node you used for this case?

jdevcs (approved):

packages/web3-core/src/web3_request_manager.ts
nazarhussain: It was tested for `eth_getTransactionReceipt` with ganache for a transaction hash which is not part of the block yet. 

https://github.com/ethereum/execution-apis/blob/main/src/eth/transaction.yaml#L42

PR #5333, opened 2022-08-10
Enable subscribing to `error` event at `WebSocketProvider` underlying connection

Muhammad-Altabba:
## Description
Implementation of #(https://github.com/ChainSafe/web3.js/issues/5330)

## Type of change

## Checklist:


avkos (approved):

avkos (changes requested):
Great job! 
just need to fix e2e tests

avkos (approved):

packages/web3-providers-ws/src/index.ts
```diff
@@ -257,6 +259,20 @@ export default class WebSocketProvider<
...
 		this._webSocketConnection?.addEventListener('open', this._onOpenHandler);
 		this._webSocketConnection?.addEventListener('close', this._onCloseHandler);
+
+		let errorListeners: unknown[] | undefined;
+		try {
+			errorListeners = this._webSocketConnection?.listeners('error');
+		} catch (error) {
+			// At some cases (at GitHub pipeline) there is an error raised when trying to access the listeners
+			//	However, no need to do take any action in this case
+			console.error(error);
```
jdevcs: Instead of console.error we should use same mechanism to notify user if any user is relying on error event. 
```suggestion
			this._wsEventEmitter.emit('error', error);
```

jdevcs (changes requested):

packages/web3-providers-ws/src/index.ts
Muhammad-Altabba: Yes, but the point is that error is not in the underlying connection. So, it should not be fired in the `error` event I thought. However, I thought about it in a different way now and enhanced the code even more :) please check now...

jdevcs (approved):

PR #5336, opened 2022-08-11
refactor: eip1559 defaults

sirpy:
## Description
The current 2x default for maxFeePerGas and 2.5x for priority fee are both not customizable as defaults and I think too high as a default.
If a user has just enough to pay gas fees at near current rates he might not have the 2x amount that needs to be reserved for the TX.

- change defaults to 1.3x for basefee and 1gwei for priority fee
- add ability to change those defaults via the options object

## Type of change

## Checklist:


nikoulai:
Thanks for reaching out to us. Currently, we are focused on the 4.x rewrite. However, feel free to open a PR.
Changing the default values would be a big change for a lot of users. Creating defaults (like `defaultCommon`) that could be set for all transactions in an instance could work.

jdevcs:
@sirpy +There are many tests failing in CI

packages/web3-core-method/src/index.js
```diff
@@ -890,11 +892,10 @@ function _handleTxPricing(method, tx) {
                         maxFeePerGas = tx.gasPrice;
                         delete tx.gasPrice;
                     } else {
-                        maxPriorityFeePerGas = tx.maxPriorityFeePerGas || '0x9502F900'; // 2.5 Gwei
+                        maxPriorityFeePerGas = tx.maxPriorityFeePerGas || method.defaultMaxPriorityFeePerGas
```
jdevcs, line 895: users can always override default with maxPriorityFeePerGas by explicitly providing it via options and also there is no activity on this PR so closing it.

PR #5338, opened 2022-08-11
call `Error.captureStackTrace(...)` only if available

Muhammad-Altabba:
## Description

Fixes #([issue](https://github.com/ChainSafe/web3.js/issues/5337))

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-errors/src/web3_error_base.ts
+++ b/packages/web3-errors/src/web3_error_base.ts
@@ -61,10 +61,5 @@ export abstract class InvalidValueError extends Web3Error {
 	public constructor(value: unknown, msg: string) {
 		super(`Invalid value given "${Web3Error.convertToString(value, true)}". Error: ${msg}.`);
 		this.name = this.constructor.name;
-		if (typeof Error.captureStackTrace === 'function') {
-			Error.captureStackTrace(this, Web3Error);
-		} else {
-			this.stack = new Error().stack;
-		}
 	}
 }
--- a/packages/web3-errors/src/web3_error_base.ts
+++ b/packages/web3-errors/src/web3_error_base.ts
@@ -61,6 +61,10 @@ export abstract class InvalidValueError extends Web3Error {
 	public constructor(value: unknown, msg: string) {
 		super(`Invalid value given "${Web3Error.convertToString(value, true)}". Error: ${msg}.`);
 		this.name = this.constructor.name;
-		Error.captureStackTrace(this, Web3Error);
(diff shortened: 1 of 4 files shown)


packages/web3-errors/src/web3_error_base.ts
```diff
@@ -61,6 +61,10 @@ export abstract class InvalidValueError extends Web3Error {
 	public constructor(value: unknown, msg: string) {
 		super(`Invalid value given "${Web3Error.convertToString(value, true)}". Error: ${msg}.`);
 		this.name = this.constructor.name;
-		Error.captureStackTrace(this, Web3Error);
+		if (typeof Error.captureStackTrace === 'function') {
+			Error.captureStackTrace(this, Web3Error);
+		} else {
+			this.stack = new Error().stack;
+		}
```
jdevcs: We are capturing stackTrace in constructor of `Web3Error` base class so we can remove it from child class,

jdevcs (changes requested):

nazarhussain (approved):

packages/web3-validator/src/errors.ts
```diff
@@ -54,7 +54,11 @@ export class Web3ValidatorError extends Error {
 			this.errors.length
 		} error[s]:\n${this._compileErrors().join('\n')}`;
 
-		Error.captureStackTrace(this, Web3ValidatorError);
+		if (typeof Error.captureStackTrace === 'function') {
+			Error.captureStackTrace(this, Web3ValidatorError);
+		} else {
+			this.stack = new Error().stack;
+		}
```
Spacesai1or: Why not do this in `packages/web3-errors/src/web3_error_base.ts` so that all errors receive the functionality?

Muhammad-Altabba: Good question! Actually, this class in `web3-validator` takes care of an array of validation errors. And it does not have `code` number like the one in `web3-errors/src/web3_error_base.ts`. So, the classes in `web3-errors` and  `web3-validator` are different. However, there could be a different task to refactor their code if this is desired.

jdevcs: There should be same web3error base class used consistently in all packages, so could you make changes in web3 validator for using web3error base class from web3-errors package, and create error codes. ( Either use Web3Error base class here or move these error classes in Error package + use  Web3Error base class )

Muhammad-Altabba: Ok

jdevcs (approved):

PR #5280, opened 2022-07-25
Add blackbox to workflow

nikoulai:
## Description

Fixes #5261 

## Checklist:

- [ ] I have selected the correct base branch.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules.
- [ ] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [ ] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code.
- [ ] I ran `npm run build` with success.
- [ ] I have tested the built `dist/web3.min.js` in a browser.
- [ ] I have tested my code on the live network.
- [ ] I have checked the Deploy Preview and it looks correct.
- [ ] I have updated the `CHANGELOG.md` file in the root folder.

PR #5353, opened 2022-08-16
Black box tests

Spacesai1or:
Supersedes #5280, but does not include `.html` file to test `web3.min.js`


packages/web3/test/black_box/test/web3-eth-accounts/hashMessage.test.ts
```diff
@@ -0,0 +1,44 @@
...
+// TODO For some reason when running this test with a WebSocket provider
+// the test takes a long time to run afterAll
+import Web3 from 'web3';
+import {
+	closeOpenConnection,
+	getSystemTestProvider,
+	isWs,
+} from 'web3/test/shared_fixtures/system_tests_utils';
+
+jest.setTimeout(15000);
```
Spacesai1or, line 26: For only this test and `packages/web3/test/black_box/test/web3-eth-accounts/create.test.ts`, the `if (isWs) await closeOpenConnection(web3);` in `afterAll` takes longer to complete and surpasses the default timeout for Jest

packages/web3/test/black_box/package.json
```diff
@@ -0,0 +1,24 @@
...
+	"version": "1.0.0",
+	"main": "index.js",
+	"license": "MIT",
+	"scripts": {
+		"test:ganache:http": "WEB3_SYSTEM_TEST_BACKEND=ganache WEB3_SYSTEM_TEST_PROVIDER=http://localhost:8545 jest",
+		"test:ganache:ws": "WEB3_SYSTEM_TEST_BACKEND=ganache WEB3_SYSTEM_TEST_PROVIDER=ws://localhost:8545 jest",
+		"test:geth:http": "WEB3_SYSTEM_TEST_BACKEND=geth WEB3_SYSTEM_TEST_PROVIDER=http://localhost:8545 jest",
+		"test:geth:ws": "WEB3_SYSTEM_TEST_BACKEND=geth WEB3_SYSTEM_TEST_PROVIDER=ws://localhost:8545 jest",
+		"test:infura:http": "WEB3_SYSTEM_TEST_BACKEND=infura jest",
+		"test:infura:ws": "WEB3_SYSTEM_TEST_BACKEND=infura jest"
```
spacesailor24, line 12: These should use already set ENVs instead of setting it here

.github/workflows/build.yml
```diff
@@ -135,6 +135,32 @@ jobs:
       - run: yarn test:e2e:ganache:ws:${{ matrix.browser }}
         shell: bash
 
+  black-box:
```
jdevcs: It should be invoked on **tag** push and **release branch** only for saving time in normal PRs.

.github/workflows/build.yml
```diff
@@ -135,6 +135,32 @@ jobs:
...
+        INFURA_WS: ${{ secrets.INFURA_WS }}
+        INFURA_WSS: ${{ secrets.INFURA_WSS }}
+        INFURA_GOERLI_WS: ${{ secrets.INFURA_GOERLI_WS }}
+        MODE: ${{ matrix.mode }}
+      strategy:
+        fail-fast: false
+        matrix:
+          node: [14, 16]
+          mode: ['http', 'ws']
+          backend: ['ganache', 'geth', 'infura']
```
jdevcs: We already have multiple eth clients tests in e2e so for blackbox we can only test on one eth client and or Infura in github actions,  ( and should keep other Eth clients blackbox tests in scripts for local testing. )

packages/web3/src/version.ts
```diff
@@ -1 +1 @@
-/* eslint-disable header/header */ export const Web3PkgInfo = { version: '4.0.0-alpha.1' };
+/* eslint-disable header/header */ export const Web3PkgInfo = { version: '4.0.0-alpha.0' };
```
jdevcs: Its auto updated by reading https://github.com/ChainSafe/web3.js/blob/7a55b62bfa1c34b78a2014cba97e2e02dff353f9/packages/web3/package.json#L3 so web3 package;s package.json should also be updated

packages/web3/test/black_box/package.json
```diff
@@ -0,0 +1,24 @@
+{
+	"name": "web3_black_box_tests",
+	"version": "1.0.0",
+	"main": "index.js",
+	"license": "MIT",
```
jdevcs: license should be same as root package.json

scripts/verdaccio.sh
```diff
@@ -0,0 +1,97 @@
...
+
+    git add .
+    git commit -m "Comitting for black box publish"
+}
+
+lernaPublish() {
+    lerna publish from-package \
+        --no-git-tag-version \
+        --no-push \
+        --registry http://localhost:4873 \
```
jdevcs, line 66: We are using local registry here but an extra step to avoid risk , 
add some tag other then default `latest` , may be `alpha` or `rc`

jdevcs (changes requested):
lgtm, except some minor changes.

nikoulai (approved):

jdevcs (approved):

luu-alex (approved):

PR #5363, opened 2022-08-19
Block Schema Fix

Spacesai1or:
`blockSchema.transactions` can either be an array of transaction hashes, or an array of transaction objects. Before this PR the schema only handled an array of transaction hashes. The fix that was applied here is very specific to handling `oneOf` for transactions, so I'm not confident in this being a good general solution for the use of `oneOf` in schemas and this code should be refactored to be more general

### DIFF
--- a/packages/web3-utils/src/formatter.ts
+++ b/packages/web3-utils/src/formatter.ts
@@ -145,7 +145,6 @@ export const convert = (
 	schema: JsonSchema,
 	dataPath: string[],
 	format: DataFormat,
-	knownSchemaProp?: JsonSchema,
 ) => {
 	// If it's a scalar value
 	if (!isObject(data) && !Array.isArray(data)) {
@@ -156,7 +155,7 @@ export const convert = (
 
 	for (const [key, value] of Object.entries(object)) {
 		dataPath.push(key);
-		const schemaProp = knownSchemaProp ?? findSchemaByDataPath(schema, dataPath);
+		const schemaProp = findSchemaByDataPath(schema, dataPath);
 
 		// If value is a scaler value
 		if (isNullish(schemaProp)) {
@@ -181,7 +180,7 @@ export const convert = (
 			// a schema using oneOf. This chunk of code was intended to handle
 			// BlockSchema.transactions
 			// TODO BlockSchema.transactions are not being formatted
-			if (schemaProp?.oneOf) {
+			if (schemaProp?.oneOf !== undefined) {
 				for (const oneOfSchemaProp of schemaProp.oneOf) {
 					if (
 						!Array.isArray(schemaProp?.items) &&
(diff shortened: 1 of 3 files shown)


packages/web3-utils/src/formatter.ts
```diff
@@ -174,7 +175,29 @@ export const convert = (
 
 		// If value is an array
 		if (Array.isArray(value)) {
-			if (isNullish(schemaProp?.items)) {
+			let _schemaProp = schemaProp;
+
+			// TODO This is a naive approach to solving the issue of
+			// a schema using oneOf. This chunk of code was intended to handle
+			// BlockSchema.transactions
+			// TODO BlockSchema.transactions are not being formatted
```
Spacesai1or, line 182: The formatter should format each transaction object within `blockSchema.transactions`, currently it doesn't

packages/web3-utils/src/formatter.ts
```diff
@@ -174,7 +175,29 @@ export const convert = (
...
+			if (schemaProp?.oneOf) {
+				for (const oneOfSchemaProp of schemaProp.oneOf) {
+					if (
+						!Array.isArray(schemaProp?.items) &&
+						((typeof value[0] === 'object' &&
+							((oneOfSchemaProp as JsonSchema)?.items as JsonSchema)?.type ===
+								'object') ||
+							(typeof value[0] === 'string' &&
+								((oneOfSchemaProp as JsonSchema)?.items as JsonSchema)?.type !==
+									'object'))
```
Spacesai1or, line 197: This is the fix that's specific to `blockSchema.transactions`, and needs to be refactored to be more general.
It's checking if the first value in the `transactions` array is a hash or a transaction object and uses the corresponding schema

packages/web3-utils/src/formatter.ts
Spacesai1or: Created #5370 to track this

packages/web3-utils/src/formatter.ts
Spacesai1or: Created #5370 to track this

luu-alex (approved):

nikoulai (approved):

jdevcs (approved):

jdevcs (commented):
Current approach will only apply `oneOf` ( from multiple possible schema ) in case of array data in future we will need to extend it for non array data items if required.
So with this requirement applying schema decision is not only dependent on `dataPath` ( in schema ) for provided item but also at data type provided, so it should be handled in https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-utils/src/formatter.ts#L75 and we can also pass data to this function for schema selection decision.
In that way `findSchemaByDataPath` function will be able to return correct schema for oneOf regardless of data is array or not, we can rename it as well.

Issue #5345, opened 2022-08-12
4.x alpha - blackbox web


jdevcs:
blackbox web is failing :

https://github.com/ChainSafe/web3.js/pull/5280/files#diff-8073e12aa862bc8504ad313f190286826cb0e057aed20f7c384882b8a27a166dR23 

```
Uncaught ReferenceError: Web3 is not defined
    at index.html:23:15
```

jdevcs:
Thanks @luu-alex for testing minified build https://github.com/ChainSafe/web3.js/pull/5365#issuecomment-1221382421

PR #5364, opened 2022-08-19
Change webpack configuration

nikoulai:
## Description

#5345

### DIFF
--- a/webpack.base.config.js
+++ b/webpack.base.config.js
@@ -27,9 +27,10 @@ function getWebPackConfig(packagePath, filename, library) {
 		entry: path.resolve(packagePath, 'src/index.ts'),
 		output: {
 			path: path.resolve(packagePath, 'dist'),
-			filename: filename,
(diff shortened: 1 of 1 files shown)


webpack.base.config.js
```diff
@@ -27,9 +27,10 @@ function getWebPackConfig(packagePath, filename, library) {
 		entry: path.resolve(packagePath, 'src/index.ts'),
 		output: {
 			path: path.resolve(packagePath, 'dist'),
-			filename: filename,
-			library: library,
-			libraryExport: 'default',
+			filename,
+			library: { name: 'Web3', type: 'var', umdNamedDefine: true }, // todo derive name from packagePath
```
nazarhussain, line 31: The `getWebPackConfig` is used in different packages, If required we should override this configuration only in the `Web3` packages.

nazarhussain (changes requested):

Issue #5327, opened 2022-08-09
Responses get mixed up due to conflicting payload IDs


abrindam:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using this library in conjunction with Truffle's `HDWalletProvider`, when requesting the current network ID via `web3.eth.net.getId()`, rarely (1 out of 20 or so) you will get back the block height instead of the network ID.

### Expected Behavior

If you request the network ID, you should get back the network ID and not another random piece of information.

### Steps to Reproduce

The following node script reproduces the issue at a frequency of approximately 1/20 times against a local blockchain. The root cause is a network race condition so using a remote blockchain might help reproduce more easily.

```
const Web3 = require('web3')
const HDWalletProvider = require("@truffle/hdwallet-provider");

async function main() {

    const nodeAddress = "ws://127.0.0.1:8545"
    const accountKey = "<valid private key here>"
                    
    provider = new HDWalletProvider({
        privateKeys: [accountKey],
        providerOrUrl: nodeAddress
    });

    const web3 = new Web3(provider)
    console.log(await web3.eth.net.getId())

}

main().catch(console.log).then(() => process.exit(0))
```

### Web3.js Version

1.7.4

### Environment

- Operating System: MacOS 12.4
- Browser: N/A
- Node.js Version: v16.15.1 
- NPM Version: 8.11.0
- `@truffle/hdwallet-provider`: 2.0.13
- Ganache:  v7.4.0 

### Anything Else?

I did a very deep dive into this issue to understand what is going on. It looks like the WebsocketProvider uses transaction IDs to correlate request messages with response messages, but there is no guarantee that transaction messages are unique. In particular, in my case, there was a conflict between the ID generation mechanism of `jsonrpc.js` (https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-core-requestmanager/src/jsonrpc.js#L40) in this package, and the `eth-block-tracker` package which is brought in by `HDWalletProvider`.

The issue with `eth-block-tracker` issue was much worse as they were always using id=1. They have already fixed this in a newer version (https://github.com/MetaMask/eth-block-tracker/pull/42).

The library is using sequential ID generation, which is better, but if two different libraries were to use that strategy, there would be a high likelihood of collision.

[...]

PR #5371, opened 2022-08-22
Start incrementing jsonrpc message id from random number

Muhammad-Altabba:
## Description

Fixes #([issue](https://github.com/ChainSafe/web3.js/issues/5327))

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [ ] I have tested my code.

### DIFF
--- a/packages/web3-core-requestmanager/src/jsonrpc.js
+++ b/packages/web3-core-requestmanager/src/jsonrpc.js
@@ -27,8 +27,9 @@
 // Initialize Jsonrpc as a simple object with utility functions.
 var Jsonrpc = {
     // This is the starting counter for the Jsonrpc.id
-    // Pick a random number between 0 and half of the max safe integer to stay below the max when incrementing
-    messageId: Math.floor(Math.random() * Number.MAX_SAFE_INTEGER / 2)
+    // Pick a random number between 0 and (the maximum safe integer minus the max unsigned 32 bit integer)
+    // Actually, to insure staying in the safe range while incrementing, we insure making the range less than the max unsigned 32 bit integer
+    messageId: Math.floor(Math.random() * (Number.MAX_SAFE_INTEGER - Math.pow( 2, 32 )))
 };
 
 /**
--- a/packages/web3-core-requestmanager/src/jsonrpc.js
+++ b/packages/web3-core-requestmanager/src/jsonrpc.js
@@ -26,9 +26,8 @@
 
 // Initialize Jsonrpc as a simple object with utility functions.
 var Jsonrpc = {
-    // This is the starting counter for the Jsonrpc.id
-    // Pick a random number between 0 and (the maximum safe integer minus the max unsigned 32 bit integer)
-    // Actually, to insure staying in the safe range while incrementing, we insure making the range less than the max unsigned 32 bit integer
+    // This is the starting counter for the Jsonrpc.id. And, to insure staying in the safe range while incrementing later,
+    //  pick a random number between 0 and (the maximum safe integer minus the max unsigned 32 bit integer)
     messageId: Math.floor(Math.random() * (Number.MAX_SAFE_INTEGER - Math.pow( 2, 32 )))
 };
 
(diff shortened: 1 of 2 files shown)


luu-alex:
Looks good 👍  is this something we want to consider for 4.x as well?

packages/web3-core-requestmanager/src/jsonrpc.js
```diff
@@ -26,7 +26,9 @@
 
 // Initialize Jsonrpc as a simple object with utility functions.
 var Jsonrpc = {
-    messageId: 0
+    // This is the starting counter for the Jsonrpc.id. And, to insure staying in the safe range while incrementing later,
+    //  pick a random number between 0 and (the maximum safe integer minus the max unsigned 32 bit integer)
+    messageId: Math.floor(Math.random() * (Number.MAX_SAFE_INTEGER - Math.pow( 2, 32 )))
```
nazarhussain: I had a hunch that this will not fix the actual issue. 

> Responses get mixed up due to conflicting payload IDs

The response will still be mixed up if the payload Ids are conflicting, starting point does not matter and randomizing starting point would minimize but not fixed the problem. 

For once instance of the `web3` there will be no conflicting payload IDs in any case.

nazarhussain (changes requested):

packages/web3-core-requestmanager/src/jsonrpc.js
nikoulai: As per here: https://www.jsonrpc.org/specification

```
An identifier established by the Client that MUST contain a String, Number, or NULL value if included. 
```
`uuid` is already used in 1.x, maybe we could use it, also, for request id`s. We won't need a starting point and other libraries use numbers for ids.

nikoulai (approved):
PR looks good, just added a comment in the conversation.

packages/web3-core-requestmanager/src/jsonrpc.js
Muhammad-Altabba: Yes, there will be no issue if one instance is running and one library is used. And, yes, this solution will minimize the possibility of a conflict to some grade that is neglected but the possibility is still there.
But using a uuid (that would be handled as a string) instead of a number would be a breaking change. Even the tests for 1.x would need to be modified accordingly. 

However, I proposed to handle this in a different way in v4.x as there we can do a breaking change by making the message id as a string that contains a UUID or another format. The new issue is here: https://github.com/ChainSafe/web3.js/issues/5373

And if you have a better suggestion to solve this in v1.x, please advise.

Thanks,

luu-alex (approved):

avkos (approved):

jdevcs (approved):

packages/web3-core-requestmanager/src/jsonrpc.js
nazarhussain: We can proceed with this approach but considering it's not a permanent fix. 

Also it does not matter if there is more than one library using the same id for the same endpoint. For the HTTP, the RPC server will be responding synchronously. For WS based RPC server it will be responding per connection. So if there are two libraries, both would be having different connections and that will not conflict the ids.

nazarhussain (approved):

packages/web3-core-requestmanager/src/jsonrpc.js
Muhammad-Altabba: Thanks @nazarhussain ,
As you see this as "not a permanent fix", do you have a better suggestion for 1.x that does not cause a breaking change?

abrindam: > So if there are two libraries, both would be having different connections and that will not conflict the ids.

@nazarhussain I just wanted to note in my original report #5327, I saw a situation where two libraries were using the same connection, which resulted in an ID conflict. I'm not certain how this happened because I'm completely unfamilar with your codebase, but it's somehow possible.

In my case, it was `eth-block-tracker` that was sending a message that somehow bypassed `jsonrpc.js` but still was being processed by the same websocket connection that `jsonrpc.js` was using, which resulted in a conflict.

Maybe someone more experienced than me could figure out how that happened? There might be another lurking bug there.

PR #5377, opened 2022-08-24
Nikos/5071/investigate signtransaction testcases

nikoulai:
## Description

#5071 part 1:   r and v values in testcases don't align with `signTransaction` results.

In two testcases r and v were wrong (they were even the same between the tests). `signTransaction` returns the right values (cross checked with external libraries, too)

Discussion:
r and s are `bigint`s in `ethereumjs` and we cast them to strings. This leads to some values with smaller values to be of smaller length when casted to string (missing leading 0s). I think, we should prepend them, but, this could be a breaking change for some users and better leave them as they are.

### DIFF
--- a/test/eth.accounts.signTransaction.js
+++ b/test/eth.accounts.signTransaction.js
@@ -133,8 +133,9 @@ var tests = [
             common: common
         },
         // expected r and s values from signature
-        r: "0x22f17b38af35286ffbb0c6376c86ec91c20ecbad93f84913a0cc15e7580cd9",
-        s: "0x83d6e12e82e3544cb4439964d5087da78f74cefeec9a450b16ae179fd8fe20",
+        r: "0x74dcecc6b8ad09ca09882ac1088eac145e799f56ea3f5b5fe8fcb52bbd3ea4f7",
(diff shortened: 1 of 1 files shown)


luu-alex:
bigints casted into strings should be fine though no? I think the issue is that some part where the bigint will be turned into a  number, thats when we get smaller values. If thats the issue it wont be breaking

avkos (approved):

test/eth.accounts.signTransaction.js
```diff
@@ -133,8 +133,9 @@ var tests = [
             common: common
         },
         // expected r and s values from signature
-        r: "0x22f17b38af35286ffbb0c6376c86ec91c20ecbad93f84913a0cc15e7580cd9",
-        s: "0x83d6e12e82e3544cb4439964d5087da78f74cefeec9a450b16ae179fd8fe20",
+        r: "0x74dcecc6b8ad09ca09882ac1088eac145e799f56ea3f5b5fe8fcb52bbd3ea4f7",
```
jdevcs, line 136: did you get correct r and s values from functionality of our lib or any 3rd party lib for creation of these new test case data?

nikoulai: The results of the libraries were aligned with these values.

jdevcs (approved):

PR #5386, opened 2022-08-25
Init CODEOWNERS File

Spacesai1or:
This will automatically add members of the `@ChainSafe/web3-js` team as reviewers when a PR is opened for the `4.x` branch. One deficiency of using the `web3-js` team is we have members that are not code owners that will be added as reviewers. I can think of two solutions for this:

1. Create another Github team: `web3-js-codeowners` and only add active maintainers of Web3.js
2. List each maintainer's Github username explicitly (I'd refrain from doing this as this `CODEOWNERS` file will have to be maintained as devs transition on and off the team)


nikoulai:
1. seems to be a good solution

mpetrunic:
> This will automatically add members of the `@ChainSafe/web3-js` team as reviewers when a PR is opened for the `4.x` branch. One deficiency of using the `web3-js` team is we have members that are not code owners that will be added as reviewers. I can think of two solutions for this:
> 
> 1. Create another Github team: `web3-js-codeowners` and only add active maintainers of Web3.js
> 2. List each maintainer's Github username explicitly (I'd refrain from doing this as this `CODEOWNERS` file will have to be maintained as devs transition on and off the team)

I think you have a third option, which I believe is totally fine: 
- remove non-active maintainers from the web3.js team. Afaik those are only Greg and Phil

Spacesai1or:
> > This will automatically add members of the `@ChainSafe/web3-js` team as reviewers when a PR is opened for the `4.x` branch. One deficiency of using the `web3-js` team is we have members that are not code owners that will be added as reviewers. I can think of two solutions for this:
> > 
> > 1. Create another Github team: `web3-js-codeowners` and only add active maintainers of Web3.js
> > 2. List each maintainer's Github username explicitly (I'd refrain from doing this as this `CODEOWNERS` file will have to be maintained as devs transition on and off the team)
> 
> I think you have a third option, which I believe is totally fine:
> 
> * remove non-active maintainers from the web3.js team. Afaik those are only Greg and Phil

@mpetrunic I was thinking this would remove admin privileges for Greg and Phil, but I see now that there's a separate team for that

avkos (approved):

nikoulai (approved):

Muhammad-Altabba (approved):

Issue #5091, opened 2022-05-31
Research Plug-in base architecture


mconnelly8:

jdevcs:
For additional features in Web3 4.x, there can be several ways of instantiating new features on demand as additional plugin in web3.js.

Consider a scenario, we want to have chainlink plugin with chainlink `getPrice()` functionality support.

- Offering plugin instantiation via additional call `web3.eth.registerPlugin(ChainlinkPlugin)` Internally, it should do required instantiation. In this way necessary plugin will be registered in Eth package and that plugin will have access to web3 context , request manager, config params ,..etc for performing its functionality. 

- and then making available plugin functionality to lib users like in above mentioned case following call `web3.eth.chainlink.getPrice(ChainlinkFeeds.ETH-USD) ` will be available.

jdevcs:
Following can be rough idea as starting point of discussion:

Part of our mono repo:
```

export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscription> {


    public registerPlugIn( pluginTitle : string, plugin : Web3PluginBase){
        
        // add plugin in Eth class as property
        Object.assign(this, {pluginTitle: self.use(plugin) } );  // via web3Context use()

        ..... other instantiating etc
    }

.... Eth functionality

}


export abstract Web3PluginBase extends Web3Context {
    // additional specifications for plugins to implement
}
```

In independent solution:
```
class ChainlinkPlugin extends Web3PlugInBase {
    // will have access to request manager, context, and config params
 ......
}


```

Spacesai1or:
[Related](https://github.com/ChainSafe/web3.js/discussions/4653) discussion

mpetrunic:
Some ideas which I think you should consider:
- compatibility with web3 versions - I think plugin should declare version range of Web3.js so we can error on unsupported versions
- plugins using other plugins - ability for plugin to declare dependency on other plugins

Example:
```
interface Web3Plugin {
   plugin: Plugin, //for example Layer2Service
   web3Version: string, // ">=4 <5" or "4.x"
   needs?: string[] // ["web3-layer2@1.x"]
}
```

Spacesai1or:
After talking with @nazarhussain there are decisions to be made:

- What exactly are the requirements for the plugin system? 
    - The `.use` and `.link` methods that exist on any class extending `Web3Context` already allow them to make use of an already instantiated provider
- How do we make appending a plugin to the `Web3` object typesafe?
[...]

PR #5393, opened 2022-08-30
Example Plugin Implementation 

Spacesai1or:
Example implementation for #5091

### Unresolved Issues

1. #5490
2. #5491
3. #5492

### Points of Discussion

- ~I believe the `Web3APISpec` on [this](https://github.com/ChainSafe/web3.js/blob/d0d388b6f26b739d787653285f1ba8a88eb4fcfe/packages/web3-plugin-example/src/index.ts#L44) line should be `ChainlinkPluginAPI`, but when making the change the following error appears for [this](https://github.com/ChainSafe/web3.js/blob/d0d388b6f26b739d787653285f1ba8a88eb4fcfe/packages/web3-plugin-example/src/index.ts#L55)~
    - ~Attempted fixes for this include: #5416 and #5480~

```ts
Argument of type 'this' is not assignable to parameter of type 'Web3Context<EthExecutionAPI, { logs: typeof LogsSubscription; newHeads: typeof NewHeadsSubscription; newBlockHeaders: typeof NewHeadsSubscription; }>'.
  Type 'ChainlinkPlugin' is not assignable to type 'Web3Context<EthExecutionAPI, { logs: typeof LogsSubscription; newHeads: typeof NewHeadsSubscription; newBlockHeaders: typeof NewHeadsSubscription; }>'.
    Types of property '_requestManager' are incompatible.
      Type 'Web3RequestManager<ChainlinkPluginAPI>' is not assignable to type 'Web3RequestManager<EthExecutionAPI>'.
        Type 'ChainlinkPluginAPI' is missing the following properties from type 'EthExecutionAPI': eth_getBlockByHash, eth_getBlockByNumber, eth_getBlockTransactionCountByHash, eth_getBlockTransactionCountByNumber, and 44 more.ts(2345)
```
- ~In order to get the [declare module](https://github.com/ChainSafe/web3.js/blob/d0d388b6f26b739d787653285f1ba8a88eb4fcfe/packages/web3-plugin-example/src/index.ts#L38) to work, I needed to update the [Web3 package](https://github.com/ChainSafe/web3.js/blob/d0d388b6f26b739d787653285f1ba8a88eb4fcfe/packages/web3/src/web3.ts) from using the `default` export to a named one - I'm not sure if/how this affects anything~ [This](https://github.com/ChainSafe/web3.js/pull/5393/commits/755d6d65fb490493b3fbf529589f27069e6e7f59) commit updates `Web3` to be exported using `default` and named export
    - ~Related to this change, importing `Web3` like [this](https://github.com/ChainSafe/web3.js/blob/d0d388b6f26b739d787653285f1ba8a88eb4fcfe/packages/web3-plugin-example/src/index.ts#L22) is the only way I could get the module augmentation to work~ The above mentioned commit also updates this to use `web3` to augment module instead of `web3/dist/web3`
- ~The return type for [getPrice](https://github.com/ChainSafe/web3.js/blob/d0d388b6f26b739d787653285f1ba8a88eb4fcfe/packages/web3-plugin-example/src/index.ts#L54) is inferred from the contract ABI instead of `ChainlinkPluginAPI`~
    - ~This doesn't matter so much because the type the end user will see when calling `web3.chainlink.getPrice` is defined by `ChainlinkPluginAPI` (i.e. user will see `getPrice` returns `Promise<Price>`)~


packages/web3-plugin-example/test/unit/plugin.test.ts
```diff
@@ -0,0 +1,39 @@
...
+*/
+import Web3 from 'web3';
+
+import { ChainlinkPlugin } from '../../src/index';
+import { AggregatorV3InterfaceABI } from '../../src/aggregator_v3_interface_abi';
+
+const aggregatorAddress = '0xECe365B379E1dD183B20fc5f022230C044d51404';
+
+describe('Chainlink Plugin Tests', () => {
+	it('should do something', async () => {
```
Spacesai1or:
```suggestion
	it('should register ChainlinkPlugin and make the getPrice call', async () => {
```

packages/web3-core/src/web3_context.ts
```diff
@@ -232,6 +232,15 @@ export class Web3Context<
 		});
 	}
 
+	// eslint-disable-next-line no-use-before-define
+	public registerPlugin(plugin: Web3PluginBase<any>) {
+		const _pluginObject = {
+			[plugin.pluginNamespace]: plugin,
+		};
+		_pluginObject[plugin.pluginNamespace].link(this);
+		Object.assign(this, _pluginObject);
```
mpetrunic, line 244: you should probably check and error if something in the same namespace is already registered

packages/web3-plugin-example/package.json
```diff
@@ -0,0 +1,63 @@
...
+		"eslint-config-prettier": "^8.5.0",
+		"eslint-config-web3-base": "0.1.0",
+		"eslint-plugin-import": "^2.26.0",
+		"jest": "^28.1.3",
+		"jest-extended": "^3.0.1",
+		"prettier": "^2.7.1",
+		"ts-jest": "^28.0.7",
+		"typescript": "^4.7.4"
+	},
+	"dependencies": {
```
mpetrunic: should those be peerDependencies?

packages/web3/src/index.ts
```diff
@@ -321,6 +321,6 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 /**
  * This comment _supports3_ [Markdown](https://marked.js.org/)
  */
-import Web3 from './web3';
+import { Web3 } from './web3';
```
mpetrunic: Why not do both module export and named export instead of just module export?
Named export generally works better with IDE (auto-import and type hinting)

packages/web3-core/src/web3_context.ts
```diff
@@ -276,3 +285,7 @@ export type TransactionBuilder<
 	web3Context: Web3Context<API>;
 	privateKey?: HexString | Buffer;
 }) => Promise<ReturnType>;
+
+export abstract class Web3PluginBase<API extends Web3APISpec> extends Web3Context<API> {
```
mpetrunic: I assume this is how the rest of the modules in web3 are built. The only issue I see with this is once we have plugins that depend on other plugins, types won't work because we are augmenting Web3 class and plugins are using Web3Context.

Alternative of top of my head would be:
```
export const plugin1 = {
   namespace,
   init: (web3: Web3) => Plugin1Api
}

declare module web3{
  	interface Web3 {
		plugin1: Plugin1Api;
	}
}


```
and if you require plugin1 in plugin2:
```

import plugin2 from "plugin2"

export const plugin1 = {
   namespace,
   needs: [plugin1.namespace]
   // because we imported plugin2, Web3 class here will be augmented with plugin1 api
   init: (web3: Web3) => Plugin1Api
}
```

packages/web3-plugin-example/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+	startedAt: string;
+	updatedAt: string;
+	answeredInRound: string;
+}
+
+interface ChainlinkPluginAPI extends Web3APISpec {
+	getPrice: () => Promise<Price>;
+}
+
+declare module 'web3/dist/web3' {
```
mpetrunic: Have you tested if end user imports Web3 from 'web3' (instead of 'web3/dist/web3') will they see augmented API?

packages/web3/src/web3.ts
```diff
@@ -31,7 +31,7 @@ import { initAccountsForContext } from './accounts';
 import { Web3EthInterface } from './types';
 import { Web3PkgInfo } from './version';
 
-export default class Web3 extends Web3Context<EthExecutionAPI> {
```
jdevcs, line 34: This will break minified web3.js build https://github.com/ChainSafe/web3.js/issues/5345

packages/web3-plugin-example/package.json
```diff
@@ -0,0 +1,63 @@
...
+		"test:ci": "jest --coverage=true --coverage-reporters=json --verbose",
+		"test:watch": "npm test -- --watch",
+		"test:unit": "jest --config=./test/unit/jest.config.js",
+		"test:integration": "jest --config=./test/integration/jest.config.js --runInBand --forceExit",
+		"test:blackbox:ganache:http": "./scripts/black_box_test.sh ganache http",
+		"test:blackbox:ganache:ws": "./scripts/black_box_test.sh ganache ws",
+		"test:blackbox:geth:http": "./scripts/black_box_test.sh geth http",
+		"test:blackbox:geth:ws": "./scripts/black_box_test.sh geth ws",
+		"test:blackbox:infura:http": "./scripts/black_box_test.sh infura http",
+		"test:blackbox:infura:ws": "./scripts/black_box_test.sh infura ws"
```
jdevcs: I think these are not required in example plugin dir as each plugin might have its own set of tests.

`"test:integration": "jest --config=./test/integration/jest.config.js --runInBand --forceExit",
		"test:blackbox:ganache:http": "./scripts/black_box_test.sh ganache http",
		"test:blackbox:ganache:ws": "./scripts/black_box_test.sh ganache ws",
		"test:blackbox:geth:http": "./scripts/black_box_test.sh geth http",
		"test:blackbox:geth:ws": "./scripts/black_box_test.sh geth ws",
		"test:blackbox:infura:http": "./scripts/black_box_test.sh infura http",
		"test:blackbox:infura:ws": "./scripts/black_box_test.sh infura ws"`

packages/web3-plugin-example/webpack.config.js
```diff
@@ -0,0 +1,20 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+const { getWebPackConfig } = require('../../webpack.base.config');
+
+module.exports = getWebPackConfig(__dirname, 'web3.min.js', 'Web3', 'src/web3.ts');
```
jdevcs:
```suggestion
module.exports = getWebPackConfig(__dirname, 'web3-plugin-example.min.js', 'web3-plugin-example', 'src/index.ts');
```

jdevcs (changes requested):

jdevcs (changes requested):
If we want to keep this as example for others to implement plugins, :
`web3-plugin-example` example package should be placed under [tools](https://github.com/ChainSafe/web3.js/tree/4.x/tools)

packages/web3/src/web3.ts
mpetrunic: https://github.com/ChainSafe/web3.js/pull/5393?notification_referrer_id=NT_kwDOAIbUcrI0Mjk0NjMxMTc2Ojg4MzYyMTA#discussion_r958191421

packages/web3-plugin-example/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export class ChainlinkPlugin extends Web3PluginBase<Web3APISpec> {
+	public pluginNamespace = 'chainlink';
+
+	protected readonly _contract: Contract<typeof AggregatorV3InterfaceABI>;
+
+	public constructor(abi: ContractAbi, address: Address) {
+		super();
+		this._contract = new Contract(abi, address);
```
jdevcs: We should leverage `context` coming via inheritance `Web3PluginBase`  and use that to instantiate contract. Ideally with every usage there should not be need to check :
`if (this._contract.currentProvider === undefined) this._contract.link(this);` if context is used while instantiation.

packages/web3-plugin-example/package.json
Spacesai1or: Removed in [this](https://github.com/ChainSafe/web3.js/pull/5393/commits/6514f1352645cb6793e9670fe7968c95176e5713) commit

packages/web3-plugin-example/webpack.config.js
Spacesai1or: Removed the file in [this](https://github.com/ChainSafe/web3.js/pull/5393/commits/9b4896e5cacc8e83b93999cc16492813389f38d5) commit, was an erroneous copy

Spacesai1or:
> If we want to keep this as example for others to implement plugins, : `web3-plugin-example` example package should be placed under [tools](https://github.com/ChainSafe/web3.js/tree/4.x/tools)

@jdevcs My thinking was to create this package with unit and black box tests that can be ran as apart of our test suite for releases to verify the plugin system continues to work as expected. I could move this under `tools/`, but we'd have to setup Lerna and such to build and execute tests in `tools/` and `packages/`

packages/web3-plugin-example/package.json
spacesailor24: I believe `web3` can be a peer dependency, but the other packages are actually being used by the plugin - perhaps this will change a little after I implement @jdevcs's [feedback](https://github.com/ChainSafe/web3.js/pull/5393#discussion_r958579938)

Spacesai1or: I'm not sure they can be since the plugin is actually [importing the packages](https://github.com/ChainSafe/web3.js/blob/db296ff7f47723980cb520f2bf98340dde096803/packages/web3-plugin-example/src/index.ts#L22)?

```ts
import { ContractAbi } from 'web3-eth-abi';
import Contract from 'web3-eth-contract';
import { Web3PluginBase } from 'web3-core';
import { Address, Web3APISpec } from 'web3-types';
// @ts-expect-error 'Web3' is declared but its value is never read.
import { Web3 } from 'web3/dist/web3';
```

packages/web3-core/src/web3_context.ts
Spacesai1or: Addressed via [this](https://github.com/ChainSafe/web3.js/pull/5393/commits/dc36d4b087c011516a5b8a19da8d71148e845fb4) commit

packages/web3-plugin-example/package.json
mpetrunic: That doesn't really matter. The only difference here is if you put peer dependencies and set a broad range like ">=4.0.0 < 5" it will warn the user if he is supplying an invalid version, otherwise it would use his version. On the other hand, if you leave it as regular dep, it will install the plugin version of web3 along with a user-supplied version of web3.

packages/web3/src/index.ts
Spacesai1or: In [this](https://github.com/ChainSafe/web3.js/pull/5393/commits/755d6d65fb490493b3fbf529589f27069e6e7f59) commit I've updated `Web3` to be exported via `default` and named

packages/web3-plugin-example/src/index.ts
Spacesai1or: After [this](https://github.com/ChainSafe/web3.js/pull/5393/commits/755d6d65fb490493b3fbf529589f27069e6e7f59) commit, `Web3` is now exported as a named export which means this is now:

```ts
import { Web3 } from 'web3';

declare module 'web3'
```

packages/web3/src/web3.ts
Spacesai1or: [This](https://github.com/ChainSafe/web3.js/pull/5393/commits/755d6d65fb490493b3fbf529589f27069e6e7f59) commit updates `Web3` to be exported via `default` and named. I've created a [test example](https://github.com/spacesailor24/web3-5393-min-test) that uses the `web3.min.js` and calls various functions to ensure the minified build works

Muhammad-Altabba:
Hello @spacesailor24,
I opened an MR (https://github.com/ChainSafe/web3.js/pull/5416) for addressing the discussion points mentioned at [#5393 (comment)](https://github.com/ChainSafe/web3.js/pull/5393#issue-1355072205). More details are at its description. Please feel free to check and discuss.
Thanks,

jdevcs:
@spacesailor24 this PR is still draft and waiting from few sprints, 
are there any pending items or required discussion?

packages/web3-plugin-example/package.json
Spacesai1or: [This](https://github.com/web3/web3.js/pull/5393/commits/7642f41e41f78f9470fecb32ad43f70ab187239a) commit adds the dependencies as both `peerDependencies` and `devDependencies` as per [this](https://github.com/lerna/lerna/issues/375#issuecomment-326397282)

packages/web3-plugin-example/src/index.ts
Spacesai1or: #5492 tracks this and there seems to be a bug with updating a context that was passed upon instantiation of `Contract`

packages/web3-core/src/web3_context.ts
Spacesai1or: #5490 tracks this

packages/web3-plugin-example/test/unit/custom_rpc_methods.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import { CustomRpcMethodsPlugin } from '../../src/custom_rpc_methods';
+
+describe('CustomRpcMethodsPlugin Tests', () => {
+	it('should register CustomRpcMethodsPlugin plugin', () => {
+		const web3 = new Web3('http://127.0.0.1:8545');
+		web3.registerPlugin(new CustomRpcMethodsPlugin());
+		// Both CustomRpcMethodsPlugin and ContractMethodWrappersPlugin
+		// redeclare the web3 module, and this seems to confuse the TypeScript server
+		// @ts-expect-error  Unsafe call of an `any` typed value
```
Spacesai1or: As mentioned in the comment, I believe this is caused by both plugin classes re-declaring the `web3` module:

```typepscript
declare module 'web3' {
	interface Web3 {
		pluginNamespace: pluginClass;
	}
}
```

As commenting out the `ContractMethodWrappersPlugin` class causes these errors to disappear

packages/web3-plugin-example/src/ERC20Token.ts
```diff
@@ -0,0 +1,141 @@
...
+
+web3.js is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+export const ERC20TokenAbi = [
```
jdevcs: I think we should use existing abi + bytecode from 4x/web3.js/fixtures/build/ERC20Token.ts

packages/web3-plugin-example/src/index.ts
```diff
@@ -0,0 +1,16 @@
+/*
```
jdevcs: why we need this empty file?

jdevcs (changes requested):
Can you add changelog entry for plugin feature.

jdevcs:
Please check that this PR is not causing any actions failing in CI.+ Just need some minor changes, and we can merge this PR. 

As discussed in call for remaining bugs, improvements and more features we should do in another PR.

packages/web3-plugin-example/src/ERC20Token.ts
Spacesai1or: The reason a new file was created because the ABI is used [here](https://github.com/web3/web3.js/blob/3ad7e1dac7536b880d306433dbba78d360eff4dd/packages/web3-plugin-example/src/contract_method_wrappers.ts#L24) for the `ContractMethodWrappersPlugin` class, and while I think it's okay for tests to reach outside of the package directory, I don't think the package implementation code should depend on files outside of the package scope

packages/web3-plugin-example/src/index.ts
Spacesai1or: Originally it was because the `main` entry file in the `package.json` was pointing to this file, but I've moved the `web3-plugin-example` to the `tools/` directory, as it shouldn't ever be published to NPM and thus the `main` entry is not needed (and this file was removed)

Spacesai1or:
> Please check that this PR is not causing any actions failing in CI.+ Just need some minor changes, and we can merge this PR.
> 
> As discussed in call for remaining bugs, improvements and more features we should do in another PR.

@jdevcs It looks like failing tests are related to a [validation error](https://github.com/web3/web3.js/actions/runs/3224435810/jobs/5275597865#step:4:154) in `web-eth-contract` that is unrelated to the changes introduced in this PR

jdevcs (approved):

packages/web3/src/index.ts
```diff
@@ -323,4 +323,5 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
  */
 import Web3 from './web3';
 
-export = Web3;
+export * from './web3';
```
Muhammad-Altabba: Those changes revert back the modifications made at: https://github.com/web3/web3.js/pull/5221/files#diff-bef53270c59945a684e63b962974a96318e07549c207ea92bbc9c369e35bc476R326
And it will cause backward incompatibility with version 1.x, right?

tools/web3-plugin-example/CHANGELOG.md
```diff
@@ -0,0 +1,36 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+<!-- EXAMPLE
+
+## [1.0.0]
+
+### Added
```
Muhammad-Altabba, line 12: Could this list the functionality that is added in this initial version?

packages/web3-core/CHANGELOG.md
```diff
@@ -40,3 +40,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 -   If the response error was `execution reverted`, raise `ContractExecutionError` and pass the response error to it in order to be set as `innerError` (this innerError will be decoded at web3-eth-contract if its ABI was provided according to EIP-838). (#5434)
+-   `registerPlugin` method to `Web3Context` (#5393)
```
Muhammad-Altabba, line 43: Could all the changes to `CHANGELOG.md` files be duplicated on the one at the root?

Muhammad-Altabba (changes requested):
Looks great beside tiny stuff in addition to the already mentioned points by other reviewers.

packages/web3/src/index.ts
Spacesai1or: It doesn't break anything as mentioned [here](https://github.com/web3/web3.js/pull/5393#discussion_r960143755)

tools/web3-plugin-example/CHANGELOG.md
Spacesai1or: I believe that would be better suited for documentation, not a `CHANGELOG` - we also didn't do this for any other package

packages/web3-types/CHANGELOG.md
```diff
@@ -35,10 +35,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -->
 
-## [Unreleased]
+## [Unreleased]Z
```
Spacesai1or:
```suggestion
## [Unreleased]
```

packages/web3-core/CHANGELOG.md
Spacesai1or: Updated via [this](https://github.com/web3/web3.js/pull/5393/commits/bba74e69e1985e06d4a971701b8e5598a8636522) commit

packages/web3/src/index.ts
Muhammad-Altabba: Actually, I am referring to how web3 will be imported inside javascript and typescript files.

So the question is: After this MR, can the user still use the old way?
```typescript
import Web3 from 'web3';  // without the curly brackets around Web3
```

tools/web3-plugin-example/CHANGELOG.md
Muhammad-Altabba: It seems like this thing was missed early. So, if you would not add it in this MR, could you please ensure requesting to write this along with writing the documentation?  And then resolve this discussion point directly.
Thanks,,

CHANGELOG.md
```diff
@@ -771,6 +775,10 @@ should use 4.0.1-alpha.0 for testing.
 
 ### Changed
 
+#### web3
+
+-   `Web3` is now a default and named export (#5393)
+
```
Muhammad-Altabba, line 787:
```suggestion

#### web3-core

-   Default value for `API` generic for `Web3ContextObject` from `any` to `unknown` (#5393)
-   Default value for `API` generic for `Web3ContextInitOptions` from `any` to `unknown` (#5393)

```

CHANGELOG.md
```diff
@@ -789,6 +797,10 @@ should use 4.0.1-alpha.0 for testing.
 
 -   Return `BigInt` instead of `string` when decoding function parameters for large numbers, such as `uint256`. (#5435)
 
+#### web3-types
+
+-   `Web3APISpec`, `Web3APIMethod`, and `Web3APIParams` not support `unknown` APIs (#5393)
```
Muhammad-Altabba:
```suggestion
-   `Web3APISpec`, `Web3APIMethod`, and `Web3APIParams` now support `unknown` APIs (#5393)
```

CHANGELOG.md
```diff
@@ -771,6 +775,10 @@ should use 4.0.1-alpha.0 for testing.
```
Muhammad-Altabba, line 780: Something like the following (ideally above `#### web3-utils` so titles are sorted alphabetically)
```suggestion

#### web3-plugin-example
- The package `web3-plugin-example` has been added with the basic functionality. (#5393)
```

packages/web3-types/CHANGELOG.md
```diff
@@ -42,3 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   `Web3EthExecutionAPI` export (#5441)
 -   `Web3NetAPI` export (#5441)
 -   `EthPersonalAPI` export (#5441)
+
+### Changed
+
+-   `Web3APISpec`, `Web3APIMethod`, and `Web3APIParams` not support `unknown` APIs (#5393)
```
Muhammad-Altabba:
```suggestion
-   `Web3APISpec`, `Web3APIMethod`, and `Web3APIParams` now support `unknown` APIs (#5393)
```

packages/web3/src/index.ts
Spacesai1or: Yes, on line `327`, `Web3` is still the `default` export of the package:

```typescript
export default Web3;
```

An example of this working is in the [contract_method_wrappers tests](https://github.com/web3/web3.js/blob/6093e5bb5aee9fd87f00c5fbfcbe35f3b5d1ddd0/tools/web3-plugin-example/test/unit/contract_method_wrappers.test.ts#L17)

packages/web3-types/CHANGELOG.md
Spacesai1or:
```suggestion
-   `Web3APISpec`, `Web3APIMethod`, and `Web3APIParams` now supports `unknown` APIs (#5393)
```

CHANGELOG.md
```diff
@@ -789,6 +802,10 @@ should use 4.0.1-alpha.0 for testing.
 
 -   Return `BigInt` instead of `string` when decoding function parameters for large numbers, such as `uint256`. (#5435)
 
+#### web3-types
+
+-   `Web3APISpec`, `Web3APIMethod`, and `Web3APIParams` not support `unknown` APIs (#5393)
```
```suggestion
-   `Web3APISpec`, `Web3APIMethod`, and `Web3APIParams` now supports `unknown` APIs (#5393)
```

tools/web3-plugin-example/CHANGELOG.md
Spacesai1or: Updated via [this](https://github.com/web3/web3.js/pull/5393/commits/c939cbc428f00151ff392588b110d1cab12621a5) commit

CHANGELOG.md
Spacesai1or: Updated via [this](https://github.com/web3/web3.js/pull/5393/commits/c939cbc428f00151ff392588b110d1cab12621a5) commit

packages/web3/src/index.ts
Muhammad-Altabba: I created a simple javascript project with the following code:
```javascript
const Web3 = require('web3'); 

const web3 = new Web3();
```

And only when using this branch, the following error occurs:
> TypeError: Web3 is not a constructor

So, could you please revert back any changes made to how Web3 is exported? Or else, if it is necessary to have those changes for the plugin, then please let me know where exactly this is needed so we can investigate more on this matter...

Spacesai1or: By changing Web3 to be exported like so

```typescript
import Web3 from './web3';

export * from './web3';

// export default Web3; <- How it was
module.exports = Web3;
```

The following works

```typescript
const Web3 = require('web3');
const web3 = new Web3('http://127.0.0.1:8545');
```

And importing Web3 in the browser seems to still work as well

image

Spacesai1or: The above give us the following eslint error though

```bash
web3.js/packages/web3/src/index.ts
  324:1  error    Cannot use import declarations in modules that export using CommonJS (module.exports = 'foo' or exports.bar = 'hi')  import/no-import-module-exports
```

Muhammad-Altabba: Do you face any problems reverting this MR's changes to the Web3 export statement?
It was `export = Web3;` not `module.exports = Web3;`
It was:
```typescript
import Web3 from './web3';

export = Web3;
```

Spacesai1or: Yes, TypeScript module augmentation seems to be designed to work with named ES modules. So only exporting `Web3` as a CommonJS module results in the following error

```bash
test/unit/contract_method_wrappers.test.ts:27:16 - error TS2664: Invalid module name in augmentation, module 'web3' cannot be found.

    27 declare module 'web3' {
```

I've tried many different ways to work around this, but it boils down to us not being able to mix CommonJS and ES modules, the latter being needed to provide module augmentation

However, we can use

```typescript
import Web3 from './web3';

export * from './web3';
export default Web3;
```

and the user would still be able to do

```typescript
const Web3 = require('web3').default;
```

but of course this still results in a breaking change and module augmentation doesn't work

jdevcs: @spacesailor24 Limitation is for module augmentation there must be named export as mentioned in ts docs Module Augmentation part: 
"image"


so If we do named export of Web3 in current CJS settings module augmentation works fine.

As discussed earlier for supporting legacy applications, and not introducing breaking change, I am in favour of keeping current Web3 4x export settings ( Also mentioned by [Muhammad-Altabba](https://github.com/Muhammad-Altabba) ) and there is a workaround for module augmentation for now:

so in user plugin project where web3 augmentation is required: if Web3 is imported and reexported as named, users can augment. 

**web3.ts** in user project
```
import Web3 from 'web3';
export { Web3 };
```

and **pluginConsumer.ts** using named reexported web3 from user's web3.ts
```
import { Web3 } from './web3';
import { Web3Context } from 'web3-core';
import { EthExecutionAPI } from 'web3-types';

import { ContractMethodWrappersPlugin } from './contract_method_wrappers';
import { ERC20TokenAbi } from './ERC20Token';

declare module './web3' {
    interface Web3 extends Web3Context<EthExecutionAPI> {
        contractMethodWrappersPlugin: ContractMethodWrappersPlugin;
    }
}
```

Spacesai1or: @Muhammad-Altabba @jdevcs [This](https://github.com/web3/web3.js/pull/5393/commits/56bf6b7e00bc3f0e21a5aed30525a88c285b46bf) commit adds the work around mentioned above by Junaid

CHANGELOG.md
```diff
@@ -765,12 +769,26 @@ should use 4.0.1-alpha.0 for testing.
...
 #### web3-utils
 
 -   Added and exported three reusable utility functions: `pollTillDefined`, `rejectIfTimeout` and `rejectIfConditionAtInterval` which are useful when dealing with promises that involves polling, rejecting after timeout or rejecting if a condition was met when calling repeatably at every time intervals.
 
 ### Changed
 
+#### web3
+
+-   `Web3` is now a default and named export (#5393)
+
```
Muhammad-Altabba: With the last commits, there are no changes to the `Web3` export, right?
(As there are no changes to the exports `index.ts`. So, there are no changes for the library user, are there?)
```suggestion
```

packages/web3/CHANGELOG.md
```diff
@@ -52,3 +52,9 @@ But this internal behavior is not exposed any further. Though you can achieve sa
 ```ts
 web3.currentProvider.disconnect();
 ```
+
+## [Unreleased]
+
+### Changed
+
+-   `Web3` is now a default and named export (#5393)
```
Muhammad-Altabba: With the last commits, there are no changes to the `Web3` export, right?
(As there are no changes to the exports `index.ts`. So, there are no changes for the library user, are there?)
```suggestion
```

packages/web3/src/index.ts
Muhammad-Altabba: Thanks @jdevcs  for the great workaround.
And thanks @spacesailor24 for the great implementation.
I think the MR is ready to be merged after updating the CHANGELOG.md.

tools/web3-plugin-example/test/unit/contract_method_wrappers.test.ts
```diff
@@ -0,0 +1,125 @@
...
+			});
+		});
+
+		it('should call CustomRpcMethodsPlugin.customRpcMethodWithParameters with expected RPC object', async () => {
+			const expectedGasPrice = '0x1ca14bd70';
+			const expectedTransactionHash =
+				'0xc41b9a4f654c44552e135f770945916f57c069b80326f9a5f843e613491ab6b1';
+
+			requestManagerSendSpy.mockResolvedValueOnce(expectedGasPrice);
+			// Not sure what's being mocked here
```
Muhammad-Altabba: Could we, please, have an issue for investigating the matter and enhancing this piece of code?

Muhammad-Altabba (approved):

tools/web3-plugin-example/test/unit/contract_method_wrappers.test.ts
Spacesai1or: Addressed by [this](https://github.com/web3/web3.js/pull/5393/commits/4375a5d1382a7e1417f6cd223924baf684fa4f99) commit

PR #5394, opened 2022-08-30
Update contract class to not mutate options object

### LINKED ISSUE #5304: the contract methods call function mutates its options argument

### Current Behavior

A simple options argument passed to the call function acquires new keys.

```js
{ from: '0xD28D1f59...' }
```

becomes

```js
{
  from: '0xd28d1f59...',
  data: '0x5e615a6b',
  gasPrice: undefined,
  gas: undefined,
  to: '0xb97da33a...'
}
```

after the call.

### Expected Behavior

This function shouldn't mutate its configuration arguments. This cause problems when you want to reuse them elsewhere.

### Steps to Reproduce

See my writeup at https://karmanivero.us/blog/unexpected-mutations/

### Web3.js Version

1.7.4

### Environment

- Operating System: Windows
- Browser: Chrome (but this is happening on the back end)
- Node.js Version: 18.2.0
- NPM Version: 8.9.0

### Anything Else?

_No response_

Spacesai1or:
closes #5304

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -593,3 +593,4 @@ Released with 1.0.0-beta.37 code base.
 ### Fixed
 -  Browser builds support polyfills (#5031) (#5053) (#4659) (#4767)
 -  Start incrementing jsonrpc.id from random number (#5327)
+-  `web3-eth-contract`'s `call` and `send` methods no longer mutate `options` argument (#5394)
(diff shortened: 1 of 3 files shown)


nikoulai (approved):
Nice! The PR solves the problem, but, should this be happening? Is it behavior we want?

Spacesai1or:
> Nice! The PR solves the problem, but, should this be happening? Is it behavior we want?

@nikoulai 

From my understanding, you should never mutate arguments given to a function (unless that's what the function is specifically designed to do), you should always return the data and let the function caller decide what to do it with it. In the case of `_getOrSetDefaultOptions`, I think you could argue that mutating the `options` objects was be design, but it leads to frustrating behavior for the user that isn't straightforward to debug (as [mentioned](https://karmanivero.us/blog/unexpected-mutations/) by the user whom opened the issue), so I see this as an improvement for the library whether or not it's actually a bug

luu-alex (approved):

PR #5407, opened 2022-08-31
ENS system tests

nikoulai:
## Description

#5039 

## Type of change

System tests for ENS package

## Checklist for 4.x:

- [ ] I have tested my code.


packages/web3-eth-ens/src/abi/ens/ENSRegistry.ts
```diff
@@ -0,0 +1,395 @@
...
+
+web3.js is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+export const ENSRegistryAbi = [
```
jdevcs, line 19: We already have registry https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-eth-ens/src/abi/registry.ts , why do we need this new file?

packages/web3-eth-ens/src/abi/ens/PublicResolver.ts
```diff
@@ -0,0 +1,894 @@
...
+
+web3.js is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+export const PublicResolverAbi = [
```
jdevcs, line 19: Same for resolver, there is already resolver ABI https://github.com/ChainSafe/web3.js/blob/4.x/packages/web3-eth-ens/src/abi/resolver.ts

packages/web3-eth-ens/src/abi/ens/ENSRegistry.ts
nikoulai: For running the tests, I needed the bytecode of the contracts so I compiled the ones mentioned in the headers. Initially, I wanted to be sure that I had the abi of the exactly same version.

However, when I tried to restore to the initial abis (the ones you mention) I was getting an error (TypeError: abis is not iterable). I'll keep investigating the error.

packages/web3-eth-ens/package.json
```diff
@@ -22,11 +22,12 @@
 		"test:ci": "jest --coverage=true --coverage-reporters=json --verbose",
 		"test:watch": "npm test -- --watch",
 		"test:unit": "jest --config=./test/unit/jest.config.js",
-		"test:integration": "jest --config=./test/integration/jest.config.js --passWithNoTests --runInBand",
+		"test:integration": "jest --config=./test/integration/jest.config.js --runInBand --forceExit",
```
avkos: Is it possible to make tests without `--forceExit`?
if you need to close a connection we have a beautiful function `closeOpenConnection` from `system_tests_utils`

packages/web3-eth-ens/src/abi/ens/ENSRegistry.ts
avkos: If you need bytecode only for the tests - please add them to `test/fixtures` folder

avkos (changes requested):

packages/web3-eth-ens/src/abi/ens/DummyNameWrapper.ts
```diff
@@ -0,0 +1,40 @@
...
+
+web3.js is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+export const DummyNameWrapperApi = [
```
Muhammad-Altabba: If you could move this, please, to `packages/web3-eth-ens/test/fixtures` similar to how it is done in other packages. Like for example: `packages/web3-eth-contract/test/fixtures/erc721.ts`.

nikoulai: In this package, the abis are used also in the class. The whole `ens` class is a wrapper of these smart contracts.
 (cc: @avkos )

Muhammad-Altabba: Thanks for the clarification.
And so I hope to have a better name for that. Because the word "Dummy" is not very instructive. Or what do you think?

nikoulai: You are right about that. The reason I gave that name, is to point to the right [contract](https://github.com/ensdomains/ens-contracts/tree/master/contracts/resolvers/mocks) (which I used) and avoid any misunderstandings. What do you think, should I rename it?

packages/web3-eth-ens/package.json
nikoulai: I missed closing some providers, thanks! Fixed!

packages/web3-eth-ens/src/abi/ens/DummyNameWrapper.ts
Muhammad-Altabba: Yes, please, as I think the name Dummy miss-leaded me a little :blush:

Muhammad-Altabba (approved):
Besides the other review comments, the MR looks great to me.

avkos (approved):

packages/web3-eth-ens/test/fixtures/ens/bytecode/PublicResolverBytecode.ts
```diff
@@ -0,0 +1,18 @@
...
+(at your option) any later version.
+
+web3.js is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
```
jdevcs, line 16: we should add link to source code of contract here, from which this byte code was generated. ( for making it easier to debug in future )

packages/web3-eth-ens/test/fixtures/ens/bytecode/NameWrapperBytecode.ts
```diff
@@ -0,0 +1,18 @@
...
+
+web3.js is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+export const NameWrapperBytecode =
```
jdevcs, line 19: we should add link to source code of contract here, from which this byte code was generated. ( for making it easier to debug in future )

packages/web3-eth-ens/test/fixtures/ens/bytecode/ENSRegistryBytecode.ts
```diff
@@ -0,0 +1,18 @@
...
+
+web3.js is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+export const ENSRegistryBytecode =
```
jdevcs, line 19: we should add link of source code of contract here in comments, from which this byte code was generated. ( for making it easier to debug in future )

jdevcs (commented):
LGTM, just one item to be changed:
There are already registry.ts, and resolver.ts having ABIs, we should have one version of each contract ABI instead of two versions of each. Also in ABI + byte code files, add contract source link in comments.

PR #5410, opened 2022-09-01
Merge block tags support 5199

jdevcs:
## Description

https://github.com/ChainSafe/web3.js/issues/5199

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [ ] I have tested my code.


avkos (approved):

luu-alex (approved):

nikoulai (approved):

jdevcs:
Discussion point: `finalized` is not made as default block tag in 1.x as it will be breaking change and also this lib might be used with other POW networks. This and other breaking changes are and will be included in next major bump ( 4.x ), so we can make finalized as default block tag in 4.x. 1.x users can set this via https://web3js.readthedocs.io/en/v1.7.5/web3-eth.html#id14

packages/web3-utils/src/index.js
```diff
@@ -328,36 +328,52 @@ var toChecksumAddress = function (address) {
...
+            return -1;
+        }else {
+            return 1;
+        }
+    } else if(b === "safe") {
+        if( a === "finalized" || a === "latest" ||  a === "pending" ){
+            return -1;
+        }else {
+            return 1;
+        }
```
Muhammad-Altabba: Sorry, I think this condition has a typo. However, the code could never come to this section using our test cases because there are already redundant checks at pervious `if`s.  So, I think we need to rewrite those `if`s that comes before. Or refactor the last 2 `else if`. What do you think?

```suggestion
        if( a === "finalized" || a === "latest" ||  a === "pending" ){
            return 1;
        }else {
            return -1;
        }
```

test/utils.compareBlockNumbers.js
```diff
@@ -37,13 +41,32 @@ const tests = [
...
+    { input: {a: finalized, b: pending}, result: -1 },
+    { input: {b: finalized, a: 0}, result: -1 },
+    { input: {b: finalized, a: new BN(1)}, result: -1 },
+    { input: {b: finalized, a: pending}, result: 1 },
+    // safe
+    { input: {a: safe, b: 0}, result: 1 },
+    { input: {a: safe, b: pending}, result: -1 },
+    { input: {b: safe, a: 0}, result: -1 },
+    { input: {b: safe, a: pending}, result: 1 },
+
```
Muhammad-Altabba, line 73: I think this test case would fail, won't it? Or what is the expected value for such a comparison?
```suggestion
    { input: {b: safe, a: 2}, result: -1 },
```

packages/web3-utils/src/index.js
```diff
@@ -328,36 +328,52 @@ var toChecksumAddress = function (address) {
  * @returns {Number} -1, 0, or 1
  */
 var compareBlockNumbers = function(a, b) {
+    // Increasing order: (genesis = earliest), safe, (finalized ~ latest), pending
```
Muhammad-Altabba, line 331:
```suggestion
    // Increasing order: (genesis = earliest), NUMBERS or safe, (finalized ~ latest), pending
```

packages/web3-utils/src/index.js
```diff
@@ -328,36 +328,52 @@ var toChecksumAddress = function (address) {
...
+    // safe vs block-num cant be compared as block number provided can be on left or right side of safe tag, until safe tag block number is extracted and compared
```
Muhammad-Altabba, line 332: So, what would this method return if `a` or `b` was `safe` and the other one was a number? Throw an exception or return a string `unknown`? Or what do you think?

Muhammad-Altabba (commented):
Looks great!
But I have some questions.

test/utils.compareBlockNumbers.js
jdevcs: This test was added at start of PR after that I think I added that this function doesn't support num vs safe tag in comments.

packages/web3-utils/src/index.js
jdevcs: This list is only for order of block tags .

packages/web3-utils/src/index.js
jdevcs: I think we should throw error in that case.

Muhammad-Altabba: We can do that. But it could be a little too aggressive. What about returning `undefined` as the function is not able to decide?

packages/web3-utils/src/index.js
Muhammad-Altabba: Yes, it just took me a while till I cought where the numbers were exactly ordered in this list. So, I thought about fitting this in the comment somehow :blush:

test/utils.compareBlockNumbers.js
jdevcs: it should fail yes, as function will return undefined now in this case.

packages/web3-utils/src/index.js
jdevcs: now function will return undefined for safe vs non zero num

packages/web3-utils/src/index.js
```diff
@@ -328,36 +328,55 @@ var toChecksumAddress = function (address) {
...
+        }
+    } else if(b === "safe") {
+        if( a === "finalized" || a === "latest" ||  a === "pending" ){
+            return 1;
+        }else if(a === 0){
+            return -1;
+        }else {
+            return undefined;
+        }
+    }
```
Muhammad-Altabba: Actually the previous `if` conditions till this point already handled all the cases  of `0`, `genesis`, `earliest`, `finalized`, `latest`, and `pending`. So, I think we can just simply have the following suggested code:

```suggestion
    } else if(a === "safe" || b === "safe") {
        // either a or b is "safe" and the other one did not fall into any of the conditions above, so the other one is a number
        return undefined;
    }
```

Or what do you think?

test/utils.compareBlockNumbers.js
```diff
@@ -37,20 +41,43 @@ const tests = [
...
+    { input: {a: genesis, b: safe}, result: -1 },
     // latest
     { input: {a: latest, b: 0}, result: 1 },
     { input: {a: latest, b: new BN(1)}, result: 1 },
     { input: {a: latest, b: pending}, result: -1 },
     // pending 
     { input: {a: pending, b: 0}, result: 1 },
     { input: {a: pending, b: new BN(1)}, result: 1 },
+    { input: {a: finalized, b: 0}, result: 1 },
+    { input: {a: finalized, b: new BN(1)}, result: 1 },
```
Muhammad-Altabba, line 56: This is duplicated I think.
```suggestion
```

Muhammad-Altabba (approved):
Looks great!

PR #5426, opened 2022-09-13
Init `web3-test-utils` Package

Spacesai1or:


packages/web3-core/src/web3_context.ts
```diff
@@ -104,6 +104,7 @@ export class Web3Context<
 			| SupportedProviders<API>
 			| Web3ContextInitOptions<API, RegisteredSubs>,
 	) {
+		console.log(providerOrContext)
```
jdevcs: I think its for local debugging so should be removed?

packages/web3-eth-accounts/package.json
```diff
@@ -37,7 +37,8 @@
 		"jest-when": "^3.5.1",
 		"prettier": "^2.7.1",
 		"ts-jest": "^28.0.7",
-		"typescript": "^4.7.4"
+		"typescript": "^4.7.4",
+		"web3-test-utils": "0.1.0-alpha.0"
```
jdevcs, line 41: This will create cyclic dependency? as testing utils will use accounts pkg for account creation.

packages/web3-eth/package.json
```diff
@@ -42,7 +42,8 @@
 		"typescript": "^4.7.4",
 		"web3-eth-abi": "^4.0.1-alpha.0",
 		"web3-eth-accounts": "^4.0.1-alpha.0",
-		"web3-providers-http": "^4.0.1-alpha.0"
+		"web3-providers-http": "^4.0.1-alpha.0",
+		"web3-test-utils": "0.1.0-alpha.0"
```
jdevcs, line 46: same here, cyclic dependency question?

packages/web3-test-utils/src/accounts_utils.ts
```diff
@@ -0,0 +1,78 @@
...
+import { DEFAULT_RETURN_FORMAT } from 'web3-utils';
+
+import { getTestClient, getTestProvider } from './get_env';
+
+export const createAccount = () => {
+    const newAccount = create();
+    return { address: newAccount.address, privateKey: newAccount.privateKey };
+}
+
+const fundGethAccount = async (fundeeAddress: Address, fundAmount: BigInt) => {
```
jdevcs: Mostly logic for funding test accounts will remain same, we just can pass `from` address ( or coinbase in case of geth, or one of 10 pre funded ganache acct) or any provided by user. so can we omit diff implementation for diff nodes.

packages/web3-core/src/web3_context.ts
Spacesai1or: Yes this PR is still a draft and not ready for review

Issue #4235, opened 2021-08-13
EIP-838 Error Codes


TomiOhl:
#### Expected behavior
It would be nice to be able to get the new, custom error codes with their arguments, introduced in Solidity 0.8.4.
Ethers.js already supports them: https://github.com/shapeshift/ethers.js/blob/master/packages/tests/src.ts/test-contract-interface.ts#L636

#### Actual behavior

If I try to catch the error as in the link above, I get the following error message:  
`Error: Returned error: VM Exception while processing transaction: revert`  
Error code, error name, etc. are all undefined

#### Steps to reproduce the behavior

1. Send a transaction to a contract that will revert with errors
2. Try to catch the error

#### Environment

Web3.js v1.5.1
Node v12.18.4
elementary OS 5.1.7

philknows:
This will be addressed in the 4.x rewrite.

frangio:
@philknows Is there a timeline for that? :slightly_smiling_face:

frangio:
Still important. OpenZeppelin Contracts is blocked from adopting custom errors until Web3.js supports them.

frangio:
Not stale! Waiting patiently. 😁

TomiOhl:
> @philknows Is there a timeline for that? slightly_smiling_face

Seems like they plan to release version 4.x in April: [RELEASE.md#long-term-support-lts](https://github.com/ChainSafe/web3.js/blob/1.x/RELEASE.md#long-term-support-lts)

TomiOhl:
gm. Dear bot remove the stale label. 😃

AyoAlfonso:
@TomiOhl is this issue still active

TomiOhl:
Yes, it very much is.Haven't tried the one on the 4.x branch yet, but it isn't addressed in the latest 1.x release afaik

TomiOhl:
Well, 4.x definitely wasn't released in April. Waiting

mconnelly8:
Hey @TomiOhl, we recently released 4.x Alpha last week.  Still working diligently now on our Beta release for later this year.

CruzMolina:
Please @Muhammad-Altabba 🙏🙏🙏

Muhammad-Altabba:
Sorry I got busy with other tasks and was also sick for a couple of days :smiling_face_with_tear:  . However, I will start working on it very soon :blush:...
And nice to see you here @CruzMolina :bouquet:

DinizSa:
[...]

PR #5434, opened 2022-09-13
Decoding EIP-838 error codes

Muhammad-Altabba:
## Description

Implements the logic and a test case for #([4235](https://github.com/web3/web3.js/issues/4235)). However, another MR will be done to add more tests and further improvements.

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [ ] I have tested my code.


Muhammad-Altabba:
The errors in the Pipeline are not related to this MR and they are addressed here: https://github.com/web3/web3.js/issues/5456.

packages/web3-eth-contract/test/integration/contract_methods_errors.test.ts
```diff
@@ -0,0 +1,80 @@
...
+						{ internalType: 'uint256', name: 'value', type: 'uint256' },
+					],
+					name: 'testError1',
+					outputs: [{ internalType: 'bool', name: '', type: 'bool' }],
+					stateMutability: 'pure',
+					type: 'function',
+				},
+			] as const;
+			const contract = new Contract(abi, addr);
+			contract.setProvider('https://ropsten.infura.io/v3/49a0efa3aaee4fd99797bfa94d8ce2f1');
```
nikoulai, line 55: Can't we deploy the contract locally?

packages/web3-eth-contract/src/encoding.ts
```diff
@@ -233,3 +235,27 @@ export const decodeMethodReturn = (abi: AbiFunctionFragment, returnValues?: HexS
...
+			const errorAbi = errorsAbi.find(abi => encodeErrorSignature(abi).startsWith(errorSha));
+
+			if (errorAbi?.inputs) {
+				errorName = errorAbi.name;
+				errorSignature = jsonInterfaceMethodToString(errorAbi);
+				// decode abi.inputs according to EIP-838
+				errorArgs = decodeParameters([...errorAbi.inputs], error.data.substring(10));
+			}
+		} catch (err) {
+			console.error(err);
```
jdevcs, line 255: should throw error if decoding fails for any reason

packages/web3-eth-contract/src/errors.ts
```diff
@@ -1,30 +0,0 @@
...
-GNU Lesser General Public License for more details.
-
-You should have received a copy of the GNU Lesser General Public License
-along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
-*/
-
-import { TransactionReceipt } from 'web3-types';
-import { ERR_CONTRACT, Web3Error } from 'web3-errors';
-
-export class Web3ContractError extends Web3Error {
```
jdevcs, line 21: This is breaking change so it should also be mentioned in changelog of contracts package,

packages/web3-eth-contract/test/integration/contract_methods_errors.test.ts
jdevcs: `ropsten` will be shutdown in Q4 this year, so this test should be done on local geth.

packages/web3-eth-abi/CHANGELOG.md
```diff
@@ -39,7 +39,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-#### web3-eth-abi
+### Added
 
--   Return `BigInt` instead of `string` when decoding function parameters for large numbers, such as `uint256`.
 -   If an error happens when decoding a value, preserve that exception at `innerError` inside the `AbiError`.
+-   Add basic functionality that is used, by `web3-eth-contract`, when decoding error data according to EIP-838.
+
+### Fixed
+
+-   Return `BigInt` instead of `string` when decoding function parameters for large numbers, such as `uint256`.
```
jdevcs: For this and all other changelog entries could you add PR number as: 
```suggestion
-   Return `BigInt` instead of `string` when decoding function parameters for large numbers, such as `uint256`. (#5434)
```

jdevcs (changes requested):

packages/web3-eth-contract/test/integration/contract_methods_errors.test.ts
Muhammad-Altabba: Thanks for telling, I will remove this connection to ropsten, when I write the tests in the next MR. As this MR is for the main logic. And there will be another one for tests that will involve deploying a contract locally also.

packages/web3-eth-contract/src/encoding.ts
Muhammad-Altabba: Actually, the ABI would not be available in all cases. And even it could appear to be conflicting with the raised error. The reason for this is that the smart contract function could call another function(s) that exists in a different smart contract(s), which could throw the error. In this case, the exact ABI for that subsequent call(s) is not available. Furthermore, in such a case, the error that could raise could even have for example the same error name but with different parameters or parameters' types.
For this, we need to be tolerant and just pass the error data even if an exception has been raised while we are trying to do so.

avkos (approved):

jdevcs (approved):

Issue #4900, opened 2022-03-29
`hexToNumber` to check MIN_SAFE_INTEGER


nazarhussain:
`hexToNumber` actually checks if:

```typescript
if (num > Number.MAX_SAFE_INTEGER) {
		return negative ? -num : num;
	}
```

so if we update it to also check for `Number.MIN_SAFE_INTEGER`, than my concern should be covered

_Originally posted by @spacesailor24 in https://github.com/ChainSafe/web3.js/pull/4859#discussion_r837727682_

nikoulai:
Why do we need this check?  `-` is subtracted from the hex string in the beginning of the function, will `num` ever be negative?

jdevcs:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/ChainSafe/web3.js/4900?planning-poker) with ZenHub @avkos @luu-alex @nikoulai @spacesailor24

PR #5440, opened 2022-09-14
check min safe integer in hexToNumber

nikoulai:
## Description

#4900


nikoulai:
I cannot really find any test cases that could lead to this path, because I think `num` will always be positive at this point. I would be great to hear your feedback on this.

avkos (approved):

jdevcs (approved):

Issue #5428, opened 2022-09-13
web3-eth-ens - Unit Test Coverage


mconnelly8:
Current coverage is at 29.82%.  Need to get to above 80%.

PR #5509, opened 2022-10-06
web3 eth ens unit test coverage

nikoulai:
## Description

#5428 Ens unit tests coverage


packages/web3-eth-ens/test/unit/registry.test.ts
```diff
@@ -0,0 +1,870 @@
...
+				} as unknown as NonPayableMethodObject<any, any>);
+
+			await expect(async () => {
+				await registry.getResolver(ENS_NAME);
+			}).rejects.toThrow(new Error());
+			expect(resolverMock).toHaveBeenCalledWith(namehash(ENS_NAME));
+			expect(call).toHaveBeenCalled();
+		});
+	});
+	// it('Doesn"t support interface', async () => {
```
jdevcs: Why these tests are commented? if there is any thing pending `.skip` should be used else these should be removed?

nikoulai: I copied tests from `resolver.test.ts` and worked on them. I forgot to delete them, thanks for the catch!

jdevcs (approved):

Muhammad-Altabba (approved):

Issue #5535, opened 2022-10-13
Fix failing test at: contract › erc721 › contract instance › methods › should award item


Muhammad-Altabba:
The issue is present sometimes at the Pipeline. And it could disappear when re-try.

e.g:https://github.com/web3/web3.js/actions/runs/3223942245/jobs/5274588157

Muhammad-Altabba:
This error came because the value for a `uint256` variable is `undefined` sometimes:
https://github.com/web3/web3.js/actions/runs/3248090366/jobs/5343739424#step:4:1522

```
web3-eth-contract:     {
web3-eth-contract:       message: 'must pass "uint256" validation',
web3-eth-contract:       keyword: 'eth',
web3-eth-contract:       params: { value: undefined },
web3-eth-contract:       instancePath: '/0',
web3-eth-contract:       schemaPath: '#/items/0/eth'
web3-eth-contract:     }
```

However, this issue was not present because the tests used to run in sequence in the past and after they are changed to run in parallel, this issue started to come.

And unfortunately, I am still not sure of the reason. But it seems Oleksii has some clues about it: https://github.com/web3/web3.js/pull/5544

Issue #5523, opened 2022-10-11
4.x CI tests fixes


jdevcs:
- [x] #5535
https://github.com/web3/web3.js/actions/runs/3223942245/jobs/5274588157
- [x] https://github.com/web3/web3.js/issues/5510

Issue #5510, opened 2022-10-06
Cypress Firefox test failing


luu-alex:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

[Running](https://github.com/web3/web3.js/actions/runs/3191896317/jobs/5208797098#step:5:138) into error:
```
web3-providers-http: Still waiting to connect to Firefox, retrying in 1 second (attempt 56/62)
web3-providers-http: Still waiting to connect to Firefox, retrying in 1 second (attempt 57/62)
web3-providers-http: Still waiting to connect to Firefox, retrying in 1 second (attempt 58/62)
web3-providers-http: Still waiting to connect to Firefox, retrying in 1 second (attempt 59/62)
web3-providers-http: Still waiting to connect to Firefox, retrying in 1 second (attempt 60/62)
web3-providers-http: Still waiting to connect to Firefox, retrying in 1 second (attempt 61/62)
web3-providers-http: Still waiting to connect to Firefox, retrying in 1 second (attempt 62/62)
web3-providers-http: Cypress failed to make a connection to Firefox.
web3-providers-http: 
web3-providers-http: This usually indicates there was a problem opening the Firefox browser.
```

### Expected Behavior

Should not be erroring and tests be able to run properly

### Steps to Reproduce

Run a github ci test

### Web3.js Version

4.x

### Environment

- Operating System: 
- Browser: 
- Node.js Version: 
- NPM Version:

PR #5544, opened 2022-10-17
4.x CI tests fixes

avkos:
## Description
#5523

- [x] #5535
https://github.com/web3/web3.js/actions/runs/3223942245/jobs/5274588157
- [x] https://github.com/web3/web3.js/issues/5510 (@luu-alex)

## Type of change

## Checklist for 4.x:

- [ ] I have tested my code.

### DIFF
--- a/packages/web3-eth-contract/test/integration/contract_erc721.test.ts
+++ b/packages/web3-eth-contract/test/integration/contract_erc721.test.ts
@@ -15,7 +15,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { utf8ToHex } from 'web3-utils';
+import { toBigInt, utf8ToHex } from 'web3-utils';
 import { Contract, EventLog } from '../../src';
 import { ERC721TokenAbi, ERC721TokenBytecode } from '../shared_fixtures/build/ERC721Token';
 import {
@@ -111,7 +111,9 @@ describe('contract', () => {
 							pkAccount.privateKey,
 						);
 						const logs = await contractDeployed.getPastEvents('Transfer');
-						const tokenId = (logs[0] as EventLog)?.returnValues?.tokenId as string;
+						const tokenId = toBigInt(
+							(logs[0] as EventLog)?.returnValues?.tokenId as string,
+						);
 						expect(
 							toUpperCaseHex(
 								(await contractDeployed.methods
@@ -137,7 +139,9 @@ describe('contract', () => {
 							pkAccount.privateKey,
 						);
 						const logs = await contractDeployed.getPastEvents('Transfer');
-						const tokenId = (logs[0] as EventLog)?.returnValues?.tokenId as string;
+						const tokenId = toBigInt(
+							(logs[0] as EventLog)?.returnValues?.tokenId as string,
+						);
 						await signAndSendContractMethod(
 							contract.provider,
 							contractDeployed.options.address as string,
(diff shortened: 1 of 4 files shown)


packages/web3-eth-contract/test/integration/contract_erc721.test.ts
```diff
@@ -70,6 +71,15 @@ describe('contract', () => {
 				contractDeployed = await contract.deploy(deployOptions).send(sendOptions);
 			});
 
+			const getTokenId = async (res: Receipt) => {
+				const topic = (res.logs as EventLog[])[0]?.topics[0];
+				const logs = await contractDeployed.getPastEvents('Transfer');
+
+				return toBigInt(
+					(logs.find(l => (l as EventLog).topics.includes(topic)) as EventLog)
```
Muhammad-Altabba: It seems that this error: https://github.com/web3/web3.js/actions/runs/3269305208/jobs/5386153617 happens when the value is undefined. 

So, do you think it is a good solution to try to fetch the past event of 'Transfer' again if  `tokenId` was undefined?

avkos (commented):
in this test, we deploy the contract in the section beforeEach. So it means every new test has a new contract instance. In this case `tokenId` will be `0n` always

avkos (commented):
Also, I've fixed the timeout problem with the IPC provider in syncing event while deploying the contract

avkos:
#5510 is not reproduced. I suggest to close it

packages/web3-eth-contract/test/integration/contract_erc721.test.ts
```diff
@@ -136,8 +133,8 @@ describe('contract', () => {
 							),
 							pkAccount.privateKey,
 						);
-						const logs = await contractDeployed.getPastEvents('Transfer');
-						const tokenId = (logs[0] as EventLog)?.returnValues?.tokenId as string;
+						// eslint-disable-next-line
```
Spacesai1or: The linter should be complaining about this, but you should provide the rule you're intending to disable

packages/web3-eth-contract/test/integration/setup.js
```diff
@@ -19,6 +19,6 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 require('../config/setup');
 
-const jestTimeout = 15000;
+const jestTimeout = String(process.env.WEB3_SYSTEM_TEST_PROVIDER).includes('ipc') ? 25000 : 15000;
```
Spacesai1or, line 22: I would think creating a `JEST_TIMEOUT` (defaulting to `15000`) ENV would be a better approach than this as it's more explicit to the developer that the environment for IPC testing is different

Spacesai1or (changes requested):
Everything else looks good, thank you for solving this issue!

packages/web3-eth-contract/test/integration/contract_erc721.test.ts
avkos: done

Spacesai1or (approved):
We can create an issue to track my [suggestion](https://github.com/web3/web3.js/pull/5544/files#r1000000384) if the team decides to move forward with it

jdevcs (approved):

packages/web3-eth-contract/test/integration/setup.js
avkos: it is only for contract tests that's why I use this solution

PR #5545, opened 2022-10-18
Init `web3-eth-tx-utils` Base

Spacesai1or:
Creates the base for the `web3-eth-tx-utils` package


CHANGELOG.md
```diff
@@ -794,3 +798,4 @@ should use 4.0.1-alpha.0 for testing.
 #### web3-eth
 
 -   Moved the errors' classes from `web3-eth/src/errors.ts` to `web3-errors/src/errors/transaction_errors.ts` (#5462)
+-   Moved transaction related utils from `web3-eth` to the new package `web3-eth-tx-utils` (#5545)
```
spacesailor24, line 801:
```suggestion
```

```suggestion
```

avkos (approved):

packages/web3-eth-tx-utils/CHANGELOG.md
```diff
@@ -0,0 +1,36 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+<!-- EXAMPLE
+
+## [1.0.0]
+
+### Added
```
Muhammad-Altabba, line 12: Could we please have a list of the features that have been added to this initial version of the package?
Thanks,

Muhammad-Altabba: The package is totally empty :)

Muhammad-Altabba (approved):

Muhammad-Altabba:
Do we really need to have this separate package? I mean, why to move the functionality from the sub-package: web3-eth into what seems to me as a sub-sub-package?

Spacesai1or:
> Do we really need to have this separate package? I mean, why to move the functionality from the sub-package: web3-eth into what seems to me as a sub-sub-package?

@Muhammad-Altabba We've discussed this on calls before, but this is necessary to avoid  circular dependency issues for `web3-test-utils`. Additionally there was a user whom request access to some of the tx utils methods that we currently aren't exporting and didn't require the rest of `Web3Eth` (i.e. RPC method wrappers)

jdevcs:
> > Do we really need to have this separate package? I mean, why to move the functionality from the sub-package: web3-eth into what seems to me as a sub-sub-package?
> 
> @Muhammad-Altabba We've discussed this on calls before, but this is necessary to avoid circular dependency issues for `web3-test-utils`. Additionally there was a user whom request access to some of the tx utils methods that we currently aren't exporting and didn't require the rest of `Web3Eth` (i.e. RPC method wrappers)

If one testing method is not applicable due to some constraints, instead of (structural refactors) changing product to be tested, best approach is to change testing method. @spacesailor24

PR #5546, opened 2022-10-18
Init `web3-eth-tx-utils`

Spacesai1or:
### Changed

-   Type of `starterBlockNumber` and `numberOfBlocks` from `number` to `Numbers` for `TransactionBlockTimeoutError` (#5546)

### Added

-   `web3-eth-tx-utils` dependency (#5546)

### Added

-   `web3-eth-tx-utils` dependency (#5546)

### Added

-   `web3-eth-tx-utils` dependency (#5546)

### Removed

-   Moved `defaultTransactionTypeParser`, `detectTransactionType`, `detectRawTransactionType`, `formatTransaction`, `prepareTransactionForSigning`, `getTransactionFromAttr`, `getTransactionNonce`, `getTransactionType`, `defaultTransactionBuilder, `transactionBuilder`exports to`web3-eth-tx-utils` (#5546)
-   Moved exported schemas to `web3-eth-tx-utils` (#5546)
-   Moved exported validation methods to `web3-eth-tx-utils` (#5546)
-   Moved exported `LogsSubscription`, `NewPendingTransactionsSubscription`, `NewHeadsSubscription`, `SyncingSubscription` to `web3-eth-tx-utils` (#5546)

### Added

-   `web3-eth-tx-utils` dependency (#5546)


avkos (approved):

jdevcs:
I think we shouldn't Proceed in direction of refactoring our lib at this point. For having clean testing helper package, I'll suggest to create independent package dependent on 1.x (like we have geth-dev-assistant ). As if a testing methodology is not working instead of changing product being tested , optimal approach is to select testing methodology or tool that best fits product and not the opposite. Also as we have now independent versioning so increasing packages it will be [messy](https://semver.org/#introduction) to go with this approach,

Issue #5470, opened 2022-09-24
eth.getBlock with hydrated = true returning empty transaction objects


AsafY:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

calling eth.getBlock(blockNumber, hydrated) returns a transactions array with empty objects.
![image](https://user-images.githubusercontent.com/890215/192090745-0911abe9-186e-415e-89cc-801fd4c0f743.png)

### Expected Behavior

transactions should return complete transactions objects array

### Steps to Reproduce

call eth.getBlock(15602045, true);

### Web3.js Version

4.0.1-alpha.0

### Environment

- Operating System: Windows 10
- Browser: NA
- Node.js Version: 16.14.2 
- NPM Version: 8.5.0

### Anything Else?

trying to debug,  the results seems to be returning from the RPC call but dropped in the parsing step.

AsafY:
After some more digging I think that the schema parser is not handling "oneOf" correctly.

PR #5556, opened 2022-10-24
eth.getBlock with hydrated = true returning empty transaction objects

nikoulai:
## Description

Solution for #5470 

`eth.getBlock` was returning an empty object for `block.transactions`. 

The error was in https://github.com/web3/web3.js/blob/bfce27fbc27b0db0261e874044c7df5d338023e6/packages/web3-utils/src/formatter.ts#L158 inside `findSchemaByDataPath`, which doesn't handle the `oneOf` type ( and couldn't since no info is passed in order to choose the right one.)

My proposed solution is to save the options of `oneOf` and pass them to `findSchemaByDataPath`.

This was my testcase
```ts
                        const web3 = new Web3Eth(
				'https://mainnet.infura.io/v3/6120b2e72c304b4eadafe2bf33862ac4',
			);
			const blocks = await web3.getBlock(15602045, true);
			console.log(blocks.transactions);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -781,6 +781,10 @@ should use 4.0.1-alpha.0 for testing.
 
 -   Corrected the error code for `JSONRPC_ERR_UNAUTHORIZED` to be `4100` (#5462)
 
+#### web3-eth
+
+-   Fix `getBlock` returning empty transactions object on `hydrated` true (#5556)
+
 #### web3-eth-contract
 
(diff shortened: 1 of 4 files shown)


jdevcs (changes requested):
In existing integration test , there should be additional check of transactions list present in case of hydrated true. or we can add a new test for querying a block with txs on some existing testnet.

luu-alex (approved):

avkos (approved):

jdevcs (approved):

PR #5574, opened 2022-10-28
fix: web3-validator ajv memory leak

### LINKED ISSUE #5515: Memory leak for the `fromWei` utility function for stirng input

### Current Behavior

If the input value for the `fromWei` function is a string type, it leaks the memory and heap continuously increased. 

### Expected Behavior

There should not be a memory leak regardless of the type of input. 

### Steps to Reproduce

```js
// test.js
const { fromWei } = require('web3-utils');

const limit = 2_000_000;

const toKb = num => (num / 1024).toFixed();

new Array(limit).fill('10000').forEach((d, i) => {
	if (i % 10000 === 0) {
		const mem = process.memoryUsage();
		console.log(`heap: ${toKb(mem.heapUsed)}, total: ${toKb(mem.heapTotal)} rss: ${toKb(mem.rss)}`);
	}
	fromWei(d, 'ether');
});
```

Start the node process with following options. 

```bash
NODE_OPTIONS=--max-old-space-size=500 node test.js
```

### Web3.js Version

4.x

### Environment

- Operating System: Mac/Linux
- Node.js Version: 16.6
- NPM Version: 8.11

### Anything Else?

This behaviour was not seen with the 1.x.

mpetrunic:
## Description

Ajv from version 7 onwards uses object references to compile and cache schema. Since web3.js is passing a new object/array for every single validation, this causes every object to be compiled and attached to the schema cache which results in a massive memory leak.

This PR removes that leak by stringifying schemas and using the stringified form as the cache key. But since we now store both string, and compiled schema, memory usage is higher. 

Also, we still have leak if the user starts using one-time validation schemas. Which we would need to remove or they end up permanently attached to the default validator.

resolves #5515

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [x] I have selected the correct 4.x base branch.
- [x] Within the description, the feature or issue is discussed in detail or an issue is linked.   
- [x] I have performed a self-review of my own code
- [x] I ran `yarn` successfully
- [x] I ran `yarn lint` successfully
- [x] I ran `yarn test:unit` successfully
- [ ] I have tested my code.

### DIFF
--- a/packages/web3-validator/src/web3_validator.ts
+++ b/packages/web3-validator/src/web3_validator.ts
@@ -16,6 +16,8 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 import Ajv from 'ajv';
+import { sha256 } from 'ethereum-cryptography/sha256';
+import { utf8ToBytes, toHex } from 'ethereum-cryptography/utils';
 import { ethAbiToJsonSchema } from './utils';
 import { ValidationSchemaInput, Web3ValidationErrorObject, Web3ValidationOptions } from './types';
 import { ethKeyword } from './keywords/eth';
@@ -98,7 +100,7 @@ export class Web3Validator {
 			]);
 		}
 
-		const schemaKey = JSON.stringify(jsonSchema);
+		const schemaKey = toHex(sha256(utf8ToBytes(JSON.stringify(jsonSchema))));
 		if (!this._validator.getSchema(schemaKey)) {
 			this._validator.addSchema(jsonSchema, schemaKey);
 		}
--- a/packages/web3-validator/src/web3_validator.ts
+++ b/packages/web3-validator/src/web3_validator.ts
@@ -98,7 +98,11 @@ export class Web3Validator {
 			]);
 		}
 
-		if (!this._validator.validate(jsonSchema, data)) {
+		const schemaKey = JSON.stringify(jsonSchema);
+		if (!this._validator.getSchema(schemaKey)) {
+			this._validator.addSchema(jsonSchema, schemaKey);
+		}
+		if (!this._validator.validate(schemaKey, data)) {
 			const errors = this._validator.errors as Web3ValidationErrorObject[];
 
 			if (options?.silent) {
--- a/packages/web3-validator/src/web3_validator.ts
+++ b/packages/web3-validator/src/web3_validator.ts
@@ -16,7 +16,7 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 import Ajv from 'ajv';
-import { sha256 } from 'ethereum-cryptography/sha256';
+import { blake2b } from 'ethereum-cryptography/blake2b';
 import { utf8ToBytes, toHex } from 'ethereum-cryptography/utils';
 import { ethAbiToJsonSchema } from './utils';
 import { ValidationSchemaInput, Web3ValidationErrorObject, Web3ValidationOptions } from './types';
@@ -100,7 +100,7 @@ export class Web3Validator {
 			]);
 		}
 
-		const schemaKey = toHex(sha256(utf8ToBytes(JSON.stringify(jsonSchema))));
+		const schemaKey = toHex(blake2b(utf8ToBytes(JSON.stringify(jsonSchema))));
 		if (!this._validator.getSchema(schemaKey)) {
 			this._validator.addSchema(jsonSchema, schemaKey);
 		}


jdevcs (approved):

luu-alex (approved):

packages/web3-validator/src/web3_validator.ts
```diff
@@ -98,7 +98,11 @@ export class Web3Validator {
 			]);
 		}
 
-		if (!this._validator.validate(jsonSchema, data)) {
+		const schemaKey = JSON.stringify(jsonSchema);
```
Muhammad-Altabba: What do you think about using the hash of the `jsonSchema` as the key? 
We can use a simple hashing function that we need to add like this for example:
```javascript
String.prototype.hashCode = function() {
  var hash = 0,
    i, chr;
  if (this.length === 0) return hash;
  for (i = 0; i < this.length; i++) {
    chr = this.charCodeAt(i);
    hash = ((hash << 5) - hash) + chr;
    hash |= 0; // Convert to 32bit integer
  }
  return hash;
}

```

And then the code will be:
```suggestion
		const schemaKey = JSON.stringify(jsonSchema).hashCode()
```

@jdevcs & @luu-alex your recommendation is required. Do you think we should use a simple hashing function as above (that is great for the task)? Or should we utilize `keccak256` from `ethereum-cryptography/keccak` (we would need to add it to the dependencies of web3-validator)? Or should we not using any hash for this?

Thanks all,

Muhammad-Altabba (changes requested):
Thanks for your contribution.
It seems great!
However, I think it would be better to use a hash for the key instead of the full JSON. And I added a comment regarding this.
Thanks,

packages/web3-validator/src/web3_validator.ts
jdevcs: Good suggestion, Thanks @Muhammad-Altabba . We can definitely use hash as key for adding or querying schema in ajv. Currently there is already dependency of eth-cryptography in validator package https://github.com/web3/web3.js/blob/fe037a09e9da7882c5a2e1707a23e778c3558a74/packages/web3-validator/package.json#L31 so I'll suggest to use some existing hash algo ( for avoiding collisions ) instead of simple one.

mpetrunic:
I don't think I caused tests to fail, they seem to be flaky but I cannot rerun them^^

Muhammad-Altabba:
> I don't think I caused tests to fail, they seem to be flaky but I cannot rerun them^^

Sorry @mpetrunic but it seems to be caused by the last update as the pipeline issues keep showing when I re-run it 3 times. And it is only shown after the latest update to use the hash function. And so it would be so nice of you to count for the time of the hash function and to count how many times this hash function is called (to check if there is something wrong with the code that causes unneeded calls for example). 
Thanks,

mpetrunic:
> > I don't think I caused tests to fail, they seem to be flaky but I cannot rerun them^^
> 
> Sorry @mpetrunic but it seems to be caused by the last update as the pipeline issues keep showing when I re-run it 3 times. And it is only shown after the latest update to use the hash function. And so it would be so nice of you to count for the time of the hash function and to count how many times this hash function is called (to check if there is something wrong with the code that causes unneeded calls for example). Thanks,

Do you suggest increasing the timeout on those tests or reverting the last commit? 

I mean converting to bytes, hashing, and converting to hex has some cost and this is quite a hot path as it gets called for every validation.

Muhammad-Altabba:
> > > I don't think I caused tests to fail, they seem to be flaky but I cannot rerun them^^
> > 
> > 
> > Sorry @mpetrunic but it seems to be caused by the last update as the pipeline issues keep showing when I re-run it 3 times. And it is only shown after the latest update to use the hash function. And so it would be so nice of you to count for the time of the hash function and to count how many times this hash function is called (to check if there is something wrong with the code that causes unneeded calls for example). Thanks,
> 
> Do you suggest increasing the timeout on those tests or reverting the last commit?
> 
> I mean converting to bytes, hashing, and converting to hex has some cost and this is quite a hot path as it gets called for every validation.

Many thanks @mpetrunic
And what do you think about using a cheaper hash function like the one called `hashCode` here:
https://github.com/web3/web3.js/pull/5574#discussion_r1010172367

mpetrunic:
> > > > I don't think I caused tests to fail, they seem to be flaky but I cannot rerun them^^
> > > 
> > > 
> > > Sorry @mpetrunic but it seems to be caused by the last update as the pipeline issues keep showing when I re-run it 3 times. And it is only shown after the latest update to use the hash function. And so it would be so nice of you to count for the time of the hash function and to count how many times this hash function is called (to check if there is something wrong with the code that causes unneeded calls for example). Thanks,
> > 
> > 
> > Do you suggest increasing the timeout on those tests or reverting the last commit?
> > I mean converting to bytes, hashing, and converting to hex has some cost and this is quite a hot path as it gets called for every validation.
> 
> Many thanks @mpetrunic And what do you think about using a cheaper hash function like the one called `hashCode` here: [#5574 (comment)](https://github.com/web3/web3.js/pull/5574#discussion_r1010172367)

It generates a 32bit number as the hash. I don't think that has enough collision resistance. And collision here would be a nightmare to debug.

Ideally, we would use md5 or sha1 but they are not available so I switched now to blake2b which should be the fastest.

Muhammad-Altabba (approved):

jdevcs:
Merged this in to 4.x even there is release branch out there as: https://github.com/web3/web3.js/pull/5580#issuecomment-1299012378

Issue #5504, opened 2022-10-04
[4.x] Contract call with tuple is missing param names


mpetrunic:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature Request

Either figure out how to fix output type so that it contains param names

or

https://github.com/web3/web3.js/blob/4.x/packages/web3-eth-contract/src/types.ts#L230
change into `call<O = Outputs>(...): Promise<O>` so people can override return type

### Use Cases

https://github.com/ChainSafe/web3.js-plugin-chainlink/blob/master/src/index.ts#L39

### Are you willing to implement this feature?

- [ ] Yes

PR #5613, opened 2022-11-14
Contract call with tuple is missing param names

avkos:
## Description

#5504 

## Checklist for 4.x:

- [ ] I have tested my code.


packages/web3-eth-contract/src/types.ts
```diff
@@ -227,7 +227,11 @@ export interface NonPayableMethodObject<Inputs = unknown[], Outputs = unknown[]>
 	 * @param block - If you pass this parameter it will not use the default block set with contract.defaultBlock. Pre-defined block numbers as `earliest`, `latest`, and `pending` can also be used. Useful for requesting data from or replaying transactions in past blocks.
 	 * @returns - The return value(s) of the smart contract method. If it returns a single value, it’s returned as is. If it has multiple return values they are returned as an object with properties and indices.
 	 */
-	call(tx?: NonPayableCallOptions, block?: BlockNumberOrTag): Promise<Outputs>;
+
+	call<SpecialOutput = undefined>(
```
mpetrunic: why not just default to Outputs?

mpetrunic (commented):
This works sort of, I was hoping you would solve this: https://github.com/web3/web3.js/blob/4.x/packages/web3-eth-abi/src/types.ts#L193 😜

avkos:
> This works sort of, I was hoping you would solve this: https://github.com/web3/web3.js/blob/4.x/packages/web3-eth-abi/src/types.ts#L193 😜

I've tried to figure it out but seems typescript not supporting dynamic names for the tuples right now.

packages/web3-eth-contract/src/types.ts
avkos: because later I use conditional type and check if it is undefined then use Outputs

mpetrunic: I understand, but wouldn't my suggestion be simpler? 😄

avkos: fixed. thanks

Muhammad-Altabba (dismissed):

mpetrunic:
@avkos Here is an idea https://github.com/web3/web3.js/pull/5624 ^^

nikoulai (approved):

avkos:
> @avkos Here is an idea #5624 ^^

Thanks a lot. I'll investigate it.

jdevcs (commented):
Its causing many tests to fail in https://github.com/web3/web3.js/actions/runs/3490219406/jobs/5841458348#step:4:176 all these tests should be passing before merging PR.

mpetrunic:
> Its causing many tests to fail in https://github.com/web3/web3.js/actions/runs/3490219406/jobs/5841458348#step:4:176 all these tests should be passing before merging PR.

Even though types should be fixed to support that, it makes more sense to use output param name instead of array index in those tests as it's much more type safer.

It might make sense to introduce type only tests for web3-eth-abi package: https://medium.com/front-end-weekly/testing-typescript-types-1595c641d7c7^^

mpetrunic:
@avkos https://github.com/web3/web3.js/pull/5647 adds type tests and support for single output tuple

avkos:
-  add the possibility to see types in the output types

image

- if single output returns exactly one type
image

- for overloaded functions shows all variants of input parameters
image

mpetrunic:
> * add the possibility to see types in the output types
> 
> image
> 
> * if single output returns exactly one type
> 
> image
> 
> * for overloaded functions shows all variants of input parameters
> 
> image

Think you should check my PR as I've extended tests so you can cover all this cases with tests to make sure everything works^^

avkos:
I've fixed the types. Now they work in both ways by index and by name. Also, type tests were added. @mpetrunic

mpetrunic:
> I've fixed the types. Now they work in both ways by index and by name. Also, type tests were added. @mpetrunic

@jdevcs @mconnelly8 Would it make sense to open an issue to create a guide for users explaining how to extract result types from abi (as they will probably need to do something with the result)? Like `type GetBalanceResult = SomethingSomething<typeof abi[1]>`

Spacesai1or (changes requested):
Please add `CHANGELOG.md` entries (for root, `web3-eth-abi`, and `web3-eth-contract`)

avkos:
> Please add `CHANGELOG.md` entries (for root, `web3-eth-abi`, and `web3-eth-contract`)

Changelogs were added

Spacesai1or (approved):

Issue #4016, opened 2021-05-01
One particular block cannot be read over websocket: "CONNECTION ERROR: Provider started to reconnect"


joeytwiddle:
Could we document the various reasons why this error may occur? To help developers who experience this to debug their issue.

We saw this error once in the past, when our light node had crashed/frozen (due to running out of disk space). The solution that time was to bring back the node, and restart our app. (Perhaps discarding the old web3 provider and creating a new one would have been sufficient, instead of a full restart, as suggested in [another issue](https://github.com/ChainSafe/web3.js/issues/1354).)

However today we had it again, and chronically. We could create a new provider and connect, and successfully request the client version and block height. But then, sometimes within 1 second of connecting, we would get this error.

This was happening on both light and fast nodes. Restarting them, and upgrading from 1.10.1 to 1.10.2 didn't help.

We have been using web3 v 1.2.11 for 6 months, but this is the first time seeing this happen. **Edit:** It also happens with web3 v 1.3.5.

We can actually make the following requests without any error:

```ts
      const clientVersion = await web3.eth.getNodeInfo();
      const blockNumber = await web3.eth.getBlockNumber();
```

But the error was triggered when we try to fetch the block data:

```ts
      const blockNumber = 12344370;
      const blockData = await web3.eth.getBlock(blockNumber, true);
```

This was the error we received:

```
Error: CONNECTION ERROR: Provider started to reconnect before the response got received!
at Object.PendingRequestsOnReconnectingError (/app/node_modules/web3-core-helpers/src/errors.js:69:16)
at /app/node_modules/web3-providers-ws/src/index.js:395:37
at Map.forEach (<anonymous>)
at WebsocketProvider.reconnect (/app/node_modules/web3-providers-ws/src/index.js:394:28)
at WebsocketProvider._onClose (/app/node_modules/web3-providers-ws/src/index.js:174:14)
at W3CWebSocket._dispatchEvent [as dispatchEvent] (/app/node_modules/yaeti/lib/EventTarget.js:115:12)
at W3CWebSocket.onClose (/app/node_modules/websocket/lib/W3CWebSocket.js:228:10)
at WebSocketConnection.<anonymous> (/app/node_modules/websocket/lib/W3CWebSocket.js:201:17)
    at WebSocketConnection.emit (events.js:314:20)
at WebSocketConnection.drop (/app/node_modules/websocket/lib/WebSocketConnection.js:475:14)
[...]

Issue #3886, opened 2021-01-26
CONNECTION ERROR: Provider started to reconnect before the response got received!


snitovets:
I looking for workaround how to handle this error within my app. It was a surprise why websocket not store pending request to resolve after reconnection.

/app/node_modules/web3-core-helpers/lib/errors.js:63
return new Error('CONNECTION ERROR: Provider started to reconnect before the response got received!');
 Error: CONNECTION ERROR: Provider started to reconnect before the response got received!
     at Object.PendingRequestsOnReconnectingError (/app/node_modules/web3-core-helpers/lib/errors.js:63:16)
     at /app/node_modules/web3-providers-ws/lib/index.js:332:37
    at Map.forEach (<anonymous>)
     at WebsocketProvider.reconnect (/app/node_modules/web3-providers-ws/lib/index.js:331:28)
     at WebsocketProvider._onClose (/app/node_modules/web3-providers-ws/lib/index.js:149:14)
     at W3CWebSocket._dispatchEvent [as dispatchEvent] (/app/node_modules/yaeti/lib/EventTarget.js:115:12)
     at W3CWebSocket.onClose (/app/node_modules/websocket/lib/W3CWebSocket.js:228:10)
     at WebSocketConnection.<anonymous> (/app/node_modules/websocket/lib/W3CWebSocket.js:201:17)
     at WebSocketConnection.emit (node:events:379:20)
     at WebSocketConnection.EventEmitter.emit (node:domain:470:12)
     at WebSocketConnection.handleSocketClose (/app/node_modules/websocket/lib/WebSocketConnection.js:389:14)
     at TLSSocket.emit (node:events:391:22)
     at TLSSocket.EventEmitter.emit (node:domain:470:12)
     at node:net:667:12
     at TCP.done (node:_tls_wrap:573:7)

_Originally posted by @Snitovets in https://github.com/ChainSafe/web3.js/issues/1354#issuecomment-765899097_

GregTheGreek:
From my understanding, you're hoping that web3js acts stateful and maintains a cache of the web socket events?

snitovets:
Yes. I hoped ws natively caching pending requests. Should I have to manually store pending requests while provider reconnects? And how to handle this error? provider.on('close', ...) ?

GregTheGreek:
[...]

PR #5622, opened 2022-11-15
handled "provider started to reconnect error"

kr-pankil:
## Description
I would like to know the use case of condition: "if (this.responseQueue.size > 0)",  in web3-ws-provider package.
As public RPC nodes are breaking web-socket connections frequetly, we will get the error "provider started to reconnect before response for received.". 
And after reconnect we are getting receipt of pending txs. then what is use of line "_this.responseQueue.delete(key);" 
And this type of error restart our servers or and due to for-each loop we cant handle in our code, so need to handle it from package. 
please correct me if i missed here something.

issues: #3886 and #4016

## Type of change

## Checklist for 1.x:


packages/web3-providers-ws/src/index.js
```diff
@@ -392,8 +392,12 @@ WebsocketProvider.prototype.reconnect = function () {
 
     if (this.responseQueue.size > 0) {
         this.responseQueue.forEach(function (request, key) {
-            request.callback(errors.PendingRequestsOnReconnectingError());
-            _this.responseQueue.delete(key);
+            try {
+                request.callback(errors.PendingRequestsOnReconnectingError());
+            } catch (e) {
+                _this.responseQueue.delete(key);
```
jdevcs, line 398: Once request is sent to eth node , its moved from requestQueue to responseQueue, and with new message/response received its id is matched with responseQueue and result is sent back to user by web3 lib.

This change will not clear `responseQueue` on reconnects in normal flow(if there is no error thrown by callback), so for any requests sent to eth node web3.js will continue keeping requests in queue.

kr-pankil: so with-out this change, if get the error "provider started to reconnect ...." then i should not get the receipt of pending txs right?  as "_this.responseQueue.delete(key);" will clear response queue. but after reconnect i am getting the receipt of tx. i wonder how.

kr-pankil: any update on this ? @jdevcs

jdevcs: fixed in https://github.com/web3/web3.js/pull/5820

PR #5627, opened 2022-11-17
Manually export `web3-core` exports

Spacesai1or:
Replaces use of `export * from 'file'` to:

```typescript
import { x, y, z } from 'file';
export { x, y, z };
```

### DIFF
--- a/packages/web3-core/src/index.ts
+++ b/packages/web3-core/src/index.ts
@@ -14,18 +14,117 @@ GNU Lesser General Public License for more details.
 You should have received a copy of the GNU Lesser General Public License
 along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
+/* eslint deprecation/deprecation: 0 */
 
-export * from './web3_config';
-export * from './web3_request_manager';
-export * from './web3_subscription_manager';
-export * from './web3_subscriptions';
-export * from './web3_context';
-export * from './web3_batch_request';
-export * from './utils';
-export * from './types';
-export * from './formatters';
-export * from './web3_promi_event';
-export * from './web3_event_emitter';
+import { Web3ConfigOptions, Web3ConfigEvent, Web3Config } from './web3_config';
+import { Web3RequestManagerEvent, Web3RequestManager } from './web3_request_manager';
+import { Web3SubscriptionManager } from './web3_subscription_manager';
+import { Web3Subscription, Web3SubscriptionConstructor } from './web3_subscriptions';
+import {
+	Web3ContextObject,
+	Web3ContextInitOptions,
+	Web3ContextConstructor,
+	Web3ContextFactory,
+	Web3Context,
+	TransactionBuilder,
+	Web3PluginBase,
+	Web3EthPluginBase,
+} from './web3_context';
+import { DEFAULT_BATCH_REQUEST_TIMEOUT, Web3BatchRequest } from './web3_batch_request';
+import {
+	isWeb3Provider,
+	isLegacyRequestProvider,
+	isEIP1193Provider,
+	isLegacySendProvider,
+	isLegacySendAsyncProvider,
+	isSupportedProvider,
+	isSupportSubscriptions,
+} from './utils';
+import { TransactionTypeParser } from './types';
+import {
+	inputStorageKeysFormatter,
+	outputProofFormatter,
+	outputBigIntegerFormatter,
+	inputBlockNumberFormatter,
+	inputDefaultBlockNumberFormatter,
+	inputAddressFormatter,
+	txInputOptionsFormatter,
+	inputCallFormatter,
+	inputTransactionFormatter,
+	inputSignFormatter,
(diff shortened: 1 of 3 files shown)


packages/web3-core/src/index.ts
```diff
@@ -14,18 +14,117 @@ GNU Lesser General Public License for more details.
...
+	outputPostFormatter,
+	outputSyncingFormatter,
+	PromiseExecutor,
+	Web3PromiEvent,
+	Web3EventMap,
+	Web3EventKey,
+	Web3EventCallback,
+	Web3Emitter,
+	Web3EventEmitter,
+};
```
Muhammad-Altabba, line 127: First of all, we can use `export` `from`  to avoid having two statements for every module (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#re-exporting_aggregating)

```suggestion
export { Web3ConfigOptions, Web3ConfigEvent, Web3Config } from './web3_config';
export { Web3RequestManagerEvent, Web3RequestManager } from './web3_request_manager';
export { Web3SubscriptionManager } from './web3_subscription_manager';
export { Web3Subscription, Web3SubscriptionConstructor } from './web3_subscriptions';
export {
	Web3ContextObject,
	Web3ContextInitOptions,
	Web3ContextConstructor,
	Web3ContextFactory,
	Web3Context,
	TransactionBuilder,
	Web3PluginBase,
	Web3EthPluginBase,
} from './web3_context';
export { DEFAULT_BATCH_REQUEST_TIMEOUT, Web3BatchRequest } from './web3_batch_request';
export {
	isWeb3Provider,
	isLegacyRequestProvider,
	isEIP1193Provider,
	isLegacySendProvider,
	isLegacySendAsyncProvider,
	isSupportedProvider,
	isSupportSubscriptions,
} from './utils';
export { TransactionTypeParser } from './types';
export {
	inputStorageKeysFormatter,
	outputProofFormatter,
	outputBigIntegerFormatter,
	inputBlockNumberFormatter,
	inputDefaultBlockNumberFormatter,
	inputAddressFormatter,
	txInputOptionsFormatter,
	inputCallFormatter,
	inputTransactionFormatter,
	inputSignFormatter,
	outputTransactionFormatter,
	inputTopicFormatter,
	inputLogFormatter,
	outputLogFormatter,
	outputTransactionReceiptFormatter,
	outputBlockFormatter,
	inputPostFormatter,
	outputPostFormatter,
	outputSyncingFormatter,
} from './formatters';
export { PromiseExecutor, Web3PromiEvent } from './web3_promi_event';
export {
	Web3EventMap,
	Web3EventKey,
	Web3EventCallback,
	Web3Emitter,
	Web3EventEmitter,
} from './web3_event_emitter';
```

However, `export *` and export by name is the same for all cases except if there is a name duplicate or if we need to `export {default as SOMETHING, ...} from ...` (as in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#re-exporting_aggregating).
And if there is an error of `undefined` it should indicate that there are 2 objects having the same name (as we can see in the documentation). So, I suggest investigating what is the exact modules that have the issue. And then we can probably suggest a better solution.
And in all cases, if we really needed to reexport using the names. I suggest doing this only for the problematic object. i.e. to use `export *`  for all modules except for the ones that have the issue. And actually figuring out the problematic object would lead to a better solution, I think.

Or what do you think?

Muhammad-Altabba (changes requested):

avkos (changes requested):
I do not understand the problem. What is the issue we close with these changes?

packages/web3-core/src/index.ts
Spacesai1or: @Muhammad-Altabba These changes were tested by building the package, deploying to a local NPM registry, installing in a test project, building the test project with Webpack, then attempting to run the code within a browser instance

Using `export { x, y, z } 'file'` and `export * from 'file'` appears to not be the same as the doing the import and export separately, as it was still causing the import error that was seen by the original code (`export * from 'file'`)

```bash
Uncaught TypeError: web3_validator_1 is undefined
    js validation.ts:37
    factory react refresh:6
    Webpack 23
validation.ts:37
    js validation.ts:37
    factory react refresh:6
    Webpack 23
```

Additionally, I could test making the change for exporting an exact module, ~but that could potentially just be a band-aid fix as a user could import from any of the modules and potentially receive the same import issue~ - After doing some tests, it seems we can revert the export changes for everything but the `web3_context.ts` exports without receiving the import issue

Spacesai1or:
> I do not understand the problem. What is the issue we close with these changes?

@avkos The issue was discussed in Slack, but I was going to go into detail about this during stand up. Basically, our use of `export * from 'file'` seems to be problematic when a user imports our package, builds with Webpack, then attempts to use the module within the browser. I tried debugging this with multiple fixes, and this solution was the only thing that worked

PR #5640, opened 2022-11-18
Bump @types/bn.js dependency to 5.1.1

wbt:
## Description

This PR sets a higher minimum version for the @types/bn.js dependency.  In version 5.1.1, a new [divmod function was added](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/3cc33671f2ed6208baaf1d819ad4bd825f056560#diff-911dd559f2680937e491afc54d76879c8f7e6624f1fc0549e158271c7d38c33a) which is a breaking change on a patch version number release: a type defined from an import of @types/bn.js@v5.1.1 is not compatible with the BN type if defined from an import of @types/bn.js@v5.1.0.  

Because the current loose specification allows either version to be used, sometimes NPM picks different versions for different web3 package or even for the same package directly vs. when included in another web3 package.  This incompatibility causes code that previously worked fine due to type consistency to now fail compilation.  Pinning all to .0 exactly would be less likely to cause breaking changes elsewhere or in the future, as maintainers of the typing package seem to not recognize this breaking change as such and could do something similar in the future.  However, pinning to .0 would likely cause greater incompatibilities unless all projects which use web3 and have type interaction with the BN type also pin to .0 specifically, and the `^` notation is more common.  As long as everything is consistent, it works, and failing forward makes a little more sense. 

## Type of change

## Checklist for 1.x:

- [n/a] I have made corresponding changes to the documentation.
- [ ] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code. [n/a and ERROR: The filename, directory name, or volume label syntax is incorrect.] 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -592,7 +592,7 @@ Released with 1.0.0-beta.37 code base.
 
 ### Fixed
 -  Browser builds support polyfills (#5031) (#5053) (#4659) (#4767)
--  Update node version on actions to fix breaking mosaic test (#5354) 
+-  Update node version on actions to fix breaking mosaic test (#5354)
 -  Start incrementing jsonrpc.id from random number (#5327)
 -  `web3-eth-contract`'s `call` and `send` methods no longer mutate `options` argument (#5394)
 -  Improvement using provided gas options in web3-eth-accounts for eip-1559 transactions (#5012)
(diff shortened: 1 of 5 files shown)


wbt:
The error is a timeout in connecting to Infura, and I don't (on this project) have privileges to force a re-run of a job with the same code, nor do I have a high opinion of adding extraneous commits to PRs that make it less focused.

wbt:
Again, it's only stale because I don't have permissions on this project to retrigger a job that failed due to an Infura timeout.

jdevcs:
merged https://github.com/web3/web3.js/pull/5885 into 1.x for these changes

PR #5659, opened 2022-11-28
Nikos/5482/implement tests for eip 838 errors with local contract deployment

### LINKED ISSUE #5482: Implement tests for EIP-838 errors with local contract deployment

### Feature Request

As a follow-up to https://github.com/web3/web3.js/issues/4235 implements more tests (and replace the current one).
Additionally,  some cases need to be handled. According to: https://blog.soliditylang.org/2021/04/21/custom-errors/ there are special selectors: `0x82b42900` which is the ‘selector’ for `Unauthorized()` and `0x08c379a0` is the ‘selector’ of `Error(string)`.

### Use Cases

Deploying a smart contract locally enables better testing. And handling for `Unauthorized()` and `Error(string)` would be good as they are special selectors that the user would not need to provide the ABI for.

### Implementation Ideas

_No response_

_No response_

- [X] Yes

nikoulai:
## Description

closes #5482 

## Checklist for 4.x:

- [ ] I have tested my code.


packages/web3-eth-contract/test/integration/contract_methods.test.ts
```diff
@@ -69,11 +75,32 @@ describe('contract', () => {
...
+				itIf(getSystemTestBackend() === 'ganache')(
+					'should returns the expected revert reason string',
+					async () => {
+						// eslint-disable-next-line jest/no-standalone-expect
+						return expect(contractDeployed.methods.reverts().call()).rejects.toThrow(
+							'REVERTED WITH REVERT',
+						);
+					},
+				);
 			});
```
Muhammad-Altabba: I think it is important to make our libarary should behave the same for all providers.

jdevcs (commented):
I think Package-lock.json~1.x shouldn't be part of this PR,

packages/web3-eth/src/types.ts
```diff
@@ -32,6 +32,7 @@ export type SendTransactionEvents<ReturnFormat extends DataFormat> = {
 		latestBlockHash: FormatType<Bytes, ReturnFormat>;
 	};
 	error: TransactionError<FormatType<TransactionReceipt, ReturnFormat>>;
+	contractExecutionError: ContractExecutionError | ResponseError<any>;
```
jdevcs, line 35: Considering high cohesive, if we need to add `contract` execution error in `SendTransactionEvents` type, then this type should be moved to web3-types package instead of keeping in eth package.

packages/web3-eth/src/types.ts
```diff
@@ -45,6 +46,7 @@ export type SendSignedTransactionEvents<ReturnFormat extends DataFormat> = {
...
```
jdevcs, line 49: same here.

jdevcs (changes requested):

packages/web3-eth-contract/test/unit/contract.test.ts
```diff
@@ -145,27 +146,54 @@ describe('Contract', () => {
 			const arg = 'Hello';
 			const contract = new Contract(GreeterAbi);
 
+			// eslint-disable-next-line @typescript-eslint/no-empty-function
```
luu-alex, line 149: do we still need this?

luu-alex (approved):

packages/web3-eth/src/types.ts
nikoulai: `SendTransactionEvents` has a dependency on `FormatType`, imported from `web3-utils`, which is not a dependency of `web3-types`. This would make it difficult to move it. What are your thoughts?

jdevcs (approved):

avkos (approved):

PR #5669, opened 2022-11-30
Remove erroneous set provider code for Contract constructor

### LINKED ISSUE #5433: Setting the provider at the constructor of `Contract` does not always work

### Current Behavior

In some cases, setting the provider at the constructor of `Contract` does not work (possibly there are other related cases when setting it outside the constructor of `Contract`.

```
web3-eth-contract:       ProviderError: Provider not available. Use `.setProvider` or `.provider=` to initialize the provider.
web3-eth-contract:           at Web3RequestManager.<anonymous> (/home/maltabba/repos/web3v4.x/packages/web3-core/src/web3_request_manager.ts:153:10)
web3-eth-contract:           at Generator.next (<anonymous>)
web3-eth-contract:           at /home/maltabba/repos/web3v4.x/packages/web3-core/dist/web3_request_manager.js:8:71
web3-eth-contract:           at new Promise (<anonymous>)
web3-eth-contract:           at Object.<anonymous>.__awaiter (/home/maltabba/repos/web3v4.x/packages/web3-core/dist/web3_request_manager.js:4:12)
web3-eth-contract:           at Web3RequestManager._sendRequest (/home/maltabba/repos/web3v4.x/packages/web3-core/dist/web3_request_manager.js:95:16)
web3-eth-contract:           at Web3RequestManager.<anonymous> (/home/maltabba/repos/web3v4.x/packages/web3-core/src/web3_request_manager.ts:129:31)
web3-eth-contract:           at Generator.next (<anonymous>)
web3-eth-contract:           at /home/maltabba/repos/web3v4.x/packages/web3-core/dist/web3_request_manager.js:8:71
web3-eth-contract:           at new Promise (<anonymous>) {
web3-eth-contract:         code: 600
```

### Expected Behavior

To have the provider set with no errors.

### Steps to Reproduce

```
const addr = '0xbd0B4B009a76CA97766360F04f75e05A3E449f1E';
const abi = [
	{
		inputs: [
			{ internalType: 'address', name: 'addr', type: 'address' },
			{ internalType: 'uint256', name: 'value', type: 'uint256' },
		],
		name: 'TestError1',
		type: 'error',
	},
	{
		inputs: [
			{ internalType: 'bool', name: 'pass', type: 'bool' },
			{ internalType: 'address', name: 'addr', type: 'address' },
			{ internalType: 'uint256', name: 'value', type: 'uint256' },
		],
		name: 'testError1',
		outputs: [{ internalType: 'bool', name: '', type: 'bool' }],
		stateMutability: 'pure',
		type: 'function',
	},
] as const;
const contract = new Contract(abi, addr, {
	provider: 'https://ropsten.infura.io/v3/49a0efa3aaee4fd99797bfa94d8ce2f1',
});
The following like makes it work!
// contract.setProvider('https://ropsten.infura.io/v3/49a0efa3aaee4fd99797bfa94d8ce2f1');

await contract.methods.testError1(false, addr, 42).call(sendOptions);
```

### Web3.js Version

4.x

### Environment

_No response_

### Anything Else?

_No response_

Spacesai1or:
Removes erroneous `addressOrOptionsOrContext !== 'string'` check  in Contract constructor

closes #5433 

When instantiating `Contract` with the following, `addressOrOptionsOrContext` was something other than `string` causing the ternary to use the falsey path setting `undefined` for the `provider`

```typescript
const provider = 'http://127.0.0.1:8545';
const contract = new Contract([], '', {
    provider,
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -935,4 +935,4 @@ should use 4.0.1-alpha.0 for testing.
 
 #### web3-eth-contract
 
--   Erroneous set `provider` logic in Contract constructor (#5669)
+-   Remove erroneous `addressOrOptionsOrContext !== 'string'` check  in Contract constructor (#5669)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -932,3 +932,7 @@ should use 4.0.1-alpha.0 for testing.
 #### web3-eth-accounts
 
 -   These types were moved from `web3-eth-accounts` to `web3-types` package: Cipher, CipherOptions, ScryptParams, PBKDF2SHA256Params, KeyStore (#5581 )
+
+#### web3-eth-contract
+
+-   Erroneous set `provider` logic in Contract constructor (#5669)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -914,6 +914,7 @@ should use 4.0.1-alpha.0 for testing.
 
 -   Emit past contract events based on `fromBlock` when passed to `contract.events.someEventName` (#5201)
 -   Use different types for `ContractOptions` -> `jsonInterface` setter and getter (#5474)
+-   An issue within the `Contract` constructor where `provider` wasn't being set when provided within the `optionsOrContextOrReturnFormat` argument (#5669)
 
 #### web3-types
 
@@ -932,7 +933,3 @@ should use 4.0.1-alpha.0 for testing.
 #### web3-eth-accounts
 
 -   These types were moved from `web3-eth-accounts` to `web3-types` package: Cipher, CipherOptions, ScryptParams, PBKDF2SHA256Params, KeyStore (#5581 )
-
-#### web3-eth-contract
-
--   Remove erroneous `addressOrOptionsOrContext !== 'string'` check  in Contract constructor (#5669)
--- a/packages/web3-eth-contract/CHANGELOG.md
+++ b/packages/web3-eth-contract/CHANGELOG.md
@@ -190,3 +190,7 @@ const transactionHash = receipt.transactionHash;
 
 -   Emit past contract events based on `fromBlock` when passed to `contract.events.someEventName` (#5201)
 -   Use different types for `ContractOptions` -> `jsonInterface` setter and getter (#5474)
+
+### Removed
+
+-   Erroneous set `provider` logic in Contract constructor (#5669)
--- a/packages/web3-eth-contract/CHANGELOG.md
+++ b/packages/web3-eth-contract/CHANGELOG.md
@@ -193,4 +193,4 @@ const transactionHash = receipt.transactionHash;
 
 ### Removed
 
--   Erroneous set `provider` logic in Contract constructor (#5669)
+-   Remove erroneous `addressOrOptionsOrContext !== 'string'` check  in Contract constructor (#5669)
--- a/packages/web3-eth-contract/CHANGELOG.md
+++ b/packages/web3-eth-contract/CHANGELOG.md
@@ -190,7 +190,4 @@ const transactionHash = receipt.transactionHash;
 
 -   Emit past contract events based on `fromBlock` when passed to `contract.events.someEventName` (#5201)
 -   Use different types for `ContractOptions` -> `jsonInterface` setter and getter (#5474)
-
-### Removed
-
--   Remove erroneous `addressOrOptionsOrContext !== 'string'` check  in Contract constructor (#5669)
+-   An issue within the `Contract` constructor where `provider` wasn't being set when provided within the `optionsOrContextOrReturnFormat` argument (#5669)
--- a/packages/web3-eth-contract/src/contract.ts
+++ b/packages/web3-eth-contract/src/contract.ts
@@ -363,18 +363,16 @@ export class Contract<Abi extends ContractAbi>
 		) {
 			provider = addressOrOptionsOrContext.provider;
 		} else if (
-			provider === undefined &&
 			typeof optionsOrContextOrReturnFormat === 'object' &&
 			'provider' in optionsOrContextOrReturnFormat
 		) {
 			provider = optionsOrContextOrReturnFormat.provider;
 		} else if (
-			provider === undefined &&
 			typeof contextOrReturnFormat === 'object' &&
 			'provider' in contextOrReturnFormat
 		) {
 			provider = contextOrReturnFormat.provider;
-		} else if (provider === undefined && Contract.givenProvider !== undefined) {
+		} else {
 			provider = Contract.givenProvider;
 		}
 
--- a/packages/web3-eth-contract/src/contract.ts
+++ b/packages/web3-eth-contract/src/contract.ts
@@ -347,25 +347,40 @@ export class Contract<Abi extends ContractAbi>
 		contextOrReturnFormat?: Web3ContractContext | DataFormat,
 		returnFormat?: DataFormat,
 	) {
+		let contractContext;
+		if (isWeb3ContractContext(addressOrOptionsOrContext)) {
+			contractContext = addressOrOptionsOrContext;
+		} else if (isWeb3ContractContext(optionsOrContextOrReturnFormat)) {
+			contractContext = optionsOrContextOrReturnFormat;
+		} else {
+			contractContext = contextOrReturnFormat;
+		}
+
+		let provider;
+		if (
+			typeof addressOrOptionsOrContext === 'object' &&
+			'provider' in addressOrOptionsOrContext
+		) {
+			provider = addressOrOptionsOrContext.provider;
+		} else if (
+			provider === undefined &&
+			typeof optionsOrContextOrReturnFormat === 'object' &&
+			'provider' in optionsOrContextOrReturnFormat
+		) {
+			provider = optionsOrContextOrReturnFormat.provider;
+		} else if (
+			provider === undefined &&
+			typeof contextOrReturnFormat === 'object' &&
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -932,3 +932,7 @@ should use 4.0.1-alpha.0 for testing.
 #### web3-eth-accounts
 
 -   These types were moved from `web3-eth-accounts` to `web3-types` package: Cipher, CipherOptions, ScryptParams, PBKDF2SHA256Params, KeyStore (#5581 )
+
+#### web3-eth-contract
+
+-   Erroneous set `provider` logic in Contract constructor (#5669)
```
```suggestion
-   Remove erroneous `addressOrOptionsOrContext !== 'string'` check  in Contract constructor (#5669)
```

packages/web3-eth-contract/CHANGELOG.md
```diff
@@ -190,3 +190,7 @@ const transactionHash = receipt.transactionHash;
 
 -   Emit past contract events based on `fromBlock` when passed to `contract.events.someEventName` (#5201)
 -   Use different types for `ContractOptions` -> `jsonInterface` setter and getter (#5474)
+
+### Removed
+
+-   Erroneous set `provider` logic in Contract constructor (#5669)
```
Spacesai1or:
```suggestion
-   Remove erroneous `addressOrOptionsOrContext !== 'string'` check  in Contract constructor (#5669)
```

luu-alex (approved):

packages/web3-eth-contract/CHANGELOG.md
```diff
@@ -190,3 +190,7 @@ const transactionHash = receipt.transactionHash;
 
 -   Emit past contract events based on `fromBlock` when passed to `contract.events.someEventName` (#5201)
 -   Use different types for `ContractOptions` -> `jsonInterface` setter and getter (#5474)
+
+### Removed
+
+-   Remove erroneous `addressOrOptionsOrContext !== 'string'` check  in Contract constructor (#5669)
```
Muhammad-Altabba: Actually the CHANGELOG.MD file is for the web3.js users. We mention what a user should know from his perspective (and for us, the web3.js developers, we put the issue number). 
And actually, this change is a `Fix` from the user's point of view. And how we fixed the issue internally is none of the user's interest. i.e. The user does not care if we added pieces of code or if we remove them. And so what we have to mention here should be in the `Fixed` section. And written to tell the user what he can find as changed... 

> "A changelog is a file that contains a curated, ordered list of notable changes for each versioned release of a project. Its purpose is to make it easier for consumers (and to a lesser extent contributors) to see precisely what changes have been made between two releases."
Ref: https://common-changelog.org/

So, I would suggest:
```suggestion
-   Fixed the `Contract` initialization issue with the string provider (#5669)
```
Or something like (and still under the `Fixed` section):
```suggestion
-   The `Contract` class works well now when initialized with a string provider (#5669)
```

CHANGELOG.md
```diff
@@ -932,3 +932,7 @@ should use 4.0.1-alpha.0 for testing.
 #### web3-eth-accounts
 
 -   These types were moved from `web3-eth-accounts` to `web3-types` package: Cipher, CipherOptions, ScryptParams, PBKDF2SHA256Params, KeyStore (#5581 )
+
+#### web3-eth-contract
+
+-   Remove erroneous `addressOrOptionsOrContext !== 'string'` check  in Contract constructor (#5669)
```
Muhammad-Altabba: Same as the below comment.

packages/web3-eth-contract/test/unit/contract.test.ts
```diff
@@ -72,6 +72,18 @@ describe('Contract', () => {
 
 			expect(contract).toBeInstanceOf(Contract);
 		});
+
+		it('should set the provider upon instantiation', () => {
+			const provider = 'http://127.0.0.1:8545';
```
Muhammad-Altabba: We are better to avoid string constants.

```suggestion
			const provider = getSystemTestProvider();
```
(you would need to import `getSystemTestProvider` from the file `system_test_utils`)

packages/web3-eth-contract/src/contract.ts
```diff
@@ -357,17 +357,15 @@ export class Contract<Abi extends ContractAbi>
...
-					: undefined,
+				(addressOrOptionsOrContext as Web3ContractContext)?.provider ??
+				// eslint-disable-next-line no-nested-ternary
+				(typeof optionsOrContextOrReturnFormat === 'object' &&
+				'provider' in optionsOrContextOrReturnFormat
+					? optionsOrContextOrReturnFormat.provider
+					: typeof contextOrReturnFormat === 'object' &&
+					  'provider' in contextOrReturnFormat
+					? contextOrReturnFormat?.provider
+					: Contract.givenProvider),
```
Muhammad-Altabba: It will be great if you also refactor those to use `if`s to infer the provider object above and then use it here, instead of this Nullish Coalescing operator with the 2 nested ternary operators. As I think the code would then be simpler to read and debug for any issues later. 
(I think eslint prevented nested-ternary for a reason and it is better to not disable this rule :smile: )

This is not in the scope of this issue. But would be nice to be refactored, if you think so.

packages/web3-eth-contract/test/unit/contract.test.ts
Spacesai1or: Addressed by [this](https://github.com/web3/web3.js/pull/5669/commits/16b6bee9ad1efc53db6e4a1427be2b2e78972d78) commit

packages/web3-eth-contract/src/contract.ts
Spacesai1or: Addressed by [this](https://github.com/web3/web3.js/pull/5669/commits/81ec99b1c71a01a8e6eef24a81458c2b0ec0c880) commit. There was a comment from Nazar saying there was a TypeScript error being thrown by not having the `super` call be the very first thing inside the constructor, but I don't seem to receive such an error

packages/web3-eth-contract/CHANGELOG.md
Spacesai1or: Addressed by [this](https://github.com/web3/web3.js/pull/5669/commits/6b1023f6db88959ff4f2d3dda7abc2a522bd3b7d) commit

CHANGELOG.md
Spacesai1or: Addressed by [this](https://github.com/web3/web3.js/pull/5669/commits/6b1023f6db88959ff4f2d3dda7abc2a522bd3b7d) commit

packages/web3-eth-contract/src/contract.ts
Muhammad-Altabba: Thanks @spacesailor24 ,
This is interesting. And when I search for it, I found that the error when not calling `super ` first is: `A 'super' call must be the first statement in the constructor when a class contains initialized properties or has parameter properties`. 
So, there will be no error if there is no properties initialization in the constructor. And it could be that at some point in time there were some properties initialization that had been deleted later. But the comment was not removed :smile:.

packages/web3-eth-contract/src/contract.ts
```diff
@@ -347,25 +347,40 @@ export class Contract<Abi extends ContractAbi>
...
+			provider = optionsOrContextOrReturnFormat.provider;
+		} else if (
+			provider === undefined &&
+			typeof contextOrReturnFormat === 'object' &&
+			'provider' in contextOrReturnFormat
+		) {
+			provider = contextOrReturnFormat.provider;
+		} else if (provider === undefined && Contract.givenProvider !== undefined) {
+			provider = Contract.givenProvider;
+		}
```
Muhammad-Altabba: Actually, `provider === undefined` is always `true` inside those `else if` statements. Since the variable is never being initialized before those conditions. 
Another point is that the last condition for `else if` is useless. And it could be just `else`: 

```suggestion
		} else if (
			typeof optionsOrContextOrReturnFormat === 'object' &&
			'provider' in optionsOrContextOrReturnFormat
		) {
			provider = optionsOrContextOrReturnFormat.provider;
		} else if (
			typeof contextOrReturnFormat === 'object' &&
			'provider' in contextOrReturnFormat
		) {
			provider = contextOrReturnFormat.provider;
		} else {
			provider = Contract.givenProvider;
		}
```

Except if we need to do something if `Contract.givenProvider` also `undefined`:
```suggestion
		} else if (
			typeof optionsOrContextOrReturnFormat === 'object' &&
			'provider' in optionsOrContextOrReturnFormat
		) {
			provider = optionsOrContextOrReturnFormat.provider;
		} else if (
			typeof contextOrReturnFormat === 'object' &&
			'provider' in contextOrReturnFormat
		) {
			provider = contextOrReturnFormat.provider;
		} else if (Contract.givenProvider !== undefined) {
			provider = Contract.givenProvider;
		} else {
		    //... in case there is something we want to do in this case...
		}
```

Spacesai1or: Thank you, I was thinking about this as if they were just if statements

I think 

```typescript
else {
    provider = Contract.givenProvider;
}
```

should be acceptable as I believe that was what the code before the refactor was intending

Spacesai1or: Addressed by [this](https://github.com/web3/web3.js/pull/5669/commits/d1b0b83adb6845ff0b72650048154a9ffb24fd0a) commit

Muhammad-Altabba (approved):

nikoulai (approved):

avkos (approved):

PR #5719, opened 2023-01-04
Let gas value be a number string instead of a number to store very large numbers

rajesh1158:
I faced an issue in BSC chain where I kept getting an error when I called eth_getBlockByNumber with value as 24463105. The error was `Error: Number can only safely store up to 53 bits`. I found the issue to be the 'gas' value being a very large number (19 digits) set in some transactions.

## Fix

To fix the issue, I changed the line 244 in `packages/web3-core-helpers/src/formatters.js`
    tx.gas = utils.hexToNumber(tx.gas);
to
    tx.gas = utils.hexToNumberString(tx.gas);


Spacesai1or:
Hi thank you for your contribution!

Unfortunately we can't make this change to version `1.x` of the library because it's considered a breaking change. I'll chat with the team to see if maybe we can implement a check to test if `gas` will be over max int and cast it to a number string if so, but we're limiting updates to `1.x` to only include serious bug fixes and security updates

Fortunately, the rewrite of the library, version `4.x`, will return BigInts by default but the data return type can be specified by the user to be BigInt, HexString, NumberString, or Number

rajesh1158:
Even in version `4.x`, the function `outputTransactionFormatter` in `packages/web3-core/src/formatters.ts` uses `modifiedTx.gas = hexToNumber(tx.gas);` and not `hexToNumberString`.

Issue #5035, opened 2022-05-17
exception should expose error code


drortirosh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

the RPC API [https://eips.ethereum.org/EIPS/eip-1474] defines error conditions with codes and semi-standard error messages.
The error messages themselves are not standardised, and different nodes return different values.
e.g. if one make a call to an RPC call that might not exist (e.g. eth_feeHistory which was added lately), there is no "canonical" way to know if an exception was generated by "method not found" or some other reason, since the error code (-32601) is not exposed in the exception, and different nodes return different error string.

### Expected Behavior

When processing an RPC error response, in addition to "`message`" and "`data`", the thrown Error object should also include the "`code`" from the rpc response.

### Steps to Reproduce

```js
//the following sample attempt to reference nonexistent method.
// the underlying "-32601" error code is hidden from the application
// the string result differs based on ethereum node version

 Web3 = require( 'web3')
 const web3 = new Web3('http://localhost:8545')
 const web3 = new Web3('https://dai.poa.network')
web3.eth.extend({methods:[{name: 'test', call: 'eth_test'}]})

web3.eth.test()
.catch(e=>console.log(e))
.finally(process.exit)
```

### Web3.js Version

1.7.3

nazarhussain:
Hi @drortirosh Thank for reaching out to us. I think that's a valid point you raised the error code should be included as part of the response. As we are currently we are investing all efforts to the 4.x rewrite, so we will consider this to be done for that release as well.

PR #5735, opened 2023-01-07
5035/exception code

luu-alex:
#5035 
## Description
From this [feature request](https://github.com/web3/web3.js/issues/5035) An Error that happens in the rpc will now create an rpc error which will include the error code, message and jsonrpc. 
It will include standard error codes and unstandard error codes from [eip 1747](https://eips.ethereum.org/EIPS/eip-1474)

Code | Message | Meaning | Category
-- | -- | -- | --
-32700 | Parse error | Invalid JSON | standard
-32600 | Invalid request | JSON is not a valid request object | standard
-32601 | Method not found | Method does not exist | standard
-32602 | Invalid params | Invalid method parameters | standard
-32603 | Internal error | Internal JSON-RPC error | standard
-32000 | Invalid input | Missing or invalid parameters | non-standard
-32001 | Resource not found | Requested resource not found | non-standard
-32002 | Resource unavailable | Requested resource not available | non-standard
-32003 | Transaction rejected | Transaction creation failed | non-standard
-32004 | Method not supported | Method is not implemented | non-standard
-32005 | Limit exceeded | Request exceeds defined limit | non-standard
-32006 | JSON-RPC version not supported | Version of JSON-RPC protocol is not supported | non-standard

<br class="Apple-interchange-newline">

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [ ] I have tested my code.


luu-alex:
Non-standard error codes may confuse users if a node doesn't use the error-codes given by eip 1747. 
For example using local ganache, the error code `-32000` is `method not supported `when in the eip 1747 it is considered a `Invalid input`.
I believe we can put a check on both the errorcode and the error message from the node to make sure that an rpc error is truly following eip 1747. Thoughts?

packages/web3-errors/src/errors/rpc_errors.ts
```diff
@@ -0,0 +1,176 @@
...
+	'-32003': {
+		message: 'Transaction rejected',
+	},
+	'-32004': {
+		message: 'Method not supported',
+	},
+	'-32005': {
+		message: 'Limit exceeded',
+	},
+	'-32006': {
```
jdevcs: I think create dictionary/map with directly using existing error codes ( for avoiding any error code duplication and future maintenance) . 

```
    export const rpcErrorsMap = new Map<number, object>();
    rpcErrorsMap.set(ERR_RPC_INVALID_JSON, ParseError);
    rpcErrorsMap.set(ERR_RPC_INVALID_REQUEST, InvalidRequestError);
    //.....
```

packages/web3-utils/src/json_rpc.ts
```diff
@@ -28,9 +28,71 @@ import {
...
+	MethodNotSupported,
+	LimitExceededError,
+	VersionNotSupportedError,
+	RpcError,
+	rpcErrorValues,
+} from 'web3-errors';
 import { uuidV4 } from './uuid';
 
+export const getRpcError = (code: number) => {
+	const codeStr = code.toString();
```
jdevcs: if we use map as shared above we can avoid these error codes duplication as well and can directly query that map where required as:
`rpcErrorsMap.get(-32700) ` // in web3-core request manager directly

packages/web3-utils/src/json_rpc.ts
```diff
@@ -28,9 +28,71 @@ import {
...
+	if (codeStr === '-32700') {
+		return ParseError;
+	}
+	if (codeStr === '-32600') {
```
jdevcs: There shouldn't be multiple copies of these error codes in web3js lib multiple places.  we should use error_codes.ts.

packages/web3-rpc-methods/src/eth_rpc_methods.ts
```diff
@@ -44,9 +44,16 @@ export async function getSyncing(requestManager: Web3RequestManager) {
...
+	return requestManager.send({
+		method: 'test',
+		params: [],
+	});
+}
+
 export async function getCoinbase(requestManager: Web3RequestManager) {
 	return requestManager.send({
-		method: 'eth_coinbase',
+		method: 'test',
```
jdevcs: this is for debugging only?

jdevcs:
This PR lgtm, but some of tests are not passing in CI, pls check these. Thanks Alex, 
Also create issue for integration/e2e tests of these error codes.

luu-alex:
Thank you @jdevcs, created the issue and still debugging.  Errors were happening as ganache and geth dont follow the [unstandard error codes](https://github.com/web3/web3.js/pull/5735/files#diff-65dc5d05081c69f1865f621e1549f8254caa34650e13756fcbd9a4374eaa9e98R421), they would give different error messages. For example error code `-32000` in Geth is `gas too low`, but according to eip-1474 it is `invalid input`. I created a [check](https://github.com/web3/web3.js/pull/5735/files#diff-90d726c886e913b9fbd11b83e2b4160d084b075f1ca99ab809bf1baa8a743700R330) that error code and message has to match to be an eip 1474 error.

jdevcs:
> Thank you @jdevcs, created the issue and still debugging. Errors were happening as ganache and geth dont follow the [unstandard error codes](https://github.com/web3/web3.js/pull/5735/files#diff-65dc5d05081c69f1865f621e1549f8254caa34650e13756fcbd9a4374eaa9e98R421), they would give different error messages. For example error code `-32000` in Geth is `gas too low`, but according to eip-1474 it is `invalid input`. I created a [check](https://github.com/web3/web3.js/pull/5735/files#diff-90d726c886e913b9fbd11b83e2b4160d084b075f1ca99ab809bf1baa8a743700R330) that error code and message has to match to be an eip 1474 error.

Its because https://eips.ethereum.org/EIPS/eip-1474 EIP is not approved and its not necessary that Eth clients implement this EIP. May be add some flag ( via web3 config ) for turning on this RPC errors feature in web3, instead of keep it always on?

nikoulai (approved):

jdevcs (approved):

Aheyachy (approved):

PR #5739, opened 2023-01-10
Update Builds Directories to Make Use of Both `lib` and `dist`

Spacesai1or:
This PR moves the `tsc` compiled files from `dist/` to `lib/` for all packages that weren't already doing so. Additionally, `web3` and `web3-validator` will compile their web builds to `dist/`. These changes are to address an issue found when developing the [Chainlink plugin](https://github.com/ChainSafe/web3.js-plugin-chainlink) that needs to build it's code for the browser. It was discovered that there was a filename conflict between files such as the `tsc` compiled `web3_validator.ts` and it's webpack compiled version. By putting `tsc` compiled files in `lib/` and webpack compiled files in `dist/`, we completely avoid this issue for all packages, and users are able to pull in published `4.x` Web3.js packages and build using webpack

### DIFF
--- a/packages/web3-core/package.json
+++ b/packages/web3-core/package.json
@@ -2,7 +2,7 @@
 	"name": "web3-core",
 	"version": "4.0.1-alpha.3",
 	"description": "Web3 core tools for sub-packages. This is an internal package.",
-	"main": "dist/index.js",
+	"main": "lib/index.js",
 	"repository": "https://github.com/ChainSafe/web3.js",
 	"author": "ChainSafe Systems",
 	"license": "LGPL-3.0",
--- a/packages/web3-core/package.json
+++ b/packages/web3-core/package.json
@@ -7,7 +7,7 @@
 	"author": "ChainSafe Systems",
 	"license": "LGPL-3.0",
 	"files": [
-		"dist/**/*"
+		"lib/**/*"
 	],
 	"scripts": {
 		"clean": "rimraf dist",
--- a/packages/web3-core/package.json
+++ b/packages/web3-core/package.json
@@ -10,8 +10,8 @@
 		"lib/**/*"
 	],
 	"scripts": {
-		"clean": "rimraf dist",
-		"prebuild": "rimraf dist",
+		"clean": "rimraf dist && rimraf lib",
+		"prebuild": "yarn clean",
 		"build": "tsc --build",
 		"build:check": "node -e \"require('./dist')\"",
 		"lint": "eslint --ext .js,.ts .",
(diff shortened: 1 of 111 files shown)


avkos (approved):

packages/web3-core/package.json
```diff
@@ -2,16 +2,16 @@
...
-		"dist/**/*"
+		"lib/**/*"
 	],
 	"scripts": {
-		"clean": "rimraf dist",
-		"prebuild": "rimraf dist",
+		"clean": "rimraf dist && rimraf lib",
+		"prebuild": "yarn clean",
 		"build": "tsc --build",
 		"build:check": "node -e \"require('./dist')\"",
```
luu-alex: I think we need to update `build:check` as well

packages/web3-core/package.json
```diff
@@ -2,16 +2,16 @@
...
 	"version": "4.0.1-alpha.3",
 	"description": "Web3 core tools for sub-packages. This is an internal package.",
-	"main": "dist/index.js",
+	"main": "lib/index.js",
 	"repository": "https://github.com/ChainSafe/web3.js",
 	"author": "ChainSafe Systems",
 	"license": "LGPL-3.0",
 	"files": [
-		"dist/**/*"
+		"lib/**/*"
```
jdevcs, line 10: For npm whitelisting only `lib` , this will filter out `dist` form including in published lib. 
This will impact CDN as current CDN is configured to pick minified build through npm ( as per Greg info, I discussed with him last time ). 

I personally like it, as it will reduce lib size , so users node_modules will be of less size ( by excluding minified builds) but we will need to configure explicitly CDN for 4.x.

jdevcs (commented):
tsconfig of some of package is not updated ( web3-erros eth-ens iban http ws rpc-methods )?, all packages should have consistent lib output.

Spacesai1or:
> tsconfig of some of package is not updated ( web3-erros eth-ens iban http ws rpc-methods )?, all packages should have consistent lib output.

@jdevcs These packages have a [symlink](https://github.com/web3/web3.js/blob/4.x/packages/web3-errors/tsconfig.json) for the `tsconfig.json`, so updating the template covers these packages

packages/web3-core/package.json
Spacesai1or: Created #5743 to track this

packages/web3-core/package.json
Spacesai1or: This [commit](https://github.com/web3/web3.js/pull/5739/commits/add5c73428b402dee57dd992eead977c73f7b6b8) addresses this

luu-alex (approved):

PR #5754, opened 2023-01-12
Remove `browser` entry from `package.json` and init blackbox browser tests

Spacesai1or:


jdevcs (commented):
@spacesailor24 This is test PR? browser entry change is already merged. so we can close this PR ?

PR #5756, opened 2023-01-13
Nikos/5603/some default properties cannot set from contract class

### LINKED ISSUE #5603: Some default properties cannot be set from `Contract` class

### Current Behavior

During writing some tests for contract `defaults`, noticed that `defaultCommon` and `defaultHardfork` didn't set properly from the `Contract` class and updated to the instance of the class.

### Expected Behavior

The values setted in the `Contract` class should be present in the instance, too.

### Steps to Reproduce

Run the two commented tests in `packages/web3-eth-contract/test/integration/contract_defaults_extra.test.ts`

### Web3.js Version

4.0.1-alpha.1

### Environment

- Operating System:  macOs
- Browser: 
- Node.js Version: v16.14.2
- NPM Version: 8.5.0

### Anything Else?

_No response_

nikoulai:
## Description

closes #5603

## Checklist for 4.x:

- [ ] I have tested my code.


packages/web3-eth-contract/src/contract.ts
```diff
@@ -427,127 +363,299 @@ export class Contract<Abi extends ContractAbi>
...
 
+	/**
+	 * Can be used to set {@link Contract.defaultAccount} for all contracts.
+	 */
+	public static get defaultAccount() {
+		return (this.constructor as typeof Contract).defaultAccount;
+	}
+
+	public static set defaultAccount(value: Address | undefined) {
+		(this.constructor as typeof Contract).defaultAccount = value;
```
jdevcs: Why property is being set at class level in static method if we have a static obj holding these properties as well: contractWeb3Config

packages/web3-eth-contract/src/contract.ts
```diff
@@ -427,127 +363,299 @@ export class Contract<Abi extends ContractAbi>
...
+		if (value === undefined) delete Contract.contractWeb3Config.defaultAccount;
+		else Contract.contractWeb3Config.defaultAccount = value;
+	}
 	public get defaultAccount() {
-		return (this.constructor as typeof Contract).defaultAccount ?? super.defaultAccount;
+		return super.defaultAccount ?? (this.constructor as typeof Contract).defaultAccount;
 	}
 
 	public set defaultAccount(value: Address | undefined) {
 		super.defaultAccount = value;
```
jdevcs: so currently we have defaults at two levels?
1. contracts shared level ( static )
2. global ( coming via inheritance context )

 we should have defaults as:

1. contract instance level ( copy when contract init from global to instance level first time) and always use these first in whole contract. ( change at contract instance level should **NOT** be reflected back to global level )

2. global level ( any change at global level should be reflected down to all contract instance levels )

packages/web3-eth-contract/src/contract.ts
nikoulai: The `contractWeb3Config`was meant to be `private`, just to keep tracked of updated properties and be passed in `setConfig()` in `constructor` during initialization.
I kept the distinct static properties, to keep the same "interface" for the class. But, yes we could keep only `contractWeb3Config`.

packages/web3-eth-contract/src/contract.ts
nikoulai: I believe 1 is fine in the implementation. 

Regarding 2. When an instance is initialized it gets the properties from the shared level 
https://github.com/web3/web3.js/pull/5756/files#diff-7fa8f782e68ce845dbb6e8446b86041697fa427db5485b9c63bf689d0b32ed81R367

However, if an instance is initialized and a static property is updated afterward it is not reflected down. I'm not sure how this could be done without having access to an `EventEmitter` ( I think it's not possible to use one in static methods).

packages/web3-eth-contract/src/contract.ts
jdevcs: If we decide to keep defaults at contracts level shared, I think we should only: 

>  But, yes we could keep only contractWeb3Config.

packages/web3-eth-contract/src/contract.ts
jdevcs: so in that case we can remove **static** from `contractWeb3Config` and just have:

1. instance level defaults ( no upward sync )
2. global level defaults  coming via config  ( with downward sync support in contracts , it will be very easy to create a flag in contract class and make it optional downward sync as new feature for end users)

and can sync down global level defaults just by subscribing `Web3ConfigEvent.CONFIG_CHANGE`  https://github.com/web3/web3.js/blob/498f07764a65e45fda2a902b8b7bd61aa70663c9/packages/web3-core/src/web3_config.ts#L519, contract can subscribe in its constructor to this event and sync its instance level defaults with that.

jdevcs:
```
//in constructor 
if( new_feature_flag)
    super.on(Web3ConfigEvent.CONFIG_CHANGE, event => {
          //sync instance level defaults with globals 
}
```

nikoulai: > so in that case we can remove **static** from `contractWeb3Config` and just have:
> 
> 1. instance level defaults ( no upward sync )
So in the setter functions in `Contract` class, shouldn't use existing implementation
```ts
super.defaultValue = value
```
(Like in defaultHardfork here:https://github.com/web3/web3.js/pull/5756/files#diff-7fa8f782e68ce845dbb6e8446b86041697fa427db5485b9c63bf689d0b32ed81L453)

because it will use the setter of the `config` and it will trigger a `CONFIG_CHANGE` event which will be caught by all `context`/`config` instances.

> 2. global level defaults  coming via config  ( with downward sync support in contracts , it will be very easy to create a flag in contract class and make it optional downward sync as new feature for end users)
> 
> and can sync down global level defaults just by subscribing `Web3ConfigEvent.CONFIG_CHANGE`
> 
> https://github.com/web3/web3.js/blob/498f07764a65e45fda2a902b8b7bd61aa70663c9/packages/web3-core/src/web3_config.ts#L519
> 
> , contract can subscribe in its constructor to this event and sync its instance level defaults with that.

As it is now, we user `super.defaultValue`, so from what I get we are already subscribed.

I see three cases:

- contracts shared level ( static )
- contract instance level 
- global ( coming via inheritance context )

Now there is no difference between global and instance-level variables.
Do you think we should distinguish them?
At the instance level, how we should prioritize them (which value to use for sending transaction etc.) since we could have (two/ or three if we distinguish them)

jdevcs: Thanks @Nikos for looking this in detail. 

yes there should be difference between global and instance level defaults. 
and first priority should be given to instance level in all cases during usage ( sending or preparing txs etc )

There are usecases for this, as a user can have multiple contracts with different defaults ( like default account etc )

packages/web3-eth-contract/src/contract.ts
```diff
@@ -427,44 +384,64 @@ export class Contract<Abi extends ContractAbi>
...
 	}
 
 	public get defaultBlock() {
-		return (this.constructor as typeof Contract).defaultBlock ?? super.defaultBlock;
+		return (
+			super.defaultBlock ??
```
Muhammad-Altabba: Actually, at the time of initialization `super.defaultBlock` will be set to `'latest'`. And therefore, this method will always return that value and will never return the static value except if `this.defaultBlock` was explicitly set to `undefined`.
The same goes for all other properties that was initialized at the parent class to a default value...

packages/web3-eth-contract/src/contract.ts
Muhammad-Altabba: I propose to have 3 places for the configurations:
The first place is to have the default config (that could be read at the start from .env file). This default configuration will have the following values:
```js
{
		handleRevert: false,
		defaultBlock: 'latest',
		transactionBlockTimeout: 50,
		transactionConfirmationBlocks: 24,
		transactionPollingInterval: 1000,
		transactionPollingTimeout: 750 * 1000,
		transactionSendTimeout: 750 * 1000,
		blockHeaderTimeout: 10,
		maxListenersWarningThreshold: 100,
		defaultChain: 'mainnet',
		defaultHardfork: 'london',
		defaultTransactionType: '0x0',
		defaultMaxPriorityFeePerGas: toHex(2500000000),
		enableExperimentalFeatures: {
			useSubscriptionWhenCheckingBlockTimeout: false,
			useRpcCallSpecification: false,
		}
}
```
The second place is the contract static values (all set to `undefined` by default).
The third place is the instance values (all set to `undefined` by default).


Any value of the previous 3 places could be changed. However, when reading the values the priorities are from the third place (instance) then the second (the static), then the first (the default values).
So, when the user would read a property's value, the code would check if the third place (at the instance level) had a value that is different from `undefined`, if yes it will return it. And if not, it will check the value at the second place (at the static properties). If the value is not `undefined` it will return it. If not, the code will reach to the value at the default configuration and return it.

This was there would be no need for listing to `CONFIG_CHANGE` event. And we can enable the user to ignore reading the instance level and use the static ones by a flag variable. Or probably it would be better if we add a function named `resetConfigToDefault(toGlobalDefaults: boolean)` that reset the configurations either to the static values or the global default values based on the passed boolean variable (`toGlobalDefaults`).
Or better to have `resetConfigToDefault()` that just set the values again to `undefined`. And we can have this function at the instance level (at `Web3Config`) and also as static function at `Contract`.

packages/web3-eth-contract/src/contract.ts
Muhammad-Altabba: The suggestion at https://github.com/web3/web3.js/pull/5756#discussion_r1087911768 would resolve this in a good manner, I think.

packages/web3-eth-contract/src/contract.ts
Spacesai1or: A bit irrelevant to the discussion, but I believe the following is a bug:

```typescript
const web3Context = new Web3Context();
web3Context.defaultBlock = 'pending'

const contract = new Contract([], web3Context);
console.log(contract.getConfig().defaultBlock);
// > latest
```

When `Contract` is passed an instance of `Web3Context`, [this config](https://github.com/web3/web3.js/blob/4bbc030ba86f2e5b6494b395d77acce130d90c66/packages/web3-core/src/web3_context.ts#L127) is `undefined` because the expected `Web3Context.config` is actually [_config](https://github.com/web3/web3.js/blob/4bbc030ba86f2e5b6494b395d77acce130d90c66/packages/web3-core/src/web3_config.ts#L68) and `Web3Config.getConfig` is expected to be used to access the `config` object

---

Discussed long ago, the reason why `Web3Config` uses `Web3EventEmitter` is so plugins and users could subscribe to value changes and ultimately decide how to handle them

Here are some of my thoughts/understandings:

1. `Web3Context` is the global control and source for defaults
    1.a The idea with this is that a user instantiates a single instance of `Web3Context` and can pass it to any `web3` package (that can make use of it) to configure it. If each package stores the instance of `Web3Context` as a reference, then all packages will get updates to default properties without having to do anything special
2. If `Contract` is instantiated with an instance of `Web3Context`, then it points to it as it's config and uses it's defaults (receiving updates as mentioned in `1.a`)
    2.a Else it defaults to the same config object `Web3Context` defaults to (configured by `Web3Config` [here](https://github.com/web3/web3.js/blob/4bbc030ba86f2e5b6494b395d77acce130d90c66/packages/web3-core/src/web3_config.ts#L68))
        2.a.1 This default object should probably be pulled into an exported object such as `Web3ConfigOptionsDefaults` from `web3-core/src/web3_config.ts`
3. We remove the `static` properties from `Contract`
4. Replace `Web3Config`'s `getConfig` and `setConfig` with a getter and setter
    4.a This would allow the config object to be accessed like `contract.config` instead of `contract.getConfig()` 

With changes `3` and `4` the following is possible:

Simplest Instantiation of `Contract`:

```typescript
const contract = new Contract(abi)
```

- `contract` has it's own independent `Web3Config` copied from the newly exported `Web3ConfigOptionsDefaults` export (`2.a.1`)
- Default values are only updated by directly setting on instance e.g. `contract.defaultBlock = 'pending'`

---

Instantiation with Global `Web3Context`:

```typescript
const web3Context = new Web3Context();
const contract = new Contract(abi, web3Context);
```

- `contract.config` is a reference to `web3Context.config`
- Any changes to `web3Context`'s properties are automatically reflected in `contract`
- Directly setting values on the instance overrides `web3Context`'s settings, but does not change `web3Context`'s values
    - This decouples the specific property from `web3Context`, all other properties are still references
    - Decoupled properties can be re-coupled via `Web3Context.link`

```typescript
const web3Context = new Web3Context();
web3Context.defaultBlock; // latest

const contract = new Contract(abi, web3Context);
contract.defaultBlock // latest

web3Context.defaultBlock = 'pending';
web3Context.defaultBlock; // pending
contract.defaultBlock // pending

contract.defaultBlock = 'earliest';
web3Context.defaultBlock; // pending
contract.defaultBlock // earliest

web3Context.defaultBlock = 'latest';
web3Context.defaultBlock; // latest
// contract.defaultBlock is now decoupled from web3Context updates, but all other properties are still linked
contract.defaultBlock // earliest
```

---

Linking the Configs of Several Instances of `Contract`

```typescript
const globalWeb3Context = new Web3Context('myprovider');
const configuredWeb3Context = new Web3Context({
    provider: globalWeb3Context.provider,
    config: { defaultBlock: 'earliest' }
});

const contract = new Contract(abi, configuredWeb3Context);
const contract2 = new Contract(abi, configuredWeb3Context);
const contract3 = new Contract(abi, configuredWeb3Context);

const contractWithGlobalContext = new Contract(abi, globalWeb3Context);

contract.defaultBlock // earliest
contract2.defaultBlock // earliest
contract3.defaultBlock // earliest

contractWithGlobalContext.defaultBlock // latest

configuredWeb3Context.defaultBlock = 'pending';

contract.defaultBlock // pending
contract2.defaultBlock // pending
contract3.defaultBlock // pending

contractWithGlobalContext.defaultBlock // latest

globalWeb3Context.defaultBlock = 'earliest'

contract.defaultBlock // pending
contract2.defaultBlock // pending
contract3.defaultBlock // pending

contractWithGlobalContext.defaultBlock // earliest
```

I don't think this is the most user friendly way to do this, but it means we don't have to make any major code changes and reinforces the idea of `Web3Context` being a grouping of desired web3 config details. This method would replace the use of `static` properties in `Contract`, so if a user wants several instances of `Contract` to share config details then they simply create a `Web3Context` configured as desired and pass it to the instances. Because instances of `Contract` will store references to the given `Web3Context`, changes will be updated automatically and the decoupling feature mentioned in `Instantiation with Global Web3Context` section still applies for each instance

packages/web3-eth-contract/src/contract.ts
nikoulai: I tried to gather all changes in class level in `contractWeb3Config` and use `setConfig` during initialization ,so config is updated
https://github.com/web3/web3.js/pull/5756/files#diff-7fa8f782e68ce845dbb6e8446b86041697fa427db5485b9c63bf689d0b32ed81R401

packages/web3-eth-contract/src/contract.ts
nikoulai: sounds good to me!  thanks for the detailed comment @spacesailor24  !
let's hear what the others think (@jdevcs @Muhammad-Altabba ).

jdevcs: Thanks Wyatt and Muhammad .

This PR should add minimum functionality around defaults, so:
- remove static layer of defaults 

and only have:

1. global defaults
2. contract instance defaults

For all new features discussed in this thread like:
- reset defaults function ( by using defaults .ts file instead of .env file )
- decoupling global default with contract instance

@nikoulai please create new issue and add **feature request** label for discussion and adding in 4.x after RC.0.

packages/web3/src/web3.ts
```diff
@@ -92,10 +92,29 @@ export class Web3 extends Web3Context<EthExecutionAPI> {
...
 
+				self.on(Web3ConfigEvent.CONFIG_CHANGE, event => {
+					for (const contract of ContractBuilder._contracts) {
+						contract.emit(Web3ConfigEvent.CONFIG_CHANGE, event);
+					}
+				});
+
+				// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+				this._eventEmitter.on('ContractCloned', (contract: Contract<any>) => {
+					ContractBuilder._contracts.push(contract);
```
jdevcs: contracts list is removed in another PR https://github.com/web3/web3.js/pull/5792/files#diff-55c687488c8bc965581ea1915c7fa86698f5c0d6bf74ae9f41fafc6d0f8281f2L72  , shouldn't we just directly subscribe to change events inside contracts by directly referring super.on and avoid this change in main web3 package

nikoulai: done

jdevcs (approved):
LGTM, just need some minor additions before merging: 
add changelog and some doc comments of new functions it will be very helpful. Thanks @nikoulai

packages/web3-eth-contract/src/contract.ts
Spacesai1or: Created #5801 to track this

packages/web3-eth-contract/src/contract.ts
```diff
@@ -211,69 +210,9 @@ export class Contract<Abi extends ContractAbi>
...
-	/**
-	 * Can be used to set {@link Contract.blockHeaderTimeout} for all contracts.
-	 */
-	public static blockHeaderTimeout?: number;
-
-	/**
-	 * Can be used to set {@link Contract.handleRevert} for all contracts.
-	 */
-	public static handleRevert?: boolean;
+	public static sync_with_globals = false;
```
Spacesai1or: What's the reasoning for this being a `static` property? I believe it would make more sense for it to be a public property for each instance with an option to set the initial value by setting `ContractInitOptions.syncWithContext` upon instantiation

Additionally, we don't use snake casing for properties, only constants

packages/web3-eth-contract/src/contract.ts
```diff
@@ -1265,4 +1088,16 @@ export class Contract<Abi extends ContractAbi>
 			return sub;
 		};
 	}
+
+	protected suscribeToGlobalEvents<T extends Web3Context>(context: T): void {
```
Spacesai1or:
```suggestion
	protected subscribeToGlobalEvents<T extends Web3Context>(context: T): void {
```

Spacesai1or:
I think the following needs to be added to the `constructor` in order to properly sync with a `Web3Context` provided upon instantiation

```typescript
if (contractContext !== undefined) {
    this.suscribeToGlobalEvents(contractContext);
}
```

packages/web3-eth-contract/src/contract.ts
```diff
@@ -1004,7 +826,8 @@ export class Contract<Abi extends ContractAbi>
 			return validatorUtils.transformJsonDataToAbiFormat(abi.inputs ?? [], params);
 		} catch (error) {
 			throw new Web3ContractError(
-				`Invalid parameters for method ${abi.name}: ${(error as Error).message}`,
+				// todo undo formatting, vscode bug
```
spacesailor24, line 829: We should remove this right? And just add it back if we have to work on this file before VSCode fixes the issue?

packages/web3-eth-contract/src/contract.ts
```diff
@@ -1004,7 +826,8 @@ export class Contract<Abi extends ContractAbi>
...
+				`Invalid parameters for method ${abi.name}: ${(error as Error).message}\``,
```
Spacesai1or: We should remove this right? And just add it back if we have to work on this file before VSCode fixes the issue?

```suggestion
				`Invalid parameters for method ${abi.name}: ${(error as Error).message}`,
```

packages/web3/src/web3.ts
```diff
@@ -88,6 +88,17 @@ export class Web3 extends Web3Context<EthExecutionAPI> {
...
 				}
+
+				super.suscribeToGlobalEvents(self);
+			}
+
+			public static set sync_with_globals(value: boolean) {
+				Contract.sync_with_globals = value;
+			}
+
+			public static get sync_with_globals() {
```
Spacesai1or: These shouldn't be snake cased either and I don't think they're needed if we make `syncWithContext` (or `SyncWithGlobals`) a public property of a `Contract` instance configurable via constructor arguments

Spacesai1or (changes requested):
I would also adds tests for `subscribeToGlobalEvents` in `web3-eth-contract` package

And if we decided to rename `sync_with_globals` to `syncWithContext`, this method should be renamed to `subscribeToContextEvents`

packages/web3-eth-contract/src/contract.ts
nikoulai: Being in `ContractInitOptions` isn't sufficient?  Why should we have it in two places( init options and public value)?

nikoulai:
> I think the following needs to be added to the `constructor` in order to properly sync with a `Web3Context` provided upon instantiation
> 
> ```ts
> if (contractContext !== undefined) {
>     this.suscribeToGlobalEvents(contractContext);
> }
> ```

I think `contractContext` is just a copy of the context, which comes from `self.getContextObject()` [here](https://github.com/web3/web3.js/blob/16a8adeb97d486e7f19229234bdd240c3c936874/packages/web3/src/web3.ts#L80) (any case), and subscribing doesn't have the effect we want.

Spacesai1or:
> I think `contractContext` is just a copy of the context, which comes from `self.getContextObject()` [here](https://github.com/web3/web3.js/blob/16a8adeb97d486e7f19229234bdd240c3c936874/packages/web3/src/web3.ts#L80) (any case), and subscribing doesn't have the effect we want.

@nikoulai `contractContext` is not a copy, it's a reference to the `Web3Context` the user is providing

```typescript
const web3Context = new Web3Context('http://127.0.0.1:8545');
web3Context.defaultBlock = 'pending';

const contract = new Contract([], { syncWithContext: true }, web3Context);
console.log(contract.defaultBlock); // latest

web3Context.defaultBlock = 'earliest';

console.log(contract.defaultBlock); // earliest
```
(the `console.log(contract.defaultBlock); // latest` is a bug, as `contract.defaultBlock` should be `pending` as it's supposed to be set by `web3Context`)

The above code works as expected, but only if the following

```typescript
if (contractContext !== undefined) {
    this.suscribeToGlobalEvents(contractContext);
}
```

occurs after the new line:

```typescript
this.syncWithContext = (options as ContractInitOptions)?.syncWithContext ?? false;
```

Spacesai1or:
@nikoulai I created an issue (#5817) to track calling `suscribeToContextEvents` within `Contract`'s `constructor` as the needed fixes are not apart of the scope of the PR

packages/web3-eth-contract/CHANGELOG.md
```diff
@@ -216,8 +216,13 @@ const transactionHash = receipt.transactionHash;
...
 ## [Unreleased]
 
-### Changed
+### Fixed
+
+-   Fix contract defaults (#5756)
+
+### Fixed
 
+-   Fix contract defaults (#5756)
```
Spacesai1or:
```suggestion
### Fixed

-   Fix contract defaults (#5756)

### Changed
```

packages/web3-eth-contract/CHANGELOG.md
```diff
@@ -216,8 +216,11 @@ const transactionHash = receipt.transactionHash;
 
 ## [Unreleased]
 
-### Changed
+### Fixed
 
+-   Fix contract defaults (#5756)
+
+### Changed
```
Spacesai1or, line 223:
```suggestion
### Changed

```

packages/web3-eth-contract/src/contract.ts
```diff
@@ -1259,6 +1084,7 @@ export class Contract<Abi extends ContractAbi>
 				}
 
 				const newContract = this.clone();
+				// newContract.subscribeToContextEvents(this.context as Web3Context);
```
Spacesai1or: Should be removed?

Spacesai1or (approved):
G2g after above comment is addressed

PR #5763, opened 2023-01-17
Make providers WS and IPC reliable

### LINKED ISSUE #5632: Make providers WS and IPC reliable
Make providers reliable:

- [x] Add logic to check if the connection is failed, then try to reconnect automatically.
- [x] Add config for auto reconnect such as: `reconnectTimeout`, `maxAttempts`
- [x] Add tests for these features

avkos:
## Description

fixes #5632 

## Checklist for 4.x:

- [ ] I have tested my code.


packages/web3-providers-ipc/package.json
```diff
@@ -22,7 +22,7 @@
 		"test:ci": "jest --coverage=true --coverage-reporters=json --verbose",
 		"test:watch": "npm test -- --watch",
 		"test:unit": "jest --config=./test/unit/jest.config.js",
-		"test:integration": "jest --config=./test/integration/jest.config.js --passWithNoTests"
+		"test:integration": "jest --config=./test/integration/jest.config.js --passWithNoTests --forceExit"
```
luu-alex: Is `--forceExit` on for debugging?

packages/web3-providers-ipc/test/fixtures/helpers.ts
```diff
@@ -0,0 +1,101 @@
...
+	port: number,
+): Promise<{ pid: number; path: string; close: () => Promise<void> }> => {
+	const prevPid = await getPid(port);
+	if (prevPid > 0) {
+		// close previous server
+		await execPromise(`kill -9 ${prevPid}`);
+	}
+
+	await execPromise(
+		`cd ../../ \n
```
luu-alex: This should probably available as a constant somewhere in the codebase rather than hardcoded. nevermind didnt see this was in helpers.

luu-alex:
Looks good! 👍

packages/web3-providers-ipc/package.json
avkos: Locally it is not stuck, but in GitHub Action, sometimes yes. I need to investigate it deeply

jdevcs: If its not fixed in current PR pls create issue pointing to this discussion.

packages/web3-providers-ipc/src/index.ts
```diff
@@ -41,8 +39,9 @@ export default class IpcProvider<API extends Web3APISpec = EthExecutionAPI> exte
 	private _connectionStatus: Web3ProviderStatus;
 	// Message handlers. Due to bounding of `this` and removing the listeners we have to keep it's reference.
 	protected _socketConnection?: Socket;
-	public constructor(socketPath: string) {
-		super(socketPath);
+
+	public constructor(socketPath: string, options?: object, reconnectOptions?: object) {
```
jdevcs: Is doc updated for this?

packages/web3-providers-ipc/src/index.ts
```diff
@@ -52,19 +51,33 @@ export default class IpcProvider<API extends Web3APISpec = EthExecutionAPI> exte
...
 			this._socketConnection.connect({ path: this._socketPath });
 		} catch (e) {
-			throw new InvalidConnectionError(this._socketPath);
+			if (this.isReconnecting) {
+				setImmediate(() => {
+					this._reconnect();
+				});
+			} else {
+				this._connectionStatus = 'disconnected';
+				throw new InvalidConnectionError(this._socketPath);
```
jdevcs: I think we should also include more error details coming via `e` in `} catch (e) {` to `throw new InvalidConnectionError(this._socketPath)`

jdevcs (changes requested):

packages/web3-providers-ipc/src/index.ts
```diff
@@ -84,17 +97,15 @@ export default class IpcProvider<API extends Web3APISpec = EthExecutionAPI> exte
 	}
 
 	protected _onCloseEvent(event: CloseEvent): void {
-		if (
-			this._reconnectOptions.autoReconnect &&
-			(![1000, 1001].includes(event.code) || !event.wasClean)
-		) {
+		if (!event && this._reconnectOptions.autoReconnect) {
```
jdevcs: There can be normal close events are those considered as well ? 
so this condition should only be true if there is no clean close or event codes are not 1000 and 1001,

packages/web3-providers-ws/src/index.ts
```diff
@@ -76,13 +72,16 @@ export default class WebSocketProvider<
 
 			this._addSocketListeners();
 		} catch (e) {
-			throw new InvalidConnectionError(this._socketPath);
+			if (!this.isReconnecting) {
+				throw new InvalidConnectionError(this._socketPath);
```
jdevcs: pls also include more error details from e obj in rethrow

packages/web3-providers-ipc/src/index.ts
avkos: IPC providers use `import { Socket } from 'net';` to connect. 
in the close event of Socket we do not have any `code` or `wasClean` information.
That is why for IPC we have a different condition.

but when we call the `disconnect` method we create an event on our own. and in this case, we will have an event object here.

packages/web3-providers-ws/src/index.ts
avkos: done

packages/web3-providers-ipc/src/index.ts
avkos: done

packages/web3-providers-ipc/src/index.ts
avkos: ipc provider doesn't have doc

packages/web3-providers-ipc/src/index.ts
jdevcs: ok, so is this tested for normal close, its existing without reconnect?

avkos: yes, this is a test "exit without reconnection":
```ts 
                 it('should emit connect and disconnected events', async () => {
			const server = await startGethServer(8547);
			const web3Provider = new IpcProvider(server.path);
			expect(!!(await waitForEvent(web3Provider, 'connect'))).toBe(true);
			const disconnectPromise = waitForEvent(web3Provider, 'disconnect');
			web3Provider.disconnect();
			expect(!!(await disconnectPromise)).toBe(true);
			await server.close();
		});
```

packages/web3-providers-ipc/src/index.ts
```diff
@@ -52,19 +51,41 @@ export default class IpcProvider<API extends Web3APISpec = EthExecutionAPI> exte
...
+				if (e && (e as Error).message) {
+					throw new ConnectionError(
+						`Error while connecting to ${this._socketPath}. Reason: ${
+							(e as Error).message
+						}`,
+					);
+				} else {
+					throw new InvalidClientError(this._socketPath);
+				}
+			}
```
Muhammad-Altabba: I think the suggestion at https://github.com/web3/web3.js/pull/5763#discussion_r1084016771 would solve the issue of having duplicate code inside 2 sibling classes. As this exact code is also pasted at `WebSocketProvider`.

packages/web3-providers-ipc/src/index.ts
```diff
@@ -52,19 +51,41 @@ export default class IpcProvider<API extends Web3APISpec = EthExecutionAPI> exte
...
-		if (!this._socketConnection || this.getStatus() === 'disconnected') {
-			this._socketConnection = new Socket();
-		}
 		try {
+			if (!this._socketConnection || this.getStatus() === 'disconnected') {
+				this._socketConnection = new Socket();
+			}
 			this._connectionStatus = 'connecting';
 			this._addSocketListeners();
 			this._socketConnection.connect({ path: this._socketPath });
```
Muhammad-Altabba: Is there a reason for not having the same logic also for `WebSocketProvider`? 

And I think it would be better to have only one `connect` at the base class (`SocketProvider`) that has all the logic. And inside it, there would be a call to another method, for example, called `openSocketConnection`, which is another abstract method at the base class that has implementations at the subclasses (`WebSocketProvider` and `IpcProvider`)?

Or what do you think?

packages/web3-providers-ipc/src/index.ts
```diff
@@ -167,4 +164,13 @@ export default class IpcProvider<API extends Web3APISpec = EthExecutionAPI> exte
...
 	}
+
+	protected _onError(event: Error): void {
+		// do not send error while trying to reconnect
+		if (this._reconnectAttempts === 0) {
+			this._eventEmitter.emit('error', event);
+		} else {
+			this._reconnect();
+		}
+	}
```
Muhammad-Altabba: Would we do not need to check for `!this.isReconnecting` before emitting the error as in `SocketProvider`? Can we consolidate and merge the logic from here and from the base class into only one method at the base class (`SocketProvider`)?
Or what do you think?

packages/web3-providers-ws/src/types.ts
```diff
@@ -1,22 +0,0 @@
...
-
-You should have received a copy of the GNU Lesser General Public License
-along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
-*/
-
-export type ReconnectOptions = {
-	autoReconnect: boolean;
-	delay: number;
-	maxAttempts: number;
-};
```
Muhammad-Altabba, line 22: Why delete this type from web3-types and keep the version at  web3-utils? I think it would be better to delete the type definition at web3-utils in order to have all types at web3-types. Or what do you think?

avkos: I've deleted it from web3-provider-ws, because I need this type in IPC and WS provider.

packages/web3-providers-ipc/package.json
avkos: I've fixed the test. So it works without `--forceExit`

packages/web3-providers-ipc/src/index.ts
avkos: moved reconnection logic and errors to common SocketProvider class

packages/web3-providers-ipc/src/index.ts
avkos: moved reconnection logic and errors to common SocketProvider class

packages/web3-providers-ipc/src/index.ts
avkos: moved reconnection logic and errors to common SocketProvider class

packages/web3-providers-ws/src/types.ts
Muhammad-Altabba: Sorry, my bad.

packages/web3-providers-ipc/src/index.ts
jdevcs: ok

jdevcs:
LGTM, Some of WS tests are not invoked in CI, pls check before merging if there is any thing with current PR causing it. @avkos

jdevcs (approved):

luu-alex (approved):

packages/web3-providers-ipc/test/fixtures/helpers.ts
```diff
@@ -0,0 +1,82 @@
...
+	});
+};
+
+export const waitForCloseConnection = async (provider: IpcProvider) => {
+	return new Promise<ProviderRpcError>(resolve => {
+		provider.on('disconnect', ((_error, data) => {
+			resolve(data as unknown as ProviderRpcError);
+		}) as Web3ProviderEventCallback<ProviderRpcError>);
+	});
+};
```
Muhammad-Altabba: What do you think about moving those 2 helper methods along with `waitForEvent` to `web3-utils` so we can have one version of them (currently they are duplicated at `web3-providers-ws`). In addition to removing the code douplication, those methods could be used later to write unit test for `SocketProvider`. Or what do you think?

packages/web3-providers-ws/test/integration/reconnection.test.ts
```diff
@@ -0,0 +1,149 @@
...
+				getSystemTestProvider(),
+				{},
+				reconnectionOptions,
+			);
+			// @ts-expect-error-next-line
+			expect(web3Provider._reconnectOptions).toEqual(reconnectionOptions);
+			await waitForOpenConnection(web3Provider);
+			web3Provider.disconnect(1000, 'test');
+			await waitForCloseConnection(web3Provider);
+		});
```
Muhammad-Altabba: Maybe the logic of those 2 tests could be moved to `web3-utils`. We just need to keep the Provider initialization. And the rest is douplicate with IpcProvider. Or what do you think?

nikoulai (approved):

scripts/geth.sh
```diff
@@ -9,18 +9,17 @@ helpFunction() {
...
+		echo "docker run -d -p $WEB3_SYSTEM_TEST_PORT:$WEB3_SYSTEM_TEST_PORT ethereum/client-go:stable --nodiscover --nousb --ws --ws.addr 0.0.0.0 --ws.port $WEB3_SYSTEM_TEST_PORT --http --http.addr 0.0.0.0 --http.port $WEB3_SYSTEM_TEST_PORT --allow-insecure-unlock --http.api personal,web3,eth,admin,debug,txpool,net --ws.api personal,web3,eth,admin,debug,miner,txpool,net --dev"
+        docker run -d -p $WEB3_SYSTEM_TEST_PORT:$WEB3_SYSTEM_TEST_PORT ethereum/client-go:stable --nodiscover --nousb --ws --ws.addr 0.0.0.0 --ws.port $WEB3_SYSTEM_TEST_PORT --http --http.addr 0.0.0.0 --http.port $WEB3_SYSTEM_TEST_PORT --allow-insecure-unlock --http.api personal,web3,eth,admin,debug,txpool,net --ws.api personal,web3,eth,admin,debug,miner,txpool,net --dev
 		echo "Waiting for geth..."
-		npx wait-port "$WEB3_SYSTEM_TEST_PORT"
+		npx wait-port -t 10000 "$WEB3_SYSTEM_TEST_PORT"
```
Muhammad-Altabba, line 22: I think this `-t 10000` is not needed, right?

scripts/geth.sh
```diff
@@ -9,18 +9,17 @@ helpFunction() {
...
-
 	if [ -z "${ORIGARGS[1]}" ]
 	then
 		echo "Starting geth..."
-		echo "docker run -p $WEB3_SYSTEM_TEST_PORT:$WEB3_SYSTEM_TEST_PORT --name web3-geth-client ethereum/client-go --nodiscover --nousb --ws --ws.addr 0.0.0.0 --ws.port $WEB3_SYSTEM_TEST_PORT --http --http.addr 0.0.0.0 --http.port $WEB3_SYSTEM_TEST_PORT --allow-insecure-unlock --http.api personal,web3,eth,admin,debug,txpool,net --ws.api personal,web3,eth,admin,debug,miner,txpool,net --dev"
-        docker run -p $WEB3_SYSTEM_TEST_PORT:$WEB3_SYSTEM_TEST_PORT ethereum/client-go --nodiscover --nousb --ws --ws.addr 0.0.0.0 --ws.port $WEB3_SYSTEM_TEST_PORT --http --http.addr 0.0.0.0 --http.port $WEB3_SYSTEM_TEST_PORT --allow-insecure-unlock --http.api personal,web3,eth,admin,debug,txpool,net --ws.api personal,web3,eth,admin,debug,miner,txpool,net --dev
+		echo "docker run -p $WEB3_SYSTEM_TEST_PORT:$WEB3_SYSTEM_TEST_PORT ethereum/client-go:stable --nodiscover --nousb --ws --ws.addr 0.0.0.0 --ws.port $WEB3_SYSTEM_TEST_PORT --http --http.addr 0.0.0.0 --http.port $WEB3_SYSTEM_TEST_PORT --allow-insecure-unlock --http.api personal,web3,eth,admin,debug,txpool,net --ws.api personal,web3,eth,admin,debug,miner,txpool,net --dev"
```
Muhammad-Altabba, line 15: I think having a name for the docker container is something good. I mean to keep `--name web3-geth-client `

Muhammad-Altabba (commented):
Looks great!
However, I have a few comments.

scripts/geth.sh
avkos: It is better to have this option. Because in this case, if something happens with docker, our action will not stick forever. It just waits 10 sec and then start the test and will fail

scripts/geth.sh
avkos: let's make changes in docker in another PR

packages/web3-providers-ipc/test/fixtures/helpers.ts
avkos: done

packages/web3-providers-ws/test/integration/reconnection.test.ts
avkos: we can have tests here and in web3-utils just to be sure it works not only on base class but in WebsocketProvider and in IPCProvider

scripts/geth.sh
Muhammad-Altabba: Thanks for the clarification!

packages/web3-providers-ws/test/integration/reconnection.test.ts
Muhammad-Altabba: Yes, I meant to have only the provider initialization at this package: 

```
const web3Provider = new WebSocketProvider(
				getSystemTestProvider(),
				{},
				reconnectionOptions,
			);
```

And then the object `web3Provider` could be passed to a function that contains the logic (that is exactly the same for both WS and IPC):
```
			// @ts-expect-error-next-line
			expect(web3Provider._reconnectOptions).toEqual(reconnectionOptions);
			await waitForOpenConnection(web3Provider);
			web3Provider.disconnect(1000, 'test');
			await waitForCloseConnection(web3Provider);
```

scripts/geth.sh
Muhammad-Altabba: Yes, for this I recommend not changing the docker configuration related to the docker image name (`--name web3-geth-client`) in this MR :smile:

Muhammad-Altabba (approved):
Great work Oleksii,
I have just 2 minor comments that you may consider. And you may merge without considering them.

PR #5764, opened 2023-01-17
Add web3-core requestManager and subscriptionManager docs

### LINKED ISSUE #5183: Web3-Types and Web3-Core Documentation - Methods
A lot of behavior and properties are inherited from the `Web3Context` and few others to the `contracts`, `eth` and `web3` package. 

We need to document those behavior and properties from `web3-types` and `web3-core` packages. These docs must cover:

1. Request related methods from `RequestManager`
2. Subscription related methods from `SubscriptionManager`

nikoulai:
## Description

closes #5183 

## Type of change
       Documentation

## Checklist for 4.x:

- [ ] I have tested my code.


packages/web3-core/src/web3_subscription_manager.ts
```diff
@@ -114,7 +141,11 @@ export class Web3SubscriptionManager<
 		this._subscriptions.delete(id);
 		return id;
 	}
-
+	/**
+	 * Will unsubscribe all subscriptions that fullfil the condition
```
jdevcs:
```suggestion
	 * Will unsubscribe all subscriptions that fulfill the condition
```

packages/web3-core/src/web3_subscription_manager.ts
```diff
@@ -114,7 +141,11 @@ export class Web3SubscriptionManager<
...
+	 * @param condition A functions that acces and `id` and a `subscription` and return `true` or `false`
```
jdevcs:
```suggestion
	 * @param condition : A function that access and `id` and a `subscription` and return `true` or `false`
```

packages/web3-core/src/web3_subscription_manager.ts
```diff
@@ -114,7 +141,11 @@ export class Web3SubscriptionManager<
...
+	 * @returns An array of all the un-subscribed subcriptions
```
jdevcs:
```suggestion
	 * @returns An array of all the un-subscribed subscriptions
```

packages/web3-core/src/web3_subscription_manager.ts
```diff
@@ -126,10 +157,17 @@ export class Web3SubscriptionManager<
...
+	/**
+	 * Clears all subscriptions
+	 */
 	public clear() {
 		this._subscriptions.clear();
 	}
 
+	/**
+	 *
+	 * @returns `true` or `false` depending on if the current providers supports subscriptions
```
jdevcs:
```suggestion
	 * @returns `true` or `false` depending on if the current provider supports subscriptions
```

jdevcs (changes requested):
looks good to me, just need some minor changes.

luu-alex (approved):

nikoulai:
> looks good to me, just need some minor changes.

thanks for catching the typos :)

avkos (approved):

jdevcs (approved):

Muhammad-Altabba (changes requested):
LGTM,
However, a few points...

- Could you please add comments for every related public member at Web3Context (https://github.com/web3/web3.js/pull/5764/files#diff-17bb8cbf6a667d7c0c6c3657ddf941786b34ade5a0c0e1962b7ca9979006e923R164 , https://github.com/web3/web3.js/pull/5764/files#diff-17bb8cbf6a667d7c0c6c3657ddf941786b34ade5a0c0e1962b7ca9979006e923R345= and https://github.com/web3/web3.js/pull/5764/files#diff-17bb8cbf6a667d7c0c6c3657ddf941786b34ade5a0c0e1962b7ca9979006e923R340)
- Could you please add a description (in addition to @params if applicable) for every related method?
- Follow the standard (https://tsdoc.org/) for things like the description first and then the @params. And the param's name is followed by `-`.  And when you put an example use `@example`... 
- If a method is do the same as another method (when it is an alternative alias) or just call another method (when it is a wrapper), I suggest to mention that to ease things for the user. Especially, since there are already too many methods for related to provider setting and getting...
Thanks,

packages/web3-core/src/web3_context.ts
```diff
@@ -250,22 +253,85 @@ export class Web3Context<
...
+	 *
+	 * ```ts
+	 * const web3Context = new Web3Context("http://localhost:8545");
+	 * console.log(web3Context.provider);
+	 * > HttpProvider {
+	 * 	clientUrl: 'http://localhost:8545',
+	 * 	httpProviderOptions: undefined
+	 *  }
+	 * ```
+	 */
```
Muhammad-Altabba: I think it is good to tell the user that this method just does the same as the getter `provider()`.

And I think that it would be nice from you, if you call that getter inside this method and then mention that this mentioned is just a wrapper.

packages/web3-core/src/web3_context.ts
```diff
@@ -250,22 +253,85 @@ export class Web3Context<
...
+	 * web3Context.currentProvider = "ws://localhost:8545";
+	 * console.log(web3Context.provider);
+	 * > WebSocketProvider {
+	 * _eventEmitter: EventEmitter {
+	 * _events: [Object: null prototype] {},
+	 * _eventsCount: 0,
+	 * ...
+	 * }
+	 * ```
+	 */
```
Muhammad-Altabba: Same as above but for the setter: `provider(...)`

packages/web3-core/src/web3_context.ts
```diff
@@ -250,22 +253,85 @@ export class Web3Context<
...
+	 *  }
+	 * ```
+	 */
+
 	public get provider(): SupportedProviders<API> | string | undefined {
 		return this.requestManager.provider;
 	}
 
+	/**
+	 * Will set the current provider.
```
Muhammad-Altabba, line 276: Could you please add `@param` description for every parameter in this method and in all other methods that is related to RequestManager and SubscriptionManager. Including the methods that already has comments but does not have any parameter description?
Thanks,

packages/web3-core/src/web3_subscription_manager.ts
```diff
@@ -114,7 +141,11 @@ export class Web3SubscriptionManager<
 		this._subscriptions.delete(id);
 		return id;
 	}
-
+	/**
+	 * Will unsubscribe all subscriptions that fulfill the condition
+	 * @param condition : A function that access and `id` and a `subscription` and return `true` or `false`
```
Muhammad-Altabba:
```suggestion
	 * @param condition - A function that access and `id` and a `subscription` and return `true` or `false`
```
Always use `-` after the parameter name, please. Actually, we got a lint warning for every missed dash. Please apply this to all other methods' params as well...

packages/web3-core/src/web3_subscription_manager.ts
```diff
@@ -75,10 +93,19 @@ export class Web3SubscriptionManager<
...
+	 */
 	public get subscriptions() {
 		return this._subscriptions;
 	}
 
+	/**
+	 *
+	 * @param sub A {@link Web3Subscription} objects
+	 *
+	 * Adds an instance of {@link Web3Subscription} and subscribes to it
```
Muhammad-Altabba:
```suggestion
	 * Adds an instance of {@link Web3Subscription} and subscribes to it
	 *
	 * @param sub - A {@link Web3Subscription} objects
```
We need to put the description and then the list of the `@param`s.

packages/web3-core/src/web3_subscription_manager.ts
```diff
@@ -126,10 +157,17 @@ export class Web3SubscriptionManager<
...
 
+	/**
+	 * Clears all subscriptions
+	 */
 	public clear() {
 		this._subscriptions.clear();
 	}
 
+	/**
+	 *
```
Muhammad-Altabba, line 179:
```suggestion
	 * Check whether the current provider supports subscriptions.
	 *
```
Could we, please, have a description for all methods even if it is simple? This would give the reader a better experience, I think.

avkos (approved):

Muhammad-Altabba (approved):

PR #5769, opened 2023-01-20
Nikos/5451/script parses changelog.md files and updates

### LINKED ISSUE #5451: Write a script that parses all the CHANGELOG.md files exist in each package and updates the one on the root

### What's up?

Would be useful to automate the update of the main `CHANGELOG.md` file, based on the updates of each `CHANGELOG.md` in each package.
As suggested here: https://github.com/web3/web3.js/pull/5435#discussion_r972574314

### Examples/References

_No response_

nikoulai:
## Description

closes #5451 

PS.
I have included two changelog.md files, so you can test easier. I'll delete them upon approval.


Spacesai1or (changes requested):
I think it would be better to have such a complex script be written in TypeScript so that 1. it's easier for any maintainer of web3.js to maintain it (given that anyone working on web3.js should have an understanding of TS) and 2. so we can write unit tests to verify functionality instead of manually having to test (this would also ensure expected behavior as changes are made to either the script or the project)

Thank you for choosing to tackle this issue and putting effort in writing the bash script, even if we decided to go with a TS script, your effort shows how the CHANGELOG script needs to work

avkos:
@nikoulai Thank you for your work. I know it is painful to work with bash. So, in this case, I will agree with @spacesailor24 because it will be easier to maintain and add more features if the script is written in JS or TS.

Spacesai1or:
@nikoulai if you don't mind, can I take over this? I have something written in TypeScript in the works, but I need a little more time for it to be ready for a PR

nikoulai:
> @nikoulai if you don't mind, can I take over this? I have something written in TypeScript in the works, but I need a little more time for it to be ready for a PR

sure,  going with TS seems to be the right choice.

jdevcs (commented):
@spacesailor24 Once this functionality is available, Plus also add this in package.json scripts and it should be invoked at end of successful build.

PR #5771, opened 2023-01-20
Enable the user to use only web3 as dependency

### LINKED ISSUE #5508: [4.x] Use web3 as only dependency

### Feature Request

Expecting users to know what resides in which web3 package is optimistic, to say the least. It's hard for me (even though I do have extensive experience in developing apps), can't imagine how new web3 developers will find their way. 

That said, you don't need to sacrifice tree shaking to achieve that. 

### Use Cases

Importing types and functionality from a bunch of different web3 packages is hard without knowing which package contains what.

It's also making it hard to link web3 and test changes against the dev version of web3.

### Implementation Ideas

You can see solutions in existing sdks like:
- https://github.com/ethers-io/ethers.js/blob/master/packages/ethers/src.ts/ethers.ts
- https://github.com/polkadot-js/api/blob/master/packages/api/src/bundle.ts

Just do named re-export of all the required types, classes, and functions (you can even group them to make it easier for users). When a user uses named import, it will tree shake all not imported code paths.

Example usage would then be:
```
import {Web3, Contract} from "web3";

const web3 = new Web3(...)
const contract = new Contract(abi, address)
contract.link(web3) or whatever
```

- https://github.com/ethers-io/ethers.js/blob/master/packages/ethers/src.ts/ethers.ts
- https://github.com/polkadot-js/api/blob/master/packages/api/src/bundle.ts

- [ ] Yes

Muhammad-Altabba:
## Description

Fixes #5508

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [ ] I have tested my code.


mpetrunic (commented):
Generally looks good, I have two questions:

1- What do you think from DevEx perspective would it be better to group
ethAbi/ethIban/ethPersonal under `eth` namespace? Same goes for providers
So for example
```typescript
import {eth} from "web3"
eth.iban.Iban()
```

2- Maybe export types without namespace so they are autoimported for users and so user doesn't have to do something like: `export function something(): types.Web3Api`?

Muhammad-Altabba:
> 1- What do you think from DevEx perspective would it be better to group ethAbi/ethIban/ethPersonal under `eth` namespace? Same goes for providers So for example
> 
> ```ts
> import {eth} from "web3"
> eth.iban.Iban()
> ```

Well, this is a nice suggestion. But it would bring circular dependencies. For example, currently, web3-eth-ens uses web3-eth. And to achieve the proposed situation we need to use web3-eth-ens inside web3-eth in order to export it. (We need to add web3-eth-ens to the dependencies of web3-eth and then write `export * as ens from 'web3-eth-ens';`).

> 2- Maybe export types without namespace so they are autoimported for users and so user doesn't have to do something like: `export function something(): types.Web3Api`?

Actually, that would bring conflicts because we have the same name for functions and objects that exist in different packages (the list of errors are here: https://github.com/web3/web3.js/issues/5508#issuecomment-1397437571). And renaming those functions/objects or consolidating them would be nice. But it would introduce some more breaking changes.

Or what do you think?

Thanks,

Muhammad-Altabba:
> > 1- What do you think from DevEx perspective would it be better to group ethAbi/ethIban/ethPersonal under `eth` namespace? Same goes for providers So for example
> > ```ts
> > import {eth} from "web3"
> > eth.iban.Iban()
> > ```
> 
> Well, this is a nice suggestion. But it would bring circular dependencies. For example, currently, web3-eth-ens uses web3-eth. And to achieve the proposed situation we need to use web3-eth-ens inside web3-eth in order to export it. (We need to add web3-eth-ens to the dependencies of web3-eth and then write `export * as ens from 'web3-eth-ens';`).


I found a way to do it without exporting the sub-packages from inside the package (no need to write `export * as ens from 'web3-eth-ens';` at `web3-eth`. Actually, I arranged this using files inside the `web3` package). 
And I pushed the code.

@mpetrunic 
So, is it good now? Or you would like to discuss the second point or any other?
Thanks,

mpetrunic:
> > > 1- What do you think from DevEx perspective would it be better to group ethAbi/ethIban/ethPersonal under `eth` namespace? Same goes for providers So for example
> > > ```ts
> > > import {eth} from "web3"
> > > eth.iban.Iban()
> > > ```
> > 
> > 
> > Well, this is a nice suggestion. But it would bring circular dependencies. For example, currently, web3-eth-ens uses web3-eth. And to achieve the proposed situation we need to use web3-eth-ens inside web3-eth in order to export it. (We need to add web3-eth-ens to the dependencies of web3-eth and then write `export * as ens from 'web3-eth-ens';`).
> 
> I found a way to do it without exporting the sub-packages from inside the package (no need to write `export * as ens from 'web3-eth-ens';` at `web3-eth`. Actually, I arranged this using files inside the `web3` package). And I pushed the code.
> 
> @mpetrunic So, is it good now? Or you would like to discuss the second point or any other? Thanks,

Yeah! I guess only types are left, which ones are conflicting? Maybe it makes sense to rename or not export some?

Muhammad-Altabba:
> Yeah! I guess only types are left, which ones are conflicting? Maybe it makes sense to rename or not export some?

Thanks @mpetrunic, Actually, the conflicting types are listed here: https://github.com/web3/web3.js/issues/5508#issuecomment-1397437571 
And I could go through the 19 errors. And here are their cases:

1. First case is: there are 2 methods that has the same name but with different implementation. This is for `sign` and `signTransaction` functions that exist in: https://github.com/web3/web3.js/blob/4bbc030ba86f2e5b6494b395d77acce130d90c66/packages/web3-eth/src/rpc_method_wrappers.ts#L1459
And in:
https://github.com/web3/web3.js/blob/0727f26612dfe808b05a9c547650eefc36c5d8f2/packages/web3-eth-accounts/src/account.ts#L134
Hotwever, just to take a look. There is also `sign` and `signTransaction` exported from web3-rpc-methods. But it is exported inside `personalRpcMethods` in: https://github.com/web3/web3.js/blob/4bbc030ba86f2e5b6494b395d77acce130d90c66/packages/web3-rpc-methods/src/personal_rpc_methods.ts#L85

Suggestion: rename the `sign` at `web3-eth-accounts` to be `signWithPrivateKey`. And `signTransaction` to be `signTransactionWithPrivateKey`.

---

2. The other case is for a class and an interface given the same name like for `Web3Error`:
https://github.com/web3/web3.js/blob/b766282c527ffecd202f95248f7269f2d7583cb2/packages/web3-types/src/error_types.ts#L18
And
https://github.com/web3/web3.js/blob/b766282c527ffecd202f95248f7269f2d7583cb2/packages/web3-errors/src/web3_error_base.ts#L22
Similar case is for `Web3DeferredPromise` that its interface is defined in `web3-types` and its class is defined in `web3-utils`. But both have the same name.

Suggestion: Rename such interfaces to be ObjectInterface and leave the class as is.

---

3. The third case is for `AbiInput` and `AbiParameter` that has duplicate definitions in:
https://github.com/web3/web3.js/blob/b766282c527ffecd202f95248f7269f2d7583cb2/packages/web3-types/src/eth_types.ts#L226
And in:
https://github.com/web3/web3.js/blob/3f72f5293fdfdd62e5f17836c26513748dd9c08a/packages/web3-eth-abi/src/types.ts#L87
Additionally, `TypedArray` is duplicated in `web3-utils` and `web3-validator`. 

Suggestion: Move all types from `/packages/web3-eth-abi/src/types.ts` to `web3-types` package as we did this for similar cases. However, the definitions are not identical and so there will be a need to investigate and then consolidate them.

---

4. The forth case is for `checkAddressCheckSum`, `isAddress`, `isBloom`, `isContractAddressInBloom`, `isHex`, `isHexStrict`, `isInBloom`, `isNullish`, `isTopic`, `isTopicInBloom` and `isUserEthereumAddressInBloom`. Which are exported at `web3-utils` and then re-exported at `web3-validator`. 

Suggestion: Move all those to `web3-types` package.
[UPDATE] Suggestion: export at `web3-utils` inside a namespace.

---

5. The last case is for `LogsSubscription` that exists in both `web3-eth` (similar to the first case):
https://github.com/web3/web3.js/blob/e5f76aa59c9dc6fc78be29dfcc2eef6c83866a58/packages/web3-eth/src/web3_subscriptions.ts#L11
And in `web3-eth-contracts`:
https://github.com/web3/web3.js/blob/e6d4b23348d54cabd179f26b8008b9bc8a60718a/packages/web3-eth-contract/src/log_subscription.ts#L7

Suggestion: Actually, the two classes refer to two different things. And so, I would rename them to something like: `BlocksLogsSubscription` and `ContractsLogsSubscription`.

---

I will start appling the mentioned suggestions. However, please, let me know if you have any concern or suggestion.

Thanks,

mpetrunic:
Few things,
exporting many methods with different names could be confusing to developers, maybe creating more namespaces could be helpful or not exporting some stuff like why would a developer need Web3Error abstract class?

The second thing, I noticed is that every package has indexed with exports like `export * from "./somefile"`, I think it's a bad practice because:
- if you are exporting everything from every file, every change is gonna be breaking change - maybe consider exporting only stuff that developers or you actually need
- if the developer is searching for the source code of some imported class, it's gonna be a nightmare because it could be in any of these files so you can't just ctrl+click to dig through source code

Muhammad-Altabba:
Thanks @mpetrunic ,
> Few things, exporting many methods with different names could be confusing to developers, maybe creating more namespaces could be helpful or not exporting some stuff like why would a developer need Web3Error abstract class?
> 
Regarding the first suggestion that is "**maybe creating more namespaces could be helpful**". I think it is the better approach. And thinking about it. Let us apply the same thing to our exported objects at the `web3` package. i.e we should not have flat exports at web3 like `export * from 'web3-utils';` but have `export * as utils from 'web3-utils';`. Do you agree with this?

For the second point "not exporting some stuff". I agree that if there is something very local and not good to export we are better to keep it local. However, for the `Web3Error` abstract class specifically, this class is already used in another package. Moreover, we do not know if some users would need to implement it, especially that now we support plugins :)

> The second thing, I noticed is that every package has indexed with exports like `export * from "./somefile"`, I think it's a bad practice because:
> 
> * if you are exporting everything from every file, every change is gonna be breaking change - maybe consider exporting only stuff that developers or you actually need
> * if the developer is searching for the source code of some imported class, it's gonna be a nightmare because it could be in any of these files so you can't just ctrl+click to dig through source code

I created an [issue for this suggestion](https://github.com/web3/web3.js/issues/5794) and I hope to discuss it after finishing the current one :)

Muhammad-Altabba:
Hello @mpetrunic 
I am sorry that I was super confused because I misunderstood one of your comments. I reread the comments and I think I am back-to-track and aligned with you.

So, I pushed a commit that does a flat exports for `web3-types` without a namespace (in addition to being available at `types` namespace). To enable the user to write: `function something(): Web3Api` without the need for `types.Web3Api`. And the same for `web3-errors`. Because this package contains error classes and constants.

Additionally, I already implemented the suggestions 1, 2 and 3 from https://github.com/web3/web3.js/pull/5771#issuecomment-1408323377. I think 2 and 3 are good in all cases. And I am neutral about 1. So, please, if you think 1 (or 2 or 3) should be revoked, let me know.

I think the MR should be ready for review. Upon your final comment or agreement. I will update the docs and set it as ready for review. 

Thanks,

packages/web3-errors/src/web3_error_base.ts
```diff
@@ -17,9 +17,9 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 
 /* eslint-disable max-classes-per-file */
 
-import { Web3Error as ErrorInterface } from 'web3-types';
+import { Web3ErrorInterface } from 'web3-types';
 
-export abstract class Web3Error extends Error implements ErrorInterface {
+export abstract class Web3Error extends Error implements Web3ErrorInterface {
```
mpetrunic: small nit: maybe it would make more sense to name this BaseWeb3Error and leave Web3Error to be interface as users would you this as a type in catch clauses. No strong preference though

mpetrunic (commented):
LGTM. Great job. 
But also, just wanna be clear, you don't have to take my every comment as the final decision, I would very much like to hear your (and other's) thoughts on this and even suggestions :)

packages/web3-eth-accounts/src/account.ts
```diff
@@ -131,7 +131,8 @@ export const hashMessage = (message: string): string => {
  * }
  * ```
  */
-export const sign = (data: string, privateKey: Bytes): SignResult => {
+
+export const signWithPrivateKey = (data: string, privateKey: Bytes): SignResult => {
```
jdevcs: This will add extra step for users to migrate to 4.x, and we should make it least effort for end users instead of additional steps. so I think this change should be reverted.

packages/web3-eth-accounts/src/account.ts
```diff
@@ -240,7 +241,7 @@ export const sign = (data: string, privateKey: Bytes): SignResult => {
  * }
  * ```
  */
-export const signTransaction = async (
+export const signTransactionWithPrivateKey = async (
```
jdevcs: same here: This will add extra step for users to migrate to 4.x, and we should make it least effort for end users instead of additional steps. so I think this change should be reverted.

jdevcs (changes requested):

packages/web3-errors/src/web3_error_base.ts
```diff
@@ -17,9 +17,9 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 
 /* eslint-disable max-classes-per-file */
 
-import { Web3Error as ErrorInterface } from 'web3-types';
+import { Web3Error } from 'web3-types';
 
-export abstract class Web3Error extends Error implements ErrorInterface {
+export abstract class BaseWeb3Error extends Error implements Web3Error {
```
jdevcs, line 22: `BaseWeb3Error` represents base error more accurately, Thanks .but its out of scope for issue referred in this PR, in future should we create new issue and PR for such changes. this will make PRs small and concise to review saving time,

packages/web3-eth-contract/src/constants.ts
```diff
@@ -15,7 +15,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { AbiEventFragment } from 'web3-eth-abi';
+import { AbiEventFragment } from 'web3-types';
```
jdevcs, line 18: Pls also include change log for these breaking changes of types moved to new package.

packages/web3-eth-accounts/src/account.ts
jdevcs: for solving same method name issue, I think we can create a wrapper here `signWithPrivateKey` for exporting through web3 package and also keep `sign` here for old users least effort.

jdevcs:
@Muhammad-Altabba good work, Over all this PR LGTM. I just added few points + there are file conflicts to check.

packages/web3/src/index.ts
```diff
@@ -321,6 +321,41 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 /**
  * This comment _supports3_ [Markdown](https://marked.js.org/)
  */
+
 import Web3 from './web3';
 
-export = Web3;
+export default Web3;
+
+/**
```
jdevcs, line 330: This will cause breaking change :
`const Web3 = require('web3').default;` instead of `const Web3 = require('web3');`
and I prefer if we can avoid it and as solution shared  by Marin we can use [conditional export / separate entry point](https://nodejs.org/api/packages.html#conditional-exports).

mpetrunic: For the sake of our users, we should do that breaking change as importing like this is such a bad habbit and antipattern that everyone should avoid in modern js..

packages/web3-errors/src/web3_error_base.ts
Muhammad-Altabba: As agreed, I will keep it. And sorry to include changes out of scope especially since it affects lots of files.

packages/web3-eth-accounts/src/account.ts
Muhammad-Altabba: I reverted the renaming.
Thanks,

packages/web3-errors/src/web3_error_base.ts
jdevcs: Thanks for your great work.

jdevcs (approved):

Spacesai1or (approved):

PR #5823, opened 2023-02-08
5811 merge block tags

jdevcs:
## Description

https://github.com/web3/web3.js/issues/5811

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [ ] I have tested my code.


Aheyachy (approved):
[A spam message has been deleted]

packages/web3-types/src/eth_types.ts
```diff
@@ -45,8 +45,11 @@ export enum BlockTags {
 	EARLIEST = 'earliest',
 	LATEST = 'latest',
 	PENDING = 'pending',
+	SAFE = `safe`,
+	FINALIZED = `finalized`,
```
luu-alex:
```suggestion
	SAFE = 'safe',
	FINALIZED = 'finalized',
```
should stick to one format

luu-alex (approved):

packages/web3-eth/CHANGELOG.md
```diff
@@ -98,3 +98,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 -   Added `createAccessList` functionality ( #5780 )
+-   Added support of `safe` and `finalized` block tags (#5823)
```
Spacesai1or, line 101: For `web3-eth` and `web3-rpc-methods` there wasn't any code changes to add support for the new tags, they just inherited support from other packages that were updated

Muhammad-Altabba: @spacesailor24 
I think the whole idea of the CHANGELOG.md is to inform the user about any change in the usage experience. And for the user, it does not make any difference if these changes came from updating some code in the package, or if the changes came from updating a dependency. So, if there was any user-experience change that is introduced in a specific version, then those must be mentioned in the CHANGELOG.md. 

In short, from the user perspective, before this update, it was not possible to pass `safe` or `finalized` block tags. And this change in behavior is introduced in this specific version. And this has to be documented here.

packages/web3-utils/src/validation.ts
```diff
@@ -109,46 +110,37 @@ export const isTopicInBloom = isTopicInBloomValidator;
...
-	if (
-		typeof blockA === 'string' &&
-		!(
-			blockA === 'genesis' ||
-			blockA === 'earliest' ||
-			blockA === 'pending' ||
-			blockA === 'latest'
-		)
-	)
+	if (typeof blockA === 'string' && !(isBlockTag(blockA) || blockA === 'genesis'))
```
Muhammad-Altabba: I think we need to remove `genesis` as in https://github.com/web3/web3.js/pull/3594/files

packages/web3-utils/src/validation.ts
```diff
@@ -109,46 +110,36 @@ export const isTopicInBloom = isTopicInBloomValidator;
  */
 export const compareBlockNumbers = (blockA: Numbers, blockB: Numbers) => {
```
Muhammad-Altabba: What about using the type `BlockNumberOrTag` here?
```suggestion
export const compareBlockNumbers = (BlockNumberOrTag: Numbers, BlockNumberOrTag: Numbers) => {
```

packages/web3-types/src/eth_types.ts
```diff
@@ -45,8 +45,11 @@ export enum BlockTags {
 	EARLIEST = 'earliest',
 	LATEST = 'latest',
 	PENDING = 'pending',
+	SAFE = 'safe',
+	FINALIZED = 'finalized',
 }
-export type BlockTag = 'earliest' | 'latest' | 'pending';
+export type BlockTag = `${BlockTags}`;
+
 export type BlockNumberOrTag = Numbers | BlockTag;
```
Muhammad-Altabba, line 53: Since `Numbers` could be any string. I suggest we use this:
```suggestion
export type BlockNumberOrTag = number | bigint | BlockTag;
```

packages/web3-utils/src/validation.ts
```diff
@@ -109,46 +110,36 @@ export const isTopicInBloom = isTopicInBloomValidator;
...
-	if (blockA === 'latest') {
+	}
+	if (blockA === 'earliest') {
+		// b !== a, thus a < b
+		return -1;
+	}
+	if (blockB === 'earliest') {
+		// b !== a, thus a > b
+		return 1;
+	}
```
Muhammad-Altabba:
```suggestion
	if (blockA === 'earliest' || blockA === 0) {
		// b !== a, thus a < b
		return -1;
	}
	if (blockB === 'earliest' || blockB === 0) {
		// b !== a, thus a > b
		return 1;
	}
```

packages/web3-utils/src/validation.ts
```diff
@@ -162,8 +153,22 @@ export const compareBlockNumbers = (blockA: Numbers, blockB: Numbers) => {
 	if (blockB === 'pending') {
 		return -1;
 	}
+	if (blockA === 'safe' && blockB === 0) {
+		return 1;
+	}
+	if (blockA === 0 && blockB === 'safe') {
+		return -1;
+	}
```
Muhammad-Altabba: I think, those could be removed if we apply the suggestion above at lines 127 and 131.
```suggestion
```

packages/web3-utils/src/validation.ts
```diff
@@ -109,46 +110,36 @@ export const isTopicInBloom = isTopicInBloomValidator;
...
-			blockB === 'latest'
-		)
-	)
-		throw new InvalidBlockError(blockB);
+	if (blockA === 'genesis' || blockB === 'genesis')
+		throw new InvalidBlockError('Genesis tag not supported'); // for more specific error message
+	if (typeof blockA === 'string' && !isBlockTag(blockA)) throw new InvalidBlockError(blockA);
+	if (typeof blockB === 'string' && !isBlockTag(blockB)) throw new InvalidBlockError(blockB);
+
+	// Increasing order:  earliest, safe, (finalized ~ latest), pending
```
Muhammad-Altabba: It seems this is the order according to https://www.alchemy.com/overviews/ethereum-commitment-levels and https://hackmd.io/@prysmaticlabs/finality and so, the conditions below need to be updated (I think the conditions for `finalized`  needs to be modified.)
```suggestion
	// Increasing order:  earliest, finalized, safe (justified), latest, pending
```

packages/web3-utils/src/validation.ts
```diff
@@ -109,46 +110,36 @@ export const isTopicInBloom = isTopicInBloomValidator;
...
+	// safe vs block-num cant be compared as block number provided can be on left or right side of safe tag, until safe tag block number is extracted and compared
```
Muhammad-Altabba, line 119:
```suggestion
	// Note: Any block tag vs block-num cant be compared as the block number provided can be on the left or the right side of that block tag, until that tag's block number is extracted and compared. 
	//   For example, `pending` is larger than all the produced blocks. But it is smaller than any number of a block that is not yet produced. ex. `pending` is smaller than `Number.MAX_VALUE`.
```

packages/web3-utils/src/validation.ts
```diff
@@ -162,8 +153,22 @@ export const compareBlockNumbers = (blockA: Numbers, blockB: Numbers) => {
...
+	}
+	if (blockA === 0 && blockB === 'safe') {
+		return -1;
+	}
+	if (blockA === 'safe' || blockB === 'safe') {
+		// either a or b is "safe" and the other one did not fall into any of the conditions above, so the other one is a number
+		throw new InvalidBlockError(
+			`Cannot compare safe tag with ${blockA === 'safe' ? blockB : blockA}`,
+		);
+	}
```
Muhammad-Altabba: I think, any block-tag vs block-num cant be compared as the block number provided can be on the left or the right side of that block tag, until that tag's block number is extracted and compared. 
For example, `pending` is larger than all the produced blocks. But it is smaller than any number of a block that is not yet produced. ex. `pending` is smaller than `Number.MAX_VALUE`.

```suggestion
	if (isBlockTag(blockA) || isBlockTag(blockB)) {
		// either a or b is a blocktag but the other one did not fall into any of the conditions above, so the other one is a number
		throw new InvalidBlockError(
			`Cannot compare a block tag with a number. blockA: ${blockA}, blockB: ${blockB}`,
		);
	}
```

Muhammad-Altabba (commented):
I have a few points, please.

packages/web3-types/src/eth_types.ts
jdevcs: Numbers already includes bigint https://github.com/web3/web3.js/blob/f3b8ca88eb34dbf73f8e3dc9f5995c4c1a54fa87/packages/web3-types/src/primitives_types.ts#L20 so this change is not required.

avkos (approved):

packages/web3-utils/src/validation.ts
jdevcs: you mean: `compareBlockNumbers = (blockA: BlockNumberOrTag, blockB: BlockNumberOrTag)` and not `(BlockNumberOrTag: Numbers, BlockNumberOrTag: Numbers) ` , done

packages/web3-types/src/eth_types.ts
jdevcs: Since BlockNumberOrTag is used in many other places of lib and uses can pass hex string to other functions so this shouldnt be changed.

packages/web3-utils/src/validation.ts
jdevcs: this was already catered in cases below, in case A or B is 0 and other was valid block tag,. but good to check earlier in logic.

packages/web3-utils/src/validation.ts
jdevcs: modified

packages/web3-utils/src/validation.ts
jdevcs: This change is not required, as in some case block tag vs number cases are supported. like earliest vs any number etc

packages/web3-utils/src/validation.ts
jdevcs: > I think, any block-tag vs block-num cant be compared as the block number provided can be on the left or the right side of that block tag, until that tag's block number is extracted and compared.

in some cases we can compare tag with number as:
earliest vs num
pending vs num, ( as pending means it will always come after mined block )

jdevcs: https://github.com/web3/web3.js/issues/5836

PR #5824, opened 2023-02-09
Fix few issues with new Web3().eth.contract

### LINKED ISSUE #5815: A couple of issues with `new Web3().eth.Contract`

### Current Behavior

The `ContractBuilder` that is later assigned to eth.Contract inside the class `Web3`:
https://github.com/web3/web3.js/blob/ab31eb08b3373292dbc65e744685fb84f0941ac3/packages/web3/src/web3.ts#L69 
It is supposed to be used with the following:
```js
const Contract = new Web3().eth.Contract;
const contract = new Contract(...);
``` 

But it seems to have many issues:
* The constructor does not accept receiving the second parameter as an object of `ContractInitOptions`. For example, the following will throw an error: `new Contract([{}], {})`.
This is because the type `Web3ContractConstructor` is not compatible
https://github.com/web3/web3.js/blob/ab31eb08b3373292dbc65e744685fb84f0941ac3/packages/web3/src/types.ts#L44
Suggestion: the type `ContractInitOptions` should be compatible with the class `Contract` that is defined in `web3-eth-contract`.
* After fixing the above, we would be able to pass (abi, address, options) but the options would not be used. At the line: https://github.com/web3/web3.js/blob/ab31eb08b3373292dbc65e744685fb84f0941ac3/packages/web3/src/web3.ts#L80
Actually, currently, the first 2 `if` and `else if` have the same code block.
* The condition `else if (!isNullish(addressOrOptions) && isNullish(options))` will never be met. Except if the second parameter passed was not a string or an object (it would fall into the previous conditions if it was a string or an object). And so it is a useless condition. And it would only be used in case the user misused the types and passed something like a function. Where the method should throw an exception in that case.
* At the function call there is an unneeded check with `??` https://github.com/web3/web3.js/blob/ab31eb08b3373292dbc65e744685fb84f0941ac3/packages/web3/src/web3.ts#L84
and at https://github.com/web3/web3.js/blob/ab31eb08b3373292dbc65e744685fb84f0941ac3/packages/web3/src/web3.ts#L86 because it each is already checked to not be null or undefined in there each previous condition.

Suggestion for the last 3 points: replace all if conditions with something like the following:
```js
        if (typeof addressOrOptions === 'object' && typeof options === 'object') {
		throw new Error(
			'Unsupported passed parameters. Should not provide options at both 2nd and 3rd parameters',
		);
	}
	if (isNullish(addressOrOptions)) {
		super(jsonInterface, options, self.getContextObject());
	} else if (typeof addressOrOptions === 'object') {
		super(jsonInterface, addressOrOptions, self.getContextObject());
	} else if (typeof addressOrOptions === 'string') {
		super(jsonInterface, addressOrOptions, options ?? {}, self.getContextObject());
	} else {
		throw new Error('Unsupported passed parameters');
	}
```

### Expected Behavior

To have a consistent bug-free constructor for `new Web3().eth.Contract`

### Steps to Reproduce

Take a look at https://github.com/web3/web3.js/blob/ab31eb08b3373292dbc65e744685fb84f0941ac3/packages/web3/src/web3.ts#L69

### Web3.js Version

4.x

### Environment

- Operating System: 
- Browser: 
- Node.js Version: 
- NPM Version: 

### Anything Else?

_No response_

Muhammad-Altabba:
## Description
Fixes #5815

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [ ] I have tested my code.


packages/web3-eth/package.json
```diff
@@ -53,7 +53,6 @@
 		"setimmediate": "^1.0.5",
 		"web3-core": "^4.0.1-alpha.5",
 		"web3-errors": "^0.1.1-alpha.4",
-		"web3-eth-accounts": "^4.0.1-alpha.5",
```
Spacesai1or, line 57: This should be reverted as  `privateKeyToAddress` from `web3-eth-accounts` [transaction_builder.ts](https://github.com/web3/web3.js/blob/4cecc7a92cd2054c1fd3a57e8da6c1f8da078e38/packages/web3-eth/src/utils/transaction_builder.ts#L41)

Muhammad-Altabba: Thanks for the note. 
Interestingly, the build succeeded but lint detected that.

packages/web3-errors/src/errors/generic_errors.ts
```diff
@@ -45,6 +46,21 @@ export class InvalidNumberOfParamsError extends BaseWeb3Error {
 	}
 }
 
+export class InvalidMethodParamsError extends BaseWeb3Error {
```
jdevcs, line 49: changelog update is also required.

jdevcs (approved):
lgtm, update changelog before merging.

avkos (approved):

Issue #5838, opened 2023-02-15
Inconsistent thrown errors between provider protocols


Spacesai1or:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```typescript
const transaction: Transaction = {
	from: tempAcc.address,
	to: '0x0000000000000000000000000000000000000000',
	value: BigInt(99999999999999999999999999999999999999999999999999999999999999),
}
try {
	await web3Eth.sendTransaction(transaction).on('error', error => console.log(error));
} catch (error) {
	console.log(error)
}
```

Throws for `HTTP`:

```
InvalidResponseError: Returned error: insufficient funds for transfer
      {
        innerError: { code: -32000, message: 'insufficient funds for transfer' },
        code: 101,
        data: undefined
      }
```

Throws for `WS` or `IPC`:

```
ResponseError: Returned error: insufficient funds for transfer
      {
        innerError: undefined,
        code: 100,
        data: undefined
      }
```

### Expected Behavior

Thrown errors to be same

### Steps to Reproduce

```typescript
const transaction: Transaction = {
	from: tempAcc.address,
	to: '0x0000000000000000000000000000000000000000',
	value: BigInt(99999999999999999999999999999999999999999999999999999999999999),
}
try {
	await web3Eth.sendTransaction(transaction).on('error', error => console.log(error));
} catch (error) {
	console.log(error)
}
```

### Web3.js Version

4.x

### Environment

- Operating System: 
- Browser: 
- Node.js Version: 
- NPM Version:

Issue #5840, opened 2023-02-15
`Web3RequestManager.send` should return errors for `Web3Provider`s instead of throwing them


Spacesai1or:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature Request

With the error message received from the transaction sent in `1.`, there is no opportunity for the web3.js developer to capture the RPC error and repackage it as, for example, `InsufficientFundsError` that has a specific error code that users could look up in the docs to learn more about it. I think inside web3.js we'd want to return errors that can be parsed and repackaged as needed up until the point where there is nothing the library can do and we throw it

I'm proposing we update [_processJsonRpcResponse](https://github.com/web3/web3.js/blob/f3b8ca88eb34dbf73f8e3dc9f5995c4c1a54fa87/packages/web3-core/src/web3_request_manager.ts#L312) to return errors instead of throwing them. When [Web3Providers](https://github.com/web3/web3.js/blob/f3b8ca88eb34dbf73f8e3dc9f5995c4c1a54fa87/packages/web3-core/src/web3_request_manager.ts#L178) use the `requestManager` they'll be able to receive the RPC error for further processing while other providers will continue to throw the errors as their supposed to (e.g. EIP-1193 providers)

### Use Cases

Would be used to to get a contract revert reason. Right now it's not possible to parse revert reason in `sendTransaction` because the thrown error is from the `Web3RequestManager` and is thrown before `sendTransaction` has the ability to catch the error for further processing

### Are you willing to implement this feature?

- [ ] Yes

Spacesai1or:
While working on this, I found that we can wrap the [trySendTransaction](https://github.com/web3/web3.js/blob/a5aa3220733f8af8aca7a09ebe02224979153f2d/packages/web3-eth/src/rpc_method_wrappers.ts#L1133) in a try catch allowing us to parse the error thrown by the request manager for further processing. I started adding checks for the possible error messages, throwing new errors with specific codes like so:

```typescript
const parsedError = parseTransactionError(error);

if (typeof parsedError === 'string') {
	switch (parsedError) {
		case 'intrinsic gas too low':
			throw new IntrinsicGasTooLowError();
		case 'insufficient funds for transfer':
			throw new InsufficientFundsError();
		case 'unknown account':
			throw new UnknownAccountError();
	
		default:
			throw error
	}
}
```

[...]

Issue #5839, opened 2023-02-15
Missing `receipt` when getting revert reason for failed sent transaction


Spacesai1or:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```typescript
const contractFunctionCall = '0xa4136862000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000023432000000000000000000000000000000000000000000000000000000000000';
const transaction: Transaction = {
	from: tempAcc.address,
	to: greeterContractAddress,
	data: contractFunctionCall,
	gas: 34000
};
web3Eth.handleRevert = true;
try {
	await web3Eth.sendTransaction(transaction);
} catch (error) {
	console.log(error)
}
```

yields the thrown error:

```
TransactionRevertError: Transaction has been reverted by the EVM:
       undefined
       {
        innerError: undefined,
        reason: 'Returned error: err: intrinsic gas too low: have 34000, want 53496 (supplied gas 34000)',
        signature: undefined,
        receipt: undefined,
        code: 402
      }
```

I believe part of the issue is [getRevertReason](https://github.com/web3/web3.js/blob/f3b8ca88eb34dbf73f8e3dc9f5995c4c1a54fa87/packages/web3-eth/src/rpc_method_wrappers.ts#L1086) is called before the transaction is submitted, and while it would be nice to throw before submitting the transaction (so the user's gas isn't spent), `1.x` explicitly says for [sendTransaction](https://web3js.readthedocs.io/en/v1.8.2/web3-eth.html#sendtransaction) that the receipt would be available if the network rejects the transaction. So I believe we'd want to call `getRevertReason` only after the transaction has failed (e.g. `transactionReceipt.status === 0`)

### Expected Behavior

Receipt to be apart of thrown error as is in `1.x`

### Steps to Reproduce

```typescript
const contractFunctionCall = '0xa4136862000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000023432000000000000000000000000000000000000000000000000000000000000';
const transaction: Transaction = {
	from: tempAcc.address,
	to: greeterContractAddress,
	data: contractFunctionCall,
	gas: 34000
};
web3Eth.handleRevert = true;
try {
	await web3Eth.sendTransaction(transaction);
} catch (error) {
	console.log(error)
}
```

### Web3.js Version

4.x

### Environment

- Operating System: 
- Browser: 
- Node.js Version: 
[...]

PR #5844, opened 2023-02-16
Refactor `getRevertReason` and add tests

### LINKED ISSUE #5841: `getRevertReason` shouldn't throw the reason
The method is supposed to return the revert reason, so failure (cause to throw an error) would be if no revert reason was received or the actual request to the provider failed. Instead this method should return the reason so errors received by the user like in `3.` can be thrown as:

```
TransactionOutOfGasError: Transaction ran out of gas:
	{
	reason: 'Returned error: err: intrinsic gas too low: have 34000, want 53496 (supplied gas 34000)',
	signature: undefined,
	receipt: { ...transactionReceipt },
	code: 402
	}
```

Spacesai1or:
#### Added

-   `decodeErrorData` from `web3-eth-contract` is now exported from this package and was renamed to `decodeContractErrorData` (#5844)

#### Removed

-   `decodeErrorData` is no longer exported (method was moved to `web3-eth-abi` and renamed `decodeContractErrorData`) (#5844)

#### Removed

-   `getRevertReason` is no longer exported (#5844)

#### Changed

-   `SocketProvider` abstract class now resolves JSON RPC response errors instead of rejecting them (#5844)

---

closes #5841 #5838


packages/web3-eth-abi/src/decode_contract_error_data.ts
```diff
@@ -0,0 +1,50 @@
...
+*/
+
+import { Eip838ExecutionError } from 'web3-errors';
+import { AbiErrorFragment } from 'web3-types';
+
+import { encodeErrorSignature } from './api/errors_api';
+import { decodeParameters } from './api/parameters_api';
+import { jsonInterfaceMethodToString } from './utils';
+
+export const decodeContractErrorData = (
```
Spacesai1or, line 25: This method was moved from `web3-eth-contract` package to `web3-eth-abi` to avoid a circular dependency issue between `web3-eth-contract` and `web3-eth` since it's now being used inside `web3-eth`'s `getRevertReason` util method. It was also renamed from `decodeErrorData`

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1926,22 +1922,6 @@ export async function getFeeHistory<ReturnFormat extends DataFormat>(
 
 	return format(feeHistorySchema, response as unknown as FeeHistory, returnFormat);
 }
-/**
- *
- * @param web3Context
- * @param transaction
- */
-async function getRevertReason<ReturnFormat extends DataFormat>(
```
Spacesai1or, line 1934: Method was moved to `utils/get_revert_reason.ts` because it's not a RPC method

packages/web3-utils/src/socket_provider.ts
```diff
@@ -333,12 +332,13 @@ export abstract class SocketProvider<
...
+			if (
+				jsonRpc.isBatchResponse(response) ||
+				jsonRpc.isResponseWithResult(response) ||
+				jsonRpc.isResponseWithError(response)
+			) {
 				this._eventEmitter.emit('message', undefined, response);
 				requestItem.deferredPromise.resolve(response);
-			} else {
-				this._eventEmitter.emit('message', response, undefined);
-				requestItem?.deferredPromise.reject(new ResponseError(response));
```
spacesailor24, line 341: This is up for discussion. I removed this code and the tests from `packages/web3-providers-ws/test/unit/web_socket_provider.test.ts` because I couldn't get an error to actually trigger this code because any error resulting from a failed network request was emitted by [_onError](https://github.com/web3/web3.js/blob/a54d63796e5cde9496af91522737a2418820108f/packages/web3-utils/src/socket_provider.ts#L197) and any RPC response with an error is now being resolved. As the only tests written for this code were for RPC response errors, I don't think this code is needed anymore

Spacesai1or, line 341: I removed this code and the tests from `packages/web3-providers-ws/test/unit/web_socket_provider.test.ts` because I couldn't get an error to actually trigger this code because any error resulting from a failed network request was emitted by [_onError](https://github.com/web3/web3.js/blob/a54d63796e5cde9496af91522737a2418820108f/packages/web3-utils/src/socket_provider.ts#L197) and any RPC response with an error is now being resolved. As the only tests written for this code were for RPC response errors, I don't think this code is needed anymore

packages/web3/test/integration/handle_revert.test.ts
```diff
@@ -28,7 +28,7 @@ import { BasicAbi, BasicBytecode } from '../shared_fixtures/build/Basic';
 
 Error.stackTraceLimit = Infinity;
 
-describe('eth', () => {
+describe.skip('eth', () => {
```
Spacesai1or, line 31: In order for these tests to pass with the expected error, #5839 and #5840 need to be addressed. This is because previously when `sendTransaction` was called `getRevertReason` we were throwing the reason making the `sendTransaction` promise to `reject` with the revert reason. Now that we're returning the reason, `sendTransaction` rejects at the `requestManager` (#5840) giving `sendTransaction` no opportunity to throw an error with the retrieved revert reason - we will need to re-enable and refactor these tests after #5840 and #5839

For context, the following is the result of running these tests with this PR's changes:

```typescript
$ jest --config=./test/integration/jest.config.js --forceExit test/integration/handle_revert.test.ts
 FAIL  test/integration/handle_revert.test.ts
  ● eth › handleRevert › should get revert reason

    expect(received).rejects.toThrow(expected)

    - Expected message  - 2
    + Received message  + 1

    - Transaction has been reverted by the EVM:
    -  undefined
    + Error happened while trying to execute a function inside a smart contract

          440 | 		// However, more processing will happen at a higher level to decode the error data,
          441 | 		//	according to the Error ABI, if it was available as of EIP-838.
        > 442 | 		if (error?.message.includes('revert')) throw new ContractExecutionError(error);
              | 		                                             ^
          443 |
          444 | 		return false;
          445 | 	}

          at Function._isReverted (../web3-core/src/web3_request_manager.ts:442:48)
          at Web3RequestManager._processJsonRpcResponse (../web3-core/src/web3_request_manager.ts:371:35)
          at Web3RequestManager.<anonymous> (../web3-core/src/web3_request_manager.ts:216:16)
          at fulfilled (../web3-core/lib/web3_request_manager.js:5:58)

      89 | 		it('should get revert reason', async () => {
      90 | 			contract.handleRevert = true;
    > 91 | 			await expect(contract.methods.reverts().send({ from: accounts[0] })).rejects.toThrow(
         | 			                                                                             ^
      92 | 				new TransactionRevertError(
      93 | 					'Returned error: execution reverted: REVERTED WITH REVERT',
      94 | 				),

      at Object.toThrow (../../node_modules/expect/build/index.js:241:22)
      at test/integration/handle_revert.test.ts:91:81
      at test/integration/handle_revert.test.ts:8:71
      at Object.<anonymous>.__awaiter (test/integration/handle_revert.test.ts:4:12)
      at Object.<anonymous> (test/integration/handle_revert.test.ts:89:45)

  ● eth › handleRevert › should get revert reason for eth tx

    expect(received).rejects.toThrow(expected)

    - Expected message  - 2
    + Received message  + 1

    - Transaction has been reverted by the EVM:
    -  undefined
    + Returned error: invalid argument 0: json: cannot unmarshal invalid hex string into Go struct field TransactionArgs.data of type hexutil.Bytes

          370 | 				}
          371 | 			} else if (!Web3RequestManager._isReverted(response)) {
        > 372 | 				throw new InvalidResponseError<ErrorType>(response);
              | 				      ^
          373 | 			}
          374 | 		}
          375 |

          at Web3RequestManager._processJsonRpcResponse (../web3-core/src/web3_request_manager.ts:372:11)
          at Web3RequestManager.<anonymous> (../web3-core/src/web3_request_manager.ts:216:16)
          at fulfilled (../web3-core/lib/web3_request_manager.js:5:58)

      112 | 					s: '0x39f77e0b68d5524826e4385ad4e1f01e748f32c177840184ae65d9592fdfe5c',
      113 | 				}),
    > 114 | 			).rejects.toThrow(
          | 			          ^
      115 | 				new TransactionRevertError(
      116 | 					'Returned error: invalid argument 0: json: cannot unmarshal invalid hex string into Go struct field TransactionArgs.data of type hexutil.Bytes',
      117 | 				),

      at Object.toThrow (../../node_modules/expect/build/index.js:241:22)
      at test/integration/handle_revert.test.ts:114:14
      at test/integration/handle_revert.test.ts:8:71
      at Object.<anonymous>.__awaiter (test/integration/handle_revert.test.ts:4:12)
      at Object.<anonymous> (test/integration/handle_revert.test.ts:98:56)

Test Suites: 1 failed, 1 total
Tests:       2 failed, 1 passed, 3 total
Snapshots:   0 total
Time:        1.873 s
Ran all test suites matching /test\/integration\/handle_revert.test.ts/i.
Force exiting Jest: Have you considered using `--detectOpenHandles` to detect async operations that kept running after all tests finished?
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

packages/web3-utils/src/socket_provider.ts
Spacesai1or: This change also fixes #5838

nikoulai (approved):

packages/web3-eth/src/utils/get_revert_reason.ts
```diff
@@ -0,0 +1,84 @@
...
+
+		if (
+			error instanceof InvalidResponseError &&
+			!Array.isArray(error.innerError) &&
+			error.innerError !== undefined
+		) {
+			return error.innerError.message;
+		}
+
+		return error instanceof Error ? error.message : error;
```
jdevcs: In case of generic error ( that is not due to failed tx reverted ) , we should throw it and remove `string` from return type`Promise<undefined | RevertReason | RevertReasonWithCustomError | string | unknown>`. 

Because for `getRevertReason` function call with eth_call rpc in case of generic errors: 
1. neither tx eth_call was reverted with reason.
2. nor it succeeded with out any revert reason, 
but request for looking any of above answers failed, that user should catch. It will also make return types homogeneous.

jdevcs (changes requested):

packages/web3-eth/src/rpc_method_wrappers.ts
jdevcs: get revert reason is wrapper on `eth_call` rpc method, so it was under `rpc_method_wrappers.ts`,

jdevcs (commented):
Overall looks good, Thanks Wyatt. I posted a change for getRevertReason return behavior.

packages/web3/test/integration/handle_revert.test.ts
jdevcs: It should be part of RC.0 @mconnelly8

packages/web3-eth/src/utils/get_revert_reason.ts
Spacesai1or: Updated via [this commit](https://github.com/web3/web3.js/pull/5844/commits/28a3d848830a0e17aa706b5092c1e917e1b87f02), however `string` is still a possible return type because of

```typescript
if (
	error instanceof InvalidResponseError &&
	!Array.isArray(error.innerError) &&
	error.innerError !== undefined
) {
	return error.innerError.message;
}
```

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: What I meant is that `getRevertReason` isn't a wrapper of it's own RPC method such as `eth_getRevertReason`. Other reasons I removed it from `rpc_method_wrappers.ts` is because it's no longer exported to the user (so that we can make changes to it as needed without it being a breaking change) and the function is a utility method to be used by other methods, not on it's own

packages/web3-eth/src/utils/get_revert_reason.ts
jdevcs: if InvalidResponseError is also not categorizable under 1 and 2, shouldn't it also be thrown?

Spacesai1or: @jdevcs The code snippet above is for errors such as:

```typescript
InvalidResponseError: Returned error: err: intrinsic gas too low: have 0, want 21544 (supplied gas 0)
          at Web3RequestManager._processJsonRpcResponse (/home/anon/Public/code/ChainSafe/git-repos/web3.js/packages/web3-core/src/web3_request_manager.ts:372:11)
          at Web3RequestManager.<anonymous> (/home/anon/Public/code/ChainSafe/git-repos/web3.js/packages/web3-core/src/web3_request_manager.ts:216:16)
          at fulfilled (/home/anon/Public/code/ChainSafe/git-repos/web3.js/packages/web3-core/lib/web3_request_manager.js:5:58)
          at processTicksAndRejections (node:internal/process/task_queues:95:5) {
        innerError: {
          code: -32000,
          message: 'err: intrinsic gas too low: have 0, want 21544 (supplied gas 0)'
        },
        code: 101,
        data: undefined
      }
```

keep in mind that this method is now an internal method to be used by devs to further process why transactions failed before returning an error to the end user with more context

jdevcs: also it can be generic error like : https://github.com/web3/web3.js/blob/b948ea4f3c6d48509a7aa726a2d051a8c83ebc96/packages/web3-utils/src/chunk_response_parser.ts#L67.  ( I think in that case at least , it should be modified in chunkresponseparser to be some different error type and code ) and we can keep returning InvalidResponseError as it is here.

jdevcs (approved):

packages/web3-eth/src/utils/get_revert_reason.ts
Spacesai1or: @jdevcs That's a good point. I think the error in the chunk parser should remain `InvalidResponseError`, but errors like the gas too low from above should be renamed to something like `RpcError`. The gas too low isn't an `InvalidResponse` it's just an error returned from the EVM because the transaction was invalid

If you agree, I can create an issue for it, merge this PR, then make the error name change?

jdevcs: sure. Good work @spacesailor24

avkos (approved):
Great Job, Thx!

Issue #5848, opened 2023-02-18
JSON RPC Response Errors Shouldn't be Returned as `InvalidResponseErrors`


Spacesai1or:
@jdevcs That's a good point. I think the error in the chunk parser should remain `InvalidResponseError`, but errors like the gas too low from above should be renamed to something like `RpcError`. The gas too low isn't an `InvalidResponse` it's just an error returned from the EVM because the transaction was invalid

If you agree, I can create an issue for it, merge this PR, then make the error name change?

_Originally posted by @spacesailor24 in https://github.com/web3/web3.js/pull/5844#discussion_r1110327172_

mconnelly8:
This will cause a breaking change, so closing issue.

Issue #5837, opened 2023-02-15
Inconsistent thrown errors for `web3Eth.sendTransaction`


Spacesai1or:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```typescript
const transaction: Transaction = {
	from: tempAcc.address,
	to: '0x0000000000000000000000000000000000000000',
	value: BigInt(99999999999999999999999999999999999999999999999999999999999999),
}
try {
	await web3Eth.sendTransaction(transaction).on('error', error => console.log(error));
} catch (error) {
	console.log(error)
}
```

The error received by the event listener would be:

```
TransactionError: Returned error: insufficient funds for transfer
      {
        innerError: undefined,
        receipt: undefined,
        code: 400
      }
```

While the error caught by the `try/catch` would be:

```
InvalidResponseError: Returned error: insufficient funds for transfer
      {
        innerError: { code: -32000, message: 'insufficient funds for transfer' },
        code: 101,
        data: undefined
      }
```

### Expected Behavior

Thrown errors to be same

### Steps to Reproduce

```typescript
const transaction: Transaction = {
	from: tempAcc.address,
	to: '0x0000000000000000000000000000000000000000',
	value: BigInt(99999999999999999999999999999999999999999999999999999999999999),
}
try {
	await web3Eth.sendTransaction(transaction).on('error', error => console.log(error));
} catch (error) {
	console.log(error)
}
```

### Web3.js Version

4.x

### Environment

- Operating System: 
- Browser: 
- Node.js Version: 
- NPM Version:

PR #5854, opened 2023-02-22
`sendTransaction` and `sendSignedTransaction` Error Refactors

Spacesai1or:
#### web3-errors

-   `request` property to `ResponseError` (#5854)
-   `data` property to `TransactionRevertInstructionError` (#5854)
-   `TransactionRevertWithCustomError` was added to handle custom solidity errors (#5854)

#### web3-eth

-   `contractAbi` option to `SendTransactionOptions` and `SendSignedTransactionOptions` to added the ability to parse custom solidity errors (#5854)

#### web3-errors

-   Renamed `TransactionRevertError` to `TransactionRevertInstructionError` to remain consistent with `1.x` (#5854)

#### web3-eth

-   `sendTransaction` and `sendSignedTransaction` now errors with (and `error` event emits) the following possible errors: `TransactionRevertedWithoutReasonError`, `TransactionRevertInstructionError`, `TransactionRevertWithCustomError`, `InvalidResponseError`, or `ContractExecutionError` (#5854)


packages/web3-errors/src/errors/response_errors.ts
```diff
@@ -36,14 +41,19 @@ const isResponseWithError = <Error = unknown, Result = unknown>(
...
-	public constructor(response: JsonRpcResponse<unknown, ErrorType>, message?: string) {
+	public constructor(
+		response: JsonRpcResponse<unknown, ErrorType>,
+		message?: string,
+		request?: JsonRpcPayload<RequestType>,
+	) {
 		super(
 			message ??
-				`Returned error: ${
+				`${
```
spacesailor24, line 56: I don't think prefixing every JSON response error with `Returned error: ` provides any benefit, and removing it allows errors to be identified by using `error.message` without having to parse the actual error by removing prefix every time

packages/web3-errors/src/errors/response_errors.ts
```diff
@@ -55,16 +65,24 @@ export class ResponseError<ErrorType = unknown> extends BaseWeb3Error {
 				? response.map(r => r.error?.data as ErrorType)
 				: response?.error?.data;
 		}
+
+		this.request = request;
```
Spacesai1or, line 69: I've added request to the error object as I wanted to add sent transaction details to failed transaction errors to give the user more context (e.g. the sending account when the error `Unknown account` or `Insufficient funds` is thrown) on what exactly was sent to provider

packages/web3-errors/src/errors/transaction_errors.ts
```diff
@@ -84,25 +85,54 @@ export class RevertInstructionError extends BaseWeb3Error {
...
 		public reason: string,
 		public signature?: string,
-		public receipt?: TransactionReceipt,
+		public receipt?: ReceiptType,
+		public data?: string,
 	) {
-		super(
-			`Transaction has been reverted by the EVM:\n ${JSON.stringify(receipt, undefined, 2)}`,
-		);
+		super(reason);
```
spacesailor24, line 97: Removed `Transaction has been reverted by the EVM:\n ` because the error is already called `TransactionRevertError` and the prefix doesn't add anymore benefit to the user. As mentioned above with `ResponseError`, removing the prefix allows the actual revert reason to be more accessible to the end user to be used within their error handling code

Also removed `JSON.stringify(receipt, undefined, 2)` because receipts are not always available when this error is thrown, makes more sense just to leave it as a property of error object

packages/web3-errors/src/errors/transaction_errors.ts
```diff
@@ -84,25 +85,54 @@ export class RevertInstructionError extends BaseWeb3Error {
...
+			...super.toJSON(),
+			reason: this.reason,
+			signature: this.signature,
+			receipt: this.receipt,
+			data: this.data
+		};
+	}
+}
+
+export class TransactionRevertWithCustomError<ReceiptType = TransactionReceipt> extends TransactionRevertError<ReceiptType> {
```
spacesailor24, line 111: This error is specifically for transactions that reverted when calling a contract method that had a custom error. I believe `web3-eth-contract` already has the logic, but this adds support for `web3Eth.sendTransaction`

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1078,14 +1080,6 @@ export function sendTransaction<
...
 						);
 
-						if (web3Context.handleRevert) {
-							// eslint-disable-next-line no-use-before-define
-							await getRevertReason(
-								web3Context,
-								transactionFormatted as TransactionCall,
-							);
-						}
-
```
Spacesai1or: This is part of addressing #5839, the revert reason is retrieved after the transaction has been processed by the network and when `transactionReceipt.status === BigInt(0)`

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1177,17 +1171,17 @@ export function sendTransaction<
...
+							);
+
 							if (promiEvent.listenerCount('error') > 0) {
-								promiEvent.emit(
-									'error',
-									new TransactionError(
-										'Transaction failed',
-										transactionReceiptFormatted,
-									),
-								);
```
Spacesai1or, line 1187: This was the cause of #5837

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1206,17 +1200,18 @@ export function sendTransaction<
 							);
 						}
 					} catch (error) {
-						if (error instanceof ContractExecutionError) {
-							promiEvent.emit('contractExecutionError', error);
-						}
```
Spacesai1or, line 1211: I don't think there's a reason to have an `contractExecutionError` event when we already have a generic `error` event that every other error uses (@nikoulai please let me know if there was a specific reason for it since you added it in #5659)

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1206,14 +1201,10 @@ export function sendTransaction<
...
-						if (promiEvent.listenerCount('error') > 0) {
-							promiEvent.emit(
-								'error',
-								new TransactionError((error as Error).message),
-							);
+						if (error instanceof InvalidResponseError &&
+							promiEvent.listenerCount('error') > 0
+						) {
+							promiEvent.emit('error', error);
```
spacesailor24, line 1207: When this code executes, `error` is one of two things:

1. It's an instance of `InvalidResponseError` (the error will be renamed to something like `RpcError` (#5848)), meaning the transaction failed to send, but it was rejected by the RPC provider and didn't even make it to the Ethereum network. This happens for errors like insufficient funds or unknown account

2. Literally every other possible error that could occur, meaning it could be an error that we haven't anticipated or just an error resulting from a bug we have in the code. In these instances, I don't think it makes sense to cast this error as `TransactionError` because it implies there's an error in the transaction the user is trying to send, but this error could be anything. If we don't cast it though then it makes `error` of type `unknown` and we'd have to include `unknown` as a part of the possible emission types for [SendTransactionEvents](https://github.com/web3/web3.js/blob/36c9b0d950f1b882dcbb28839bda5cd83a9ae5bb/packages/web3-eth/src/types.ts#L42) which makes the entire type show as `unknown` to the user reducing type safety. So, I'm suggesting we only emit an error for expected errors i.e. anything that extends `InvalidResponseError` as all other expected transaction related errors will be handled within the `try` block as the result of the transaction actually being included in a block on chain

packages/web3-eth/src/utils/get_revert_reason.ts
```diff
@@ -25,6 +25,45 @@ import { DataFormat, DEFAULT_RETURN_FORMAT } from 'web3-utils';
 import { call } from '../rpc_method_wrappers';
 import { RevertReason, RevertReasonWithCustomError } from '../types';
 
+export const parseTransactionError = (error: unknown, contractAbi?: ContractAbi) => {
```
Spacesai1or, line 28: Moved to it's own function so it could make `getRevertReason` more concise and because the new util method `getTransactionError` [uses it](https://github.com/web3/web3.js/blob/6c9a4a45ac2658b2e76565ee659d3a39ef71d149/packages/web3-eth/src/utils/get_transaction_error.ts#L41)

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: I see now that it's being using in [web3-eth-contract](https://github.com/web3/web3.js/blob/f6ff6ce0099c32c44b7e169d0bfdb4a1be842ca6/packages/web3-eth-contract/src/contract.ts#L1056)

Spacesai1or: `web3-eth-contract` was updated to check for `error instanceof ContractExecutionError` instead of relying on an emission of `contractExecutionError` event

packages/web3-errors/src/errors/transaction_errors.ts
```diff
@@ -84,16 +85,21 @@ export class RevertInstructionError extends BaseWeb3Error {
 	}
 }
 
-export class TransactionRevertError extends BaseWeb3Error {
+export class TransactionRevertInstructionError<
```
Spacesai1or, line 88: Renaming to keep parity with [1.x](https://github.com/web3/web3.js/blob/8d369a9e76267df9ce666283521c1fd6d3f77fc4/packages/web3-core-helpers/src/errors.js#L87)

packages/web3-eth-contract/src/contract.ts
```diff
@@ -1053,7 +1053,7 @@ export class Contract<Abi extends ContractAbi>
 		const transactionToSend = sendTransaction(this, tx, DEFAULT_RETURN_FORMAT);
 
 		// eslint-disable-next-line no-void
-		void transactionToSend.on('contractExecutionError', (error: unknown) => {
+		void transactionToSend.on('error', (error: unknown) => {
```
Spacesai1or, line 1059: I removed `contractExecutionError` event emission from `sendTransaction` and instead used the existing `error` event that emits an instance of `ContractExecutionError`

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1068,42 +1085,34 @@ export function sendTransaction<
...
-									transactionFormatted,
-									web3Context,
-									ETH_DATA_FORMAT,
-								)),
-							};
-						}
+								ETH_DATA_FORMAT,
+							)),
+						};
+					}
```
Spacesai1or: Removed from `try/catch` to make `transactionFormatted` available to `getTransactionError` in `catch`

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1325,22 +1349,17 @@ export function sendSignedTransaction<
...
-							{ eth: 'bytes' },
-							signedTransaction,
-							ETH_DATA_FORMAT,
-						);
+					// Formatting signedTransaction to be send to RPC endpoint
+					const signedTransactionFormattedHex = format(
+						{ eth: 'bytes' },
+						signedTransaction,
+						ETH_DATA_FORMAT,
+					);
```
Spacesai1or, line 1384: Removed from `try` block to make available to 

```typescript
const unSerializedTransaction = TransactionFactory.fromSerializedData(
    bytesToBuffer(hexToBytes(signedTransactionFormattedHex)),
    ).toJSON();
```

in `catch` block

packages/web3-eth/src/types.ts
```diff
@@ -45,17 +62,23 @@ export type SendSignedTransactionEvents<ReturnFormat extends DataFormat> = {
...
 
 export interface SendTransactionOptions<ResolveType = TransactionReceipt> {
 	ignoreGasPricing?: boolean;
 	transactionResolver?: (receipt: TransactionReceipt) => ResolveType;
+	contractAbi?: ContractAbi;
 }
 
 export interface SendSignedTransactionOptions<ResolveType = TransactionReceipt> {
 	transactionResolver?: (receipt: TransactionReceipt) => ResolveType;
+	contractAbi?: ContractAbi;
```
Spacesai1or: Used for decoding when creating `TransactionRevertWithCustomError`

CHANGELOG.md
```diff
@@ -1211,10 +1215,12 @@ should use 4.0.1-alpha.0 for testing.
 #### web3-errors
 
 -   The abstract class `Web3Error` is renamed to `BaseWeb3Error` (#5771)
+-   Renamed TransactionRevertError to TransactionRevertInstructionError to remain consistent with 1.x
```
```suggestion
-   Renamed `TransactionRevertError` to `TransactionRevertInstructionError` to remain consistent with `1.x` (#5854)
```

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1068,42 +1085,34 @@ export function sendTransaction<
...
-								...transaction,
-								from: getTransactionFromOrToAttr('from', web3Context, transaction),
-								to: getTransactionFromOrToAttr('to', web3Context, transaction),
-							},
-							ETH_DATA_FORMAT,
-						);
-
-						if (web3Context.handleRevert) {
-							// eslint-disable-next-line no-use-before-define
-							await getRevertReason(
```
jdevcs, line 1083: Whole idea of doing `eth_call ` / tx simulation on evm before sending actual transaction is to save gas in case there is error in transaction and to detect error earlier. So calling `getRevertReason` after transaction status 0x0 ( fail ) at line 1189 via `getTransactionError` is not useful. This should be changed.

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1177,17 +1186,19 @@ export function sendTransaction<
...
-										'Transaction failed',
-										transactionReceiptFormatted,
-									),
-								);
+								promiEvent.emit('error', error);
 							}
-							reject(transactionReceiptFormatted as unknown as ResolveType);
-							return;
+
+							reject(error);
```
jdevcs, line 1228: This will result in inconsistent function behavior. If transaction status is even 0x0 and tx receipt is **available** we should reject with receipt obj and only emit error. as 0x0 can transaction is **mined** but failed.

packages/web3-eth/src/utils/get_transaction_error.ts
```diff
@@ -0,0 +1,84 @@
...
+	transactionFormatted?: TransactionCall,
+	transactionReceiptFormatted?: FormatType<TransactionReceipt, ReturnFormat>,
+	receivedError?: unknown,
+	contractAbi?: ContractAbi,
+) {
+	let reason: string | RevertReason | RevertReasonWithCustomError | undefined;
+
+	if (receivedError !== undefined) {
+		reason = parseTransactionError(receivedError);
+	} else if (web3Context.handleRevert && transactionFormatted !== undefined) {
```
jdevcs: As shared above, eth_call with tx obj should be sent before actual tx and this function is used after sending tx.

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: @jdevcs  `1.x` doesn't send an `eth_call` before every sent transaction, and `eth_call` will not return an error for transactions such as a send to this method:

```solidity
function revertSend(string memory revertString) public {
    require(keccak256(abi.encodePacked(revertString)) != keccak256(abi.encodePacked("revert")), "This is a send revert");
}
```

so relying on getting the revert reason without sending the transaction isn't a complete solution. Instead, I would suggest checking for a revert reason before sending the transaction only if `options.checkRevertBeforeSend` (a new transaction option) is `true`, and we can set it to default to `true` since this would be helpful to new ETH users and more experienced devs will just turn it off if unwanted. Then checking a mined transaction's revert reason if `status === '0x0` and `web3Context.handleRevert === true` (as currently done in this refactor)

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: @jdevcs Rejecting with only the receipt provides no context to the user as to why their transaction failed (or even that it failed if the user doesn't know to check for the `status`). Additionally, users may not be using an event listener for `error` events, so we shouldn't force that as the only option to obtain why the transaction failed. `1.x` also doesn't return the receipt for failed transaction, it throws an error like so:

```typescript
Error: Returned error: execution reverted: This is a send revert
    at Object.ErrorResponse (node_modules/web3-core-helpers/lib/errors.js:28:19)
    at node_modules/web3-core-requestmanager/lib/index.js:300:36
    at node_modules/web3-providers-http/lib/index.js:127:13
  data: '0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000155468697320697320612073656e64207265766572740000000000000000000000'
}
```

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: There also appears to be a discrepancy between providers with returning revert errors when using `eth_call`:

Sending the transaction:

```typescript
{
    from: tempAcc.address,
    value: BigInt('999999999999999999999999999999999999999999999999999999999'),
}
```

yields `err: insufficient funds for gas * price + value: address` when submitted to Geth using `eth_call`, but Ganache doesn't return an error until the transaction is actually submitted

aokiyap78:
66f6938850ac843982d286f93dee0a96cb179d30

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: Addressed by [this](https://github.com/web3/web3.js/pull/5854/commits/bf6d717b5edb77a3a26c9d0bbbccde73a993c3dc) commit

jdevcs: This will not secure users 100% of times but where applicable, if it can save user in few scenarios even in that case it should not be skipped before sending tx, and its more good to check it based on defined flag. Thanks for making this change. LGTM.

packages/web3-eth/src/rpc_method_wrappers.ts
jdevcs: Thanks Wyatt for these discussion points. 

> or even that it failed if the user doesn't know to check for the status

 >users may not be using an event listener for error events
 
 
 I think above mentioned are few assumptions and we dnt have any user survey data for these?
 
How will user get tx hash in case of reject-error? : via `receipt` event listener may be. That falls under:
  >users may not be using an event listener for <del>error</del> events
 
 When a tx is mined (fails or succeeds), the most primary thing to Directly point is **tx hash** ( by default we can return it via receipt which also has status flag), and it can be used for making any number of further queries later explicitly. we have opportunity here for improvements as compared to 1.x due to major bump.
 
 If tx is not mined we can throw error.

jdevcs (commented):
over all LGTM. except Tx reject error points.

avkos (approved):

packages/web3-eth/src/rpc_method_wrappers.ts
Spacesai1or: @jdevcs A brand new user to ETH that's creating their first Dapp is most likely not going to know all the intricacies of an ETH transaction, and I think explicitly erroring with an `Error` is more helpful than returning the transaction receipt the same way we'd return for a successful transaction

For all the errors that `getTransactionError` can return for failed mined transactions, the receipt is a property on the error object and can be accessed if the error is caught

jdevcs (commented):
There are list of tests failing in CI , that needs fix/updates before merging. Thanks

packages/web3-eth/src/rpc_method_wrappers.ts
jdevcs: > For all the errors that getTransactionError can return for failed mined transactions, the receipt is a property on the error object and can be accessed if the error is caught

that can be helpful.

luu-alex (approved):

jdevcs (approved):

Issue #5864, opened 2023-02-22
Disconnect function doesn't exists in types of currentProvider


avkos:
related [discussion](https://github.com/web3/web3.js/discussions/5829)

#### Expected behavior

The function `disconnect` should exist in `web3.currentProvider` type.

Also, to successfully close the connection, user must pass code 1000 to the disconnect function. I think it should be by default

#### Actual behavior
```ts
import Web3 from 'web3';
const web3 = new Web3('ws://127.0.0.1:8545');

(async () => {
	console.log('1', await web3.eth.getBlockNumber());
	console.log('2', await web3.eth.getBalance('0x7ed0e85b8e1e925600b4373e6d108f34ab38a401'));

	// @ts-ignore
	console.log('3', web3.currentProvider.disconnect()); // <-- function doesn't exists in types
})()

```

PR #5865, opened 2023-02-22
Disconnect function doesn't exists in types of currentProvider

avkos:
## Description

#5864

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [ ] I have tested my code.


packages/web3-utils/src/socket_provider.ts
```diff
@@ -216,11 +216,12 @@ export abstract class SocketProvider<
 	 * @param data - The data to be sent to the server
 	 */
 	public disconnect(code?: number, data?: string): void {
+		const disconnectCode = code ?? 1000;
```
Muhammad-Altabba: Where this 1000 came from? Could we please reference a resource to its origin?
And I think it is ideally to put this 1000 in a constant and export it from `web3-types`.

Muhammad-Altabba (changes requested):
I see that `web3.currentProvider?.disconnect()` still gives an error, right? And this is because at web3-core the Web3Context.currentProvider wrongly returns `string` and it has to be converted from 
```ts
public get currentProvider(): SupportedProviders<API> | string | undefined {
}
```
To:
```ts
public get currentProvider(): SupportedProviders<API> | undefined {
}
```
Or what do you think?

packages/web3-utils/src/socket_provider.ts
avkos: ok I will add this as a comment https://github.com/Luka967/websocket-close-codes

avkos:
> I see that `web3.currentProvider?.disconnect()` still gives an error, right? And this is because at web3-core the Web3Context.currentProvider wrongly returns `string` and it has to be converted from
> 
> ```ts
> public get currentProvider(): SupportedProviders<API> | string | undefined {
>     return this.requestManager.provider;
> }
> ```
> 
> To:
> 
> ```ts
> public get currentProvider(): SupportedProviders<API> | undefined {
>     return this.requestManager.provider;
> }
> ```
> 
> Or what do you think?

yes. this is first that I tried. but it doesn't help. if we remove `string` currentProvider somehow related to legacy provider. So  I think we need to have conditional type here

jdevcs (commented):
pls can you also add a test for `web3.currentProvider?.disconnect();`

packages/web3/test/integration/web3.test.ts
```diff
@@ -75,6 +75,16 @@ describe('Web3 instance', () => {
 		expect(() => new Web3()).not.toThrow();
 	});
 
+	it('check disconnect function', async () => {
```
luu-alex, line 82: can you test other `supportedproviders` on top of `clienturl` and check functions exist 👍

avkos: done

packages/web3-providers-ws/test/fixtures/proxy.ts
```diff
@@ -48,14 +48,26 @@ export const createProxy = async (
 		});
 		closeFunc = async () => {
 			await new Promise(resolve => {
+				const timeOut = setTimeout(() => {
+					// eslint-disable-next-line no-console
+					console.log('timeout1');
```
luu-alex: remove console.log

packages/web3-providers-ws/test/fixtures/proxy.ts
```diff
@@ -48,14 +48,26 @@ export const createProxy = async (
...
 					ws.removeAllListeners();
+					clearTimeout(timeOut);
 					resolve(true);
 				});
 				ws.terminate();
 			});
 			await new Promise(resolve => {
+				const timeOut = setTimeout(() => {
+					// eslint-disable-next-line no-console
+					console.log('timeout2');
```
luu-alex: remove debug logs

luu-alex (approved):

luu-alex:
looks good to me, just remove console logs

packages/web3-providers-http/src/index.ts
```diff
@@ -112,7 +112,7 @@ export default class HttpProvider<
 
 	/* eslint-disable class-methods-use-this */
 	public disconnect() {
-		throw new MethodNotImplementedError();
+		// NO OP
```
Muhammad-Altabba: Why to stop throwing on only this specific method on HttpProvider?
I think we should keep this method as it was, similar to other methods in this class.

jdevcs (approved):
a part from changes required ( existing feedback from other members ) , lgtm .

packages/web3-providers-http/src/index.ts
jdevcs: we should throw error in http instead of no action for disconnect.

avkos: ok

avkos: done

Muhammad-Altabba (approved):
Thanks,

PR #5875, opened 2023-02-27
hotfix (BatchResponseHandler) pass the inner error to the response ca…

tenbits:
Hello Team, 

here is the patch-fix to Batch Response Handler - as for now, it **hides** the initial **Error** passed to the callback, and a generic `InvalidResponse` is passed to the outer listener. This is due to the flow:

- when **`err`** is not null - the `results` is null
- `sortResponses` still creates the array with undefined responses for each request
- `map` method creates **Empty Object**  for each result
-  **Empty Object** is not the valid response
- not helpful error is generated `Invalid JSON RPC response: {}` 

Current snippet:

```javascript
this.requestManager.sendBatch(requests, function (err, results) {
        results = sortResponses(results);
        requests.map(function (request, index) {
            return results[index] || { };
        }).forEach(function (result, index) {
            if (requests[index].callback) {
                if (result && result.error) {
                    return requests[index].callback(errors.ErrorResponse(result));
                }

                if (!Jsonrpc.isValidResponse(result)) {
                    return requests[index].callback(errors.InvalidResponse(result));
                }

                try {
                    requests[index].callback(null, requests[index].format ? requests[index].format(result.result) : result.result);
                } catch (err) {
                    requests[index].callback(err);
                }
            }
        });
    });
```

We should add the `error` field to the object, so it can be handled in `ErrorResponse` and passed back to the caller.

### DIFF
--- a/packages/web3-core-requestmanager/src/batch.js
+++ b/packages/web3-core-requestmanager/src/batch.js
@@ -52,7 +52,7 @@ Batch.prototype.execute = function () {
     this.requestManager.sendBatch(requests, function (err, results) {
         results = sortResponses(results);
         requests.map(function (request, index) {
-            return results[index] || {};
+            return results[index] || { error: err };
         }).forEach(function (result, index) {
             if (requests[index].callback) {
                 if (result && result.error) {
@@ -74,7 +74,7 @@ Batch.prototype.execute = function () {
 };
 
 // Sort responses
-Batch.prototype._sortResponses = function (responses) {    
+Batch.prototype._sortResponses = function (responses) {
     return (responses || []).sort((a, b) => a.id - b.id);
 }
 


Muhammad-Altabba:
Many thanks @tenbits for your contribution,
It would be great to accept your MR. However, this would introduce a breaking change. There could be some project that uses web3.js and check for the output to be exactly `Invalid JSON RPC response: {}`. It is considered a breaking change because, if we update this, then that project would break if they upgraded web3.js to the latest. And so, for any breaking change the version would need to be increased.

And currently, we are working on v4.x. The alpha is out and the RC0 is very near. So, I encourage you to check version 4 and if you find any similar issues. Please make an MR there and we can check it freely.

tenbits:
Hi Muhammad, thanks for looking into it, I understand your point. I'll close the PR then and come to the RC as soon as it is out. 

Meanwhile, I have **workaround** the issue with a single additional request. Usually the **`err`** in the callback would mean some network or server issues (_as if a single request in the batch will fail, the error will be in the result_), so after I get the `invalid response` error, I call the first request from the batch as a normal single request, and receive detailed error explanation.

PR #5884, opened 2023-03-02
Nikos/5835/websocket provider keeps important error message back

### LINKED ISSUE #5835: WebsocketProvider keeps important error messages back.

### Current Behavior

It would be great if we would get the errors with their description from the underlying ``websocket`` package shown to us. 

### Expected Behavior

I do get the underlying errors displayed.

### Steps to Reproduce

Try to connect to a WS endpoint where the upgrade is failing. Basically, all error cases where [this](https://github.com/theturtle32/WebSocket-Node/blob/master/lib/WebSocketClient.js#L335) method is used would provide helpful details. Same for [this case](https://github.com/theturtle32/WebSocket-Node/blob/master/lib/WebSocketClient.js#L225).

### Web3.js Version

<=1.8.x

### Environment

- Operating System: 
- Browser: 
- Node.js Version: 
- NPM Version: 

### Anything Else?

_No response_

nikoulai:
## Description

closes #5835 

## Checklist for 1.x:

### DIFF
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -73,6 +73,10 @@ module.exports = {
         if (event) {
             error.code = event.code;
             error.reason = event.reason;
+            console.log("the event description", event.description)
+            if(event.description) {
+                error.description = event.description;
+            }
         }
 
         return error;
--- a/packages/web3-core-helpers/src/errors.js
+++ b/packages/web3-core-helpers/src/errors.js
@@ -73,7 +73,6 @@ module.exports = {
         if (event) {
             error.code = event.code;
             error.reason = event.reason;
-            console.log("the event description", event.description)
             if(event.description) {
                 error.description = event.description;
             }
--- a/packages/web3-providers-ws/src/index.js
+++ b/packages/web3-providers-ws/src/index.js
@@ -162,12 +162,14 @@ WebsocketProvider.prototype._onConnect = function () {
 
 WebsocketProvider.prototype._onConnectFailed = function (event) {
     this.connectFailedDescription = event.toString().split('\n')[0];
-    console.log("The description of the error is: " + this.connectFailedDescription);
     var _this = this;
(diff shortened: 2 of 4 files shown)


packages/web3-core-helpers/src/errors.js
```diff
@@ -73,6 +73,10 @@ module.exports = {
         if (event) {
             error.code = event.code;
             error.reason = event.reason;
+            console.log("the event description", event.description)
```
jdevcs: console.log should be removed

jdevcs (commented):
also changelog update is required

miraclezrus:
Pardon?

miraclezrus:
Ok new at this I'll update it later

packages/web3-providers-ws/src/index.js
```diff
@@ -207,7 +247,11 @@ WebsocketProvider.prototype._addSocketListeners = function () {
     this.connection.addEventListener('message', this._onMessage.bind(this));
     this.connection.addEventListener('open', this._onConnect.bind(this));
     this.connection.addEventListener('close', this._onClose.bind(this));
-};
+    if(this.connection._connection){
+        this.connection._client.removeAllListeners('connectFailed'); //Override the internal listeners, so they don't trigger a `close` event. We want to trigger `_onClose` manually with a description.
```
jdevcs, line 251: As browsers use the native [WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) object + doesn't use/have ws lib client and node uses ws client. so can we more specifically check `_client` here?

jdevcs (approved):

avkos (approved):
LGTM

packages/web3-providers-ws/src/index.js
Muhammad-Altabba: And do we also need to handle the same for the browser WebSocket? Or is it only an issue that we need to handle just for node?

PR #5888, opened 2023-03-05
refactor: Remove the unnecessary chainId parameter

### LINKED ISSUE #5949: refactor: Remove the unnecessary chainId parameter

## Description

The **toChecksumAddress()** function had an unnecessary **chainId** parameter.
In fact, the **chainId** parameter is not used within the **toChecksumAddress()** function, so I've removed it because it might raise questions about why it exists when you want to use that function.
(web3.js/packages/web3-utils/types/index.d.ts)

### Examples/References

_No response_

in74mz:
## Description

Fixes #5949 

The **toChecksumAddress()** function had an unnecessary **chainId** parameter.
In fact, the **chainId** parameter is not used within the **toChecksumAddress()** function, so I've removed it because it might raise questions about why it exists when you want to use that function.
(web3.js/packages/web3-utils/types/index.d.ts)


luu-alex:
Thank you for this, it looks like this was added mistakenly.

luu-alex:
This looks like it'll be a breaking change but its also a redundant parameter that effects the d.ts file.

avkos (approved):

nikoulai:
> This looks like it'll be a breaking change but its also a redundant parameter that effects the d.ts file.

Yep, breaking change. But, in any case, we want to merge it, lint actions should be passing.

in74mz:
> > This looks like it'll be a breaking change but its also a redundant parameter that effects the d.ts file.
> 
> Yep, breaking change. But, in any case, we want to merge it, lint actions should be passing.

Is there anything else I need to work on?

luu-alex:
@in74mz nope, i think you are all good to go. I'll debug this when I have the time. Can you make an issue for this PR and make link it to this PR?

in74mz:
> @in74mz nope, i think you are all good to go. I'll debug this when I have the time. Can you make an issue for this PR and make link it to this PR?

I have just created [this issue](https://github.com/web3/web3.js/issues/5949) and linked it with this PR.

Spacesai1or:
Because `chainId` isn't actually an expected parameter of [toChecksumAddress](https://github.com/web3/web3.js/blob/b4d12726f80f9a887be270d9c2328c5837685c21/packages/web3-utils/src/index.js#L294) I think this is fine to be merged. `chainId` [also never seems](https://github.com/search?q=repo%3Aweb3%2Fweb3.js%20toChecksumAddress&type=code) to be passed to `toChecksumAddress`

luu-alex:
@spacesailor24 Agreed. Will merge after solving  test issue, theres a [lint failure in the tests ](https://github.com/web3/web3.js/actions/runs/4560520082/jobs/8045550764?pr=5888)
Even though the [test](https://github.com/web3/web3.js/pull/5888/files#diff-342990d20631ea7e76d82880317b08d5ed908b58a4add8e9606830f398aac595L28) is removed it still is run for an unknown reason.

https://github.com/web3/web3.js/pull/4841#issuecomment-1065780787 I believe it is the same issue this PR faces
> The tests are pulling expected types being compared to from what is published on NPM, possibly due to its attempt at linearizing a circular dependency, instead of pulling the type definitions from the current package.

I think we should merge this and the other pr I linked. it'll break lint tests until next release, so we should merge them before a cut-off. thoughts?

Spacesai1or:
@luu-alex Yes seems to be an erroneously reported failing test. I'm less inclined to merge #4841 though as it has a higher likelihood of breaking typescript user's code for no real benefit. I understand #4841 is a matter of correctness, but no other users seems to have reported the issue and it isn't an issue in [4.x](https://github.com/web3/web3.js/blob/5391d8a47db5e21f2476fc4c62e02dcb55844c1e/packages/web3-types/src/apis/eth_execution_api.ts#L205)

Muhammad-Altabba:
So, I think this MR should be closed as it is fixed in 4.x version, and because changing it at version 1.x could trouble some users if they updated their npm package after this change came while they expect non-breaking changes to their TypeScript code.

(By the way, there is an EIP (https://eips.ethereum.org/EIPS/eip-1191) for using the chain Id when calculating the checksum. This is the reason for this parameter addition. Even though, it was not implemented in web3.js later.)

Spacesai1or:
> So, I think this MR should be closed as it is fixed in 4.x version, and because changing it at version 1.x could trouble some users if they updated their npm package after this change came while they expect non-breaking changes to their TypeScript code.
> 
> (By the way, there is an EIP (https://eips.ethereum.org/EIPS/eip-1191) for using the chain Id when calculating the checksum. This is the reason for this parameter addition. Even though, it was not implemented in web3.js later.)

@Muhammad-Altabba 1.x users shouldn't really be affected by this change as the actual method [toChecksumAddress](https://github.com/web3/web3.js/blob/4e5afa169a5b0e404b19a8b8d9320189a508935d/packages/web3-utils/src/index.js#L294) doesn't accept the `chainId` parameter

luu-alex (approved):

Muhammad-Altabba:
> > So, I think this MR should be closed as it is fixed in 4.x version, and because changing it at version 1.x could trouble some users if they updated their npm package after this change came while they expect non-breaking changes to their TypeScript code.
> > (By the way, there is an EIP (https://eips.ethereum.org/EIPS/eip-1191) for using the chain Id when calculating the checksum. This is the reason for this parameter addition. Even though, it was not implemented in web3.js later.)
> 
> @Muhammad-Altabba 1.x users shouldn't really be affected by this change as the actual method [toChecksumAddress](https://github.com/web3/web3.js/blob/4e5afa169a5b0e404b19a8b8d9320189a508935d/packages/web3-utils/src/index.js#L294) doesn't accept the `chainId` parameter

Actually in 1.x the `toChecksumAddress` do accept `chainId`. But it does not use it. So, there could be some users who used to pass this parameter (that does not change anything in the calculation). And those users will start to get an error if they updated 1.x to the latest version if we accept the change in this MR.

However, it is Ok if we deal with it as a bug fix. But I think it would be even more beneficial to keep the parameter and implement it according to https://eips.ethereum.org/EIPS/eip-1191. And then the bug we are dealing with would be the issue: "the `chainId` is not taken into account when calculating the checksum".

However, I am Ok with either one of the 2 scenarios, if the other team members think so :smile: .

jdevcs (approved):

jdevcs:
merged in 1.x via https://github.com/web3/web3.js/pull/6057

PR #5902, opened 2023-03-08
Support xdc prefix address for xinfin blockchain

gzliudan:
## Description

The [xinfin blockchain](https://www.xinfin.org/) use `xdc` as address prefix, not `0x`. Below is an example:

```shell
curl -s -X POST https://rpc.apothem.network/ -H "Content-Type: application/json" -d '
{
  "jsonrpc": "2.0",
  "method": "eth_getBlockByNumber",
  "params": ["latest", true],
  "id":1
}' | jq
```

This PR support xdc prefix address by extend the function of `web3-utils` and `web3-core-helpers` packages.

## Type of change

## Checklist for 1.x:

- [X] I ran `npm run build` with success.
- [X] I have tested the built `dist/web3.min.js` in a browser.


Muhammad-Altabba:
Hello @gzliudan 
I am just wondering about the reason why xinfin chose a non-standard prefix for hex strings. As you know there is `chainId` that differentiates Ethereum-compatible networks from each other. You can count over 460 networks at https://chainlist.org/ and the list is still expanding...

And so, just imagine if every EVM-based network will use its own nonstandard prefix for hex strings. After a few years, we will end up with thousands of possible prefixes for addresses that are at the end just hexadecimal strings. And this is something that would not be good to be handled by web3.js and similar libraries. Or what do you think?

jdevcs:
@gzliudan Thanks for your interest and contribution, Web3.js 1.x is currently in minimum maintenance phase and any new features should be added in [4.x. ](https://github.com/web3/web3.js/tree/4.x) .

Issue #5241, opened 2022-07-13
ESM and CJS builds


jdevcs:
We should have support of `ESM` builds along with `CJS` in future 4.x ( beta TBD). 
Currently we have `CJS` build in 4.x and should be offering separate `ESM` and `CJS` builds.

mconnelly8:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/web3js-553a980ab285bae538012a40/issues/ChainSafe/web3.js/5241?planning-poker) with ZenHub @avkos @jdevcs @luu-alex @nazarhussain @nikoulai @spacesailor24

PR #5904, opened 2023-03-08
 ESM CJS builds

jdevcs:
## Description

https://github.com/web3/web3.js/issues/5241

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [ ] I have tested my code.


jdevcs:
node users will need to use `--experimental-specifier-resolution=node` if they want to use our ESM builds, as we are not importing files in our packages with file extension. (This is for ESM node apps, using our ESM build )

packages/web3-eth-accounts/src/account.ts
```diff
@@ -15,7 +15,8 @@ You should have received a copy of the GNU Lesser General Public License
 along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { TransactionFactory, TypedTransaction } from '@ethereumjs/tx';
+import { TypedTransaction } from '@ethereumjs/tx';
+import defaultImport, * as fullImport from '@ethereumjs/tx';
```
jdevcs, line 19: While local testing in ESM app and importing web3 ESM build there was problem in importing `TransactionFactory` from `@ethereumjs/tx`, reason with ESM `@ethereumjs/tx` is not exposing named export but exporting under default but with CJS `@ethereumjs/tx` is exporting it with named export. Following fix is applied in eth and accounts package and it imports both default and all ( named + default ) making sure that we will not miss `TransactionFactory` in either esm or cjs.

jdevcs: This export problem is fixed in @ethereumjs/tx in 4.x and we can upgrade to that in our 4.x after reviewing breaking changes in another PR.

tsconfig.base.json
```diff
@@ -3,8 +3,6 @@
 		"resolveJsonModule": true,
 		"forceConsistentCasingInFileNames": true,
 		"target": "es2016",
-		"module": "commonjs",
```
jdevcs, line 6: Now we have base config for building in root with common configuration items.

packages/web3-core/tsconfig.cjs.json
```diff
@@ -0,0 +1,9 @@
+{
+	"extends": "../../tsconfig.base.json",
```
jdevcs, line 2: each package has three configs for building CommonJS, ESM and type defs, these are extending from base config in root

packages/web3-core/package.json
```diff
@@ -16,7 +24,10 @@
 	"scripts": {
 		"clean": "rimraf dist && rimraf lib",
 		"prebuild": "yarn clean",
-		"build": "tsc --build",
+		"build": "yarn build:cjs && yarn build:esm && yarn build:types",
+		"build:cjs": "tsc --build tsconfig.cjs.json && echo '{\"type\": \"commonjs\"}' > ./lib/commonjs/package.json",
+		"build:esm": "tsc --build tsconfig.esm.json && echo '{\"type\": \"module\"}' > ./lib/esm/package.json",
```
jdevcs, line 29: It was explicitly required to have package json in build dir ( lib ) to explicitly override module type for correctly loading ESM when required so I am generating here a package json in lib

jdevcs, line 29: It was explicitly required to have package json in build dir ( lib ) to explicitly override module type for correctly loading ESM when required so I am generating here a package json in lib

mpetrunic: Woah, this looks so wrong. I don't think this is what you should be doing. There should be only one package.json at root of library.

You might be encountering typescript problem, I think each package.json needs to also have:
```
"main": "lib/cjs/index.js",
"module": "lib/esm/index.js",
```

packages/web3-core/tsconfig.esm.json
```diff
@@ -0,0 +1,8 @@
+{
+	"extends": "../../tsconfig.base.json",
+	"compilerOptions": {
+		"outDir": "./lib/esm",
+		"module": "es2020"
```
mpetrunic, line 5: shouldnt this be "esnext"?

packages/web3-core/package.json
```diff
@@ -2,7 +2,15 @@
...
 	"version": "4.0.1-rc.0",
 	"description": "Web3 core tools for sub-packages. This is an internal package.",
-	"main": "lib/index.js",
+	"main": "./lib/commonjs/index.js",
+	"module": "./lib/esm/index.js",
+	"exports": {
+		".": {
+			"import": "./lib/esm/index.js",
+			"require": "./lib/commonjs/index.js",
+			"types": "./lib/types/index.d.ts"
```
mpetrunic: `types` must always be first in list and `default` last: https://nodejs.org/api/packages.html#community-conditions-definitions

packages/web3-core/package.json
mpetrunic: I don't think you need that for cjs dir

packages/web3-core/tsconfig.esm.json
jdevcs: `esnext` may have [proposals](https://github.com/tc39/proposals/blob/main/finished-proposals.md) that may not ratified in formal release, so its I think more safe to not use esnext in production apps/libs. ( and ok to use in dev versions like alpha beta etc)

packages/web3-core/package.json
jdevcs: Thanks

packages/web3-core/package.json
jdevcs: By default its not required, but I kept it to always override any higher level package config  ( in client app) for surely giving error in case if any module type is loaded / referred by mistake. 
Do you think it will cause any issue in client side if we keep this explicit overriding file?

mpetrunic: No idea tbh

packages/web3-eth-accounts/package.json
```diff
@@ -16,7 +24,10 @@
 	"scripts": {
 		"clean": "rimraf dist && rimraf lib",
 		"prebuild": "yarn clean",
-		"build": "tsc --build",
+		"build": "yarn build:cjs && yarn build:esm && yarn build:types",
```
avkos: We can use `npm concurrently` to run builds in parallel. what do you think?

avkos (changes requested):

packages/web3-eth/src/utils/decode_signed_transaction.ts
```diff
@@ -14,12 +14,14 @@ GNU Lesser General Public License for more details.
 You should have received a copy of the GNU Lesser General Public License
 along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
-import { TransactionFactory } from '@ethereumjs/tx';
+import defaultImport, * as fullImport from '@ethereumjs/tx';
```
Spacesai1or, line 17: Doing this doesn't increase the build size?

packages/web3-eth-accounts/src/account.ts
mpetrunic: Hmm maybe you should try again, now that types are first in export. Maybe also publish repro repo^^

packages/web3-core/package.json
mpetrunic: @jdevcs Maybe it would be better to use nodejs script for this. Afaik this excludes developers on windows to build library^^

packages/web3-eth/src/utils/decode_signed_transaction.ts
jdevcs: No, it shouldn't, because there is one copy of this dependency in node_modules.

packages/web3-eth-accounts/src/account.ts
jdevcs: It isnt issue in web3 exporting, but in ethereumjs/tx lib. I tested in another proj as well.

packages/web3-core/package.json
jdevcs: for creating package.json? I think `echo` is also avalible in win

packages/web3-eth-accounts/package.json
jdevcs: we can do that with & as well.

nikoulai (approved):

avkos (approved):

packages/web3-eth/src/utils/decode_signed_transaction.ts
Spacesai1or: @jdevcs But would this affect tree shaking since we're now importing everything and setting it to a `const` instead of just importing and using the `TransactionFactory` export?

jdevcs: I think this may not affect treeshaking as in case of ESM usage of our lib , dependency usage tree is still construable for module being used. 
```
import defaultImport, * as fullImport from '@ethereumjs/tx';
const { TransactionFactory } = defaultImport || fullImport;
```
Even if any issue arise , the lib under discussion is decided to be removed, or we can upgrade this to latest @ethereumjs/tx where these exports are fixed as mentioned earlier:

> This export problem is fixed in @ethereumjs/tx in 4.x and we can upgrade to that in our 4.x after reviewing breaking changes in another PR.

PR #5915, opened 2023-03-11
Use `gas` and `input` instead of `gasLimit` and `data`

### LINKED ISSUE #5387: `gasLimit` vs `gas`
Originally brought up [here](https://github.com/ChainSafe/web3.js/pull/5312#issue-1328068259) in PR #5312, there are instances where our library will return `gasLimit` and other times when it returns `gas` within transaction objects. This issue tracks the decision and implementation on how the library should handle the use of `gasLimit` vs `gas`

Spacesai1or:
#### web3-core

-   If a transaction object with a `data` property is passed to `txInputOptionsFormatter`, it will now be replaced with `input` (#5915)
#### web3-errors

-   `gasLimit` is no longer accepted as a parameter for `MissingGasError` and `TransactionGasMismatchError`, and is also no longer included in error message (#5915)

#### web3-eth

-   `signTransaction` will now return `gas` instead of `gasLimit` for returned transaction object regardless of what property name the provider uses (#5915)
-   `formatTransaction` will now replace `data` transaction property with `input` (#5915)
-   `isTransactionCall` will now check if `value.input` `isHexStrict` if provided (#5915)

#### web3-eth-contract

-   `getSendTxParams` will now return `input` instead of `data` in returned transaction parameters object (#5915)
-   `Contract` constructor will now thrown new `ContractTransactionDataAndInputError` if both `data` and `input` are passed in `ContractInitOptions` for `Contract` constructor (#5915)

#### web3-types

-   `data` property in `TransactionOutput` was renamed to `input` (#5915)

#### web3-eth-contract

-   `input` is now an acceptable property for `ContractInitOptions` in place of `data` (either can be used, but `input` is used withing the
    `Contract` class) (#5915)

#### web3-eth-contract

-   `data` was removed as a property of `ContractOptions` type (#5915)

closes #5387


Spacesai1or:
So there seems to be a side effect from updating `formatTransaction` to replace `gasLimit` with `gas`: [TransactionGasMismatchError](https://github.com/web3/web3.js/blob/0a8a8ae879fd6b7489120dfc186a3cf817c65b2b/packages/web3-errors/src/errors/transaction_errors.ts#L334) and [MissingGasError](https://github.com/web3/web3.js/blob/0a8a8ae879fd6b7489120dfc186a3cf817c65b2b/packages/web3-errors/src/errors/transaction_errors.ts#L313) use `gasLimit` for their error messages and in some cases where these errors are used we call `formatTransaction` before throwing the error, so `gasLimit` is now `undefined` when it wasn't before this PR

packages/web3-eth/src/utils/transaction_builder.ts
```diff
@@ -166,15 +166,15 @@ export async function defaultTransactionBuilder<ReturnType = Record<string, unkn
...
 	}
 
-	if (isNullish(populatedTransaction.data) || populatedTransaction.data === '') {
-		populatedTransaction.data = '0x';
-	} else if (!populatedTransaction.data.startsWith('0x')) {
-		populatedTransaction.data = `0x${populatedTransaction.data}`;
+	if (isNullish(populatedTransaction.input) || populatedTransaction.input === '') {
+		populatedTransaction.input = '0x';
+	} else if (!populatedTransaction.input.startsWith('0x')) {
+		populatedTransaction.input = `0x${populatedTransaction.input}`;
```
jdevcs, line 177: Can you add unit test for newly added code. Thanks.

Spacesai1or: Yes, this wasn't meant to be marked as ready for review - need to update changelog and migration guide too

Spacesai1or: Added in [this](https://github.com/web3/web3.js/pull/5915/commits/90e708adaaaa359bcf3a2055fa39bae5dbe80b99) commit

jdevcs (approved):

luu-alex (approved):

avkos (approved):

PR #5934, opened 2023-03-17
Nikos/5931/coverage web3 abi

### LINKED ISSUE #5931: Coverage Web3 ABI
push unit tests coverage >= 80% of web3 abi https://app.codecov.io/gh/web3/web3.js/tree/4.x/packages

nikoulai:
## Description

web3-eth-abi coverage 

closes #5931


packages/web3-eth-abi/src/api/parameters_api.ts
```diff
@@ -21,33 +21,6 @@ import { HexString, AbiInput, DecodedParams } from 'web3-types';
 import ethersAbiCoder from '../ethers_abi_coder';
 import { formatParam, isAbiFragment, mapTypes, modifyParams } from '../utils';
 
-/**
- * Helper function to format the decoded object
- */
-const formatDecodedObject = (
```
jdevcs, line 27: Its not used in 4.x currently also we dnt have it in 1.x, but `formatDecodedObject` was part of last few 4.x releases so it should be added in changelog.

jdevcs (approved):
lgtm, just update changelog for `formatDecodedObject` . Thanks

avkos (approved):

packages/web3-eth-abi/src/api/parameters_api.ts
Spacesai1or: @luu-alex You added this function in #4844, was there an intended use for it?

Spacesai1or (approved):

packages/web3-eth-abi/src/api/parameters_api.ts
luu-alex: I think its a helper function I made that I never ended up using, we could get rid of it

nikoulai: mentioned in changelog

PR #5979, opened 2023-04-01
Format `transaction.type` to hex. Add empty `accessList` is `tx.type === '0x1'`

### LINKED ISSUE #4597: Bad input interpretation for `type` parameter to transaction; also exact type is not always respected

### Current Behavior

The `type` parameter to the transaction has some problems with the interpretation of its input.  Currently, these problems don't matter since the largest legal transaction type is 2, but they could in the future.

Specifically:
1. If a decimal string is given, it will be interpreted as hex rather than as decimal.
2. If a number (as `number` or `BN`; I haven't tried `BigNumber` but I assume it's the same) is given, it will be treated as if it were a decimal *string* and have its *decimal digits treated as hex*.  I.e., if you enter the number `10`, it's treated as `16`!  Worse yet, if you enter the number `0x10`, *intending* to get 16, you'll instead get `0x16`!  This one is definitely the worst part, I'd say.
3. Types greater than or equal to 0xc0 are treated as if they were 0, rather than being rejected as illegal.

Also, the exact type entered is not always respected.  For instance if you specify a type of `1` but don't specify an access list, you'll get a transaction of type 0, rather than a transaction of type 1 with an empty access list.

### Expected Behavior

1. Numbers should be treated as the numbers that they are, not have their decimal digits read as hex.
2. Decimal strings, that do not have a `"0x"` prefix, should be treated as decimal, not hex.
3. Types greater than or equal to 0xc0 should be treated as invalid, not 0.
4. The exact type entered should always be respected; type 1 should not be converted to type 0 when no access list is specified (instead the access list should be treated as empty).

### Steps to Reproduce

Examples to try:

```javascript
web3.eth.sendTransaction({ from: accounts[0], to: accounts[1], type: 10 }) //yields 16 instead of 10
web3.eth.sendTransaction({ from: accounts[0], to: accounts[1], type: "10" }) //yields 16 instead of 10
web3.eth.sendTransaction({ from: accounts[0], to: accounts[1], type: 0x10 }) //yields 0x16 instead of 0x10!
web3.eth.sendTransaction({ from: accounts[0], to: accounts[1], type: "0xc0" }) //yields 0 instead of invalid
web3.eth.sendTransaction({ from: accounts[0], to: accounts[1], type: 1 }) //yields 0 rather than 1 unless accessList is also given
```

### Web3.js Version

1.5.3

### Environment

- Operating System: Release Linux Mint 20.2 Uma 64-bit
- Node.js Version: 12.21.0 
- NPM Version: 6.14.11

### Anything Else?

That last one about type 1 should maybe be a separate issue, let me know if you want me to split this up.

Spacesai1or:
- `transaction.type` is now formatted to a hex string before being sent to provider
- When sending a transaction, if `transaction.type === '0x1' && transaction.accessList === 'undefined'`, then `transaction.accessList` is set to `[]`

Added test cases for Steps to Reproduce examples from [here](https://github.com/web3/web3.js/issues/4597#issue-1070998729). With these changes everything is formatted as expected, however, we do not throw an error for `tx.type` exceededing bounds of `0x0` to `0x7f`

closes #4597

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -666,3 +666,8 @@ Released with 1.0.0-beta.37 code base.
 ### Fixed
 
   - Improved the error propagation in `web3-providers-http` package to effectively propagate useful error infomation about failed HTTP connections (#5955)
+
+### Changed
+
+  - `transaction.type` is now formatted to a hex string before being send to provider (#5979)
+  - When sending a transaction, if `transaction.type === '0x1' && transaction.accessList === []`, then `transaction.accessList` is set to `[]` (#5979)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -670,4 +670,4 @@ Released with 1.0.0-beta.37 code base.
 ### Changed
 
   - `transaction.type` is now formatted to a hex string before being send to provider (#5979)
-  - When sending a transaction, if `transaction.type === '0x1' && transaction.accessList === []`, then `transaction.accessList` is set to `[]` (#5979)
+  - When sending a transaction, if `transaction.type === '0x1' && transaction.accessList === undefined`, then `transaction.accessList` is set to `[]` (#5979)
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -799,14 +799,13 @@ Method.prototype.buildCall = function () {
             && !!payload.params[0]
             && typeof payload.params[0] === 'object';
 
-        if (hasSendTxObject) {
-            if (
+        if (hasSendTxObject &&
                 payload.params[0].type === '0x1'
-                &&typeof payload.params[0].accessList === 'undefined'
+                && typeof payload.params[0].accessList === 'undefined'
             ) {
                 payload.params[0].accessList = [];
             }
-        }
+      
 
         // Send the actual transaction
         if (hasSendTxObject
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -795,10 +795,21 @@ Method.prototype.buildCall = function () {
             return method.requestManager.send(payload, sendTxCallback);
         };
 
-        // Send the actual transaction
-        if (isSendTx
+        const hasSendTxObject = isSendTx
             && !!payload.params[0]
-            && typeof payload.params[0] === 'object'
+            && typeof payload.params[0] === 'object';
+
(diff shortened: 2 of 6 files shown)


packages/web3-core-method/src/index.js
```diff
@@ -795,10 +795,21 @@ Method.prototype.buildCall = function () {
...
+            && typeof payload.params[0] === 'object';
+
+        if (hasSendTxObject) {
+            if (
+                payload.params[0].type === '0x1'
+                &&typeof payload.params[0].accessList === 'undefined'
+            ) {
+                payload.params[0].accessList = [];
+            }
+        }
```
jdevcs:
```suggestion
        if (hasSendTxObject &&
                payload.params[0].type === '0x1'
                && typeof payload.params[0].accessList === 'undefined'
            ) {
                payload.params[0].accessList = [];
            }
      
```

jdevcs (commented):
lgtm, except tiny refactoring.

Spacesai1or:
> lgtm, except tiny refactoring.

@jdevcs Merged, thank you

luu-alex (approved):

CHANGELOG.md
```diff
@@ -666,3 +666,8 @@ Released with 1.0.0-beta.37 code base.
 ### Fixed
 
   - Improved the error propagation in `web3-providers-http` package to effectively propagate useful error infomation about failed HTTP connections (#5955)
+
+### Changed
+
+  - `transaction.type` is now formatted to a hex string before being send to provider (#5979)
+  - When sending a transaction, if `transaction.type === '0x1' && transaction.accessList === []`, then `transaction.accessList` is set to `[]` (#5979)
```
avkos: do you mean
`transaction.type === '0x1' && transaction.accessList is undefined`  then `transaction.accessList` is set to `[]`
?

avkos (dismissed):

CHANGELOG.md
```suggestion
  - When sending a transaction, if `transaction.type === '0x1' && transaction.accessList === undefined`, then `transaction.accessList` is set to `[]` (#5979)
```

Spacesai1or: Updated via [this](https://github.com/web3/web3.js/pull/5979/commits/0771e419f76a093bc9617577a5aa7d1d2463a134) commit, thank you

Muhammad-Altabba (approved):

PR #5990, opened 2023-04-05
Init sepolia and mainnet testing

### LINKED ISSUE #5348: Eth Testnet  - 4.x RC Testing Infrastructure
4.x Beta Testing Infrastructure

### Description  

- **eth live NW tests** 
- [x] Tests on at least 1 testnet via Infura (`sepolia network` )
     - [x] read calls (block, account state, balance ..etc)
     - [x] and also setup tests for functions doing changes to be mined in blocks (run once with pre funded account on testnet)
          - [x] value transfer tx, 
          - [x] contract deployment tx, 
          - [x] contract function tx, 
          - [x] 0x0, 0x1 and 0x2 tx
( create a separate test file and use pre funded private key from env variable for live tests on sepolia testnet , do not integrate this in github CI actions)

- [x] Read calls on `mainnet` via Infura ( TBD )

Spacesai1or:
This PR adds E2E tests that read data from the Sepolia test network and mainnet, and write data to the Sepolia network

To run these tests locally:

1. Run `yarn` or manually create the file `.secrets.json` in the root of the project using the template found under `templates/.secrets.json.tmpl`
```json
{
    "SEPOLIA": {
        "HTTP": "",
        "WS": "",
        "ACCOUNT": {
            "address": "0xa127C5E6a7E3600Ac34A9a9928E52521677e7211",
            "privateKey": ""
        },
        "ALLOWED_SEND_TRANSACTION": false
    },
    "MAINNET": {
        "HTTP": "",
        "WS": "",
        "ACCOUNT": {
            "address": "0x0000000000000000000000000000000000000000",
            "privateKey": ""
        },
        "ALLOWED_SEND_TRANSACTION": false
    }
}
```
2. To run the tests that READ data from the chains, you must fill in either `SEPOLIA.HTTP`, `SEPOLIA.WS`, `MAINNET.HTTP`, or `MAINNET.WS` variable with a provider URL
    - To Run the WRITE tests, you must also fill in the `privateKey` variable and set `ALLOWED_SEND_TRANSACTION` to `true`
3. In the root of the project run the test commands
    - `yarn test:e2e:sepolia:http`
    - `yarn test:e2e:sepolia:ws`
    - `yarn test:e2e:mainnet:http`
    - `yarn test:e2e:mainnet:ws` 

---

@jdevcs Should we make the READ tests run in CI?

closes #5348


package.json
```diff
@@ -33,7 +33,7 @@
 	"scripts": {
 		"version": "yarn run bootstrap && yarn build",
 		"bootstrap": "lerna bootstrap",
-		"prebuild": "yarn clean",
+		"prebuild": "yarn clean && ts-node scripts/init.ts",
```
Spacesai1or, line 36: This will create a `.secrets.json` file from the template if it doesn't already exist when installing the project - needed so the import in `packages/web3/test/e2e/e2e_utils.ts` doesn't cause a `file doesn't exist` error

packages/web3/test/e2e/e2e_utils.ts
```diff
@@ -0,0 +1,75 @@
...
+	if (process.env.ALLOWED_SEND_TRANSACTION !== undefined) {
+		return Boolean(process.env.ALLOWED_SEND_TRANSACTION);
+		// eslint-disable-next-line no-else-return
+	} else if (getSystemTestBackend() === 'sepolia' || getSystemTestBackend() === 'mainnet') {
+		return secrets[getSystemTestBackend().toUpperCase() as 'SEPOLIA' | 'MAINNET']
+			.ALLOWED_SEND_TRANSACTION;
+	}
+
+	return false;
+};
```
Spacesai1or: These envs are how we control whether we execute the tests that spend test ether

packages/web3/test/e2e/estimate_gas.test.ts
```diff
@@ -0,0 +1,116 @@
...
+	afterAll(async () => {
+		await closeOpenConnection(web3);
+	});
+
+	it.each(
+		toAllVariants<{
+			transaction: Transaction;
+			block: 'latest' | 'pending' | 'finalized' | 'safe' | 'blockHash' | 'blockNumber';
+			format: string;
+		}>({
```
Spacesai1or, line 57: Most of the tests that read data from the blockchain and are formattable use this `toAllVariants` helper, but being that we already test that data can be formatted as expected in both our unit and integration tests, I don't think this is necessary as it bloats the number of tests significantly (turns a < 100 test to hundreds). It doesn't really hurt to have these tests anyways, it just increases execution time significantly

packages/web3/test/e2e/get_transaction_from_block.test.ts
```diff
@@ -0,0 +1,115 @@
...
+				// to: null,
+				// value: '0x0',
+				// yParity: '0x0'
```
Spacesai1or, line 99: Created #5997 to track this, but not sure what we should do

packages/web3/test/e2e/get_uncle.test.ts
```diff
@@ -0,0 +1,88 @@
...
+				'safe',
+				'finalized',
+				'blockHash',
+				'blockNumber',
+			],
+			uncleIndex: ['0x1', '1', 1, BigInt(1)],
+		}),
+	)('getUncle', async ({ block, uncleIndex }) => {
+		// TODO Returns Position Index is incorrect RPC error
+		// when the client is Nethermind, but not Geth
```
Spacesai1or, line 65: Another provider discrepancy issue

packages/web3/test/e2e/sepolia/get_past_logs.test.ts
```diff
@@ -0,0 +1,127 @@
...
+							hexToBytes(expectedLogs[0].transactionHash as string),
+						),
+						topics: expectedLogs[0].topics?.map(topic =>
+							bytesToBuffer(hexToBytes(topic)),
+						),
+						// TODO Should these be formatted?
+						// blockNumber: bytesToBuffer(hexToBytes((expectedLogs[0]).blockNumber as string)),
+						// data: bytesToBuffer(hexToBytes((expectedLogs[0]).data as string)),
+						// logIndex: bytesToBuffer(hexToBytes((expectedLogs[0]).logIndex as string)),
+						// transactionIndex: bytesToBuffer(hexToBytes((expectedLogs[0]).transactionIndex as string)),
```
Spacesai1or, line 98: I feel like these properties were missed for formatting, should we create an issue to fix this @jdevcs?

packages/web3/test/e2e/mainnet/get_transaction.test.ts
```diff
@@ -0,0 +1,88 @@
...
+			// to: null,
+			// value: '0x0',
+			// yParity: '0x0'
```
Spacesai1or, line 85: #5997

packages/web3/test/e2e/sepolia/contract.test.ts
```diff
@@ -0,0 +1,159 @@
...
+		const result = await web3.eth.sendSignedTransaction(signedTransaction.rawTransaction);
+		deployedContractAddress = result.contractAddress as string;
+
+		// TODO This should work, but throws a type error
+		// for root not being included in expected object.
+		// However, root is not apart of result object.
+		// expect(result).toMatchObject<TransactionReceipt>({
+		// eslint-disable-next-line jest/no-standalone-expect
+		expect(result).toMatchObject({
+			// root: expect.any(String),
```
Spacesai1or, line 73: A type error

packages/web3/test/e2e/mainnet/get_past_logs.test.ts
```diff
@@ -0,0 +1,119 @@
...
+					data: bytesToBuffer(hexToBytes(expectedLog.data as string)),
+					transactionHash: bytesToBuffer(
+						hexToBytes(expectedLog.transactionHash as string),
+					),
+					topics: expectedLog.topics?.map(topic => bytesToBuffer(hexToBytes(topic))),
+					// TODO Should these be formatted?
+					// blockNumber: bytesToBuffer(hexToBytes((expectedLog).blockNumber as string)),
+					// data: bytesToBuffer(hexToBytes((expectedLog).data as string)),
+					// logIndex: bytesToBuffer(hexToBytes((expectedLog).logIndex as string)),
+					// transactionIndex: bytesToBuffer(hexToBytes((expectedLog).transactionIndex as string)),
```
Spacesai1or, line 95: https://github.com/web3/web3.js/pull/5990/files#r1169423998

packages/web3/test/e2e/sepolia/get_transaction.test.ts
```diff
@@ -0,0 +1,88 @@
...
+			// to: null,
+			// value: '0x0',
+			// yParity: '0x0'
```
Spacesai1or, line 85: #5997

packages/web3/test/e2e/sepolia/send_signed_transaction.test.ts
```diff
@@ -0,0 +1,164 @@
...
+			);
+			const result = await web3.eth.sendSignedTransaction(signedTransaction.rawTransaction);
+
+			// TODO This should work, but throws a type error
+			// for root not being included in expected object.
+			// However, root is not apart of result object.
+			// expect(result).toMatchObject<TransactionReceipt>({
+			// eslint-disable-next-line jest/no-standalone-expect
+			expect(result).toMatchObject({
+				// root: expect.any(String),
```
Spacesai1or, line 65: https://github.com/web3/web3.js/pull/5990/files#r1169424808

jdevcs:
> @jdevcs Should we make the READ tests run in CI

Yes we can include read calls in CI as current Infura free account will not be issue :
`100,000 Total Requests/Day` [are allowed](https://www.infura.io/pricing).

packages/web3/test/e2e/get_transaction_from_block.test.ts
jdevcs: There should be support of Nethermind as well along with Geth in our lib. For read JSON RPC calls we can manage it in our library to be consistent with `input` field. 
image

packages/web3/test/e2e/get_uncle.test.ts
jdevcs: Could you post exact error? is it again some field rename issue?

jdevcs: btw uncle blocks are no more in POS NW, but we should support it if some one is using Eth node on different consensus NW.

packages/web3/test/e2e/sepolia/get_past_logs.test.ts
jdevcs: sure.

packages/web3/test/e2e/get_balance.test.ts
```diff
@@ -0,0 +1,107 @@
...
+	)('getBalance', async ({ block, format }) => {
+		const result = await web3.eth.getBalance(getE2ETestAccountAddress(), blockData[block], {
+			number: format as FMT_NUMBER,
+			bytes: FMT_BYTES.HEX,
+		});
+
+		switch (format) {
+			case 'NUMBER_NUMBER':
+				// eslint-disable-next-line jest/no-conditional-expect
+				expect(isNumber(result)).toBeTruthy();
```
jdevcs: Instead of just number type check, this test should be checking correct exact balance of some specific account ( for some specific block number ) on mainnet or sepolia.

packages/web3/test/e2e/get_block.test.ts
```diff
@@ -0,0 +1,95 @@
...
+		if (blockData[block] === 'pending') {
+			b.nonce = '0x0';
+			b.miner = '0x0000000000000000000000000000000000000000';
+			b.totalDifficulty = '0x0';
+		}
+		expect(validator.validateJSONSchema(blockSchema, b)).toBeUndefined();
+
+		if (hydrated && b.transactions?.length > 0) {
+			// eslint-disable-next-line jest/no-conditional-expect
+			expect(b.transactions).toBeInstanceOf(Array<Transaction>);
```
jdevcs: Same here, there should be a test for checking some specific block number and validate our lib returned block data with local copy of block data.

packages/web3/test/e2e/get_block_transaction_count.test.ts
```diff
@@ -0,0 +1,107 @@
...
+		});
+
+		switch (format) {
+			case 'NUMBER_NUMBER':
+				// eslint-disable-next-line jest/no-conditional-expect
+				expect(isNumber(result)).toBeTruthy();
+				break;
+			case 'NUMBER_HEX':
+				// eslint-disable-next-line jest/no-conditional-expect
+				expect(isHexStrict(result)).toBeTruthy();
```
jdevcs: For most of existing tests , I'll suggest to add some definite test cases as well for validating that we are getting and processing correctly in our lib. Like for getting block transaction count instead of :
`expect(isString(result)).toBeTruthy();`
there should be test of getting number of transaction for some specific block and validating that on mainnet or sp nw. 

And same for other eth functions in other test files where applicable. Thanks @spacesailor24 for your efforts.

packages/web3/test/e2e/get_uncle.test.ts
Spacesai1or: @jdevcs This is the error received

```typescript
InvalidResponseError: Returned error: Position Index is incorrect
          at Web3RequestManager._processJsonRpcResponse (web3.js/packages/web3-core/src/web3_request_manager.ts:363:11)
          at Web3RequestManager.<anonymous> (web3.js/packages/web3-core/src/web3_request_manager.ts:207:16)
          at fulfilled (web3.js/packages/web3-core/lib/commonjs/web3_request_manager.js:21:58)
          at processTicksAndRejections (node:internal/process/task_queues:95:5) {
        innerError: { code: -32602, message: 'Position Index is incorrect' },
        code: 101,
        data: undefined,
        request: {
          jsonrpc: '2.0',
          id: '8f872b5e-9a66-486b-9c74-f2699167191a',
          method: 'eth_getUncleByBlockNumberAndIndex',
          params: [ '0x317340', '0x0' ]
        }
      }
```

I'm assuming the error is because there are no longer uncles, as you've mentioned, but Geth doesn't throw an error for the same request. I don't think we should do anything special for this circumstance as it's up to the provider whether they through an error for this or not, and web3.js is simply just passing along the error response to the user

packages/web3/test/e2e/sepolia/get_past_logs.test.ts
Spacesai1or: #6027 tracks this

spacesailor24:
> > @jdevcs Should we make the READ tests run in CI
> 
> Yes we can include read calls in CI as current Infura free account will not be issue : `100,000 Total Requests/Day` [are allowed](https://www.infura.io/pricing).

I mentioned it [here](https://github.com/web3/web3.js/pull/5990#discussion_r1169422049), but because we use `toAllVariants` in these tests, reaching 100,000 requests could be done by running these tests 10 times or so

Spacesai1or:
> > @jdevcs Should we make the READ tests run in CI
> 
> Yes we can include read calls in CI as current Infura free account will not be issue : `100,000 Total Requests/Day` [are allowed](https://www.infura.io/pricing).

@jdevcs I mentioned it [here](https://github.com/web3/web3.js/pull/5990#discussion_r1169422049), but because we use `toAllVariants` in these tests, reaching 100,000 requests could be done by running these tests 10 times or so

packages/web3/test/e2e/get_block_transaction_count.test.ts
Spacesai1or: @jdevcs This is hard to do when using block tags because anything other than `earliest` will return a recent block where the properties are non-deterministic for our tests e.g. we won't know what the block transaction count will be for the `latest` block, we won't know the block number, hash, etc. of the `latest` block, and we won't know the balance of the account for these block tags

Spacesai1or: We should be able to check exact properties when we make the request using a block hash or number though, so I'll update the tests to do that

jdevcs: You are right, Thats why I requested to :

> there should be test of getting number of transaction for some `specific block` and validating that on mainnet or sp nw.

jdevcs:
> 


> > > @jdevcs Should we make the READ tests run in CI
> > 
> > 
> > Yes we can include read calls in CI as current Infura free account will not be issue : `100,000 Total Requests/Day` [are allowed](https://www.infura.io/pricing).
> 
> @jdevcs I mentioned it [here](https://github.com/web3/web3.js/pull/5990#discussion_r1169422049), but because we use `toAllVariants` in these tests, reaching 100,000 requests could be done by running these tests 10 times or so


In that case we should invoke read tests on release tag ( like we do with blackbox tests )

packages/web3/test/e2e/get_block_transaction_count.test.ts
Spacesai1or: @jdevcs Updated via [this](https://github.com/web3/web3.js/pull/5990/commits/dfd111793b4dab26eee936567e84a3f8e3d7965b) commit

Spacesai1or:
> Init e2e_network_test Github workflow

@jdevcs [Added](https://github.com/web3/web3.js/pull/5990/commits/bb082d9b89ab49cbe98f6091599152f1f496ee16) workflow for running on release branches, ~although I haven't tested it~ You can see [here](https://github.com/web3/web3.js/actions/runs/4751101517/jobs/8440056174) that the tests are passing

I also added a `E2E_TESTS_ALLOWED_SEND_TRANSACTION` secret to Github that can be changed to anything other than `false` to trigger tests that send eth to run

jdevcs:
> > Init e2e_network_test Github workflow
> 
> @jdevcs [Added](https://github.com/web3/web3.js/pull/5990/commits/bb082d9b89ab49cbe98f6091599152f1f496ee16) workflow for running on release branches, ~although I haven't tested it~ You can see [here](https://github.com/web3/web3.js/actions/runs/4751101517/jobs/8440056174) that the tests are passing
> 
> I also added a `E2E_TESTS_ALLOWED_SEND_TRANSACTION` secret to Github that can be changed to anything other than `false` to trigger tests that send eth to run

This should be added in release document, for future developers, before merging.

jdevcs (approved):
Good work @spacesailor24

luu-alex (approved):

nikoulai (approved):

Spacesai1or:
> Update RELEASE.md to include info about E2E testing

Update via [this](https://github.com/web3/web3.js/pull/5990/commits/bc626006dc044039514debb919ecc6a86cc13c57) commit

PR #6000, opened 2023-04-10
Fix error: "n.data.substring is not a function"

### LINKED ISSUE #5999: Error: "`n.data.substring` is not a function", when setting `web.eth.handleRevert = true`

### Current Behavior

When setting `web.eth.handleRevert = true` and calling any Smart Contract method that reverts with an error:
The following error is raised: "`n.data.substring` is not a function".

### Expected Behavior

To return an error that contains the actual error from the Smart Contract.

### Steps to Reproduce

The full details are at: https://github.com/web3/web3.js/issues/5677#issuecomment-1499986255

### Web3.js Version

1.8.2

### Environment

_No response_

### Anything Else?

The current version of the Infura provider seems to be updated to put the error inside `data.data`. And here is what the internal object looks like with Infura:
image
 

And so we need to update web3.js to handle this. And here is the line that needs to be changed:
https://github.com/web3/web3.js/blob/2948927a18a2424621743294a80ea3ee5bf0e1e9/packages/web3-core-method/src/index.js#L629
And it could be changed to something like:
```ts
    reasonData = (err.data.data || err.data).substring(10);
```

Muhammad-Altabba:
## Description

Fixes #5999 

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [ ] I have tested my code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -666,7 +666,7 @@ Released with 1.0.0-beta.37 code base.
 ### Fixed
 
   - Improved the error propagation in `web3-providers-http` package to effectively propagate useful error infomation about failed HTTP connections (#5955)
-
+  - Fix error: "n.data.substring is not a function", that is raised when there is a revert and `web.eth.handleRevert = true` (#6000)
 ### Changed
 
   - `transaction.type` is now formatted to a hex string before being send to provider (#5979)
--- a/packages/web3-core-method/src/index.js
+++ b/packages/web3-core-method/src/index.js
@@ -653,7 +653,7 @@ Method.prototype.buildCall = function () {
                 if (!err && method.isRevertReasonString(result)){
                     reasonData = result.substring(10);
                 } else if (err && err.data){
-                    reasonData = err.data.substring(10);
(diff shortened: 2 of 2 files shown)


avkos (changes requested):
Coverage decreased. It will be great to have unit test which cover these changes

Muhammad-Altabba:
> Coverage decreased. It will be great to have unit test which cover these changes

As a continuation of our chat: 

> The fixed issue is not presented when dealing with the local provider. And because of that the tests are passing and the issue is not detectable with our tests. But it exists when using Infura as it seems to have a slightly different data structure. And I tested this change on my local machine when connecting to Infura and it works well after the code modification.

So because we do not have a test setup with Infura and because this is not a case that can be produced with a local provider. And because it does not introduce any breaking changes (as it use `||` to get the object). I think it is fair enough to depend on a local test. Since doing a unit test (and not an integration test) for this seems to be meaningless. Or do have other thoughts?

Thanks,

avkos (approved):

nikoulai (approved):

PR #6004, opened 2023-04-11
Replace Buffer with uint8array

mpetrunic:
## Description

Alternative to #6005 

Replaces all usage of Buffer with Uint8Array.
Pros:
- even smaller build without the need for polyfill
- if we don't do it now it needs to wait for 5.x
Cons:
- some tests are falling and I don't have time to investigate those as I'm going on a vacation
- break some API

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist for 1.x:

- [ ] I have selected the correct base branch.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules.
- [ ] I ran `npm run dtslint` with success and extended the tests and types if necessary.
- [ ] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code.
- [ ] I ran `npm run build` with success.
- [ ] I have tested the built `dist/web3.min.js` in a browser.
- [ ] I have tested my code on the live network.
- [ ] I have checked the Deploy Preview and it looks correct.
- [ ] I have updated the `CHANGELOG.md` file in the root folder.

## Checklist for 4.x:

- [ ] I have selected the correct 4.x base branch.
- [ ] Within the description, the feature or issue is discussed in detail or an issue is linked.   
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have added any required tests for the changes I made 
- [ ] I have made corresponding changes to the documentation
- [ ] Any dependent changes have been merged and published in downstream modules.
- [ ] I ran `yarn` successfully
- [ ] I ran `yarn lint` successfully
- [ ] I ran `yarn build:web` successfully
- [ ] I ran `yarn test:unit` successfully
- [ ] I ran `yarn test:integration` successfully
- [ ] I ran `compile:contracts` successfully
- [ ] I have tested my code.
- [ ] I have updated the corresponding `CHANGELOG.md` file in the packages I have edited.

PR #6033, opened 2023-04-21
6004 replace buffer with uint8array

luu-alex:
## Description
Supersecedes #6004 and fixes the broken testcases 

## Type of change

## Checklist for 1.x:

## Checklist for 4.x:

- [ ] I have tested my code.


boy520520456:
> ## Description
> 
> Please include a summary of the changes and be sure to follow our [Contribution Guidelines](../blob/1.x/CONTRIBUTIONS.md).
> 
> 
> 
> ## Type of change
> 
> 
> 
> - [ ] Bug fix (non-breaking change which fixes an issue)
> - [ ] New feature (non-breaking change which adds functionality)
> - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
> 
> ## Checklist for 1.x:
> 
> - [ ] I have selected the correct base branch.
> - [ ] I have performed a self-review of my own code.
> - [ ] I have commented my code, particularly in hard-to-understand areas.
> - [ ] I have made corresponding changes to the documentation.
> - [ ] My changes generate no new warnings.
> - [ ] Any dependent changes have been merged and published in downstream modules.
> - [ ] I ran `npm run dtslint` with success and extended the tests and types if necessary.
> - [ ] I ran `npm run test:cov` and my test cases cover all the lines and branches of the added code.
> - [ ] I ran `npm run build` with success.
> - [ ] I have tested the built `dist/web3.min.js` in a browser.
> - [ ] I have tested my code on the live network.
> - [ ] I have checked the Deploy Preview and it looks correct.
> - [ ] I have updated the `CHANGELOG.md` file in the root folder.
> 
> ## Checklist for 4.x:
> 
> - [ ] I have selected the correct 4.x base branch.
> - [ ] Within the description, the feature or issue is discussed in detail or an issue is linked.   
> - [ ] I have performed a self-review of my own code
> - [ ] I have commented my code, particularly in hard-to-understand areas.
> - [ ] I have added any required tests for the changes I made 
> - [ ] I have made corresponding changes to the documentation
> - [ ] Any dependent changes have been merged and published in downstream modules.
> - [ ] I ran `yarn` successfully
> - [ ] I ran `yarn lint` successfully
> - [ ] I ran `yarn build:web` successfully
> - [ ] I ran `yarn test:unit` successfully
> - [ ] I ran `yarn test:integration` successfully
> - [ ] I ran `compile:contracts` successfully
> - [ ] I have tested my code.
> - [ ] I have updated the corresponding `CHANGELOG.md` file in the packages I have edited.
>

packages/web3-utils/src/converters.ts
```diff
@@ -173,12 +174,17 @@ export const toDecimal = hexToNumber;
  * > "0xa"
  * ```
  */
-export const numberToHex = (value: Numbers): HexString => {
+export const numberToHex = (value: Numbers, hexstrict?: boolean): HexString => {
```
luu-alex, line 176: Added an optional parameter hexstrict,

if a user tries to do `HexToBytes(numberToHex(3))` this will fail do to `0x3` not being hexstrict, 

numberToHex(3, true) will return `0x03` allowing `HexToBytes(numberToHex(3, true))`  to work

packages/web3-eth-accounts/CHANGELOG.md
```diff
@@ -84,3 +84,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   The method `signTransaction` returned by `privateKeyToAccount` is now accepting the type `Transaction` for its argument. (#5993)
 
 ## [Unreleased]
+
+### Changed
+
+-   Replaced `Buffer` for `Uint8Array` (#6004)
+-   The methods `recover`, `encrypt`, `privateKeyToAddress` does not support type `Buffer` but supports type `Uint8Array` (#6004)
+-   The method parseAndValidatePrivateKey returns a type `Uint8Array` instead of type `Buffer` (#6004)
```
Spacesai1or:
```suggestion
-   The method `parseAndValidatePrivateKey` returns a type `Uint8Array` instead of type `Buffer` (#6004)
```

packages/web3-eth-accounts/src/common/types.ts
```diff
@@ -155,55 +148,44 @@ export type PrefixedHexString = string;
...
  */
-export type BufferLike =
-	| Buffer
-	| Uint8Array
-	| number[]
-	| number
-	| bigint
-	| TransformableToBuffer
-	| PrefixedHexString;
+export type Uint8ArrayLike = Buffer | Uint8Array | number[] | number | bigint | PrefixedHexString;
```
Spacesai1or: It's okay for `Buffer` to be included here and on `Line 139`?

packages/web3-utils/src/uint8array.ts
Spacesai1or, line 1: We should add tests for these new util methods

packages/web3-validator/src/utils.ts
```diff
@@ -376,3 +376,30 @@ export const padLeft = (value: ValidInputTypes, characterAmount: number, sign =
...
+	if (value.length % 2 !== 0) {
+		throw new InvalidBytesError('hex string has odd length');
+	}
+	const bytes = new Uint8Array(Math.ceil(value.length / 2));
+	for (let i = 0; i < bytes.length; i += 1) {
+		const byte = parseInt(value.substring(i * 2, i * 2 + 2), 16);
+		bytes[i] = byte;
+	}
+	return bytes;
+}
```
Spacesai1or: We should have tests for these too, especially since the logic is non-trivial

packages/web3-eth-accounts/src/common/types.ts
mpetrunic: Nope, there is no need for that as every Buffer is Uint8Array. This will cause this type to be "any" in env where Buffer global is not available^^

packages/web3-validator/src/validation/bytes.ts
```diff
@@ -16,26 +16,21 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
 import { ValidInputTypes } from '../types';
-import { parseBaseType } from '../utils';
+import { hexToUint8Array, parseBaseType } from '../utils';
 import { isHexStrict } from './string';
 
 /**
  * checks input if typeof data is valid buffer input
  */
-export const isBuffer = (data: ValidInputTypes) => Buffer.isBuffer(data);
+export const isBuffer = (data: ValidInputTypes) => data instanceof Uint8Array;
```
jdevcs: `isBuffer` should be removed as we should be accepting Uint8Array inputs only,
 like here https://github.com/web3/web3.js/blob/2258b2cbb4a7c9e15cbe6dbd07d29f10e4da9350/packages/web3-eth-accounts/src/account.ts#L82

packages/web3-utils/src/converters.ts
```diff
@@ -70,36 +72,29 @@ export const ethUnitMap = {
...
+ * @returns - The Uint8Array representation of the input data
  *
  * @example
  * ```ts
  * console.log(web3.utils.bytesToBuffer(new Uint8Array([72, 12])));
  * > <Buffer 48 0c>
  * ```
  */
-export const bytesToBuffer = (data: Bytes): Buffer | never => {
+export const bytesToBuffer = (data: Bytes): Uint8Array | never => {
```
jdevcs:
```suggestion
export const bytesToUint8Array = (data: Bytes): Uint8Array | never => {
```
should it be renamed for more accurate info, as it will convert string or hex string to uint 8 array https://github.com/web3/web3.js/blob/2258b2cbb4a7c9e15cbe6dbd07d29f10e4da9350/packages/web3-types/src/primitives_types.ts#L19

packages/web3-utils/src/uint8array.ts
```diff
@@ -0,0 +1,50 @@
...
+		result.set(part, offset);
+		offset += part.length;
+	}
+	return result;
+}
+
+/**
+ * Returns true if the two passed Uint8Arrays have the same content
+ */
+export function uint8ArrayEquals(a: Uint8Array, b: Uint8Array): boolean {
```
jdevcs, line 34: uint8ArrayEquals and other functions should be covered with unit tests.

packages/web3-validator/src/validation/numbers.ts
```diff
@@ -74,6 +74,10 @@ export const isInt = (
 		return false;
 	}
 
+	if (typeof value === 'number' && value > Number.MAX_SAFE_INTEGER) {
```
nikoulai, line 77: Is it possible to make both conditions true at the same time? 
To make the second condition true shouldn't the value be of type `BigInt`?

packages/web3-utils/src/random.ts
```diff
@@ -28,7 +29,7 @@ import { getRandomBytesSync } from 'ethereum-cryptography/random';
  * > <Buffer a9 a2 70 ff 00 9d 0b c9 2f 9e 5f 0e 40 a4 da 4a f9 1c 6f 23 41 59 46 a6 b5 8b 99 49 72 01 68 99>
```
nikoulai: We need to update the docs here.

jdevcs (commented):
following unit test needs to be checked:
```
web3-utils: FAIL test/unit/uint8array.test.ts
web3-utils:   ● uint8Array utils › uint8ArrayConcat › uint8ArrayConcat › cases › [ [Uint8Array], [Uint8Array] ]
web3-utils:     expect(received).toEqual(expected) // deep equality
```

packages/web3-validator/src/validation/numbers.ts
luu-alex: Yes it is possible, a type number can be greater than `MAX_SAFE_INTEGER` but it may lose precision

packages/web3-utils/src/converters.ts
luu-alex: updated

packages/web3-utils/src/uint8array.ts
luu-alex: added tests for uint8array utils

Spacesai1or:
[This](https://github.com/web3/web3.js/actions/runs/4823162006/jobs/8591946987?pr=6033#step:6:727) looks like a legitimate error

Spacesai1or:
I imagine `sign` is covered by unit tests elsewhere, but we should probably create an issue to create unit tests for these methods so completeness

image

luu-alex:
@spacesailor24  This error has been present in other jobs like [integration (14, ganache, ws) ](https://github.com/web3/web3.js/actions/runs/4813268502/jobs/8569827809) on earlier commits but would pass after running a couple of times. I tried producing this issue locally but [End-to-End ganache:ws (14, firefox)](https://github.com/web3/web3.js/actions/runs/4823162006/jobs/8646772704#logs) pass.

Edit: after running the failed test a third time it works, i think there might be a chance of a 65 length string will be produced somehow and causes the tests to fail, in particular it happens to the getblock function

Spacesai1or:
> @spacesailor24 This error has been present in other jobs like [integration (14, ganache, ws) ](https://github.com/web3/web3.js/actions/runs/4813268502/jobs/8569827809) on earlier commits but would pass after running a couple of times. I tried producing this issue locally but [End-to-End ganache:ws (14, firefox)](https://github.com/web3/web3.js/actions/runs/4823162006/jobs/8646772704#logs) pass.
> 
> Edit: after running the failed test a third time it works, i think there might be a chance of a 65 length string will be produced somehow and causes the tests to fail, in particular it happens to the getblock function

@luu-alex Oh no worries then, I thought it was only the two `Build / Integration IPC (14/16, geth, ipc) (pull_request)` tests that were failing - thank you for checking though

luu-alex:
@spacesailor24  Created an issue for sign unit tests cod cov https://github.com/web3/web3.js/issues/6055

packages/web3-eth/test/integration/block/rpc.getBlock.test.ts
```diff
@@ -111,6 +111,7 @@ describe('rpc with block', () => {
 				format: Object.values(FMT_NUMBER),
 			}),
 		)('getBlock', async ({ hydrated, block, format }) => {
+			console.log(await web3Eth.getBlock(blockData[block], hydrated))
```
avkos: please remove console.log

avkos (changes requested):

packages/web3-eth/test/integration/block/rpc.getBlock.test.ts
luu-alex: removed

Spacesai1or:
> [This](https://github.com/web3/web3.js/actions/runs/4823162006/jobs/8591946987?pr=6033#step:6:727) looks like a legitimate error

After some investigation, it looks this test fails when Ganache is returning a transaction with a signature that has an `r` value of < 33 bytes (in this case the `r` value is 65 characters long), which is apparently allowed to happen. I tested a transaction with the shorter `r` value using ethers and it seems like ethers pads the value with `0` at the beginning:

```tpyescript
0x3ddadb60ab671088edd732b21ea6de0d9c073c06ca799538c2d2d153f403928    // What 4.x returns
0x3ddadb60ab671088edd732b21ea6de0d9c073c06ca799538c2d2d153f403928   // What 1.x returns
0x03ddadb60ab671088edd732b21ea6de0d9c073c06ca799538c2d2d153f403928 // What ethers@v5 returns
```

mpetrunic:
> > [This](https://github.com/web3/web3.js/actions/runs/4823162006/jobs/8591946987?pr=6033#step:6:727) looks like a legitimate error
> 
> After some investigation, it looks this test fails when Ganache is returning a transaction with a signature that has an `r` value of < 33 bytes (in this case the `r` value is 65 characters long), which is apparently allowed to happen. I tested a transaction with the shorter `r` value using ethers and it seems like ethers pads the value with `0` at the beginning:
> 
> ```
> 0x3ddadb60ab671088edd732b21ea6de0d9c073c06ca799538c2d2d153f403928    // What 4.x returns
> 0x3ddadb60ab671088edd732b21ea6de0d9c073c06ca799538c2d2d153f403928   // What 1.x returns
> 0x03ddadb60ab671088edd732b21ea6de0d9c073c06ca799538c2d2d153f403928 // What ethers@v5 returns
> ```

bytes array cannot have hex string with odd amount of characters (each byte is represented by 2 characters). It could be that it worked before because Buffer when given odd length hex string just drops last character internally.
image


Pretty sure the correct behavior is to pad with zero as ethers does.

Spacesai1or:
> > > [This](https://github.com/web3/web3.js/actions/runs/4823162006/jobs/8591946987?pr=6033#step:6:727) looks like a legitimate error
> > 
> > 
> > After some investigation, it looks this test fails when Ganache is returning a transaction with a signature that has an `r` value of < 33 bytes (in this case the `r` value is 65 characters long), which is apparently allowed to happen. I tested a transaction with the shorter `r` value using ethers and it seems like ethers pads the value with `0` at the beginning:
> > ```
> > 0x3ddadb60ab671088edd732b21ea6de0d9c073c06ca799538c2d2d153f403928    // What 4.x returns
> > 0x3ddadb60ab671088edd732b21ea6de0d9c073c06ca799538c2d2d153f403928   // What 1.x returns
> > 0x03ddadb60ab671088edd732b21ea6de0d9c073c06ca799538c2d2d153f403928 // What ethers@v5 returns
> > ```
> 
> bytes array cannot have hex string with odd amount of characters (each byte is represented by 2 characters). It could be that it worked before because Buffer when given odd length hex string just drops last character internally. image
> 
> Pretty sure the correct behavior is to pad with zero as ethers does.

@mpetrunic Interestingly, we only found this issue because Ganache returns the unpadded value. Don't believe Geth does this though as the test only fails for Ganache

jdevcs:
thanks @mpetrunic for pinning it down, and @spacesailor24 for your findings. 
There must be an even number of hex digits to converting to bytes, else discarding a digit is not valid. 

Based on team discussion we will open issue in ganache to return valid hex value ( padded ) for getblock with txs like geth does.

luu-alex:
@jdevcs was taking a look at our failing testcases and [geth fails sometimes](https://github.com/web3/web3.js/actions/runs/4855166031/jobs/8653614670) meaning its not a ganache issue. Looking around [stackexchange ](https://ethereum.stackexchange.com/questions/51673/what-is-the-size-of-the-web3-eth-sign-result/51720#51720) it seems that r value can be less than 32 bytes. I believe this change is needed to be on our side and to add padding to equal 32 bytes, ill create an issue .

Spacesai1or (approved):

avkos (approved):

PR #6047, opened 2023-04-27
Remove & Replace the AJV dependency

avkos:
## Description

## Type of change

## Checklist:


packages/web3-validator/src/formats/address.ts
Spacesai1or, line 1: What are all these `formats/` files used for? Can we not just use the functions they're wrapping directly?

packages/web3-validator/src/types.ts
```diff
@@ -63,8 +74,96 @@ export type Web3ValidationOptions = {
 	readonly silent: boolean;
 };
 
-// In `JSONSchemaType` from `ajv` the `type` is required
-// We need to make it optional
-export type JsonSchema = Optional<SomeJSONSchema, 'type'> & {
+// is-my-json-valid types
+export type Json = string | number | boolean | Array<Json> | { [id: string]: Json };
+
+export type Schema = {
```
Spacesai1or, line 80: We should document where this came from

Spacesai1or: Just for reference, `is-my-json-valid` has a [types declaration file](https://github.com/mafintosh/is-my-json-valid/blob/master/index.d.ts), but it's not very useful

avkos: Yep. It is very poor.  I do not want to use it

packages/web3-validator/src/formats/address.ts
avkos: I can move it into the validator class. they needs only there

jdevcs (commented):
some of tests are not passing , is it due to same reason mention in https://github.com/web3/web3.js/pull/6033#issuecomment-1531348928 ? @avkos

avkos:
> #6033 (comment)

I will check it today. thanks for the link

packages/web3-eth/test/integration/block/rpc.getBlock.test.ts
```diff
@@ -141,11 +141,11 @@ describe('rpc with block', () => {
 			) {
 				// eslint-disable-next-line prefer-destructuring
 				schema.properties.transactions = schema.properties.transactions.oneOf[0];
-				// @ts-expect-error add leading zeros
+				// @ts-expect-error add leading zeros remove when fixes https://github.com/web3/web3.js/issues/6060
```
jdevcs, line 144: I think let these tests to fail here, instead of temp fix, and these will be fixed with https://github.com/web3/web3.js/issues/6060 @luu-alex

packages/web3-validator/src/web3_validator.ts
```diff
@@ -15,59 +15,29 @@ You should have received a copy of the GNU Lesser General Public License
...
-			strictTuples: false,
-		});
-
-		this._validator.addKeyword(ethKeyword);
-
-		for (const formatName of Object.keys(formats)) {
-			// eslint-disable-next-line import/namespace
-			this._validator.addFormat(formatName, formats[formatName as keyof typeof formats]);
-		}
+		this._validator = new Validator();
```
jdevcs: should we have singleton Validator so we ensure that only one entry is made into its map for each schema.

jdevcs: and that there is only one instance of validator and one map across lib for managing schemas.

jdevcs (commented):
Great work @avkos , LGTM, I added few discussion points.

luu-alex (approved):

packages/web3-validator/src/web3_validator.ts
avkos: I've committed a singleton

Muhammad-Altabba (commented):
Great job @avkos,
We need to document the version of library(ies) from which some part of the code has been copied.

And the [license](https://github.com/ExodusMovement/schemasafe/blob/master/LICENSE) of the [ExodusMovements](https://github.com/ExodusMovement/schemasafe/blob/master/index.d.ts) from which the definition of `type Schema` had been copied sates:

> The above copyright notice and this permission notice shall be included in all
> copies or substantial portions of the Software.

Is this requirement fulfilled?

Thanks,

jdevcs (approved):

PR #6050, opened 2023-04-27
Fix typo in readme

Shpota:
## Description

Change "Errors Objects" to "Error Objects"


jdevcs (approved):

Muhammad-Altabba (dismissed):
Thanks @Shpota for your tries to help raise the quality of the readme file.

However, the correct word here is `Errors` and not `Error`. Because the package has multiple objects that belong to multiple errors (not just one error).

If we say "Error Objects" then it means the objects are related to only one error. But the actual case is that the package has multiple objects of errors.

You may check with a professional native English speaker if you are still not convinced. Or else, please, close this PR if you are good with  "Errors Objects". :smile: 

Thanks,

Shpota:
Thank you, @Muhammad-Altabba ! I am certainly okay with using "Errors Objects". It just caught my eye because it looks grammatically incorrect. I am not an English expert to debate on this, and I don't know any experts in English, but "Errors Objects" doesn't indicate in any way that the package has multiple objects that belong to multiple errors, nor does "Error Objects" deny it. Here is what ChatGPT says about it.
image
Let me know what you think and I will close the PR (or close it yourself).

Muhammad-Altabba:
> Thank you, @Muhammad-Altabba ! I am certainly okay with using "Errors Objects". It just caught my eye because it looks grammatically incorrect. I am not an English expert to debate on this, and I don't know any experts in English, but "Errors Objects" doesn't indicate in any way that the package has multiple objects that belong to multiple errors, nor does "Error Objects" deny it. Here is what ChatGPT says about it. image Let me know what you think and I will close the PR (or close it yourself).

Thanks @Shpota,
Unfortunately, you can never depend on ChatGPT as a reliable reference. This is because of many things and one of them is the answer depends on how you present the question. You used the term "error classes" in your question. I replaced it with "classes of errors" and got a different answer :smile:  

image

So, I will leave this to other team members to decide. And will try to consult a native English speaker who is also a developer :smile:, if I could.

All the best,

luu-alex:
I'm not sure what grammar rule this would but both words as plural, "errors objects" doesn't sound correct. 
But as Muhammad stated "Error objects" isn't accurate of what the package is about.

Looking at metamasks error [package they just state "JSON-RPC Errors"](https://github.com/MetaMask/rpc-errors#metamaskrpc-errors)
An alternative description we can use would be "A package of web3js errors", what do you think?

Shpota:
Thanks guys, closing it.

PR #6054, opened 2023-04-30
Refactor `rpc_method_wrapper` docs to point to `Web3Eth` docs

### LINKED ISSUE #5295: Web3Eth Doc Improvements
- We should implement a disclaimer as mentioned [here](https://github.com/ChainSafe/web3.js/pull/5251#discussion_r932016753) that notifies the user that methods exported from `rpc_method_wrappers.ts` are expected to not be used by most users
- We should investigate how to use `@inheritDoc` to avoid duplicate code between `Web3Eth` and `rpc_method_wrappers` as mentioned [here](https://github.com/ChainSafe/web3.js/pull/5251#discussion_r929883330)

Spacesai1or:
The issue requests to make use of `inheritDoc`, but I couldn't get it to work across files. Additionally, a disclaimer utilizing [Admonitions](https://docusaurus.io/docs/next/markdown-features/admonitions) was requested to be added to each rpc method wrapper doc, but this functionality is only available in markdown files, not as doc strings. So, the only change in this PR is that the doc string for each rpc method wrapper links to the previously duplicate docs for each method found under `Web3Eth` class docs

closes #5295


avkos (approved):

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -1737,49 +980,8 @@ export async function estimateGas<ReturnFormat extends DataFormat>(
 
 // TODO - Add input formatting to filter
 /**
- * Gets past logs, matching the provided `filter`.
- *
+ * View additional documentations here: {@link Web3Eth.getLogs}
```
jdevcs:
```suggestion
 * View additional documentations here: {@link Web3Eth.getPastLogs}
```
as`@ link Web3Eth.getLogs` navigation doesn't work

jdevcs:
LGTM, just need tiny change

luu-alex (approved):

PR #6064, opened 2023-05-04
Update `formatTransaction` to handle Nethermind edge case

### LINKED ISSUE #5997: `data` and `input` are returned for Sepolia transaction receipts

### Current Behavior

When calling `web3.eth.getTransaction('0xe3d28c23ffcd8b94b8f2fd802b4c8dd17a5f19992660acd082ac229ab410b959')` the following error is thrown:

```typescript
TransactionDataAndInputError: Invalid value given "data: 60806...008120033, input: 60806...008120033". Error: You can't have "data" and "input" as properties of transactions at the same time, please use either "data" or "input" instead..
```

### Expected Behavior

No error to be thrown

### Steps to Reproduce

`web3.eth.getTransaction('0xe3d28c23ffcd8b94b8f2fd802b4c8dd17a5f19992660acd082ac229ab410b959')`

### Web3.js Version

4.x

### Environment

- Operating System: 
- Browser: 
- Node.js Version: 
- NPM Version: 

### Anything Else?

_No response_

Spacesai1or:
- `formatTransaction` no longer throws a `TransactionDataAndInputError` if it's passed a transaction object with both `data` and `input` properties set (as long as they are the same value)
- If `transaction.data === transaction.input`, `formatTransaction` will delete the `transaction.data` property and use `transaction.input` as according to spec:

image

closes #5997


packages/web3-eth/src/utils/format_transaction.ts
```diff
@@ -40,7 +40,11 @@ export function formatTransaction<
 
 	formattedTransaction = format(options.transactionSchema, formattedTransaction, returnFormat);
 
-	if (!isNullish(formattedTransaction.data) && !isNullish(formattedTransaction.input))
+	if (
+		!isNullish(formattedTransaction.data) &&
+		!isNullish(formattedTransaction.input) &&
+		formattedTransaction.data !== formattedTransaction.input
```
jdevcs, line 46: codecov patch diff is not passing, due to following not covered.

avkos (approved):
LGTM except @jdevcs comment

nikoulai (approved):

Muhammad-Altabba (approved):

packages/web3-eth/src/utils/format_transaction.ts
Spacesai1or: The new line is covered by the e2e tests, but I can add a unit test

jdevcs (approved):

PR #6082, opened 2023-05-11
Listening events on v4 is not working

### LINKED ISSUE #5989: Listening events on v4 is not working

### Current Behavior

When contract events happen on the chain, web3 v4 does not do anything

### Expected Behavior

When contract events happen on-chain should trigger a callback.

### Steps to Reproduce

1. Open [stackblitz main.ts](https://stackblitz.com/edit/vitejs-vite-enrpyq?file=src/main.ts) there should be all run-able code
2. Start new [remix](https://remix.ethereum.org/) project with a basic template
3. Change all content from `1_Storage.sol` with one from [stackblitz Storage.sol](https://stackblitz.com/edit/vitejs-vite-enrpyq?file=Storage.sol)
4. Build a contract and deploy it with an `Injected Provider` - get a contract address
5. Update `contractAddress` inside [stackblitz main. ts](https://stackblitz.com/edit/vitejs-vite-enrpyq?file=src/main.ts) with the contract address from the remix
6. just in case save all inside stackblitz and refresh page 
7. call `store` method with some numeric value from remix
8. watch changes on stackblitz preview

eventually you will get something like this
image

### Web3.js Version

4.0.1-rc.0

### Environment

_No response_

### Anything Else?

Events are set as https://web3js.readthedocs.io/en/v1.2.11/web3-eth-contract.html#contract-events there is nothing breaking in the migration guide https://docs.web3js.org/docs/guides/web3_migration_guide/contracts_migration_guide
and in tests, seems you can listen for events without providing any options
https://github.com/web3/web3.js/blob/8463f6bd45d79874b004ea128bb9d8aec63431b8/packages/web3-eth-contract/test/integration/contract_events.test.ts#L59

avkos:
## Description

Fixes #5989 

metamask mov

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-core/src/web3_subscriptions.ts
+++ b/packages/web3-core/src/web3_subscriptions.ts
@@ -29,6 +29,7 @@ import {
 	DataFormat,
 	DEFAULT_RETURN_FORMAT,
 	JsonRpcSubscriptionResultOld,
+	EIP1193Provider,
 } from 'web3-types';
 import { jsonRpc } from 'web3-utils';
 import { Web3EventEmitter, Web3EventMap } from './web3_event_emitter';
@@ -93,8 +94,7 @@ export abstract class Web3Subscription<
 			}
 		};
 
-		// @ts-expect-error subscribe using old providers
-		if ((typeof this._requestManager.provider as Web3BaseProvider).request === 'function') {
+		if (typeof (this._requestManager.provider as EIP1193Provider<API>).request === 'function') {
 			(this._requestManager.provider as Web3BaseProvider).on<Log>('message', messageListener);
 		} else {
 			(this._requestManager.provider as Web3BaseProvider).on<Log>('data', messageListener);
--- a/packages/web3-core/src/web3_subscriptions.ts
+++ b/packages/web3-core/src/web3_subscriptions.ts
@@ -77,7 +77,7 @@ export abstract class Web3Subscription<
 				| JsonRpcSubscriptionResultOld<Log>
 				| JsonRpcNotification<Log>,
 		) => {
-			// for old providers
+			// for EIP-1193 provider
 			if (data?.data) {
 				this._processSubscriptionResult(data?.data?.result ?? data?.data);
 				return;
@@ -93,7 +93,12 @@ export abstract class Web3Subscription<
 			}
 		};
 
-		(this._requestManager.provider as Web3BaseProvider).on<Log>('message', messageListener);
+		// @ts-expect-error subscribe using old providers
+		if ((typeof this._requestManager.provider as Web3BaseProvider).request === 'function') {
+			(this._requestManager.provider as Web3BaseProvider).on<Log>('data', messageListener);
+		} else {
+			(this._requestManager.provider as Web3BaseProvider).on<Log>('message', messageListener);
+		}
 
 		this._messageListener = messageListener;
 	}
--- a/packages/web3-core/src/web3_subscriptions.ts
+++ b/packages/web3-core/src/web3_subscriptions.ts
@@ -95,9 +95,9 @@ export abstract class Web3Subscription<
 
 		// @ts-expect-error subscribe using old providers
 		if ((typeof this._requestManager.provider as Web3BaseProvider).request === 'function') {
-			(this._requestManager.provider as Web3BaseProvider).on<Log>('data', messageListener);
-		} else {
 			(this._requestManager.provider as Web3BaseProvider).on<Log>('message', messageListener);
+		} else {
+			(this._requestManager.provider as Web3BaseProvider).on<Log>('data', messageListener);
 		}
 
 		this._messageListener = messageListener;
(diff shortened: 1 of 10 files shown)


Muhammad-Altabba (commented):
Could we please have tests for the added logic?

avkos:
> Could we please have tests for the added logic?

Yes. I am working on it. but unfortunately, it will be not browser+metamask test, just unit test

packages/web3-core/src/web3_subscriptions.ts
```diff
@@ -74,7 +74,20 @@ export abstract class Web3Subscription<
 			err: Error | undefined,
 			data?: JsonRpcSubscriptionResult | JsonRpcNotification<Log>,
```
Spacesai1or: Your changes indicate that `err` is/can be the expected return data. Does anyone @web3/web3-js know why `err` is the first argument for `messageListener`?

Spacesai1or: From the [Examples section](https://eips.ethereum.org/EIPS/eip-1193#appendix-ii-examples) of EIP-1193, it would appear the a failure to subscribe would reject the promise, not that an error would be sent as a `message`

```typescript
// Example 4: Log new blocks
ethereum
  .request({
    method: 'eth_subscribe',
    params: ['newHeads'],
  })
  .then((subscriptionId) => {
    ethereum.on('message', (message) => {
      if (message.type === 'eth_subscription') {
        const { data } = message;
        if (data.subscription === subscriptionId) {
          if ('result' in data && typeof data.result === 'object') {
            const block = data.result;
            console.log(`New block ${block.number}:`, block);
          } else {
            console.error(`Something went wrong: ${data.result}`);
          }
        }
      }
    });
  })
  .catch((error) => {
    console.error(
      `Error making newHeads subscription: ${error.message}.
       Code: ${error.code}. Data: ${error.data}`
    );
  });
```

This would mean `messageListener` is never called with an `err`?

packages/web3-core/src/web3_subscriptions.ts
```diff
@@ -71,14 +72,24 @@ export abstract class Web3Subscription<
...
-			if (err) {
-				this._processSubscriptionError(err);
+
+			if (
+				data &&
+				jsonRpc.isResponseWithNotification(
+					data as unknown as JsonRpcSubscriptionResult | JsonRpcNotification<Log>,
+				)
+			) {
+				this._processSubscriptionResult(data?.params.result);
```
Spacesai1or, line 93: Judging from [this 1.x code](https://github.com/web3/web3.js/blob/13a2edc852aa2ae35cf827f8b08958c26356db4c/packages/web3-core-requestmanager/src/index.js#L109-L118), it seems like this code is to support legacy providers while `data.data.result` (what you have marked as `// for old providers`) is to support EIP-1193

However, to handle the legacy provider, it would appear that we're missing an event listener for `data` event, as `messageListener` is only triggered for `message` events

avkos: thanks. I've added support for old providers and change comment

packages/web3-core/src/web3_subscriptions.ts
```diff
@@ -71,18 +72,33 @@ export abstract class Web3Subscription<
...
 			}
 		};
 
-		(this._requestManager.provider as Web3BaseProvider).on<Log>('message', messageListener);
+		// @ts-expect-error subscribe using old providers
+		if ((typeof this._requestManager.provider as Web3BaseProvider).request === 'function') {
+			(this._requestManager.provider as Web3BaseProvider).on<Log>('data', messageListener);
+		} else {
+			(this._requestManager.provider as Web3BaseProvider).on<Log>('message', messageListener);
+		}
```
spacesailor24, line 101: This is backwards, [it should be](https://github.com/web3/web3.js/blob/13a2edc852aa2ae35cf827f8b08958c26356db4c/packages/web3-core-requestmanager/src/index.js#L100-L101)

```typescript
if ((typeof this._requestManager.provider as Web3BaseProvider).request === 'function') {
			(this._requestManager.provider as Web3BaseProvider).on<Log>('message', messageListener);
		} else {
			(this._requestManager.provider as Web3BaseProvider).on<Log>('data', messageListener);
		}
```

Spacesai1or: This is backwards, [it should be](https://github.com/web3/web3.js/blob/13a2edc852aa2ae35cf827f8b08958c26356db4c/packages/web3-core-requestmanager/src/index.js#L100-L101)

```suggestion
		if ((typeof this._requestManager.provider as Web3BaseProvider).request === 'function') {
			(this._requestManager.provider as Web3BaseProvider).on<Log>('message', messageListener);
		} else {
			(this._requestManager.provider as Web3BaseProvider).on<Log>('data', messageListener);
		}
```

Spacesai1or (approved):
Otherwise looks good, thank you

jdevcs (approved):
@avkos update change log before merging

packages/web3-core/src/web3_subscriptions.ts
avkos: omg! yes. thank you @spacesailor24 !

PR #6089, opened 2023-05-14
Add a tutorial for smart contract basic interaction

### LINKED ISSUE #6078: Contracts Tutorial

Muhammad-Altabba:
## Description

Closes #6078

## Type of change

## Checklist:


mpetrunic:
I feel like it would make a lot more sense, to write tutorial using typescript^^

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
```diff
@@ -0,0 +1,382 @@
...
+```
+npm install web3@4.0.1-rc.1
+```
+
+Note that we are installing the latest version of 4.x, at the time of this tutorial writing. You can check the latest version at https://www.npmjs.com/package/web3?activeTab=versions
+
+Next, create a new file called `index.js` in your project directory and add the following code to it:
+
+```javascript
+const { Web3 } = require('web3');
```
jdevcs: I think also mention in desc that we also support ESM imports as we have native ESM builds.

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
```diff
@@ -0,0 +1,382 @@
...
+If everything is working correctly, you should see the current block number logged to the console. However, if you got an error with the reason `connect ECONNREFUSED 127.0.0.1:7545` then double check that you are running Ganache locally on port `7545`.
+
+## Step 5: Deploy the smart contract to the Ganache network using web3.js
+
+In this step, we will use web3.js to deploy the smart contract to the Ganache network.
+
+Create a file named `deploy.js` and fill it with the following code:
+
+```javascript
+const { Web3 } = require('web3');
```
jdevcs: wt do you think about encouraging users to use individual packages instead of importing main web3, as it brings many not-required packages as well , if some one is not doing optimizations for that non-required pkgs.

mpetrunic: My 2c: tutorias are meant for people first time interacting with web3.js. Goal is to make it as simple as possible. Using individual packages would make sense in some "Advanced tutorials"

avkos: We need to keep in mind that auto-signing requests with the local wallet are not working when you use individual packages

avkos:
> I feel like it would make a lot more sense, to write tutorial using typescript^^

it will be great to have tutorials in TS and JS. so people who use JS or TS can just copy/paste the code and it should works. But I think it is another task. Because we need to change all our tutorials

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
```diff
@@ -0,0 +1,382 @@
...
+
+Note that we are installing the latest version of 4.x, at the time of this tutorial writing. You can check the latest version at https://www.npmjs.com/package/web3?activeTab=versions
+
+Next, create a new file called `index.js` in your project directory and add the following code to it:
+
+```javascript
+const { Web3 } = require('web3');
+
+// Set up a connection to the Ganache network
+const web3 = new Web3(new Web3.providers.HttpProvider('http://localhost:7545'));
```
avkos, line 185: you can just pass 'http://localhost:7545'. why do you need to create an instance of HTTP provider here?

avkos: if I am not mistaken, the default ganache port is 8545. It will be great to have it here to avoid users having problems with the port.

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
```diff
@@ -0,0 +1,382 @@
...
+	const gas = await myContract.estimateGas({
+		from: defaultAccount,
+	});
+	console.log('estimated gas:', gas);
+
+	try {
+		// Deploy the contract to the Ganache network
+		const tx = await myContract.send({
+			from: defaultAccount,
+			gas: 1000000,
```
avkos: What do you think to use estimated gas here?

avkos (changes requested):

Muhammad-Altabba:
> > I feel like it would make a lot more sense, to write tutorial using typescript^^
> 
> it will be great to have tutorials in TS and JS. so people who use JS or TS can just copy/paste the code and it should works. But I think it is another task. Because we need to change all our tutorials

Yes, I thought about using TS but this would add more steps to configure the samples for using TypeScript. So, I thought about keeping it simple as the tutorials are for absolute beginners. And for this, we may do something like a bit more advanced tutorial for TypeSecript users that would contain more bit advanced usage.

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
Muhammad-Altabba: Yes, this is essential to have the same default port the Ganache use by default, to ensure smooth execution.
And the port used by the Ganache application, which is mentioned in the tutorial prerequisite, is 7545. While the port used for ganache-cli is 8545.

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
Muhammad-Altabba: Seems a good idea. I will do it.

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
Muhammad-Altabba: Seems a good idea. I will add a hint for this.

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
Muhammad-Altabba: @jdevcs @mpetrunic 
Thanks, and yes, I kept it using one package for simplicity. But I should add a hint regarding this.

@avkos 
Thanks for sharing this interesting information. Do we have it mentioned somewhere? If not, I think we should highlight this in multiple places.

mpetrunic:
> > > I feel like it would make a lot more sense, to write tutorial using typescript^^
> > 
> > 
> > it will be great to have tutorials in TS and JS. so people who use JS or TS can just copy/paste the code and it should works. But I think it is another task. Because we need to change all our tutorials
> 
> Yes, I thought about using TS but this would add more steps to configure the samples for using TypeScript. So, I thought about keeping it simple as the tutorials are for absolute beginners. And for this, we may do something like a bit more advanced tutorial for TypeSecript users that would contain more bit advanced usage.

The thing is, that configuration and types is the main issue for the majority users. Js users can just cp typescript code and remove types so you kill 2 birds with one stone.

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
jdevcs: @avkos auto signing works if there is [wallet in context](https://github.com/web3/web3.js/blob/ae1d156e4b6cf17ab641b1e96cebf9cacd2a1e41/packages/web3-eth/src/rpc_method_wrappers.ts#L565). 

for example one of working snippit:
```
const {Wallet, privateKeyToAccount} = require('web3-eth-accounts');
const {Contract} = require('web3-eth-contract');
const {Web3Context} = require('web3-core');

const con_json = require('contract.json');

async function test() {
    
//create a context with wallet and provider
    const account = privateKeyToAccount("PRIVATE KEY");
    const context = new Web3Context(
        { 
            provider: 'http://127.0.0.1:7545',
            wallet: new Wallet(account)
        });

//pass that context to other packages instead of only provider
    const contract = new Contract(con_json.abi, context);
    
    const contractInstance = await contract.deploy(
        {
            data: con_json.bytecode
        }).send({ 
            gas: 10000,
            from: account.address });

    console.log(contractInstance.options.address);


    const receipt = await contractInstance.methods.store(12).send(
        {
            from: account.address, 
            gas: 10000
        });

    console.log(receipt);


    //.....
}


```

However as @mpetrunic said , it should not be added in current tutorial , so we will not do it.

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
```diff
@@ -0,0 +1,382 @@
+---
+sidebar_position: 1
+sidebar_label: 'Deploying and Interacting with Smart Contracts'
+---
+
+# Deploying and Interacting with Smart Contracts
+
+## Introduction
+
+In this tutorial, we will walk through the process of deploying a smart contract to the Ethereum network, generating the ABI, and interacting with the smart contract using web3.js version 4.x. We will cover the basic concepts of Ethereum, Solidity, and web3.js and provide step-by-step instructions for deploying a simple smart contract to the Ganache network.
```
nikoulai: I would change `the Ganache network` to ` a test network using Ganache`. 
What do you think?

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
```diff
@@ -0,0 +1,382 @@
...
+Run the following command to deploy the smart contract:
+
+```sh
+node deploy.js
+```
+
+If everything is working correctly, you should see something like the following:
+
+```
+Deployer account: 0xdd5F9948B88608a1458e3a6703b0B2055AC3fF1b
```
nikoulai, line 280: I guess these values here, are not going to be always the values. Should we mention that here, so new users are not confused?

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
```diff
@@ -0,0 +1,382 @@
...
+
+```
+node interact.js
+```
+
+If everything is working correctly, you should see the current counter value logged to the console, followed by the transaction receipt, and then the updated counter value. The output would like:
+
+```sh
+my number value: 1
+Transaction Hash: 0x9825e2a2115896728d0c9c04c2deaf08dfe1f1ff634c4b0e6eeb2f504372f927
```
nikoulai, line 352: same here

nikoulai (approved):

Muhammad-Altabba:
> > > > I feel like it would make a lot more sense, to write tutorial using typescript^^
> > > 
> > > 
> > > it will be great to have tutorials in TS and JS. so people who use JS or TS can just copy/paste the code and it should works. But I think it is another task. Because we need to change all our tutorials
> > 
> > 
> > Yes, I thought about using TS but this would add more steps to configure the samples for using TypeScript. So, I thought about keeping it simple as the tutorials are for absolute beginners. And for this, we may do something like a bit more advanced tutorial for TypeSecript users that would contain more bit advanced usage.
> 
> The thing is, that configuration and types is the main issue for the majority users. Js users can just cp typescript code and remove types so you kill 2 birds with one stone.

I agree. However, I hope to give a smooth experience with a strait forward copy/past then it works. But using TypeScript in the samples would also involve adding additional small steps for installing and compiling the code before running. 

So, I thought about having 2 tabs. One for JavaScript and the other for TypeScript. And I needed to update docusaurus to enable taps as they are available in new version. But the taps have issues with code blocks if they are placed inside. So, I am thinking about keeping the javascript here similar to other tutorials. And create an issue to investigate how to put code snippets inside taps with docusaurus and update all the tutorials thereafter.

mpetrunic:
> I agree. However, I hope to give a smooth experience with a strait forward copy/past then it works. But using TypeScript in the samples would also involve adding additional small steps for installing and compiling the code before running.
> 

Or ts-node :)

> So, I thought about having 2 tabs. One for JavaScript and the other for TypeScript. And I needed to update docusaurus to enable taps as they are available in new version. But the taps have issues with code blocks if they are placed inside. So, I am thinking about keeping the javascript here similar to other tutorials. And create an issue to investigate how to put code snippets inside taps with docusaurus and update all the tutorials thereafter.

Two tabs is also fine^^

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
Muhammad-Altabba: done

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
Muhammad-Altabba: Actually, I tried to clarify this by saying "something *like* the following". And I prefer to keep them as they give the user the ability to scan the code and the output before he/she would run it. As some readers may just run a none or few of the code snippets (as they sometimes come to the page from a search engine. And they may copy and use just a few lines of the provided code). But in all cases, they like to read what is the expected output.

nikoulai: I see. sounds good!
Thank you @Muhammad-Altabba

jdevcs (approved):
LGTM, except few parts where changes required by other members.

docs/docs/tutorials/deploying_and_interacting_with_smart_contracts.md
avkos: Thanks for example @jdevcs , It will be great to have this example somewhere in tutorials

avkos (approved):

PR #6102, opened 2023-05-18
`defaultTransactionTypeParser` Refactor

### LINKED ISSUE #6065: transaction type error
Invalid transaction type applied to given tx:
```
const web3 = new Web3("https://goerli.infura.io/v3/......");

web3.eth.wallet?.add("0x0f8592102198f565b00937d2da3cd00475866ebdb18053243355dd6aa0222ab4");

const res = await web3.eth.sendTransaction(
    {
        to: "0x8f3e9c1Bd65EB267d19B176A73217524DC21A5ca",
        nonce: 951,
        gasLimit: "0x67b8bf",
        gasPrice: "0x4617ee7c40",
        from: "0x76D43E68c2aE1084B340f0eDC04C325DD341039B",
      },
    undefined,
    {
        ignoreGasPricing: true,
        checkRevertBeforeSending: false
    },
);
```
[wallet.signTransaction](https://github.com/web3/web3.js/blob/1291d957ef3cccbe916cc9267535804e70980c37/packages/web3-eth/src/rpc_method_wrappers.ts#L567) receives :
```
{
  to: "0x8f3e9c1Bd65EB267d19B176A73217524DC21A5ca",
  nonce: "0x3b7",
  gasPrice: "0x4617ee7c40",
  from: "0x76D43E68c2aE1084B340f0eDC04C325DD341039B",
  gas: "0x67b8bf",
}
```

[defaultTransactionTypeParser](https://github.com/web3/web3.js/blob/1291d957ef3cccbe916cc9267535804e70980c37/packages/web3-eth/src/utils/detect_transaction_type.ts#L24) receives follow and applies type 0x2 :
```
{
  to: "0x8f3e9c1Bd65EB267d19B176A73217524DC21A5ca",
  nonce: 951n,
  gasPrice: 301049216064n,
  from: "0x76D43E68c2aE1084B340f0eDC04C325DD341039B",
  gas: 6797503n,
  value: "0x",
  input: "0x",
  chain: "mainnet",
  hardfork: "london",
  chainId: "0x1",
  networkId: "0x1",
  gasLimit: 6797503n,
}
```
and [validate transaction](https://github.com/web3/web3.js/blob/1291d957ef3cccbe916cc9267535804e70980c37/packages/web3-eth/src/utils/prepare_transaction_for_signing.ts#L139) throws :
```
"Eip1559GasPriceError: Invalid value given \"0x4617ee7c40\". Error: eip-1559 transactions don't support gasPrice."
```
seems due to :
https://github.com/web3/web3.js/blob/1291d957ef3cccbe916cc9267535804e70980c37/packages/web3-eth/src/utils/transaction_builder.ts#L192

Spacesai1or:
closes #6065


packages/web3-eth/test/integration/defaults.test.ts
Spacesai1or, line 1: The types for these tests were updated to `0x0` because they all use `gasLimit`

spacesailor24:
@web3/web3-js Should `defaultTransactionTypeParser` throw an error if type `0x2` properties (`maxFeePerGas` and `maxPriorityFeePerGas`) are used in conjunction with `gasPrice`? Or should it just return `undefined` and our tx validator should catch the pricing mismatch?

The below tx should probably be treated as invalid because we don't know which gas pricing properties to use:

```typescript
{
		from: '0xEB014f8c8B418Db6b45774c326A0E64C78914dC0',
		to: '0x3535353535353535353535353535353535353535',
		value: '0x174876e800',
		gas: '0x5208',
		maxPriorityFeePerGas: '0x49504f80', // type 0x2
		data: '0x0',
		nonce: '0x4',
		chainId: '0x1',
		gasLimit: '0x5208',
		gasPrice: '0x12' // type 0x0
	},
```

packages/web3-eth/src/utils/detect_transaction_type.ts
```diff
@@ -17,26 +17,102 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
+};
+const type0x2TransactionSchema = {
+	type: 'object',
+	properties: {
+		gasPrice: {
+			type: 'null',
+		},
+	},
+};
+
```
Spacesai1or: I don't believe these should go in [schemas.ts](https://github.com/web3/web3.js/blob/4.x/packages/web3-eth/src/schemas.ts) since they aren't complete schemas of the transaction object

packages/web3-eth/src/utils/detect_transaction_type.ts
```diff
@@ -17,26 +17,124 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
+	// supports EIP-2718 (https://eips.ethereum.org/EIPS/eip-2718)
+	// and whether we should return undefined for legacy txs,
+	// or type 0x0 for legacy txs post EIP-2718
+
+	if (!isNullish(tx.accessList)) {
+		validateTxTypeAndHandleErrors(type0x1TransactionSchema, tx, '0x1');
 		return '0x1';
+	}
+
+	if (!isNullish(tx.maxFeePerGas) || !isNullish(tx.maxPriorityFeePerGas)) {
```
jdevcs: if  `gasPrice` is given without HF , there should be a case for that as well for returning `0x0 `and validating for `validateTxTypeAndHandleErrors(type0x1TransactionSchema, tx, '0x0');`,

packages/web3-types/src/eth_types.ts
```diff
@@ -247,21 +247,51 @@ export type AccessListResult = {
...
+	| 'grayGlacier'
+	| 'bellatrix'
 	| 'merge'
-	| 'muirGlacier'
-	| 'petersburg'
-	| 'shanghai'
-	| 'spuriousDragon'
-	| 'tangerineWhistle';
+	| 'capella'
+	| 'shanghai';
```
jdevcs:
```
export type Hardfork = `${HardforksOrdered}`; 
```
this will auto generate type like 'chainstart' | 'frontier' ....

packages/web3-eth/src/utils/detect_transaction_type.ts
jdevcs: just tiny suggestion, if its Tx type schema like ( TypeSchemaFor0x1Tx , TypeSchemaFor0x2Tx ) it will more readable and can be moved to schema file.

jdevcs:
Thanks @spacesailor24

luu-alex (approved):

packages/web3-eth/src/utils/detect_transaction_type.ts
jdevcs: or will it work with out that for 0x0, are there any test cases for it?

packages/web3-errors/src/error_codes.ts
```diff
@@ -74,6 +74,7 @@ export const ERR_TX_DATA_AND_INPUT = 425;
 export const ERR_TX_POLLING_TIMEOUT = 426;
 export const ERR_TX_RECEIPT_MISSING_OR_BLOCKHASH_NULL = 427;
 export const ERR_TX_RECEIPT_MISSING_BLOCK_NUMBER = 428;
+export const ERR_TX_INVALID_PROPERTIES_FOR_TYPE = 429;
```
avkos: Is it ok that we have the same code for two different errors?
`ERR_TX_INVALID_PROPERTIES_FOR_TYPE` & `ERR_TX_LOCAL_WALLET_NOT_AVAILABLE`

avkos (changes requested):

packages/web3-eth/src/utils/detect_transaction_type.ts
Spacesai1or: @jdevcs 
I'll rename them to: `transactionType0x0Schema`, but I don't think they should be moved to `schemas.ts` because then they'll be exported from `web3-eth` package and these type shouldn't be any where else as they are an incomplete representation of what a typed transaction is (i.e. they don't contain the other fields of a transaction)

I experimented with doing something like:

```typescript
export const transactionType0x0Schema = {
	type: 'object',
	properties: {
		...transactionSchema.properties,
		accessList: {
			type: 'null',
		},
		maxFeePerGas: {
			type: 'null',
		},
		maxPriorityFeePerGas: {
			type: 'null',
		},
	}
};
```

but now the function `validateTxTypeAndHandleErrors` in `defaultTransactionTypeParser` is validating the entire transaction object which is out of scope of `defaultTransactionTypeParser`. This function is only concerned about returning the detected tx type, or throwing an error for an invalid tx object regarding types and their appropriate properties, not the rest of the tx object

I think it's okay to keep these type specified here so that they're are exported from the package and because they are only used for `defaultTransactionTypeParser`

packages/web3-errors/src/error_codes.ts
Spacesai1or: No, good catch, updated [here](https://github.com/web3/web3.js/pull/6102/commits/db3f3ee30c45025364d1f7106d2ae420ec8ec401)

packages/web3-eth/src/utils/detect_transaction_type.ts
Spacesai1or: Renamed [here](https://github.com/web3/web3.js/pull/6102/commits/cef42da04daf0db833f516c4bcfb7faf7b0884f9), also added a validation check for type `0x0` txs without `tx.type` specified

packages/web3-types/src/eth_types.ts
Spacesai1or: Addressed [here](https://github.com/web3/web3.js/pull/6102/commits/cef42da04daf0db833f516c4bcfb7faf7b0884f9), thank you for the suggestion, much cleaner

packages/web3-eth/src/utils/detect_transaction_type.ts
Spacesai1or: @jdevcs The test you're describing is [here](https://github.com/web3/web3.js/blob/db3f3ee30c45025364d1f7106d2ae420ec8ec401/packages/web3-eth/test/fixtures/detect_transaction_type.ts#L202-L212) and the detected type would be `undefined`. Without `hardfork`, we can't be sure if the network supports EIP-2718, so we return `undefined` assuming is a pre-EIP-2718 legacy tx

avkos (approved):
LGTM.
BTW. Would be great to increase test coverage, thanks

Spacesai1or:
> LGTM. BTW. Would be great to increase test coverage, thanks

@avkos This is the only line not covered by tests, and it's sorta irrelevant as it just a catch-all for any error that not a validation error

image

jdevcs (approved):

packages/web3-eth/src/utils/detect_transaction_type.ts
jdevcs: > because then they'll be exported from web3-eth package

that make sense to not export it, with current code that you added.

PR #6120, opened 2023-05-24
Eth turorial

### LINKED ISSUE #6077: Eth Tutorial

nikoulai:
## Description

closes #6077


docs/docs/tutorials/eth.md
```diff
@@ -0,0 +1,375 @@
...
+
+```
+npm install web3@4.0.1-rc.1
+```
+
+Note that we are installing the latest version of 4.x, at the time of this tutorial writing. You can check the latest version at https://www.npmjs.com/package/web3?activeTab=versions
+
+Next, create a new file called `index.js` in your project directory and add the following code to it:
+
+```javascript
```
avkos, line 67: I think we need to have consistency. I mean, it will be good to have all code written using Typescript. What do you think?

avkos (changes requested):

docs/docs/tutorials/eth.md
```diff
@@ -0,0 +1,375 @@
...
+		to: accounts[1],
+		value: web3.utils.toWei('1', 'ether'), // value should be passed in wei. For easier use and to avoid mistakes we utilize the auxiliary `toWei` function.
+	};
+
+	//send the actual transaction
+	const transactionHash = await web3.eth.sendTransaction(transaction);
+	console.log('transactionHash', transactionHash);
+
+	balance1 = await web3.eth.getBalance(accounts[0]);
+	balance2 = await web3.eth.getBalance(accounts[1]);
```
luu-alex, line 134: it would also be good to show the different return formats in one of these examples. thoughts?

docs/docs/tutorials/eth.md
nikoulai: fixed, thanks for the catch, it was a mistake.

docs/docs/tutorials/eth.md
nikoulai: I use `ETH_DATA_FORMAT` in a later example (https://github.com/web3/web3.js/pull/6120/files#diff-c99987636f6f022dcffeaf9e5779f32ee485b7f5909785fdf97134bc58914ebbR249). I can add it in more places if you think it would be better.

luu-alex: Ah I missed that. 
I think it would be nice if you can provide one more example as well with a different format,
For example 
```
const estimatedGas = await deployment.estimateGas(
		{
			from: acc,
		},
		ETH_DATA_FORMAT, // the returned data will be formatted as a hexstring 
	);

console.log(estimatedGas)

const estimatedGas = await deployment.estimateGas(
		{
			from: acc,
		},
		DEFAULT_RETURN_FORMAT, // the returned data will be formatted as a bigint
	);

console.log(estimatedGas)
```

luu-alex (approved):

avkos (approved):

docs/docs/tutorials/eth.md
Sammiegirl411: How do I start web 3 n I have matamask

PR #6167, opened 2023-06-08
Event filtering using non/indexed string arguments

### LINKED ISSUE #5807: Ensure filtering works for strings on events

### What's up?

In `1.x` seems that there is a bug and strings cannot be used to filter indexed events

Example 
```
contract.events.MyEvent({filter: {a: "string"}})
``` 
We should ensure that this works on `4.x`

### Examples/References

Reported lately: https://github.com/web3/web3.js/issues/5806
Initial issue: https://github.com/web3/web3.js/issues/3053

Spacesai1or:
This PR fixes filter contract events using non-indexed string event arguments and adds support for filtering using indexed string event arguments

closes #5807


packages/web3-eth-contract/src/contract.ts
```diff
@@ -735,22 +735,52 @@ export class Contract<Abi extends ContractAbi>
...
+		// 							? (filter[k] as Numbers[]).some(
+		// 									(v: Numbers) =>
+		// 										String(log.returnValues[k]).toUpperCase() ===
+		// 										String(v).toUpperCase(),
+		// 							  )
+		// 							: String(log.returnValues[k]).toUpperCase() ===
+		// 							  String(filter[k]).toUpperCase(),
+		// 				  ),
+		// 	  )
+		// 	: decodedLogs;
```
Spacesai1or:
```suggestion
```

packages/web3-eth-contract/src/contract.ts
```diff
@@ -735,22 +735,35 @@ export class Contract<Abi extends ContractAbi>
...
-												String(log.returnValues[k]).toUpperCase() ===
-												String(v).toUpperCase(),
-									  )
-									: String(log.returnValues[k]).toUpperCase() ===
-									  String(filter[k]).toUpperCase(),
-						  ),
-			  )
-			: decodedLogs;
+
+		if (filterKeys.length > 0) {
```
Spacesai1or, line 739: On [line 729](https://github.com/web3/web3.js/blob/ef63c3237e9155dc7f7ff78af5f224ad087bd063/packages/web3-eth-contract/src/contract.ts#L729):

```typescript
const logs = await getLogs(this, { fromBlock, toBlock, topics, address }, returnFormat);
```

we submit an RPC request to get a list of events for a period of blocks that match `topics` i.e. the event signatures we're interested in. If `topics === undefined` then the RPC response will contain all the events that occurred during the block period. Previously this check was:

```typescript
eventName === 'allEvents' && filterKeys.length > 0
```

to account for this scenario as we needed to perform the filtering on the all the events returned by the RPC response using the `filter` provided by the user. Usually in this scenario, the user is intending to get all the events within a block period with a specific event argument i.e. "I want all events during the block period where `from` event arugment was `0x0...`". Otherwise, we skip the additional filtering because the events have already been filtered using the user's `filter` by the RPC client. However, this did not work when the user's `filter` was attempting to filter using a non-indexed string event argument as done in [this](https://github.com/web3/web3.js/blob/ef63c3237e9155dc7f7ff78af5f224ad087bd063/packages/web3-eth-contract/test/integration/contract_filter_events.test.ts#L245-L262) test. By removing the check `eventName === 'allEvents'` for this line, we now perform this additional filtering on all events as long as some `filter` object was provided by the user. This adds support for filtering events using non-indexed and indexed string event arguments (i.e. the issue reported by #5807)

packages/web3-eth-contract/src/contract.ts
```diff
@@ -735,22 +735,35 @@ export class Contract<Abi extends ContractAbi>
...
+								String(log.returnValues[key]).toUpperCase() ===
+								String(v).toUpperCase(),
+						);
+					}
+
+					const inputAbi = abi.inputs?.filter(input => input.name === key)[0];
+					if (inputAbi?.indexed && inputAbi.type === 'string') {
+						const hashedIndexedString = keccak256(filter[key] as string);
+						if (hashedIndexedString === String(log.returnValues[key])) return true;
+					}
```
Spacesai1or, line 756: This added logic adds support for filtering indexed string event arguments. Because all Solidity event topics (i.e. indexed event arguments) have to be `32 bytes`, anytime a `string` is indexed for an event, it's value is represented by the `keccak256` hash of the string value

packages/web3-eth-abi/src/api/logs_api.ts
```diff
@@ -20,6 +20,9 @@ import { decodeParameter, decodeParametersWith } from './parameters_api.js';
 
 const STATIC_TYPES = ['bool', 'string', 'int', 'uint', 'address', 'fixed', 'ufixed'];
 
+const _decodeParameter = (inputType: string, clonedTopic: string) =>
+	inputType === 'string' ? clonedTopic : decodeParameter(inputType, clonedTopic);
+
```
Spacesai1or, line 25: When a indexed event input is of type string, we know that it's value is going to be the keccak256 hash of the original string value, so this change returns the hash instead of trying to decode it

avkos (approved):

Muhammad-Altabba (commented):
Thanks @spacesailor24 ,
And can we have unit tests for the added logic so the test coverage does not decrease?

jdevcs (approved):
lgtm, except some existing change requests.

PR #6177, opened 2023-06-09
Add bump-versions step after publish at RELEASE.md and bump the versions

### LINKED ISSUE #6175: canary (`dev`) versions need to have higher version number than the last release
Running `lerna publish from-package --canary ...` (while having the versions at pachage.json files not increased) causes it to not publish the `dev` version. And it provided this log:
```bash
lerna notice from-package No unpublished release found
lerna success No changed packages to publish 
```

You can check the logs here: https://github.com/web3/web3.js/actions/runs/5218529567/jobs/9419407963
Actually, this error came because we are still having the same 4.0.1 version in the package.json files (and 1.0.1 for a few of them). And learna is right to be confused. Because we should not publish the dev version after with published the release of that version.
So, there is a need to increase the versions numbers and then publish the dev versions. I think that: this is something that we missed taking care of earlier, because we did not have the canary configuration back then. 

And now as we have the configuration for canary release. I think we need to have ,as part of our release process, to increase the versions of the packages so the `dev` versions are published for that next version before its release.

i.e the flow I am suggesting will be like this, for example:
- We published 4.0.1 release.
- We increase the version to 4.0.2 as part of the release process (we need to update it).
- At each MR, if there is a need to increase the minor or major version number because of a breaking change for example, we probably need to increase it inside the MR that caused that change. And this change is compared to the last release version and not compared to the last MR merged.
- Github will publish dev versions for 4.0.2 automatically for every merge. 
- Later when 4.0.2 is ready. We publish its `release` and we increase the version at package.json files at the same `release` MR
- and so on...

Let me know what you think or if I missed something.

Muhammad-Altabba:
## Description
Fixes #6175

## Type of change

## Checklist:

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -55,3 +55,4 @@ Further details about versioning can be found in the [semver 2.0.0 specification
 17. Publish draft release created in `Step 11`
 18. Run `npx lerna publish from-package --ignore-scripts --dist-tag alpha` in the root directory to publish packages to NPM
     - lerna will not invoke life cycle scripts before publishing and this will publish all packages to NPM public registry
+19. Open a new MR that will contain bumped packages version numbers using `yarn lerna version --no-push --no-private --no-git-tag-version`. And select `Patch` when the cli asks for which version to pump to.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -33,26 +33,26 @@ Further details about versioning can be found in the [semver 2.0.0 specification
 
 1. `git checkout 4.x`: Verify you are on the `4.x` base branch
 2. `git checkout -b release/bumped-version`: Create and checkout a branch with the `bumped-version` e.g. `git checkout -b release/4.0.0-alpha.0`
-    - `bumped-version` of release branch should be of main web3 package.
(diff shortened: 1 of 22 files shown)


RELEASE.md
```diff
@@ -55,3 +55,4 @@ Further details about versioning can be found in the [semver 2.0.0 specification
 17. Publish draft release created in `Step 11`
 18. Run `npx lerna publish from-package --ignore-scripts --dist-tag alpha` in the root directory to publish packages to NPM
     - lerna will not invoke life cycle scripts before publishing and this will publish all packages to NPM public registry
+19. Open a new MR that will contain bumped packages version numbers using `yarn lerna version --no-push --no-private --no-git-tag-version`. And select `Patch` when the cli asks for which version to pump to.
```
jdevcs, line 58: Its already at step 4, why we need it again after publishing?

packages/web3-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "web3-core",
-	"version": "4.0.1",
+	"version": "4.0.2",
```
jdevcs, line 3: Plus all version bumps are done in release PRs, and major minor patch is determined based on PRs merged after last release, why we want to increase patch number here in this PR? for canary?

mpetrunic:
You can do this but it is not needed as canary release does this automatically!

Issue #5670, opened 2022-12-01
"error" PromiEvent does not catch the error from send()


yanikitat:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have a such code snippet where I'm getting the EVM exception:
```Typescript
  await liquidityStorager.methods
    .redeem(sender, 100, "0x00")
    .send({from: sender, gas: 70000})
    .on("error", (e: Error) => console.log("on.error-->", e));
```
When I'm getting EVM exception, I'm getting this as JavaScript exception, not in the `.on("error")` function (I have removed a lot of not important info from logs):
```
.../node_modules/web3-core-helpers/lib/errors.js:80
        var error = new Error('Transaction has been reverted by the EVM:\n' + JSON.stringify(receipt, null, 2));
                    ^

Error: Transaction has been reverted by the EVM:
{
 "Correct transaction receipt here"
}
    at Object.TransactionRevertInstructionError (.../node_modules/web3-core-helpers/lib/errors.js:80:21)
    at .../node_modules/web3-core-method/lib/index.js:384:65 {
  reason: 'LS001',
  signature: 'Error(String)',
  receipt: {
 "Correct transaction receipt again"
  }
}
```
But! When I'm adding the `.catch()` to the my snippet:
```Typescript
  await liquidityStorager.methods
    .redeem(sender, 100, "0x00")
    .send({from: sender, gas: 70000})
    .on("error", (e: Error) => console.log("on.error-->", e))
    .catch((e: Error) => console.log("catch -->", e));
```
I'm not getting the exception anymore, but my error is written twice: from `.on("error", ...)` and `.catch(...)`:
```
catch --> Error: Transaction has been reverted by the EVM:
{
 "Correct transaction receipt here"
}
    at Object.TransactionRevertInstructionError (.../node_modules/web3-core-helpers/lib/errors.js:80:21)
    at .../node_modules/web3-core-method/lib/index.js:384:65 {
  reason: 'LS001',
  signature: 'Error(String)',
  receipt: {
 "Correct transaction receipt again"
  }
}
on.error--> Error: Transaction has been reverted by the EVM:
{
 "Correct transaction receipt here"
}
    at Object.TransactionRevertInstructionError (.../node_modules/web3-core-helpers/lib/errors.js:80:21)
    at .../node_modules/web3-core-method/lib/index.js:384:65 {
  reason: 'LS001',
  signature: 'Error(String)',
  receipt: {
[...]

PR #6194, opened 2023-06-13
Add `.on('error')` test for reverted contract methods

Spacesai1or:
This PR just adds two tests that verify we receive the revert error using `.on('error')` when calling a contract method that reverts

related to #5670

### DIFF
--- a/packages/web3-eth-contract/test/integration/contract_methods_errors.test.ts
+++ b/packages/web3-eth-contract/test/integration/contract_methods_errors.test.ts
@@ -81,7 +81,7 @@ describe('contract errors', () => {
 				code: 310,
 				receipt: undefined,
 				innerError: {
-					code: 3,
+					code: ERR_CONTRACT_EXECUTION_REVERTED,
 					data: '0x82b42900',
 					errorName: 'Unauthorized',
 					errorSignature: 'Unauthorized()',
--- a/packages/web3-eth-contract/test/integration/contract_methods_errors.test.ts
+++ b/packages/web3-eth-contract/test/integration/contract_methods_errors.test.ts
@@ -126,7 +126,7 @@ describe('contract errors', () => {
 		it('should catch error with parameter using PromiEvent.on("error")', async () => {
 			const expectedThrownError = {
 				name: 'ContractExecutionError',
-				code: 310,
+				code: ERR_CONTRACT_EXECUTION_REVERTED,
 				receipt: undefined,
 				innerError: {
 					code: 3,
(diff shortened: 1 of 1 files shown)


packages/web3-eth-contract/test/integration/contract_methods_errors.test.ts
```diff
@@ -75,6 +75,27 @@ describe('contract errors', () => {
 			});
 		});
 
+		it('should catch Unauthorized error PromiEvent.on("error")', async () => {
+			const expectedThrownError = {
+				name: 'ContractExecutionError',
+				code: 310,
+				receipt: undefined,
+				innerError: {
+					code: 3,
```
Muhammad-Altabba, line 84:
```suggestion
					code: ERR_CONTRACT_EXECUTION_REVERTED,
```

packages/web3-eth-contract/test/integration/contract_methods_errors.test.ts
```diff
@@ -101,5 +122,27 @@ describe('contract errors', () => {
 				},
 			});
 		});
+
+		it('should catch error with parameter using PromiEvent.on("error")', async () => {
+			const expectedThrownError = {
+				name: 'ContractExecutionError',
+				code: 310,
```
Muhammad-Altabba:
```suggestion
				code: ERR_CONTRACT_EXECUTION_REVERTED,
```

packages/web3-eth-contract/test/integration/contract_methods_errors.test.ts
```diff
@@ -75,6 +75,27 @@ describe('contract errors', () => {
 			});
 		});
 
+		it('should catch Unauthorized error PromiEvent.on("error")', async () => {
+			const expectedThrownError = {
```
Muhammad-Altabba, line 79: I think it would be nice if we have this `const expectedThrownError = {...` above and we use it inside the 2 related test cases which related to the `Unauthorized()` error.

packages/web3-eth-contract/test/integration/contract_methods_errors.test.ts
```diff
@@ -101,5 +122,27 @@ describe('contract errors', () => {
 				},
 			});
 		});
+
+		it('should catch error with parameter using PromiEvent.on("error")', async () => {
+			const expectedThrownError = {
```
Muhammad-Altabba, line 127: I think it would be nice if we have this `const expectedThrownError = {...` above and we use it inside the 2 related test cases which related to the `CustomError` error.

Muhammad-Altabba (approved):
Sounds good.
Just, please, consider the comments.
Thanks,

jdevcs (approved):

luu-alex (approved):

PR #6210, opened 2023-06-18
Refactor subscription's logic

### LINKED ISSUE #6202: Subscribing to multiple blockchain events causes every listener to be fired for every registered event
Lets consider this code:

```
const headSubs = await this.web3.eth.subscribe('newHeads');

headSubs.on('data', async blockHead => {
  console.log('block', blockHead);
});

const pendingTxsSub = await this.web3.eth.subscribe('pendingTransactions');

pendingTxsSub.on('data', async(tx) => {
  console.log('tx', tx);
});
```
#### Expected behavior
I'd expect pendingTxsSub to spit just txs and headSubs to only console.log new block headers.

#### Actual behavior
```
block 0x787c1245c067b4943d9585e8021e4003a1ddd17d3e2588bc95a7b0a6332b38bd
tx 0x787c1245c067b4943d9585e8021e4003a1ddd17d3e2588bc95a7b0a6332b38bd
block 0x24ce987ea13f1ea88051167a7d888455cb658a6ffb852603b23032a6fd78b1c3
tx 0x24ce987ea13f1ea88051167a7d888455cb658a6ffb852603b23032a6fd78b1c3
block 0xdb893c7e70a6f82eef78caf0927030c39488e9fb5110542b904ede3864cc240b
tx 0xdb893c7e70a6f82eef78caf0927030c39488e9fb5110542b904ede3864cc240b
block 0xbae1bafb682d0f6f3ad6fc3506c57528f1d4a95bcaad6ab5c4b841a16eafdbe4
tx 0xbae1bafb682d0f6f3ad6fc3506c57528f1d4a95bcaad6ab5c4b841a16eafdbe4
```

1. Check code above
2. Run
3. Check console.log

#### Environment

Node 18.12.0, Web3 4.0.1

Muhammad-Altabba:
## Description
Identified issues:
-   Subscribing to multiple blockchain events causes every listener to be fired for every registered event
-   Unsubscribe at a Web3Subscription class will still have the id of the subscription at the Web3SubscriptionManager
-   A call to the provider is made for every subscription object. (This means, in case of multiple creating of subscription objects, multiple subscriptions are made at the provider for each subscription type).

Refactor subscription's logic

-   `Web3Subscription` constructor now accepts `Web3SubscriptionManager` (as an alternative to accepting `Web3RequestManager` that is now marked as deprecated)
-   `Web3Subscription` constructor overloading that accepts `Web3SubscriptionManager` is marked as deprecated
  - Moved subscribing to provider events from  `Web3Subscription` to `Web3SubscriptionManager`
  - `subscribe` and `unsubscribe` called at `Web3Subscription` now is the same as calling them on `Web3SubscriptionManager`
   - `Web3Subscription` is lined now to `Web3SubscriptionManager` instead of directly to `Web3RequestManager`
   - Added the `SimpleProvider` interface which has only `request(args)` method that is compatible with EIP-1193
-   Added the `Eip1193EventName` type that contains the possible events names according to EIP-1193
-   The `EIP1193Provider` class has now all the events (for `on` and `removeListener`) according to EIP-1193

 - update test cases ...

Fixes: #6202

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-core/src/web3_subscriptions.ts
+++ b/packages/web3-core/src/web3_subscriptions.ts
@@ -30,9 +30,12 @@ import {
 	Web3APISpec,
 } from 'web3-types';
 import { jsonRpc } from 'web3-utils';
+import { SubscriptionError } from 'web3-errors';
+
 // eslint-disable-next-line import/no-cycle
 import { Web3SubscriptionManager } from './web3_subscription_manager.js';
 import { Web3EventEmitter, Web3EventMap } from './web3_event_emitter.js';
+import { Web3RequestManager } from './web3_request_manager.js';
 
 export abstract class Web3Subscription<
 	EventMap extends Web3EventMap,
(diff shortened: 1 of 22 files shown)


luu-alex (approved):

packages/web3-core/src/web3_subscriptions.ts
```diff
@@ -17,45 +17,44 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
 	private readonly _returnFormat: DataFormat;
-	private _id?: HexString;
-	private _messageListener?: (data?: JsonRpcNotification<Log>) => void;
+	protected _id?: HexString;
 
 	public constructor(
 		public readonly args: ArgsType,
-
-		options: { requestManager: Web3RequestManager<API>; returnFormat?: DataFormat },
+		subscriptionManager: Web3SubscriptionManager,
```
jdevcs, line 50: Thanks for sending me more info on this, As discussed this will break API so I suggest to post a fix without that.

packages/web3-core/test/unit/web3_subscription.test.ts
```diff
@@ -15,25 +15,32 @@ You should have received a copy of the GNU Lesser General Public License
...
 		requestManager = {
-			send: jest.fn(),
+			send: jest.fn().mockImplementation(async () => {
+				return 'sub-id';
+			}),
 			on: jest.fn(),
 			provider: { on: jest.fn(), removeListener: jest.fn(), request: jest.fn() },
 		};
-		sub = new ExampleSubscription({ param1: 'value' }, { requestManager });
+		subscriptionManager = new Web3SubscriptionManager(requestManager, subscriptions);
```
jdevcs, line 36: lgtm, could you also add following integration tests:
- at least two subscriptions, and 
- another test with two subs where one sub is un-subd, and other should be running. 
Thanks

packages/web3-types/src/web3_base_provider.ts
```diff
@@ -94,12 +94,28 @@ export interface LegacyRequestProvider {
 	): void;
 }
 
-export interface EIP1193Provider<API extends Web3APISpec> {
+export interface SimpleProvider<API extends Web3APISpec> {
```
jdevcs, line 97: changelog for this in web3-types package is also required,.

Muhammad-Altabba: Done

packages/web3-core/test/unit/web3_subscription.test.ts
Muhammad-Altabba: done

jdevcs (approved):

Issue #5675, opened 2022-12-02
1.x build failing due to types error


luu-alex:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We ran into issues with package-lock for [example](https://github.com/web3/web3.js/actions/runs/3592528072/jobs/6048346485) with dependancies using different typescript versions.

from my understanding, what is going on is when we run npm run build lerna will hoist all the dependancies meaning all the modules will be a specific version.  babel will always be using the latest version of typescript, and when we look at the @types/babel___traverse file we see that type version is 4.1 "typeScriptVersion": "4.1" .

The packages within web3 are all using typescript:'3.9'  which will result in an error when index.d.ts is different

### Expected Behavior

build should work properly

### Steps to Reproduce

pull the latest branch and try to build

### Web3.js Version

1.8.1

### Environment

- Operating System: 
- Browser: 
- Node.js Version: 
- NPM Version:

PR #6238, opened 2023-06-27
Bump `typescript` to `4.9.5` and `ts-node` to `10.9.1`

Spacesai1or:
#5675 popped up again, and upgrading `typescript` to `4.9.5` seems to solve the issue


dist/web3.min.js.LICENSE.txt
Spacesai1or, line 1: I'm not sure why this change was made, but it was a result of running `npm i && npm run build` after bumping `typescript` to `4.9.5`

luu-alex: this license file is generated when we changed webpack 4 -> 5

luu-alex:
I believe mosiac tests have been failing for a while now

luu-alex (approved):

Spacesai1or:
> I believe mosiac tests have been failing for a while now

@luu-alex Yea, seems unrelated to these changes

#6242 to track it

avkos (approved):

PR #6243, opened 2023-06-28
Add missing `blockHeaderSchema` properties

### LINKED ISSUE #6198: web3.eth.subscribe newHeads missing keys in response
#### Expected behavior

```
New block v1: {
  number: 94119265,
  hash: '0x60f6fd56770b2d04e71e6add3d9b4d85b3599f44bcddb9e368be1c21c0c43aa2',
  parentHash: '0x0942e0e244cbba697725b45a2fbf2e1aad084bae3e3eca4327ed6f693a83e515',
  nonce: '0x0000000000000000',
  sha3Uncles: '0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347',
  logsBloom: '0x00000000000000000000000000000000000000000000000002000000000000000000000000000000000000100000000000000000000000004000000000000000000004000000000002000000000020000000000000000000000000000000001000000000008000000000000000000000000000000000000000000000000004000000000000000000000000000400000000000000000000000080000000000000000000040000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000000400000000',
  transactionsRoot: '0xb141e983c8f5f2e61239f5936d5e54c66d4fdd926ae95f76f108c808919d94de',
  stateRoot: '0xa0a67aa84a781828bd9970738415aaaac9a145babd15b6f5ea1457b3b043f895',
  receiptsRoot: '0x5dea430287fd5418724a625f899236267ccc75cfc7108edd504d82e98717c3c5',
  miner: '0x0000000000000000000000000000000000000000',
  mixHash: '0x',
  difficulty: '0',
  totalDifficulty: '0',
  extraData: '0x',
  size: 737,
  gasLimit: 4503599627370495,
  gasUsed: 43803,
  timestamp: 1686667466,
  transactions: [
    {
      blockHash: '0x60f6fd56770b2d04e71e6add3d9b4d85b3599f44bcddb9e368be1c21c0c43aa2',
      blockNumber: 94119265,
      from: '0xF2f3776B9F69A9302c897AEc26b59027e01D36Cc',
      gas: 2000000,
      gasPrice: '70000000',
      hash: '0x9a135dcf4f15dfadf995c042cdf1a39bdbb16510fb91b77bac0eeba9f6983d7e',
      input: '0xf4eb73d4000000000000000000000000c7ea819ebf08e5ff481d4708a602f92380afbb0a000000000000000000000000000000000000000000000000000000000820b3bd00000000000000000000000000000000000000000000000000000000648880a3',
      nonce: 1332762,
      v: '0x9c8a82c7',
      r: '0x9bd6a2da99ad882d8d1bf6a03fe72b53ab808fc6343f208678782a6bf5e0788',
      s: '0x6eb1711cd7a7f56810c774ac6fd81df51744ddcacf4e4546b2135731085b1ba7',
      to: null,
      type: 0,
      transactionIndex: 0,
      value: '0'
    }
  ],
  uncles: []
}
```

Response contains `hash, transactions`

#### Actual behavior

```
New block v4: {
  number: 94119300n,
  parentHash: '0x0caa8d54fc148033eb2f127abbb9c99974337dffe8fff151b95945cdb7da394d',
  nonce: 0n,
  sha3Uncles: '0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347',
  logsBloom: '0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000',
  transactionsRoot: '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421',
  stateRoot: '0x75cad9fd31e16311dfdb3efd56f46ea4335115b5aea00bafaa5ee8f4586fd4f6',
  miner: '0x0000000000000000000000000000000000000000',
  difficulty: 0n,
  extraData: '0x',
  gasLimit: 4503599627370495n,
  gasUsed: 0n,
  timestamp: 1686667509n
}
```

```
import { Web3 } from 'web3'

(async () => {
  const url = 'wss://mainnet.aurora.dev'
  const web3Lib = new Web3(
    new Web3.providers.WebsocketProvider(url)
  )

  const subscription = await web3Lib.eth.subscribe('newBlockHeaders');

  subscription.on('data', async blockhead => {
    console.log('New block v4:', blockhead);
  });
  subscription.on('error', error =>
    console.log('Error when subscribing to New block header: ', error)
  );
})()

```

#### Environment
node: v18.16.0
web3.js: 4.0.1

Spacesai1or:
closes #6198

### DIFF
--- a/packages/web3-eth/src/schemas.ts
+++ b/packages/web3-eth/src/schemas.ts
@@ -321,13 +321,37 @@ export const blockSchema = {
 	},
 };
 
+export const withdrawalsSchema = {
+	type: 'object',
+	properties: {
+		index: {
+			format: 'uint',
+		},
+		validatorIndex: {
+			format: 'uint',
+		},
+		address: {
+			format: 'bytes32',
+		},
+		amount: {
+			format: 'uint',
+		},
+	},
+};
+
 export const blockHeaderSchema = {
 	type: 'object',
 	properties: {
+		author: {
+			format: 'bytes32',
+		},
+		hash: {
+			format: 'bytes32',
+		},
 		parentHash: {
 			format: 'bytes32',
 		},
-		receiptRoot: {
+		receiptsRoot: {
 			format: 'bytes32',
 		},
 		miner: {
@@ -339,10 +363,16 @@ export const blockHeaderSchema = {
 		transactionsRoot: {
 			format: 'bytes32',
 		},
+		withdrawalsRoot: {
+			format: 'bytes32',
+		},
 		logsBloom: {
 			format: 'bytes256',
 		},
 		difficulty: {
+			format: 'uint',
+		},
+		totalDifficulty: {
 			format: 'uint',
 		},
 		number: {
@@ -365,6 +395,36 @@ export const blockHeaderSchema = {
 		},
 		sha3Uncles: {
 			format: 'bytes32',
+		},
+		size: {
+			format: 'uint',
+		},
+		baseFeePerGas: {
+			format: 'uint',
+		},
+		excessDataGas: {
+			format: 'uint',
+		},
+		mixHash: {
+			format: 'bytes32',
+		},
+		transactions: {
+			type: 'array',
+			items: {
+				format: 'bytes32',
+			},
+		},
+		uncles: {
+			type: 'array',
+			items: {
+				format: 'bytes32',
+			},
+		},
+		withdrawals: {
+			type: 'array',
+			items: {
+				...withdrawalsSchema,
+			},
 		},
 	},
 };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1699,6 +1699,7 @@ If there are any bugs, improvements, optimizations or any new feature proposal f
 
 ## [Unreleased]
 
+### Fixed
 
 #### web3-utils
 
@@ -1707,3 +1708,7 @@ If there are any bugs, improvements, optimizations or any new feature proposal f
 #### web3-rpc-methods
 
 -   Rpc method `getPastLogs` accept blockHash as a parameter https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getlogs (#6181)
+
+#### web3-eth
+
+-   Missing `blockHeaderSchema` properties causing some properties to not appear in response of `newHeads` subscription (#6243)
--- a/packages/web3-eth/CHANGELOG.md
+++ b/packages/web3-eth/CHANGELOG.md
@@ -156,3 +156,7 @@ Documentation:
 -   Dependencies updated
 
 ## [Unreleased]
+
+### Fixed
+
+-   Missing `blockHeaderSchema` properties causing some properties to not appear in response of `newHeads` subscription (#6243)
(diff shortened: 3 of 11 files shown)


luu-alex (approved):

packages/web3-eth/src/schemas.ts
```diff
@@ -321,13 +321,37 @@ export const blockSchema = {
...
+		amount: {
+			format: 'uint',
+		},
+	},
+};
+
 export const blockHeaderSchema = {
 	type: 'object',
 	properties: {
+		author: {
```
jdevcs, line 345: some of these fields are not aligned with EL specs, there can be possibilities:

- EL specs not updated
- geth specific fields

so could you add an integration test , so when we add more EL clients we can catch issue earlier.

avkos (approved):
LGTM

packages/web3-eth/src/schemas.ts
Spacesai1or: @jdevcs What EL spec are you looking at? I've been using [this](https://ethereum.github.io/execution-apis/api-documentation/) one and don't even see `eth_subscribe` listed

Regardless, these added fields don't exist on Geth's response, but are instead from Nethermind's. It's okay and actually necessary to have these fields in the `blockHeaderSchema`, because it's used to format responses. Meaning as is, any fields not mentioned in this schema will be dropped from the response before it's returned to the user. No error is caused if the fields are not present. This does, however, beg the question of how do we declare the [BlockHeaderOutput](https://github.com/web3/web3.js/blob/6992ccdd555604ff22d92c0eba0aa5f366f8b3cf/packages/web3-types/src/eth_types.ts#L154) type - do we include these properties as optional since they _could_ be there if the connected RPC client uses them? Or do we only include the common ones as the type is declared now?

The extra fields included by Nethermind that don't seem to be available using Geth:

- author
- totalDifficulty
- size
- excessDataGas
- mixHash
- transactions
- uncles
- withdrawals

jdevcs: I wasnt able to find `author` field in EL Specs https://github.com/ethereum/execution-apis/blob/4da6a02f094d9d8c826272f6ceac2d751866f132/src/schemas/block.yaml#L22 in block schema at first place, and later there is no `eth_subscribe` in EL specs ( as you found as well in specs doc). 


If we include integration tests ( we will need to have a check for Nethermind in tests and expect that these additional fields are present), if its breaking in future for Nethermind we will be able to detect. Currently we dnt have NM, so it can be skipped.

> This does, however, beg the question of how do we declare the [BlockHeaderOutput](https://github.com/web3/web3.js/blob/6992ccdd555604ff22d92c0eba0aa5f366f8b3cf/packages/web3-types/src/eth_types.ts#L154) type - do we include these properties as optional since they could be there if the connected RPC client uses them? Or do we only include the common ones as the type is declared now?

I'll suggest lets have additional types as optional and add doc comments that these are specific for NM.

Issue #6168, opened 2023-06-08
Unsafe eval


gintsmurans:
Hi! 

Congrats on the new version! Just tried updating to v4.0.1 - lucky coincidence that it was released yesterday.  :)

Anyways, I am using web3js in browser extension and after update I am getting error: "Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'wasm-unsafe-eval' 'inline-speculation-rules' ". I tracked it down to a validator.js which is using "is-my-json-valid" package, which in turn is using genfun from generate-function where the error comes from. 

This makes current version unusable in extension environment. Does anyone have any suggestions on how to solve it?

jdevcs:
@gintsmurans Thanks for checking 4.x and sharing feedback, 
is-my-json-valid is on list for [replacement](https://github.com/web3/web3.js/issues/6156).

luu-alex:
Hi there @gintsmurans , thanks for reporting this issue. We plan on removing this library, as its been causing a few issues. unfortunately we don't have a workaround this yet but if something comes up that would be great

fatchan:
> This makes current version unusable in extension environment. 

Or anybody not willing to tolerate allowing eval() in their CSP. Would be great to have this fixed 👍🏻

PR #6264, opened 2023-07-07
6156 find replacement for is my json valid

### LINKED ISSUE #6156: Find replacement for is-my-json-valid
#### Expected behavior
- library is not browser friendly as it uses dependencies that depend on nodejs "util" as well as "process" global which causes build to fail in angular. Related to #6146 

#### Actual behavior

Using browser friendly library by either forking is-my-json valid or finding some browser friendly validator

1. [First step]
2. [Second step]
3. [and so on...]

avkos:
## Description
Fixes #6156  and #6168

## Type of change

## Checklist:


packages/web3-validator/src/types.ts
```diff
@@ -143,10 +143,7 @@ export type Schema = {
 };
 export interface Validate {
 	(value: Json): boolean;
-	errors?: ValidationError[];
+	errors?: ZodIssueBase[];
```
jdevcs: This will be breaking change, as Validation interface is being exported and available to users. For this during release we will need to bump major version of validator package. so:

1. mention this in changelog
2. I think create some validation error wrapper around `ZodIssueBase` , so we are not tied to specific lib (like we were is-my-json-valid.ValidationError)

packages/web3-validator/src/types.ts
```diff
@@ -143,10 +143,7 @@ export type Schema = {
...
 }
-export type RawValidationError = ValidationError & {
```
jdevcs, line 148: (breaking change ) it was also exported and will no longer available so : changelog

packages/web3-validator/src/validator.ts
```diff
@@ -16,53 +16,212 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
 		return Validator.validatorInstance;
 	}
-	private readonly _schemas: Map<string, Validate> = new Map();
-	public getSchema(key: string) {
-		return this._schemas.get(key);
-	}
-
-	public addSchema(key: string, schema: Schema) {
-		this._schemas.set(key, this.createValidator(schema));
-	}
```
jdevcs, line 44: (its breaking change) For all public functions in this class , add in change log, that these will not be available.

jdevcs:
LGTM, just need some minor changes. 

Plus with zod lib using in this PR, could you manually test some recent issues reported for browser support due to is my json valid lib.

packages/web3-validator/src/types.ts
avkos: ValidationError wrapper was added

packages/web3-validator/src/validator.ts
avkos: Class Validator is not public to users. It is only for internal Web3Validator usage

packages/web3-validator/src/validation/boolean.ts
```diff
@@ -35,9 +35,6 @@ export const isBoolean = (value: ValidInputTypes) => {
 		return value === '0x1' || value === '0x0';
 	}
 
-	if (typeof value === 'number') {
-		return value === 1 || value === 0;
-	}
-
-	return false;
+	// type === number
+	return value === 1 || value === 0;
```
avkos, line 39: here can be only number type. we don't need extra conditions.

packages/web3-validator/src/types.ts
avkos: done

jdevcs (approved):

packages/web3-validator/src/validator.ts
jdevcs: at two places:
https://github.com/web3/web3.js/blob/ad377d16a7e9a8e82857d6102bb4efd95c543e4f/packages/web3-validator/src/default_validator.ts#L20
https://github.com/web3/web3.js/blob/ad377d16a7e9a8e82857d6102bb4efd95c543e4f/packages/web3-validator/src/index.ts#L19

and

https://github.com/web3/web3.js/blob/ad377d16a7e9a8e82857d6102bb4efd95c543e4f/packages/web3/src/index.ts#L342

This will be breaking change for main web3 package as well.

luu-alex (approved):

Issue #6261, opened 2023-07-06
4.x Web3 custom RPC plugin Tutorial 


jdevcs:
4.x Web3 plugin tutorial for custom RPC methods ( for motivating community to use plugin feature in place of web3.extend ), also link this in migration guide under `web3.extend` part.

Spacesai1or:
Essentially a guide for this [plugin example code](https://github.com/web3/web3.js/blob/4.x/tools/web3-plugin-example/src/custom_rpc_methods.ts)

PR #6279, opened 2023-07-13
custom RPC methods guide

jdevcs:
## Description

https://github.com/web3/web3.js/issues/6261

Simplified version of creating plugin, where plugin and plugin user code is consistent. 

## Type of change

## Checklist:

### DIFF
--- a/docs/docs/guides/advanced/support_additional_rpc_methods/index.md
+++ b/docs/docs/guides/advanced/support_additional_rpc_methods/index.md
@@ -7,7 +7,7 @@ sidebar_label: Add custom RPC methods
 
 #### Introduction
 
-Web3.js is a popular library for interacting with Ethereum blockchain. It provides a set of APIs to interact with Ethereum nodes via JSON-RPC calls. For adding new JSON-RPC function calls to the library, you can do so using the plugin feature in web3.js 4.x. This allows you to extend the functionality of Web3.js and add support for new JSON-RPC methods.
+Web3.js is a popular library for interacting with the Ethereum blockchain. It provides a set of APIs to interact with Ethereum nodes via JSON-RPC calls. For adding new JSON-RPC function calls to the library, you can do so using the plugin feature in web3.js 4.x. This allows you to extend the functionality of Web3.js and add support for new JSON-RPC methods.
 
 :::caution
 In Web3.js 1.x, `web3.extend()` function could be used to add new JSON-RPC methods. However, this function is not available in Web3.js 4.x. Instead, the plugin feature can be used to achieve the feature of extending web3 functionality.
--- a/docs/docs/guides/advanced/support_additional_rpc_methods/index.md
+++ b/docs/docs/guides/advanced/support_additional_rpc_methods/index.md
@@ -10,7 +10,7 @@ sidebar_label: Add custom RPC methods
 Web3.js is a popular library for interacting with the Ethereum blockchain. It provides a set of APIs to interact with Ethereum nodes via JSON-RPC calls. For adding new JSON-RPC function calls to the library, you can do so using the plugin feature in web3.js 4.x. This allows you to extend the functionality of Web3.js and add support for new JSON-RPC methods.
 
 :::caution
-In Web3.js 1.x, `web3.extend()` function could be used to add new JSON-RPC methods. However, this function is not available in Web3.js 4.x. Instead, the plugin feature can be used to achieve the feature of extending web3 functionality.
+In Web3.js 1.x, `web3.extend()` function could be used to add new JSON-RPC methods. However, this function is not available in Web3.js v4. Instead, the plugin feature can be used to achieve the feature of extending web3 functionality.
 :::
 
 Following tutorial will guide you through the process of creating a custom plugin to extend the functionality of web3.js 4.x and add support for new RPC methods.
--- a/docs/docs/guides/advanced/support_additional_rpc_methods/index.md
+++ b/docs/docs/guides/advanced/support_additional_rpc_methods/index.md
@@ -1,0 +1,114 @@
+---
+sidebar_position: 0
+sidebar_label: Add custom RPC methods
+---
+
+# Add custom RPC methods
+
+#### Introduction
+
+Web3.js is a popular library for interacting with Ethereum blockchain. It provides a set of APIs to interact with Ethereum nodes via JSON-RPC calls. For adding new JSON-RPC function calls to the library, you can do so using the plugin feature in web3.js 4.x. This allows you to extend the functionality of Web3.js and add support for new JSON-RPC methods.
+
+:::caution
+In Web3.js 1.x, `web3.extend()` function could be used to add new JSON-RPC methods. However, this function is not available in Web3.js 4.x. Instead, the plugin feature can be used to achieve the feature of extending web3 functionality.
+:::
+
+Following tutorial will guide you through the process of creating a custom plugin to extend the functionality of web3.js 4.x and add support for new RPC methods.
+
+#### Creating new RPC methods Plugin in 4 Steps
+
+1. First add web3.js as peer dependency in project´s `package.json` and create a TypeScript class for your plugin. This class should extend the `Web3Plugin` class provided by web3.js.
+
+:::info
+This will give your plugin access to [requestManager](/api/web3-core/class/Web3Context#requestManager) and [accountProvider](/api/web3-core/class/Web3Context#accountProvider).
+:::caution
+
+```ts
+import { Web3PluginBase } from 'web3';
+
+export default class CustomRpcMethodsPlugin extends Web3PluginBase {
+	// step 2
+	// ...
+}
+```
+
+2. After that add public `pluginNamespace` property. This will be used to access your plugin, as mentioned in step number 5 code example.
+
+```ts
+import { Web3PluginBase } from 'web3';
+
+export default class CustomRpcMethodsPlugin extends Web3PluginBase {
+	public pluginNamespace = 'customRpcMethods'; // step 3
+}
+```
+
+3. Once plugin class is created using above mentioned steps, its very easy to add new RPC methods like:
+
+```ts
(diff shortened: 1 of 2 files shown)


docs/docs/guides/advanced/support_additional_rpc_methods/index.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Using Web3 Custom PRC Plugin (with web3 instance)
+
+5. First add plugin in your plugin consumer project's `package.json`, create web3 and plugin instances, and after that use `.registerPlugin` method with some web3.js module (in following example its registered with main web3).
+
+Once plugin is registered its custom methods will be available to use.
+
+```ts
+import { Web3 } from 'web3';
+import CustomRpcMethodsPlugin from 'web3-plugin-example';
```
avkos, line 103: I suggest importing plugin from a local folder like this
`import CustomRpcMethodsPlugin from './path/to/plugin';`, because importing from `'web3-plugin-example'` looks like a separate npm module and can be confusing for users.

docs/docs/guides/advanced/support_additional_rpc_methods/index.md
```diff
@@ -0,0 +1,114 @@
...
+}
+```
+
+2. After that add public `pluginNamespace` property. This will be used to access your plugin, as mentioned in step number 5 code example.
+
+```ts
+import { Web3PluginBase } from 'web3';
+
+export default class CustomRpcMethodsPlugin extends Web3PluginBase {
+	public pluginNamespace = 'customRpcMethods'; // step 3
```
avkos: I think we do need to have comments like `// step 3` or `// step 4` in documentation especially comment `step 3` in step 2. It is not valuable for users. What do you think @jdevcs?

avkos (changes requested):

docs/docs/guides/advanced/support_additional_rpc_methods/index.md
jdevcs: but normally plugins are supposed to be in independent repo and npm package.

docs/docs/guides/advanced/support_additional_rpc_methods/index.md
jdevcs: updated

docs/docs/guides/advanced/support_additional_rpc_methods/index.md
```diff
@@ -0,0 +1,113 @@
+---
+sidebar_position: 0
+sidebar_label: Add custom RPC methods
+---
+
+# Add custom RPC methods
+
+#### Introduction
+
+Web3.js is a popular library for interacting with Ethereum blockchain. It provides a set of APIs to interact with Ethereum nodes via JSON-RPC calls. For adding new JSON-RPC function calls to the library, you can do so using the plugin feature in web3.js 4.x. This allows you to extend the functionality of Web3.js and add support for new JSON-RPC methods.
```
luu-alex:
```suggestion
Web3.js is a popular library for interacting with the Ethereum blockchain. It provides a set of APIs to interact with Ethereum nodes via JSON-RPC calls. For adding new JSON-RPC function calls to the library, you can do so using the plugin feature in web3.js 4.x. This allows you to extend the functionality of Web3.js and add support for new JSON-RPC methods.
```

docs/docs/guides/advanced/support_additional_rpc_methods/index.md
```diff
@@ -0,0 +1,113 @@
...
+
+:::caution
+In Web3.js 1.x, `web3.extend()` function could be used to add new JSON-RPC methods. However, this function is not available in Web3.js 4.x. Instead, the plugin feature can be used to achieve the feature of extending web3 functionality.
```
luu-alex:
```suggestion
In Web3.js 1.x, `web3.extend()` function could be used to add new JSON-RPC methods. However, this function is not available in Web3.js v4. Instead, the plugin feature can be used to achieve the feature of extending web3 functionality.
```
I think we should move from calling 4.x to v4, as we did in the article and on the [site](https://web3js.org/#/)

docs/docs/guides/advanced/support_additional_rpc_methods/index.md
avkos: yes. if use some other plugin. but if the user just creates it inside some private project he needs to user it locally.
in this guide, we do not have information on how to deploy the plugin to the npm and make it public. That's why I think it will be better to use local import

docs/docs/guides/web3_upgrade_guide/1.x/index.md
```diff
@@ -36,9 +36,10 @@ For example, the approach to subscribing-to and listening-for blockchain events
 
 However, the approach to subscribing to Provider events remains the same, utilizing callbacks as explained in the [Providers Events Listening guide](../../web3_providers_guide/events_listening.md). It is important to note that Providers have undergone some breaking changes, including the renaming of the `on('close', ...)` to `on('disconnect', ...)`.
 
-### Not Implemented or Exported
+### Not Available
+
+-   [web3.extend](https://web3js.readthedocs.io/en/v1.7.3/web3.html#extend) function is not available in web3.js 4.x, instead of this for extending web3 functionality more powerful plugin feature is available. Follow [Add custom RPC methods guide](/guides/advanced/support_additional_rpc_methods/) for this new feature.
```
Muhammad-Altabba:
```suggestion
- The [`web3.extend`](https://web3js.readthedocs.io/en/v1.7.3/web3.html#extend) function is not available in web3.js v4. Instead, for extending web3 functionality, a more powerful plugin feature is available. Follow the [Add custom RPC methods guide](/guides/advanced/support_additional_rpc_methods/) for this new feature.
```

docs/docs/guides/advanced/support_additional_rpc_methods/index.md
```diff
@@ -0,0 +1,113 @@
...
+
+#### Using Web3 Custom PRC Plugin (with web3 instance)
+
+5. First add plugin in your plugin consumer project's `package.json`, create web3 and plugin instances, and after that use `.registerPlugin` method with some web3.js module (in following example its registered with main web3).
+
+Once plugin is registered its custom methods will be available to use.
+
+```ts
+import { Web3 } from 'web3';
+import CustomRpcMethodsPlugin from 'web3-plugin-example';
```
Muhammad-Altabba, line 103: I think it will be the best to add a comment about publishing to npm registry. And to use this for import: 
```suggestion
import { CustomRpcMethodsPlugin } from 'web3-plugin-example';
```
And also mention that if the user did not publish to npm registry, he need to use:
`import { CustomRpcMethodsPlugin } from './path/to/plugin';`

docs/docs/guides/advanced/support_additional_rpc_methods/index.md
```diff
@@ -0,0 +1,113 @@
...
+
+	interface Web3Context {
+		customRpcMethods: CustomRpcMethodsPlugin;
+	}
+}
+```
+
+#### Using Web3 Custom PRC Plugin (with web3 instance)
+
+5. First add plugin in your plugin consumer project's `package.json`, create web3 and plugin instances, and after that use `.registerPlugin` method with some web3.js module (in following example its registered with main web3).
```
Muhammad-Altabba, line 97: I think that the usage of the plugin is good to be different from the steps for creating the plugin. So, I think no need for having `5. ` here.

Muhammad-Altabba (commented):
I am just thinking that actually, if the user does not ever want to write a plugin. He can simply create a child class, that extends our `Web3` and write his custom functionality like this:

```ts
import { Web3 } from 'web3';

export default class MyWeb3 extends Web3 {

	public async customRpcMethod() {
		return this.requestManager.send({
			// plugin has access to web3.js internal features like request manager
			method: 'custom_rpc_method',
			params: [],
		});
	}
}

const web3 = new MyWeb3('http://127.0.0.1:8545');

const res = web3.customRpcMethods.customRpcMethod();
```

And we still need to encourage the users especially if they are the company who provides the custom RPC functionality. And for this, I suggest writing a clear note. Something similar to: https://github.com/web3/web3.js/pull/6285/files#diff-725414e373fdb2402f2cdcaaeb39b94d0dbe9e70e69fbe423fbbaf59fa55bdefR14

jdevcs:
@Muhammad-Altabba @avkos 
From web3.js 4.x vision perspective its more recommended to create and publish plugins (for reusing, and supporting to build community of plugins). and if any user want to choose any other path there is no restriction. How ever in our documentation we will facilitate and educate users to create, and publish plugins, if any one want to extend web3.js functionality. This will increase number of plugins and help us to build ecosystem.

docs/docs/guides/advanced/support_additional_rpc_methods/index.md
jdevcs: in that case :
how user will know , "how to register a plugin"? and which function is there for this purpose? by navigating more in documentation, that will add extra friction.

avkos (approved):

docs/docs/guides/advanced/support_additional_rpc_methods/index.md
jdevcs: added:
image
regarding import, its default export in doc so can be default imported.

I think we should only encourage users to make public plugins ( repo, and on public npm ) so we have more # of plugins.

jdevcs: added info note. I'll still recommend to publish so we have more # of plugins available, instead of encouraging local usage.

Spacesai1or (approved):
Awesome job Junaid, really good resource to point users to when they have questions

docs/docs/guides/advanced/support_additional_rpc_methods/index.md
Muhammad-Altabba: Yes sure. Sorry, I meant to just remove the number and keep the section.

PR #6283, opened 2023-07-13
Replace ethereumjs-util with @ethereumjs/util

### LINKED ISSUE #6021: There are dependencies to "ethereumjs-util" and "@ethereumjs/util"

### Current Behavior

Internally the code refers to `ethereumjs-util` but due to also using `@ethereumjs/tx` and `@ehterumjs/common` but those modules referring to `ethereumjs-util` by it's new name `@ethereumjs/util` there is some code duplication inside of packaged output which also makes it harder to properly upgrade to newer versions of crypto modules.

### Expected Behavior

Ideally `ethereumjs-util` should no longer be used if the new naming is used for the other modules.

### Steps to Reproduce

```js
import "web3-eth-accounts"

// will cause both modules to be in the bundle because web3-eth-accounts requires `ethereumjs-util` and `@ethereumjs/tx` and 
// `@ethereumjs/common` which in turn include `@ethereumjs/util`
```

### Web3.js Version

1.9.0

### Environment

- Operating System: Ubuntu 22.10
- Browser: Chrome Version 111.0.5563.149 (Official Build) (64-bit)
- Node.js Version: v18.14.0
- NPM Version: yarn 3.3.0 npm 9.3.1
- Webpack: 5.79.0

### Anything Else?

The current way to prevent upgrade problems is to leave
"ethereum-cryptography": "1.2.0",
and
"@noble/secp256k1": "1.7.1",

avkos:
## Description

Fixes #6021

## Type of change

## Checklist:


package.json
```diff
@@ -106,6 +106,7 @@
         "bignumber.js": "^9.0.1",
         "bn.js": "^5.2.1",
         "browserify": "^16.5.2",
+        "browserify-zlib": "^0.2.0",
```
jdevcs, line 109: why this lib is added?

luu-alex (approved):

luu-alex:
other than junaids comment lgtm

package.json
avkos: we need this Polyfill otherwise build fails by [error](https://github.com/web3/web3.js/actions/runs/5548744411/jobs/10132070027)

jdevcs (approved):

Muhammad-Altabba (approved):

test/e2e.method.signing.js
```diff
@@ -539,6 +539,7 @@ describe('transaction and message signing [ @E2E ]', function() {
     });
 
     it('eth.accounts.sign', async function(){
+        this.timeout(30000);
```
Spacesai1or: Seems weird to require such a long timeout for a signing call

jdevcs:
@avkos , before merging tests should be passing:
` Error: Returned error: The method personal_sign does not exist/is not available`

PR #6294, opened 2023-07-19
Transaction `input` and `data` refactors

### LINKED ISSUE #6183: input parameter added when calling estimateGas for a specific contract call in v4.0.1
#### Expected behavior

When calling `estimateGas` for a contract deployment it should use the `data` parameter only to estimate gas and should not use an `input` parameter. Take the following [example on the docs site](https://docs.web3js.org/guides/basics/eth#step-3-interact-with-the-ethereum-blockchain-using-web3js):

```js
    // Example from wbe3.js docs site
    let contract = new web3.eth.Contract(abi, undefined);
    const deployment = contract.deploy({
        data: '0x...', // Insert bytecode
        // @ts-expect-error
        arguments: [1],
    });
    const estimatedGas = await deployment.estimateGas(
        {
            from: acc, 
        },
        DEFAULT_RETURN_FORMAT, // the returned data will be formatted as a bigint
    );
```

This example results in the following JSON RPC call:

```
{
    jsonrpc: '2.0',
    id: '4691de88-3948-47fc-b7fe-e8bfaef6e9ba',
    method: 'eth_estimateGas',
    params: [
      {
       // Removed actual bytecode as its quite long, but its the same for both `input` and `data`
        input: '0x...',
        data: '0x...' 
      },
      'latest'
    ]
  }
```

`eth_call` and `eth_estimateGas` accept a `data` parameter, but not an `input` parameter per the [JSON-RPC API]( https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_call).

Although this doesn't cause any issue on Ethereuem, it does cause issues on Moonbeam and i'm not sure why an `input` field is automatically added, but I don't think that this should be the default behavior. 

This issue is also new as of v4.0.1, this was not an issue in v1.x

To reproduce the issue, you can use the above example from your docs site and set the Web3 provider to Moonbeam or the Moonbeam TestNet, Moonbase Alpha:

```
  const web3 = new Web3('https://rpc.api.moonbase.moonbeam.network'); // Moonbase Alpha
```

#### Logs

The error is as follows:

```
InvalidResponseError: Returned error: unknown field `input`, expected one of `from`, `to`, `gasPrice`, `maxFeePerGas`, `maxPriorityFeePerGas`, `gas`, `value`, `data`, `nonce`, `accessList`, `type` at line 1 column 8
    at Web3RequestManager._processJsonRpcResponse (file:///Users/es/Documents/work/moonbeam-docs-test-suite/node_modules/web3-core/lib/esm/web3_request_manager.js:252:23)
    at Web3RequestManager.<anonymous> (file:///Users/es/Documents/work/moonbeam-docs-test-suite/node_modules/web3-core/lib/esm/web3_request_manager.js:150:29)
    at fulfilled (file:///Users/es/Documents/work/moonbeam-docs-test-suite/node_modules/web3-core/lib/esm/web3_request_manager.js:20:58)
  innerError: {
    code: -32602,
    message: 'unknown field `input`, expected one of `from`, `to`, `gasPrice`, `maxFeePerGas`, `maxPriorityFeePerGas`, `gas`, `value`, `data`, `nonce`, `accessList`, `type` at line 1 column 8'
  },
  code: 101,
  data: undefined,
  request: {
    jsonrpc: '2.0',
    id: '4691de88-3948-47fc-b7fe-e8bfaef6e9ba',
    method: 'eth_estimateGas',
    params: [
      {
        input: '0x...',
        data: '0x...'
      },
      'latest'
    ]
  }
}
```

#### Environment

npm v8.19.2
node v18.12.1
web3.js v4.0.1

Spacesai1or:
closes #6183


packages/web3-eth/src/utils/prepare_transaction_for_signing.ts
```diff
@@ -42,7 +42,7 @@ const getEthereumjsTxDataFromTransaction = (
 	gasLimit: transaction.gasLimit ?? transaction.gas,
 	to: transaction.to,
 	value: transaction.value,
-	data: transaction.input,
+	data: transaction.data ?? transaction.input,
```
Spacesai1or, line 45: The property name is still `data` because that's what [TransactionFactory.fromTxData](https://github.com/web3/web3.js/blob/933ef512d5c10cf092b21ab82da38c29cde575f8/packages/web3-eth/src/utils/prepare_transaction_for_signing.ts#L145) is expecting

packages/web3-types/src/eth_types.ts
```diff
@@ -409,7 +409,8 @@ export interface PopulatedUnsignedBaseTransaction {
 	gas?: Numbers;
 	gasPrice: Numbers;
 	type: Numbers;
-	input: Bytes;
+	input?: Bytes;
+	data?: Bytes;
```
Spacesai1or, line 413: This is a breaking change for `eth_types`, but `input` won't be apart of this object if the user uses `data`

luu-alex:
LGTM needs a changelog update

luu-alex (approved):

jdevcs (approved):

packages/web3-eth/test/integration/web3_eth/sign_transaction.test.ts
```diff
@@ -47,17 +48,21 @@ describe('Web3Eth.signTransaction', () => {
...
 				gasPrice: BigInt(1000000001),
 				gas: BigInt(21000),
 				value: BigInt(1),
 				to: transaction.to,
-				input: '0x',
 			},
-		});
+		};
+		if (getSystemTestBackend() === 'geth') expectedResponse.tx.input = '0x';
+		else if (getSystemTestBackend() === 'ganache') expectedResponse.tx.data = '0x';
```
Muhammad-Altabba: This indicates a breaking change.

We cannot be sure that no user used to fill `input` and then read `data` or vice versa.

Muhammad-Altabba (changes requested):
I think when we need to update our tests because they break after a change. Then this is clearly a breaking change.
And there are many changes in the tests. We cannot be sure that no user used to fill `input` and then read `data` or vice versa...

I think if we want to pass those changes we will need to go for version 5. But when thinking about the original issue, I do not see that we need to update our code to support an unsupported use that does not comply with the specs. Especially that: [they said they will update their code to comply](https://github.com/web3/web3.js/issues/6183#issuecomment-1613847426). And I expect that every similar case should be resolved by those who do not comply with the specs but not us.... Am I missing something here? Or what do you all think?

avkos (approved):

avkos (approved):

packages/web3-eth/src/utils/format_transaction.ts
```diff
@@ -17,18 +17,23 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
 	ReturnFormat extends DataFormat = typeof DEFAULT_RETURN_FORMAT,
 	TransactionType extends Transaction = Transaction,
 >(
 	transaction: TransactionType,
 	returnFormat: ReturnFormat = DEFAULT_RETURN_FORMAT as ReturnFormat,
-	options: { transactionSchema: ValidationSchemaInput | typeof transactionSchema } = {
-		transactionSchema,
+	options: {
+		transactionSchema?: ValidationSchemaInput | typeof transactionSchema;
+		fillInputAndData?: boolean;
```
Spacesai1or, line 33: A new parameter, `fillInputAndData`, was added to the `options` argument for `formatTransaction`. This method is used for formatting both user submitted and RPC response transaction objects. So, we need a flag to determine whether to leave the transaction object as given, or fill in the `input` or `data` properties if they weren't present. The `fillInputAndData` option is set in `rpc_method_wrappers.ts` for relevant methods

avkos (approved):

jdevcs (approved):

Spacesai1or:
@Muhammad-Altabba Need your re-review

packages/web3-eth/src/utils/format_transaction.ts
Muhammad-Altabba: This, and things similar, needs to be in the documentation.

Muhammad-Altabba (approved):
Good job!
All good now I think.
However, the documentation needs to be updated accordingly.

Issue #6355, opened 2023-08-14
Data is missing from methods.myMethod


luu-alex:
needs investigation, from discord user. Looks like the change from data/input has affected methods.mymethod:
Is there a way to force the format of the jsonRPC calls made by methods.myMethod to include the right fields?

Currently generates (4.1.0rc)

```
{"jsonrpc":"2.0","id":"b4c23746-97b2-4512-9dc2-2492d81151ce","method":"eth_call","params":[{"to":"0x5FbDB2315678afecb367f032d93F642f64180aa3","input":"0x70a082310000000000000000000000007a06dcfb82c84eb0e94bec11161012a3374ad98f"},"latest"]}

```

4.0.3

```
{"jsonrpc":"2.0","id":"b4c23746-97b2-4512-9dc2-2492d81151ce","method":"eth_call","params":[{"to":"0x5FbDB2315678afecb367f032d93F642f64180aa3","input":"0x70a082310000000000000000000000007a06dcfb82c84eb0e94bec11161012a3374ad98f","data":"0x70a082310000000000000000000000007a06dcfb82c84eb0e94bec11161012a3374ad98f"},"latest"]}

```

When it should be generating

```
{"jsonrpc":"2.0","id":"b4c23746-97b2-4512-9dc2-2492d81151ce","method":"eth_call","params":[{"to":"0x5FbDB2315678afecb367f032d93F642f64180aa3","data":"0x70a082310000000000000000000000007a06dcfb82c84eb0e94bec11161012a3374ad98f"},"latest"]}

```

Using either option currently gets met with:
```
{
    "jsonrpc": "2.0",
    "id": "b4c23746-97b2-4512-9dc2-2492d81151ce",
    "error": {
        "code": -32602,
        "message": "unknown field `input`, expected one of `from`, `to`, `gasPrice`, `maxFeePerGas`, `maxPriorityFeePerGas`, `gas`, `value`, `data`, `nonce`, `chainId`, `accessList`, `type`"
    }
}
```

#### Steps to reproduce the behavior

1. [First step]
2. [Second step]
3. [and so on...]

luu-alex:
[...]

PR #6377, opened 2023-08-23
6355 - allow data to be used in contract methods.

luu-alex:
## Description
Fixing issues related to the [merge](https://github.com/web3/web3.js/pull/6294) about transaction data/input. As well contains community PR https://github.com/web3/web3.js/pull/6378
Should fix these issues  #6355 #6400

After removing transactions to contain both input and data, this affected our contracts. contract methods were defaulting to sending only input in the calls which caused some people issues that required their methods to be sent with the data property. If they tried to use data in their contract options, this would error due to our [typings](https://github.com/web3/web3.js/pull/6377/files#diff-7fa8f782e68ce845dbb6e8446b86041697fa427db5485b9c63bf689d0b32ed81R378) only accepting input.

This PR: 
- allows users to provide data in their contract and contract options, instead of just defaulting to using the input property. 
- Adds a web3config `contractDataInputFill` to web3context option to allow users to pass encodedAbi through `data`, `input` or for both properties  

## Type of change

## Checklist:


packages/web3-eth-contract/src/utils.ts
```diff
@@ -191,15 +222,32 @@ export const getCreateAccessListParams = ({
...
+		txParams = {
+			...txParams,
+			input: encodeMethodABI(abi, params, txParams.data as HexString),
+		};
+	} else {
+		// if no data is specified, default to input
+		txParams = {
+			...txParams,
+			input: encodeMethodABI(abi, params, txParams.data as HexString),
+		};
```
Muhammad-Altabba: Two points here:
- The code insdie the last `else if` is identical to the code inside the `else`.
- Why we need the code inside the last `else if` and `else`? Inside those blocks the `txParams.data` is always `undefined`, right?

luu-alex: - ur right the last else is redundant.
- it should be using txParams.input and not data.

Thanks for these points

packages/web3-eth-contract/src/contract.ts
```diff
@@ -373,7 +374,8 @@ export class Contract<Abi extends ContractAbi>
 			gas: options?.gas ?? options?.gasLimit,
 			gasPrice: options?.gasPrice,
 			from: options?.from,
-			input: options?.input ?? options?.data,
+			input: options?.input,
```
jdevcs, line 393: @luu-alex  so this will use same logic at contract level. like [we have here](https://github.com/web3/web3.js/issues/6183#issuecomment-1654656753)?
as wt ever user will specify ( either `input` or `data`) that will be sent to node.

jdevcs (approved):
LGTM, dnt forget to update change log.

also check https://github.com/web3/web3.js/issues/6379

packages/web3-eth-contract/src/contract.ts
luu-alex: Yes and if theres no specification, it will default to input which is the behavior right now.

avkos (approved):
LGTM. Thanks!

packages/web3-eth-contract/src/utils.ts
```diff
@@ -31,6 +31,32 @@ import { isNullish, mergeDeep, toHex } from 'web3-utils';
...
+			...txParams,
+			data: encodeMethodABI(abi, params, txParams.data as HexString),
+		};
+	} else {
+		// default to using input
+		tx = {
+			...txParams,
+			input: encodeMethodABI(abi, params, txParams.input as HexString),
+		};
+	}
```
Muhammad-Altabba: What inside `else` is exactly what is inside the first `if`. Is this the correct logic? If yes, then it is better to merge them inside one block :smile:

packages/web3-eth-contract/src/utils.ts
```diff
@@ -191,16 +208,27 @@ export const getCreateAccessListParams = ({
...
+			...txParams,
+			data: encodeMethodABI(abi, params, txParams.data as HexString),
+		};
+	} else {
+		// default to using input
+		txParams = {
+			...txParams,
+			input: encodeMethodABI(abi, params, txParams.input),
+		};
 	}
```
Muhammad-Altabba: Sorry, is not this a doublicated logic with `dataInputEncodeMethodHelper`?

packages/web3-eth-contract/src/utils.ts
```diff
@@ -133,14 +145,19 @@ export const getEstimateGasParams = ({
...
-	txParams = {
-		...txParams,
-		input: encodeMethodABI(abi, params, txParams.input ? toHex(txParams.input) : undefined),
-	};
+	if (!isNullish(txParams.input)) {
+		txParams.input = toHex(txParams.input);
+	}
+	if (!isNullish(txParams.data)) {
+		txParams.data = toHex(txParams.data);
+	}
```
Muhammad-Altabba: Sorry, it is not clear to me. Is this 2 `if`s still needed? If yes, isn't it needed in all other functions, and then we need to move it to `dataInputEncodeMethodHelper`? 
And if this is still needed, why exactly? As with the current code, if the user provides both data and input. The `input` will be assigned with toHex(txParams) and the `data` will be assigned with the encode of its value (you may try the code and then modify).

luu-alex: Thanks for the suggestions and taking a look into this Muhammad:
I [updated](https://github.com/web3/web3.js/pull/6377/files#diff-d3f00787c6c9b9f20005a07bdad48461cb7370797a0391442e7025413204d743R40) so that they are seperate if statements for both `data` and `input` and if neither are provided, automatically default to `input`.
As well data and input are always a `hexstring`, so theres no need to use the `toHex` function and has been removed.

luu-alex:
Updated PR to include web3config contractDataInputFill to web3context option to allow users to pass encodedAbi through data, input or for both properties
re review needed

packages/web3-core/src/web3_config.ts
```diff
@@ -37,6 +37,7 @@ export interface Web3ConfigOptions {
 	transactionConfirmationPollingInterval?: number;
 	blockHeaderTimeout: number;
 	maxListenersWarningThreshold: number;
+	contractDataInputFill: 'data' | 'input' | 'both'
```
jdevcs: @luu-alex  This config is specific for contract package only, so imo should be in contract package only instead of in web3_config class.

packages/web3-eth-contract/src/contract.ts
```diff
@@ -233,6 +234,7 @@ export class Contract<Abi extends ContractAbi>
 	private readonly _overloadedMethodAbis: Map<string, AbiFunctionFragment[]>;
 	private _methods!: ContractMethodsInterface<Abi>;
 	private _events!: ContractEventsInterface<Abi>;
+	private readonly _dataInputFill?: 'data' | 'input' | 'both';
```
jdevcs, line 242: also add in docs, thanks

jdevcs: document with info for which RPCs this will be used? sending tx, eth_call, estimate gas all?

packages/web3-core/src/web3_config.ts
luu-alex: Setting it in web3_config class will allow users to only create this option once for all new contracts instantiated. If this was only in the contracts package, users with multiple contracts will need to define this option for every contract instance

docs/docs/guides/smart_contracts/deploying_and_interacting_with_smart_contracts.md
```diff
@@ -353,6 +353,56 @@ Transaction Hash: 0x9825e2a2115896728d0c9c04c2deaf08dfe1f1ff634c4b0e6eeb2f504372
 my number updated value: 2
 ```
 
+## Troubleshooting and errors
+
+If you are running into errors when executing contract methods such as `myContract.methods.call` or `myContract.deploy.estimateGas()` you might be see a contract execution revert error such as: `value transfer did not complete from a contract execution reverted`.
+
+This could be due to the node you are connected to and is expecting the `data` property to be populated in your contract instead of `input`. Web3 version >4.03 will always populate `input` when sending transactions.
```
Muhammad-Altabba: I think the version `4.03` has a typo.
And can we give an example for the provider that has this issue?

packages/web3-eth-contract/src/utils.ts
```diff
@@ -191,16 +214,13 @@ export const getCreateAccessListParams = ({
...
 	) as unknown as TransactionForAccessList;
 
-	if (!txParams.input || abi.type === 'constructor') {
-		txParams = {
-			...txParams,
-			input: encodeMethodABI(abi, params, txParams.input as HexString),
-		};
-	}
+	const dataInput = dataInputEncodeMethodHelper(txParams, abi, params, options?.dataInputFill);
+	txParams = { ...txParams, data: dataInput.data, input: dataInput.input };
```
Muhammad-Altabba, line 223: This means when data or input are not supposed to be filled we will have in json: data: undefined or input: undefined, right? I think this might cause some providers to throw when we provide for example: data: '0x...', input: undefined. Or this would not be the case?

Muhammad-Altabba (approved):
Looks good!
However, there are a few questions that I hope to answer and modify before merging.
Thanks,

packages/web3-eth-contract/src/utils.ts
luu-alex: will always have data or input filled in [dataInputEncodeMethodHelper](https://github.com/web3/web3.js/pull/6377/files/0df6e8b05ecc272448deef313ede102a5c84bca4#diff-d3f00787c6c9b9f20005a07bdad48461cb7370797a0391442e7025413204d743R48) is run (will default to input)

docs/docs/guides/smart_contracts/deploying_and_interacting_with_smart_contracts.md
luu-alex: updated, thank you

moshmage:
hey @luu-alex will we have a tagged version with this fix oooooor ?

luu-alex:
@moshmage hey, sorry didnt get to u sooner. This will be included in the next release which will be a minor version update. In the meantime you can use the dev branch which includes all the latest merges https://www.npmjs.com/package/web3?activeTab=versions latest dev version is [4.1.2-dev.e4ba45c.0](https://www.npmjs.com/package/web3/v/4.1.2-dev.e4ba45c.0) . As well updated [docs](https://docs.web3js.org/guides/smart_contracts/deploying_and_interacting_with_smart_contracts/#troubleshooting-and-errors) for this

luu-alex:
https://github.com/web3/web3.js/releases/tag/v4.1.2 @moshmage

Issue #6211, opened 2023-06-19
[4.x] Rollout own Abi coder to replace ethers abi coder


mpetrunic:
#### Feature
- own performant abi coder with good error handling

Some design choices I've made:
- use uint8array everywhere to avoid 0x prefixes nightmare and to make it more performant
- detailed error handling to enable user to locate error

Tasks:
- [ ] implement all basic encoders
- [ ] implement complex encoders
- [ ] implement all basic decoders
- [ ] implement complex decoders
- [ ] add unit tests using ethers test suite

I've started working on it in a branch `feat/abi-coder` (it was old branch that got rebased so some stuff is not aligned). I plan to work on it in my spare time. If you decide to pick it up, we can divide tasks.

shrpne:
What do you think about using https://github.com/wagmi-dev/abitype instead?

mpetrunic:
> What do you think about using https://github.com/wagmi-dev/abitype instead?

Those are just abi types not abi coder which we already have our own^^

PR #6385, opened 2023-08-28
feat: replace ethers abi coder with ours

### LINKED ISSUE #6270: Rollout own ABI coder: implement complex encoders
**Feature**
- own performant abi coder with good error handling

**Some design choices previously chosen:**
- use uint8array everywhere to avoid 0x prefixes nightmare and to make it more performant
- detailed error handling to enable user to locate error

**Task:**
- [ ] implement complex encoders

Tasks started at`feat/abi-coder` (it was old branch that got rebased so some stuff is not aligned). Marin plans to work on it in his own time, but can split tasks.

mpetrunic:
## Description

Fresh abi code implementation to replace ethers implementation. Should be checked but afaik it adds abilities to encode/decode dynamic array and tuples with dynamic children.
There are probably bugs lurking so I would suggest to add more tests.

P.S. Improved build with concurrently because using '&' means previous scripts are run in the background so there can be concurrency issues with packages that depend on build causing weird type errors from time to time.

resolves #6270 #6211 #6269 

## Type of change

## TODO
- [x] negative number decoding
- [x] figure out what does loose decoding flag means
- [x] array encoding/decoding unit tests
- [x] add coverage where missing
- [x] run against ethers abi encoder tests if possible
- [ ] optional: upgrade project to typescript 5 so we can use latest abitype version
- [x] optional: run in viem benchmark: https://github.com/wagmi-dev/viem/blob/main/src/utils/abi/encodeAbiParameters.bench.ts

## Checklist:


packages/web3-eth-abi/src/coders/base/address.ts
```diff
@@ -0,0 +1,76 @@
...
+	return {
+		dynamic: false,
+		encoded,
+	};
+}
+
+export function decodeAddress(_param: AbiParameter, bytes: Uint8Array): DecoderResult<string> {
+	const start = WORD_SIZE - ADDRESS_BYTES_COUNT;
+	const end = start + ADDRESS_BYTES_COUNT;
+	const addressBytes = bytes.subarray(start, end);
```
jdevcs:
```suggestion
	const addressBytes = bytes.subarray(start, 32);
```
end will always be equal to 32 ( due to 32 bytes encoded length of address )

packages/web3-eth-abi/src/coders/base/address.ts
```diff
@@ -0,0 +1,76 @@
...
+	const encoded = alloc(WORD_SIZE);
+	encoded.set(addressBytes, WORD_SIZE - ADDRESS_BYTES_COUNT);
+	return {
+		dynamic: false,
+		encoded,
+	};
+}
+
+export function decodeAddress(_param: AbiParameter, bytes: Uint8Array): DecoderResult<string> {
+	const start = WORD_SIZE - ADDRESS_BYTES_COUNT;
```
jdevcs: we are sure that offset is 12 for 32 byte encoded address, so should we avoid this arithmetic here and use const?

packages/web3-eth-abi/src/coders/base/address.ts
```diff
@@ -0,0 +1,76 @@
...
+			value: input,
+			name: param.name,
+			type: param.type,
+		});
+	}
+	// for better performance, we could convert hex to destination bytes directly (encoded var)
+	const addressBytes = hexToBytes(address);
+	// expand address to WORD_SIZE
+	const encoded = alloc(WORD_SIZE);
+	encoded.set(addressBytes, WORD_SIZE - ADDRESS_BYTES_COUNT);
```
jdevcs: we are sure that offset is 12 for 32 byte encoded address, so should we avoid this arithmetic ( WORD_SIZE - ADDRESS_BYTES_COUNT) ) here and use const?

packages/web3-eth-abi/src/coders/base/number.ts
```diff
@@ -0,0 +1,176 @@
...
+		throw new AbiError('provided input is greater then maximum for given type', {
+			type: param.type,
+			value: input,
+			name: param.name,
+			maximum: limit.max.toString(),
+		});
+	}
+	return {
+		dynamic: false,
+		encoded: bigIntToUint8Array(value),
```
jdevcs, line 98: It works well but a bit confusing at first look, as its padded uint8array , can rename this function? to encoded or padded suffix? so its not misused else where.

packages/web3-eth-abi/src/coders/base/bool.ts
```diff
@@ -0,0 +1,52 @@
...
+				name: param.name,
+			});
+		}
+	}
+
+	return encodeNumber({ type: 'uint8', name: '' }, Number(value));
+}
+
+export function decodeBool(_param: AbiParameter, bytes: Uint8Array): DecoderResult<boolean> {
+	const boolBytes = bytes.subarray(0, WORD_SIZE);
```
jdevcs: shouldn't encoded bool be strictly 32 bytes, and if not we should throw error? instead of getting sub

packages/web3-eth-abi/src/coders/base/number.ts
mpetrunic: It's not exported, just in this file.

packages/web3-eth-abi/src/coders/base/bool.ts
mpetrunic: When decoding we are going abi item by item, so in this case we are just decoding bool item and returning rest of the array for other decoders

mpetrunic: But to your point, this subarray is unnecessary as we are not even using it. bool is basically uint8 encoded 0 or 1. Will refactor

packages/web3-eth-abi/src/coders/base/bytes.ts
```diff
@@ -0,0 +1,100 @@
...
+		throw new AbiError('provided input is not valid bytes value', {
+			type: param.type,
+			value: input,
+			name: param.name,
+		});
+	}
+	const bytes = bytesToUint8Array(input as Bytes);
+	const [, size] = param.type.split('bytes');
+	// fixed size
+	if (size) {
```
jdevcs, line 43: For fixed bytes data type , should we throw error if size > 32

packages/web3-eth-abi/src/coders/base/bytes.ts
```diff
@@ -0,0 +1,100 @@
...
+	const partsLength = Math.ceil(bytes.length / WORD_SIZE);
+	// one word for length of data + WORD for each part of actual data
+	const encoded = alloc(WORD_SIZE + partsLength * WORD_SIZE);
+
+	encoded.set(encodeNumber({ type: 'uint32', name: '' }, bytes.length).encoded);
+	let offset = WORD_SIZE;
+	for (let i = 0; i < partsLength; i += 1) {
+		encoded.set(bytes.subarray(i * WORD_SIZE, (i + 1) * WORD_SIZE), offset);
+		offset += WORD_SIZE;
+	}
```
jdevcs:
```suggestion
            encoded.set(bytes, WORD_SIZE);
```
as its setting all bytes to right sized `encoded` , only last chunk will be 0 padded at right due to big endian.

packages/web3-eth-abi/src/coders/base/number.ts
```diff
@@ -0,0 +1,176 @@
...
+		throw new AbiError('decoded value is greater then maximum for given type', {
+			type: param.type,
+			value: numberResult,
+			name: param.name,
+			maximum: limit.max.toString(),
+		});
+	}
+	return {
+		result: numberResult,
+		encoded: bytes.subarray(WORD_SIZE),
```
jdevcs, line 131: subarray creates a new view on existing buffer, and changes on new obj will impact original and vice versa. so I think a copy should be made of bytes with required data and sent back from this function.

jdevcs: same for other places of code in this PR

packages/web3-eth-abi/src/api/parameters_api.ts
```diff
@@ -132,21 +98,21 @@ export const encodeParameter = (abi: AbiInput, param: unknown): string =>
...
 	const returnValue: DecodedParams = {
 		__length__: 0,
 	};
 
 	for (const key of Object.keys(data)) {
 		returnValue[key] =
-			Array.isArray(data[key]) && isParamRequiredToConvert(data[key] as Result)
-				? formatArrayResToObject(data[key] as Result)
+			Array.isArray(data[key]) && isParamRequiredToConvert(data[key])
+				? formatArrayResToObject(data[key] as Record<string, any>)
```
jdevcs:
```suggestion
				? formatArrayResToObject(data[key] as Record<string, unknown>)
```

packages/web3-eth-abi/src/api/parameters_api.ts
```diff
@@ -132,21 +98,21 @@ export const encodeParameter = (abi: AbiInput, param: unknown): string =>
...
 // If encoded param is an array and there are mixed on integer and string keys
-const isParamRequiredToConvert = (data: Result): boolean =>
+const isParamRequiredToConvert = (data: any): boolean =>
 	Array.isArray(data) &&
 	Object.keys(data).filter(k => Number.isInteger(+k)).length !== Object.keys(data).length;
 
 // Ethers-Encoder return the decoded result as an array with additional string indexes for named params
 // We want these to be converted to an object with named keys
-const formatArrayResToObject = (data: Result): DecodedParams => {
+const formatArrayResToObject = (data: Record<string, any>): DecodedParams => {
```
jdevcs:
```suggestion
const formatArrayResToObject = (data: Record<string, unknown>): DecodedParams => {
```

packages/web3-eth-abi/src/api/parameters_api.ts
```diff
@@ -132,21 +98,21 @@ export const encodeParameter = (abi: AbiInput, param: unknown): string =>
 	encodeParameters([abi], [param]);
 
 // If encoded param is an array and there are mixed on integer and string keys
-const isParamRequiredToConvert = (data: Result): boolean =>
+const isParamRequiredToConvert = (data: any): boolean =>
```
jdevcs:
```suggestion
const isParamRequiredToConvert = (data: unknown): boolean =>
```

packages/web3-eth-abi/src/coders/base/number.ts
mpetrunic: That was the goal as it will reduce memory and performance impact. And there is no fear here since decoders never modify any bytes. Better option would be to find a way to freeze internal array buffer but I do not know a way to do it.

jdevcs: ok, Its tradeoff between memory+performance vs safety( so data is not corrupted by maintainers misuse of it). 

I'll also lean towards UX as compared to maintainers. btw I like this approach of using uint8arry / buffers instead of padding using high level functions, its great.

packages/web3-eth-abi/src/api/parameters_api.ts
```diff
@@ -37,45 +35,13 @@ import { formatParam, isAbiFragment, mapTypes, modifyParams } from '../utils.js'
...
-	}
-};
+export const encodeParameters = (abi: ReadonlyArray<AbiInput>, params: unknown[]): string =>
+	encodeParametersInternal(abi, params);
 
 /**
  * Encodes a parameter based on its type to its ABI representation.
  * @param abi -  The type of the parameter. See the [Solidity documentation](https://docs.soliditylang.org/en/develop/types.html) for a list of types.
- * @param param - The actual parameter to encode.
+ * @param param - The actual pazrameter to encode.
```
avkos: here is a typo, I think

avkos (changes requested):
The size was decreased a lot - that's amazing!
But coverage was also decreased.
I think having as many as possible tests of new functionality will be great. Do you consider an option to copy test cases from the open-source project?

mpetrunic:
> But coverage was also decreased. I think having as many as possible tests of new functionality will be great.

Diff coverage is actually >90%. I assume total coverage decreased as we used a library before so that code had "100%" coverage^^

> Do you consider an option to copy test cases from the open-source project?

I mentioned that in PR description as todo. That said, I noticed ethers test cases uses human readable abi and I do think we will hit a lot of cases where abitype is not really parsing them well (afaik they are not standardized).

We could use https://github.com/wagmi-dev/viem/blob/main/src/_test/abis.ts instead^^

packages/web3-eth-abi/src/api/parameters_api.ts
mpetrunic:
```suggestion
 * @param param - The actual parameter to encode.
```

packages/web3-eth-abi/src/api/parameters_api.ts
```diff
@@ -172,14 +114,11 @@ export const decodeParametersWith = (
...
-			`0x${bytes.replace(/0x/i, '')}`,
-			loose,
-		);
-		return formatArrayResToObject(res);
+		return decodeParametersInternal(abis, `0x${bytes.replace(/0x/i, '')}`, loose);
 	} catch (err) {
-		throw new AbiError(`Parameter decoding error: ${(err as Error).message}`);
+		throw new AbiError(`Parameter decoding error: ${(err as Error).message}`, {
+			internalErr: err,
+		});
```
Muhammad-Altabba, line 121: Could we utilize here `innerError` at the base class `BaseWeb3Error`?
```suggestion
		throw new AbiError(`Parameter decoding error: ${(err as Error).message}`, {}, err);
```

packages/web3-errors/src/errors/generic_errors.ts
```diff
@@ -83,6 +83,12 @@ export class OperationAbortError extends BaseWeb3Error {
 
 export class AbiError extends BaseWeb3Error {
 	public code = ERR_ABI_ENCODING;
+	public readonly props: Record<string, unknown> & { name?: string };
+
+	public constructor(message: string, props?: Record<string, unknown> & { name?: string }) {
+		super(message);
```
Muhammad-Altabba, line 89: The base class `BaseWeb3Error` has `innerError` and it accepts it as an optional second parameter in the constructor.

```suggestion
	public constructor(message: string, props?: Record<string, unknown> & { name?: string }, innerError?: Error | Error[]) {
		super(message, innerError);
```

packages/web3-eth-abi/src/coders/base/array.ts
```diff
@@ -0,0 +1,120 @@
...
+export function encodeArray(param: AbiParameter, values: unknown): EncoderResult {
+	if (!Array.isArray(values)) {
+		throw new AbiError('Expected value to be array', { abi: param, values });
+	}
+	const { size, param: arrayItemParam } = extractArrayType(param);
+	const encodedParams = values.map(v => encodeParamFromAbiParameter(arrayItemParam, v));
+	const dynamic = size === -1;
+	const dynamicItems = encodedParams.length > 0 && encodedParams[0].dynamic;
+	if (!dynamic && values.length !== size) {
+		throw new AbiError("Given arguments count doesn't match array length", {
```
Muhammad-Altabba, line 36: Errors here and everywhere else in the MR seems to inturduce breaking changes, right?
Are we planning on upgreading the major version after this MR? Or else we need to keep the errors as they where, right?

packages/web3-eth-abi/src/coders/base/bool.ts
```diff
@@ -0,0 +1,54 @@
...
+		}
+	}
+
+	return encodeNumber({ type: 'uint8', name: '' }, Number(value));
+}
+
+export function decodeBool(_param: AbiParameter, bytes: Uint8Array): DecoderResult<boolean> {
+	const numberResult = decodeNumber({ type: 'uint8', name: '' }, bytes);
+	if (numberResult.result > 1 || numberResult.result < 0) {
+		throw new AbiError('Invalid boolean value encoded', {
```
Muhammad-Altabba, line 44: I noticed that we sometimes start the error message with Capital and sometime with small. I think it would be good to establish a convention to start the error message with small letter (unless beginning with an exported name, a proper noun or an acronym).

Muhammad-Altabba (commented):
Thanks for your great contributions!

However, I have a special concern regarding the breaking change that would be introduced with the new errors ([mentioned here](https://github.com/web3/web3.js/pull/6385#discussion_r1328919565)). Or, what do you think?

packages/web3-eth-abi/src/coders/base/array.ts
mpetrunic: What do you mean? its the same error instance as before. Adding new error messages is not a breaking chnage

packages/web3-errors/src/errors/generic_errors.ts
mpetrunic: Yes

packages/web3-eth-abi/src/coders/base/array.ts
Muhammad-Altabba: I think this is not just a new error message. The code used to throw an error for this case, and for others, but with a different message and a different structure, right?
The old message was according to the old deleted code:
```throw new AbiError(`Parameter encoding error`, err as Error);``` And the innerError was the error object we get from the encoder.

Comment (mpetrunic) [reply packages/web3-eth-abi/src/coders/base/array.ts:36]: Code still throws AbiError error and our api doesn't guarantee inner error type will stay the same (It will be of type Error | Error[] if present). It's user's fault if he expected something that typescript types did not guarantee.

Comment (Muhammad-Altabba) [reply packages/web3-eth-abi/src/coders/base/array.ts:36]: Thanks @mpetrunic,
Yes, it is not guaranteed by typescript types here. But still the user can even use JavaScript which has no types. I am thinking here about the actual strings and values that the user used to get and might already has `if`s checks on. And if we think it is the user's fault to depend on types we did not specify, changing them is still considered a breaking change, as far as I know. 

So, as far as I know, if we do not maintain to provide the exact message string and the exact internal properties for every error, as we used to, then we are introducing a breaking change. And yes we can add more properties to the error object. But we are not supposed to remove or update the string or a property value, without pumping the major version.  

And yes, our API has been used to guarantee the same message and properties's values for the innerError message because the early used library is following semver.

And I think keeping the error messages and properties, as they were, is possible even though it might be a bit tricky. 

Please correct me if changing an error message, or a value of an internal property inside an error object, is not considered a breaking change.

Thanks,

Comment (mpetrunic) [reply packages/web3-eth-abi/src/coders/base/array.ts:36]: Afaik, error messages are implementation details intended for the developer's eyes. No one should rely on their exact content or even worse make an if/else statement based on their content (fixing a typo in an error message would then be a breaking change). If the developer intended for you to handle those errors, errors would either be different instances or contain error codes along with the list of error codes the function might throw.

File: packages/web3-eth-abi/test/unit/encodeDecodeParams.test.ts [hunk 17]
```diff
@@ -0,0 +1,35 @@
...
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { AbiInput } from 'web3-types';
+import { decodeParameters, encodeParameters } from '../../src';
+import testsData from '../fixtures/abitestsdata.json';
+
+describe('encodeParameters decodeParameters tests should pass', () => {
+	it.each(testsData)(`unit test of encodeParameters`, async encoderTestObj => {
+		const encodedResult = encodeParameters([encoderTestObj.type], [encoderTestObj.value]);
```
jdevcs, line 25: 843 tests are passing and 181 failing (total tests 1024)

packages/web3-eth-abi/test/unit/encodeDecodeParams.test.ts
```diff
@@ -0,0 +1,35 @@
...
+		expect(encodedResult).toEqual(encoderTestObj.encoded);
+	});
+
+	it.skip.each(testsData)('unit test of decodeParameters', async decoderTestObj => {
+		const decodedResult = decodeParameters(
```
jdevcs, line 30: These tests are skipped as we cannot use data in tests cases for decoded objects , reason: we only put attribute key numbers while tests data has attribute keys having types.

packages/web3-eth-abi/test/unit/encodeDecodeParams.test.ts
mpetrunic: all are passing now :)

packages/web3-eth-abi/test/unit/encodeDecodeParams.test.ts
jdevcs: I updated data comparison for decodeParameters(), and all 1024 decoder tests are passing now.

jdevcs (approved):
As discussed, if loosedecoding is enabled it skips checking strict padding rules, due to some old solidity versions bug for event data emitted from external functions.
we are not using loosedecoding feature by passing false, so we can skip this feature in this PR.

Muhammad-Altabba (approved):

avkos (approved):

PR #6398, opened 2023-08-31
Implement `EventEmitter` compatible with browsers

### LINKED ISSUE #6371: Uncaught TypeError: Class extends value undefined is not a constructor or null
#### Expected behavior
Web3 should be run without error in vuejs.

#### Actual behavior
it displays 
image

1. Fresh installed Laravel
2. Install Jetstream
3. Install Inertiajs

#### Environment
```
"@inertiajs/vue3": "^1.0.0",
"@tailwindcss/forms": "^0.5.2",
"@tailwindcss/typography": "^0.5.2",
"@vitejs/plugin-vue": "^4.0.0",
"autoprefixer": "^10.4.7",
"axios": "^1.1.2",
"laravel-vite-plugin": "^0.8.0",
"postcss": "^8.4.14",
"tailwindcss": "^3.1.0",
"vite": "^4.0.0",
"vue": "^3.2.31"
"web3": "^4.1.0"
```

Muhammad-Altabba:
## Description

Fixes #6371

## Type of change

## Checklist:


packages/web3-utils/src/event_emitter.ts
```diff
@@ -0,0 +1,104 @@
...
+
+You should have received a copy of the GNU Lesser General Public License
+along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { EventEmitter as NodeEventEmitter } from 'events';
+
+type Callback = (params: any) => void | Promise<void>;
+
+export class InBrowserEventEmitter extends EventTarget {
```
mpetrunic: Where is this implementation comming from?

Muhammad-Altabba: Hi @mpetrunic,
This implementation comes from Stackoverflow with the help of ChatGPT :smile: .
Because `events` module is only a `node` module that does not exist for the browsers. There is a need to use something different for the browsers. And this class simply uses the class `EventTarget` available at the browsers and adds more methods to it which we need in our code.
The other alternative to this solution is to use a 3rd party library like: https://www.npmjs.com/package/eventemitter, https://www.npmjs.com/package/events and https://www.npmjs.com/package/eventemitter3. However, I did not want to add an additional dependency.
However, if you have a better approach, kindly advise.
Thanks,

mpetrunic: Event emmitters are tricky beasts, easy to cause memory leaks or annoying bugs. I would rely on some solid implementation like https://www.npmjs.com/package/eventemitter3

jdevcs: This lib is ~70kb unpack size, for now its good to use some dependable lib but also create issue as lib optimisation in future we may attempt reducing lib size then.

Muhammad-Altabba: Thanks @mpetrunic and @jdevcs, for your valid concern and good suggestions.
Because these MR changes still rely on node `EventEmitter`, when it is a node environment, there are no changes for node js users. And for the browser, it depends on the available class `EventTarget` by just wrapping it to use it similar to node `EventEmitter`. So, I see this approach as stable. And, actually, our end-to-end tests pass. But I still need to write unit tests.
However, as you suggested, I can use [`eventemitter3`](https://www.npmjs.com/package/eventemitter3) for now and create a task to write our own Event Emitter later. This is in case you still see the proposed solution in the MR is not good enough even after writing some unit tests. What do you think?

[There are failed tests for Firefox. But they are not related to the changes in this MR as they exist in other MRs as well]

packages/web3-utils/src/event_emitter.ts
```diff
@@ -0,0 +1,117 @@
...
+
+	public on(eventName: string, fn: Callback) {
+		this.addEventListener(eventName, fn);
+		return this;
+	}
+
+	public once(eventName: string, fn: Callback) {
+		const onceCallback = async (params: Callback) => {
+			await fn(params);
+			this.off(eventName, onceCallback);
```
jdevcs, line 46: This PR LGTM, only
`codecov/patch` is not passing, coverage should be increased before merging.

packages/web3-utils/src/event_emitter.ts
```diff
@@ -0,0 +1,117 @@
...
+		super.addEventListener(eventName, wrappedFn as EventListener);
+		if (!this._listeners[eventName]) {
+			this._listeners[eventName] = [];
+		}
+		this._listeners[eventName].push([fn, wrappedFn]);
+	}
+
+	public removeEventListener(eventName: string, fn: Callback) {
+		const eventListeners = this._listeners[eventName];
+		if (eventListeners) {
```
jdevcs, line 108: we should have node+browser(integration) tests for events testing,

avkos:
Thank you for the PR. LGTM, I agree with @jdevcs that we need to have browser + node tests for the new implementation

packages/web3-utils/src/event_emitter.ts
Muhammad-Altabba: done

packages/web3-utils/src/event_emitter.ts
Muhammad-Altabba: done using `@jest-environment jsdom`

Muhammad-Altabba:
There are now tests for `node` as well as for `jsdom` (browser environment).
Thanks,

avkos (approved):

jdevcs (approved):

jdevcs (commented):
@Muhammad-Altabba update changelog for all packages where required

PR #6407, opened 2023-09-06
create a new type `Eip1193Compatible`

### LINKED ISSUE #6345: The returned value of `.request(...)` method at the providers does not comply with EIP-1193
#### Expected behavior
The `.request(...)` method should return the `result` of the response, not the full response.

> If resolved, the Promise MUST resolve with a result per the RPC method's specification. The Promise MUST NOT resolve with any RPC protocol-specific response objects, unless the RPC method's return type is so defined.
Ref: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md

And this is how it is implemented for example by MetaMask as in https://docs.metamask.io/guide/ethereum-provider.html#methods:
```ts
const chainId = await ethereum.request({ method: 'eth_chainId' });
```
#### Actual behavior
The returned value is a response object like:
```js
{
    jsonrpc: '2.0',
    result: []
}
```

And this caused an issue for example at Ganache repo when using the new provider from version 4.x with the pice of code: https://github.com/trufflesuite/truffle/blob/develop/packages/encoder/lib/adapter.ts#L150

```js

const { Web3 } = require('web3');

// Set up a connection to the Ethereum network
const url = 'http://127.0.0.1:7545/';
const provider = new Web3.providers.HttpProvider(url, { keepAlive: false });

async function run() {
    try {
        const response = await provider.request({
            method: 'eth_getLogs',
            params: [ { fromBlock: 5, toBlock: NaN, address: undefined } ]
          });

        // This currently logs: { jsonrpc: '2.0', result: [] }
        //  But it should log: []
        console.log('the response should be the result:', response)

    } catch (error) {
        console.error(error);
    }
}

run();
```

#### Environment

web3.js v4.0.3

### Hint
If you face this issue, you can work around it by simply accessing `response.result`.

Muhammad-Altabba:
## Description

Fixes #6345 by allowing the user to get an instance of any Web3 Provider that has the returned value of `request` fully compatible with EIP-1193.

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-types/src/web3_base_provider.ts
+++ b/packages/web3-types/src/web3_base_provider.ts
@@ -133,6 +133,19 @@ export interface EIP1193Provider<API extends Web3APISpec> extends SimpleProvider
 	removeListener(event: 'accountsChanged', listener: (accounts: ProviderAccounts) => void): void;
 }
 
+export type Eip1193FullyCompatible<API extends Web3APISpec = EthExecutionAPI> = Omit<
+	// eslint-disable-next-line no-use-before-define
+	Omit<Web3BaseProvider, 'request'>,
+	'asEip1193FullyCompatible'
+> & {
+	request<
+		Method extends Web3APIMethod<API>,
+		ResultType = Web3APIReturnType<API, Method> | unknown,
+	>(
+		request: Web3APIPayload<API, Method>,
+	): Promise<ResultType>;
+};
+
 // Provider interface compatible with EIP-1193
 // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md
 export abstract class Web3BaseProvider<API extends Web3APISpec = EthExecutionAPI>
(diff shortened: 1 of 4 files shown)


packages/web3-types/src/web3_base_provider.ts
```diff
@@ -189,6 +202,40 @@ export abstract class Web3BaseProvider<API extends Web3APISpec = EthExecutionAPI
...
+	 *
+	 * @example
+	 * ```ts
+	 * const provider = new Web3HttpProvider('http://localhost:8545');
+	 * const fullyCompatibleProvider = provider.asEip1193FullyCompatible();
+	 * const result = await fullyCompatibleProvider.request({ method: 'eth_getBalance' });
+	 * console.log(result); // '0x0234c8a3397aab58' or something like that
+	 * ```
+	 */
+	public asEip1193FullyCompatible(): Eip1193FullyCompatible<API> {
```
jdevcs:
```suggestion
	public asEIP1193Provider(): asEIP1193Provider<API> {
```
There is no fully or partially Compatible, but only compatible or not. as its confusing that in case of not fully compatible what features are missing. just a minor name change, for making it simple.

packages/web3-types/src/web3_base_provider.ts
```diff
@@ -133,6 +133,19 @@ export interface EIP1193Provider<API extends Web3APISpec> extends SimpleProvider
 	removeListener(event: 'accountsChanged', listener: (accounts: ProviderAccounts) => void): void;
 }
 
+export type Eip1193FullyCompatible<API extends Web3APISpec = EthExecutionAPI> = Omit<
```
jdevcs: same here

packages/web3-types/src/web3_base_provider.ts
Muhammad-Altabba: done

packages/web3-types/src/web3_base_provider.ts
Muhammad-Altabba: done

luu-alex (approved):

jdevcs (approved):

avkos (approved):

PR #6410, opened 2023-09-08
Add events to receipt object

### LINKED ISSUE #6312: The returned logs of a contract event are not of an event object
Investigate why the test `"should return transaction hash, logs and receipt when using synchronised transactions"` at  https://github.com/nazarhussain/truffle/blob/656263aadf21612b17829483d7376eb7afbe63aa/packages/artifactor/test/contracts.js#L195 is failing with 4.x.

un-skip the test to investigate the error:
   "AssertionError: expected 'ExampleEvent' to equal undefined"
The issue seems to be that the returned logs are not of an event.
   (The returned log does not include the properties: `event` and `args`)
```
  logs: [
    {
      address: '0x608fcae953f6ecbb979bd670260fea9052a7e527',
      blockHash: '0x3a5f39352e0fd37c6a0b7870b16d585a10dde1fec30b0e1b64b6065caef5469a',
      blockNumber: 2n,
      logIndex: 0n,
      removed: false,
      transactionHash: '0xd9ad9f8a3d74c8bad000249977ec60d8313b2e4bbc389de06709a45cd5e0cb31',
      transactionIndex: 0n
    }
  ]
```

avkos:
## Description

Fixes #6312 

## Type of change

## Checklist:


packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -479,6 +485,256 @@ export async function getTransactionCount<ReturnFormat extends DataFormat>(
 
 	return format({ format: 'uint' }, response as Numbers, returnFormat);
 }
+class SendTxHelper<
```
jdevcs: as its another class so place it in eth [utils dir ](https://github.com/web3/web3.js/tree/4.x/packages/web3-eth/src/utils)

packages/web3-eth-contract/src/constants.ts
```diff
@@ -1,26 +0,0 @@
...
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-GNU Lesser General Public License for more details.
-
-You should have received a copy of the GNU Lesser General Public License
-along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
-*/
-
-import { AbiEventFragment } from 'web3-types';
-
-export const ALL_EVENTS = 'ALLEVENTS';
```
jdevcs, line 20: this will break contract package, so reexport it from eth.

packages/web3-eth-contract/src/types.ts
```diff
@@ -75,76 +51,6 @@ export interface ContractEventOptions {
 	topics?: string[];
 }
 
-export interface ContractOptions {
```
jdevcs, line 78: for all types moved from contracts to other packages, reexport these from contracts so its not breaking.

packages/web3-eth/src/rpc_method_wrappers.ts
avkos: it is used only in `web3-eth` package. I will move it to another file in `web3-eth`

luu-alex:
Looks like codecov is failing

jdevcs (commented):
pls there should be e2e tests for tx having receipt.events present for contract.send, and also not present ( for simple tx ) with geth.

luu-alex:
other than junaids comment, lgtm 👍

luu-alex (approved):

avkos:
> pls there should be e2e tests for tx having receipt.events present for contract.send, and also not present ( for simple tx ) with geth.

done

packages/web3-types/src/eth_types.ts
```diff
@@ -312,6 +312,23 @@ export interface LogBase<NumberType, ByteType> {
 export interface Log extends LogBase<Numbers, Bytes> {
 	readonly id?: string;
 }
+
+export interface EventLog {
```
jdevcs, line 316: pls add changelog in this package

packages/web3-eth-contract/src/utils.ts
```diff
@@ -26,10 +26,11 @@ import {
 	NonPayableCallOptions,
 	PayableCallOptions,
 	ContractInitOptions,
+	ContractOptions,
 } from 'web3-types';
 import { isNullish, mergeDeep } from 'web3-utils';
 import { encodeMethodABI } from './encoding.js';
-import { ContractOptions, Web3ContractContext } from './types.js';
+import { Web3ContractContext } from './types.js';
```
jdevcs, line 33: also add changelog in contracts that receipt.event will be avalible

jdevcs (approved):

Muhammad-Altabba (changes requested):
Thanks @avkos , this seems good.
However, still `args` are not filled. According to the raised issue: https://github.com/web3/web3.js/issues/6312
> (The returned log does not include the properties: event and args)

I am not sure about the exact properties of those args. It needs to be checked against 1.x. But they seem to be used to access at least args.eventID, args._from, args.num (See here https://github.com/nazarhussain/truffle/blob/656263aadf21612b17829483d7376eb7afbe63aa/packages/artifactor/test/contracts.js#L242 and https://github.com/nazarhussain/truffle/blob/656263aadf21612b17829483d7376eb7afbe63aa/packages/deployer/test/await.js#L134-L135)
Or what do you think?

avkos:
> Thanks @avkos , this seems good. However, still `args` are not filled. According to the raised issue: #6312
> 
> > (The returned log does not include the properties: event and args)
> 
> I am not sure about the exact properties of those args. It needs to be checked against 1.x. But they seem to be used to access at least args.eventID, args._from, args.num (See here https://github.com/nazarhussain/truffle/blob/656263aadf21612b17829483d7376eb7afbe63aa/packages/artifactor/test/contracts.js#L242 and https://github.com/nazarhussain/truffle/blob/656263aadf21612b17829483d7376eb7afbe63aa/packages/deployer/test/await.js#L134-L135) Or what do you think?

here is a web3js@1.x response
image

it doesn't have any args field. Looks like. truffle added it inside the library

avkos:
just for your information here is web3js@4.x response
image

we have logs and events fields

Muhammad-Altabba (approved):

Issue #6461, opened 2023-09-26
web3-eth-Contract validation test fails


luu-alex:
test fails inconsistently from validation, i believe its been affected by this [merge](https://github.com/web3/web3.js/pull/6435/files)
https://github.com/web3/web3.js/actions/runs/6268996538/job/17024937030
```
web3-eth-contract:  FAIL  test/integration/local_account/contract_erc721.test.ts (10.058 s)
web3-eth-contract:   ● contract › erc721 › should award item "0x2"
web3-eth-contract:     Web3ValidatorError: Web3 validator found 1 error[s]:
web3-eth-contract:     value at "/0" is required
web3-eth-contract: 
web3-eth-contract:       108 | 					return errors;
web3-eth-contract:       109 | 				}
web3-eth-contract:     > 110 | 				throw new Web3ValidatorError(errors);
web3-eth-contract:           | 				      ^
web3-eth-contract:       111 | 			}
web3-eth-contract:       112 | 		}
web3-eth-contract:       113 | 		return undefined;
web3-eth-contract: 
web3-eth-contract:       at Validator.validate (../web3-validator/src/validator.ts:110:11)
web3-eth-contract:       at Web3Validator.validate (../web3-validator/src/web3_validator.ts:66:26)
web3-eth-contract:       at Object.<anonymous> (src/contract.ts:927:15)
web3-eth-contract:       at test/integration/local_account/contract_erc721.test.ts:88:38
web3-eth-contract:       at fulfilled (test/integration/local_account/contract_erc721.test.ts:21:58)
```

#### Steps to reproduce the behavior

1. [First step]
2. [Second step]
3. [and so on...]

PR #6434, opened 2023-09-17
Fix validation uint int sizes

fullkomnun:
## Description

Previously an ABI that contained a numeric type (int/uint) of size that is not a power of 2 (such as uint40, int24) would result in the validator / check function being set on Zod as undefined and lead to a cryptic TypeError during validation.
I have setup validators for all valid numeric type sizes ([u]int8, [u]int 16, ... , [u]int256) and throw an error right away if during conversion of JSON scheme to Zod an unsupported type is encountered. 
Generated 'single parameter' ABI test cases for all valid sizes.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules.
- [x] I ran `npm run lint` with success and extended the tests and types if necessary.
- [x] I ran `npm run test:unit` with success.
- [x] I ran `npm run test:coverage` and my test cases cover all the lines and branches of the added code.
- [ ] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
- [ ] I have tested my code on the live network.
- [ ] I have checked the Deploy Preview and it looks correct.
- [ ] I have updated the `CHANGELOG.md` file in the root folder.
- [ ] I have linked Issue(s) with this PR in "Linked Issues" menu.

PR #6474, opened 2023-10-04
web3-eth-contract validation test fail

luu-alex:
## Description
investigated #6461 
```
const logs = await contractDeployed.getPastEvents('Transfer');
			const tokenId = (logs[0] as EventLog)?.returnValues?.tokenId as string;
```
in this line, logs can sometimes be undefined, creating this error.

Fix is to just use receipt and instead of `getPastEvents` 

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-eth-contract/test/integration/local_account/contract_erc721.test.ts
+++ b/packages/web3-eth-contract/test/integration/local_account/contract_erc721.test.ts
@@ -19,7 +19,6 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 import Web3 from 'web3';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { Web3Account } from 'web3-eth-accounts';
-import { EventLog } from 'web3-types';
 import { Contract } from '../../../src';
 import { ERC721TokenAbi, ERC721TokenBytecode } from '../../shared_fixtures/build/ERC721Token';
 import { getSystemTestProvider, createLocalAccount } from '../../fixtures/system_test_utils';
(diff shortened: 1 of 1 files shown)


jdevcs (approved):
> investigated https://github.com/web3/web3.js/issues/6461
> 
> const logs = await contractDeployed.getPastEvents('Transfer');
> 			const tokenId = (logs[0] as EventLog)?.returnValues?.tokenId as string;
> in this line, logs can sometimes be undefined, creating this error.
> 
> Fix is to just use receipt and instead of getPastEvents

Why logs were undefined? if tx is mined there should be logs available via getPastEvents, was there issue in tx waiting long in tests and not mined?

Muhammad-Altabba:
> > investigated #6461
> > const logs = await contractDeployed.getPastEvents('Transfer');
> > const tokenId = (logs[0] as EventLog)?.returnValues?.tokenId as string;
> > in this line, logs can sometimes be undefined, creating this error.
> > Fix is to just use receipt and instead of getPastEvents
> 
> Why logs were undefined? if tx is mined there should be logs available via getPastEvents, was there issue in tx waiting long in tests and not mined?

I also think that we need to know why the logs were undefined sometime. 

And I think either the proposed code will still fail in the same occasions where the old code fails. Or, if this would not be the case, then there is another issue that the users will face when using `const logs = await contractDeployed.getPastEvents('Transfer');` outside the erc721 context.

Could you please @luu-alex investigate more on why and when the `await contractDeployed.getPastEvents('Transfer')` is returning `undefined`. And if you like, it could be helpful also to investigate this case with a code not related to erc721. Because, this is a generic functionality that could be used anywhere.

Many thanks,

luu-alex:
sounds good @Muhammad-Altabba , this PR and i think this [one as well](https://github.com/web3/web3.js/pull/6434) was holding up release. I created an [issue](https://github.com/web3/web3.js/issues/6475) and will begin investigating. Thanks for the feedback

luu-alex:
@jdevcs im not too sure, i will need to [investigate](https://github.com/web3/web3.js/issues/6475) getPastEvents next

avkos (approved):

Muhammad-Altabba (approved):

luu-alex:
Needs more investigation, it seems to be an issue with getpastevents returning undefined due to a change from validation

PR #6486, opened 2023-10-06
Fix an issue with detecting Uint8Array

Muhammad-Altabba:
## Description
Fix error: `value "..." at "/0" must pass "bytes" validation`.

[UPDATE: sorry it tagged an unrelated issue by mistake. Now I correct it]
Suppose to close: #6393

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I have selected the correct base branch.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules.
- [ ] I ran `npm run lint` with success and extended the tests and types if necessary.
- [ ] I ran `npm run test:unit` with success.
- [ ] I ran `npm run test:coverage` and my test cases cover all the lines and branches of the added code.
- [ ] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
- [ ] I have tested my code on the live network.
- [ ] I have checked the Deploy Preview and it looks correct.
- [ ] I have updated the `CHANGELOG.md` file in the root folder.
- [ ] I have linked Issue(s) with this PR in "Linked Issues" menu.

PR #6487, opened 2023-10-06
Throw `InsecureContextError` if `window.crypto.subtle` is `undefined`

### LINKED ISSUE #6485: TypeError: Cannot read properties of undefined (reading 'importKey')
#### Description

Unclear error is thrown when calling `encrypt` from `web3-eth-accounts` in an in-secured environment. This could happen when using the library with a remote website with `http` (and not `https`).
However, I checked this using Jest with `jsdom`; And in this case it is also an in-secured environment. So, this issue was detected using `jsdom` but it applies to other scenarios (remote http websites). 

It also worth noting that it could be implemented for jsdom later, and there are polyfiles for it, see this issue for more: https://github.com/jsdom/jsdom/issues/1612 

#### Expected behavior

To throw a meaningful error in case some functionality is not available.

#### Actual behavior

Throws: `TypeError: Cannot read properties of undefined (reading 'importKey')`

1. At `test/unit/account.test.ts` inside the package `web3-eth-accounts`:

  - Add this to the very first line of the file (before the existing comment):

```ts
/**
 * @jest-environment jsdom
 */
```

  - Add polyfills for `TextEncoder` before the imports:

```ts
import { TextEncoder } from 'util';
global.TextEncoder = TextEncoder;
```

  - Search for `describe('encrypt', () => {` and replace with: `describe.only('encrypt', () => {`

2. `cd  packages/web3-eth-accounts && yarn test account.test.ts` 

#### Logs
```
FAIL  test/unit/account.test.ts (7.943 s)
 ● accounts › encrypt › valid cases › [
 '0x67f476289210e3bef3c1c75e4de993ff0a00663df00def84e73aa7411eac18a6',
  '123',
  [Object]
]

    TypeError: Cannot read properties of undefined (reading 'importKey')

      552 |     }
      553 |
    > 554 |     const cipher = await createCipheriv(
          |                          ^
      555 |             privateKeyUint8Array,
      556 |             derivedKey.slice(0, 16),
      557 |             initializationVector,

      at getBrowserKey (node_modules/ethereum-cryptography/aes.js:33:42)
      at encrypt (node_modules/ethereum-cryptography/aes.js:41:36)
      at src/account.ts:554:23
      at src/account.ts:5226:39
      at Object.<anonymous>.__awaiter (src/account.ts:5169:10)
      at encrypt (src/account.ts:487:24)
      at test/unit/account.test.ts:169:33
      at test/unit/account.test.ts:8:71
      at Object.<anonymous>.__awaiter (test/unit/account.test.ts:4:12)
      at test/unit/account.test.ts:168:60
```

Muhammad-Altabba:
## Description

Fixes #6485

## Type of change

## Checklist:


Muhammad-Altabba:
To resume working on this after merging: https://github.com/web3/web3.js/pull/6398/ and https://github.com/web3/web3.js/pull/6486

jdevcs (commented):
I am not in favor of using jsdom for browser testing, as it already need many changes in our lib. jest + jsdom is direction that we will not go. 

So we should use cypress for browser testing for now.

jdevcs:
so if using a testing env is causing changes in code to be tested, thn it should be avoided. 

and in case of requirement of SSL, I think thats ok.

Muhammad-Altabba:
> I am not in favor of using jsdom for browser testing, as it already need many changes in our lib. jest + jsdom is direction that we will not go.
> 
> So we should use cypress for browser testing for now.

Sounds good. However, the case that this PR is handling does not show in the browser for local addresses. So, it would not happen with our Cypress configuration. It happens only with remote Urls that is also not secure (http and not with https). And this case is simulated in jsdom currently. 
However, We can also not use jsdom and just set mocking the window and set window.crypto.subtle to undefined, in order to simulate the case.

PR #6492, opened 2023-10-10
getTransaction.gasPrice return type fix

neerosity:
## Description

When we import web3 with typescript, it doesn't provide gasPrice as return param for a getTransaction function. 

to reproduce issue 

```typescript
import { Web3 } from 'web3';

getWeb3Instance(blockchainId: string, wssURL: string): Web3 {
    if (this.blockchainWeb3Instances[blockchainId]) {
      return this.blockchainWeb3Instances[blockchainId];
    }

    const provider = new Web3.providers.WebsocketProvider(wssURL);
    const web3 = new Web3(provider);

    this.blockchainWeb3Instances[blockchainId] = web3;
    return web3;
  }
  
  async getTransactionInformation(
    transactionHash: string,
    wssURL: string,
    blockchainId: string,
  ) {
    const web3 = await this.getWeb3Instance(blockchainId, wssURL);

    const receipt = await web3.eth.getTransactionReceipt(transactionHash);

    if (!receipt)
      throw new Error();

    const transactionDetails = await web3.eth.getTransaction(
      receipt?.transactionHash,
    );
    // here it will throw error that gasPrice is not a property of transaction
    const gasPrice = BigInt(transactionDetails.gasPrice);
    // alternative way to solve this
    const gasPrice = BigInt((transactionDetails as any).gasPrice);
```

## Type of change

## Checklist:

- [x] I ran `npm run test:coverage` and my test cases cover all the lines and branches of the added code.


avkos:
Thank you for your contribution. I've reproduced the issue locally but the fix doesn't solve it.

packages/web3-types/src/eth_types.ts
```diff
@@ -83,6 +83,7 @@ export type TransactionOutput = {
 	readonly from?: HexString;
 	readonly input: string;
 	readonly gas?: Numbers;
+	readonly gasPrice?: Numbers;
```
jdevcs, line 86: This change will not fix issue, this type is used in formatter functions and those formatters are only for backward compatibility.

PR #6506, opened 2023-10-12
Avoid using `**` with `BigInt`

### LINKED ISSUE #6187: TypeError: Cannot convert a BigInt value to a number
#### Expected behavior
view screen on deployment

#### Actual behavior

failing to view screen on deployment because of the error, This is the error ```TypeError: Cannot convert a BigInt value to a number``` that seems to be coming from the ```converter.ts``` file. line ```40``` in the screenshot below
Screenshot 2023-06-12 at 14 40 01

1. [First step]
create-react-app
2. [second step]
install the lastest web3 package ```yarn add web3```
3. [third step]
build the project
4. [and so on...]
deploy to [netlify](app.netlify.com) or [vercel](vercel.com)

#### Logs

Screenshot 2023-06-12 at 14 34 30

#### Environment
npm ```9.6.2```
node ```v18.15.0```
web3js ```v4.0.1```
OS ```macOS```
device ```macbook```

Muhammad-Altabba:
## Description

Fixes #6187

## Type of change

## Checklist:


packages/web3-eth-abi/src/coders/base/numbersLimits.ts
```diff
@@ -0,0 +1,505 @@
...
+		{
+			min: BigInt('0'),
+			max: BigInt('255'),
+		},
+	],
+	[
+		'int8',
+		{
+			min: BigInt('-128'),
+			max: BigInt('127'),
```
jdevcs: Instead of adding this large file , why not to use function that you added `bigintPower` in existing Map code .
```
const numberLimits = new Map<string, { min: bigint; max: bigint }>();

// precalculate all the limits
for (let i = 8; i <= 256; i += 8) {
	numberLimits.set(`uint${i}`, {
		min: BigInt(0),
		max: BigInt(2) ** BigInt(i) - BigInt(1),
	});
	numberLimits.set(`int${i}`, {
		min: -(BigInt(2) ** BigInt(i - 1)),
		max: BigInt(2) ** BigInt(i - 1) - BigInt(1),
	});
}
```
and see to save some lib size: as in this PR its 

Total | 585 KB | 592 KB | 6.96 KB | 1.19%

packages/web3-eth-abi/src/coders/base/number.ts
```diff
@@ -43,25 +44,6 @@ function uint8ArrayToBigInt(value: Uint8Array, max: bigint): bigint {
 	return result - mask;
 }
 
-const numberLimits = new Map<string, { min: bigint; max: bigint }>();
-
-// precalculate all the limits
-for (let i = 8; i <= 256; i += 8) {
-	numberLimits.set(`uint${i}`, {
-		min: BigInt(0),
-		max: BigInt(2) ** BigInt(i) - BigInt(1),
```
jdevcs, line 52: Exponentiation Operator usage issue was solved by specifying min browser versions in one of issue? 
Its supported in ECMAScript 2016 and 

Chrome 52 | Edge 14 | Firefox 52 | Safari 10.1 | Opera 39
Jul 2016 | Aug 2016 | Mar 2017 | Mar 2017 | Aug 2016

so react with older browsers is causing this problem only, and if some one specify min browsers list in react config and use on ECMAScript 2016 supporting browser there is no issue?

Muhammad-Altabba: Unfortunately, the issue is present with the latest as well as any older version of React . And any other framework that uses babel (the latest version of babel and any older version as well).

And this is what happens:
1. When building React app (to generate the build folder to be used for deployment), babel is used to tans-compile the code.
2. Babel will convert every `**` to `Math.pow`.
3. When browsing the page, an exception will be thrown because `Math.pow` does not work with `BigInt`.

Currently, the only way to avoid this behavior is to add configuration to `package.json` as mentioned in: https://github.com/babel/babel/issues/13109#issuecomment-826287089 and applied in https://github.com/web3/web3.js/issues/6187#issuecomment-1601223430

And this MR propose changes so the users do not need to add configurations to their `package.json`.

packages/web3-eth-abi/src/coders/base/numbersLimits.ts
Muhammad-Altabba: I though about reducing the execution time. But, sorry, I did not notice the size. I will modify to use the added `bigintPower`.

Muhammad-Altabba: I changed it to use a loop with multiplication instead of power.

packages/web3-eth-abi/src/coders/base/number.ts
avkos: I think it is better to have our own bigint pow function right now. and wait while Babel starts supporting it

avkos (approved):
Thanks for this fix

jdevcs (approved):
ok, Good, Thanks

Issue #6535, opened 2023-10-20
ERR_UNSUPPORTED_DIR_IMPORT in web3-eth-abi/lib/esm/coders/encode.js with node >18


CelestialCrafter:
#### Expected behavior
web3-eth-abi/lib/esm/coders/encode.js should be able to import from web3-eth-abi/lib/esm/coders/base

#### Actual behavior
web3-eth-abi/lib/esm/coders/encode.js errors with ERR_UNSUPPORTED_DIR_IMPORT in nodejs versions above or equal to 18

#### Steps to reproduce the behavior

1. install nodejs 18
2. npm install on an environment that has web3-eth-abi/lib/esm/coders/encode.js included
3. invoke something from web3-eth-abi/lib/esm/coders/encode.js

#### Logs

```
Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '/app/node_modules/web3-eth-abi/lib/esm/coders/base' is not supported resolving ES modules imported from /app/node_modules/web3-eth-abi/lib/esm/coders/encode.js
    at new NodeError (node:internal/errors:405:5)
    at finalizeResolution (node:internal/modules/esm/resolve:320:17)
    at moduleResolve (node:internal/modules/esm/resolve:946:10)
    at defaultResolve (node:internal/modules/esm/resolve:1132:11)
    at nextResolve (node:internal/modules/esm/loader:163:28)
    at ESMLoader.resolve (node:internal/modules/esm/loader:835:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:424:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:77:40)
    at link (node:internal/modules/esm/module_job:76:36)
Emitted 'error' event on Worker instance at:
    at [kOnErrorMessage] (node:internal/worker:300:10)
    at [kOnMessage] (node:internal/worker:311:37)
    at MessagePort.<anonymous> (node:internal/worker:212:57)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:757:20)
    at exports.emitMessage (node:internal/per_context/messageport:23:28) {
  code: 'ERR_UNSUPPORTED_DIR_IMPORT',
  url: 'file:///app/node_modules/web3-eth-abi/lib/esm/coders/base'
}

```
Node.js v18.18.2
10.69 error Command failed with exit code 1.
10.69 info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

[...]

PR #6545, opened 2023-10-25
4.2.1 release

luu-alex:
#### web3-eth-abi

-   Bug fix of `ERR_UNSUPPORTED_DIR_IMPORT` in ABI (#6535)

#### web3-eth-contract

-   Dependencies updated

#### web3-eth

-   Dependencies updated

#### web3-eth-ens

-   Dependencies updated

#### web3-eth-personal

-   Dependencies updated


packages/web3/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "web3",
-	"version": "4.2.0",
+	"version": "4.2.1",
```
jdevcs, line 3: plus also update changelog of main web3 package.

tools/web3-plugin-example/package.json
```diff
@@ -45,10 +45,10 @@
 		"prettier": "^2.7.1",
 		"ts-jest": "^28.0.7",
 		"typescript": "^4.7.4",
-		"web3": "^4.2.0",
+		"web3": "^4.2.1",
```
jdevcs, line 48: you will need to bump patch of plugin example if modifying its package.json, also its change log

jdevcs (approved):
lgtm, only you need above change.

avkos (approved):

PR #6596, opened 2023-11-14
fix: faster hex to byte implementation

mpetrunic:
## Description

Noticed this implementation in couple of crypto repos, apparently this implementation is 6x faster.

## Type of change

## Checklist:


avkos (approved):

packages/web3-validator/src/utils.ts
```diff
@@ -449,20 +449,49 @@ export function uint8ArrayToHexString(uint8Array: Uint8Array): string {
...
 	}
-	if (value.length % 2 !== 0) {
+	if (hex.length % 2 !== 0) {
 		throw new InvalidBytesError(`hex string has odd length: ${hex}`);
 	}
-	const bytes = new Uint8Array(Math.ceil(value.length / 2));
-	for (let i = 0; i < bytes.length; i += 1) {
-		const byte = parseInt(value.substring(i * 2, i * 2 + 2), 16);
-		bytes[i] = byte;
+	const length = Math.ceil((hex.length-offset) / 2);
```
Muhammad-Altabba: I think there is no need for `Math.ceil` here. Because, `(hex.length-offset) / 2` would always return integer because there is an already an exception for when hex.length is odd. Right?
```suggestion
	const length = (hex.length - offset) / 2;
```

packages/web3-validator/src/utils.ts
```diff
@@ -449,20 +449,49 @@ export function uint8ArrayToHexString(uint8Array: Uint8Array): string {
...
+	  return char - (charCodeMap.A - 10)
+	if (char >= charCodeMap.a && char <= charCodeMap.f)
+	  return char - (charCodeMap.a - 10)
+	return undefined
+  }
+
 export function hexToUint8Array(hex: string): Uint8Array {
-	let value;
+	let offset = 0;
 	if (hex.toLowerCase().startsWith('0x')) {
```
Muhammad-Altabba: I think, it is a bit costy to convert the full string to lowercase.
```suggestion
	if (hex[0] === '0' && (hex[1] === 'x' || hex[1] === 'X')) {
```

Muhammad-Altabba (approved):
Thanks @mpetrunic for your contribution.

I have just small suggestions.

jdevcs (commented):
linting should be fixed before merging ,

jxom:
Any chance there should be a comment in the code here that this was adapted from Viem?

mpetrunic:
> Any chance there should be a comment in the code here that this was adapted from Viem?

@jxom Sure, I can open PR to add that note. But tbh, I wasn't sure where it originated from. 

I thought it came from https://github.com/paulmillr/noble-hashes/blob/main/src/utils.ts#L50 
because I've seen paul's PR on viem repo?

jxom:
Ah true! Paul made the PR to Viem for this. Didn’t realise it was also in hashes! Might make a reference note for this in Viem repo then.

mpetrunic:
> Ah true! Paul made the PR to Viem for this. Didn’t realise it was also in hashes! Might make a reference note for this in Viem repo then.

https://github.com/web3/web3.js/pull/6646 :)

packages/web3-validator/src/utils.ts
```diff
@@ -449,20 +449,49 @@ export function uint8ArrayToHexString(uint8Array: Uint8Array): string {
...
+
+  function charCodeToBase16(char: number) {
+	if (char >= charCodeMap.zero && char <= charCodeMap.nine)
+	  return char - charCodeMap.zero
+	if (char >= charCodeMap.A && char <= charCodeMap.F)
+	  return char - (charCodeMap.A - 10)
+	if (char >= charCodeMap.a && char <= charCodeMap.f)
+	  return char - (charCodeMap.a - 10)
+	return undefined
+  }
```
nazarhussain, line 470: @mpetrunic Is this function not equivalent to the native `parseInt` with difference take it take integer parameter while `parseInt` take string parameter as `charCodeToBase16(97)` is same as `parseInt('a', 16)`.

The native `parseInt` in that use case is almost 5 times more faster. 

https://runkit.com/nazarhussain/6571df9b17589500083041e5

mpetrunic: Well yes, but for parseInt, I need to substring 2 characters which is expensive (memory allocation and all).
Funny enough, benching on runkit yields totally different results than locally:
image
image

PR #6635, opened 2023-12-03
Allow the `deploy` function to accept parameters even when no ABI was provided to the Contract

### LINKED ISSUE #6634: If the contract ABI not passed, the deploy method (the smart contract constructor) does not accept parameters
If the contract ABI not passed the smart contract constructor does not accept parameters

#### Expected behavior
If the ABI was not passed (technically it is given as `[]`) then the type checking for the constructor should allow passing parameters. And also no error is thorwn.

#### Actual behavior

If the ABI was passed as `[]`, the developer can call the `deploy`; But there will be an error if parameter(s) where passed. 

If the user tried to work around this by using `// @ts-expect-error`, an error will come at runtime:   
>  Web3ContractError: The number of arguments is not matching the methods required number. You need to pass 0 arguments.

      125 |     console.log(abi);
      126 |     if (inputLength !== args.length) {
    > 127 |             throw new Web3ContractError(
          |                   ^
      128 |                     `The number of arguments is not matching the methods required number. You need to pass ${inputLength} arguments.`,
      129 |             );
      130 |     }

      at encodeMethodABI (src/encoding.ts:127:9)
      at dataInputEncodeMethodHelper (src/utils.ts:45:14)
      at dataInputEncodeMethodHelper (src/utils.ts:94:20)
      at Contract._contractMethodDeploySend (src/contract.ts:1112:14)
      at Object._contractMethodDeploySend [as send] (src/contract.ts:640:17)
      at test/unit/contract.test.ts:288:6
      at test/unit/contract.test.ts:47:71
      at Object.<anonymous>.__awaiter (test/unit/contract.test.ts:43:12)

```ts
const fromAccount = '0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1';
const contract = new web3.eth.Contract([]);

const contractDeployed = await contract
  .deploy({
    arguments: [1000], // Error: Type 'number[]' is not assignable to type 'undefined'.
  })
  .send({
    from: fromAccount,
    gas: '1000000',
  });

```
#### Environment
web3 v4

Muhammad-Altabba:
## Description

Fixes #6634 

## Type of change

## Checklist:


Muhammad-Altabba:
Once the approach is approved, I will update to fix the failing test. (That failing test is not failing on my machine so it needs some investigation.)
Thanks,

avkos (approved):

packages/web3-eth-abi/src/api/parameters_api.ts
```diff
@@ -41,8 +41,10 @@ import { encodeParameters as encodeParametersInternal } from '../coders/encode.j
  *  > 0x000000000000000000000000000000000000000000000000000000008bd02b7b0000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000748656c6c6f212500000000000000000000000000000000000000000000000000
  * ```
  */
-export const encodeParameters = (abi: ReadonlyArray<AbiInput>, params: unknown[]): string =>
-	encodeParametersInternal(abi, params);
+export const encodeParameters = (
+	abi: ReadonlyArray<AbiInput> | 'infer-types',
```
luu-alex: is there a particular reason we are passing `infer-types` for the first paramter instead of having it as a new optional parameter?

Muhammad-Altabba: No specific reason except that the caller still need to pass something for this first parameter. And using for example `undefined` used to cause to throw and I did not want to change the behavior. 
However, thanks for commenting on this ugly param. I now created a separate function to handle the infer-types logic.

luu-alex (approved):

PR #6636, opened 2023-12-03
[DRAFT] Allow smart contract methods to accept parameters when ABI was provided not `as const`

### LINKED ISSUE #6275: `web3.eth.Contract.methods.myMethod()` does not accept to insert any parameters.
#### Expected behavior
- Expect `web3.eth.Contract.methods.myMethod(myParam)` works with parameters.
- The expectation behavior is explained in comments like below.

image

#### Actual behavior
- I implemented smart contract in solidity, and already deployed it to Goerli test network.
- ABI file was extracted, and I added the file into my code repository.
- Thus, ABI file was OK. However, the below things are bothering me.
- `web3.eth.Contract.methods.setSystemAddress(address)` does not allow the parameter (see the below image).
- The error says `You need 0 arguments, but you got 1 argument.`.

image

1. Setting as below Environment.
2. Get ABI file into `web3.eth.Contract` whatever you have to generate `Contract`.
3. Call `web3.eth.Contract.methods.yourMethod` with some parameters.
4. The syntax does not allow you insert any parameters.

#### Logs

Sorry, the compile cannot be processed, any error logs cannot be provided.

#### Environment

- Use `yarn` (classic v1) package manager 
- Use `NestJS` boilerplate with Node.js version `19.16.1` (also in `16.15.1`)
- Use `web3` package with version `4.0.2` (maybe recent version)
image
- MacOS (M1 ChipSet)

Muhammad-Altabba:
## Description

Fixes #6275

If the developer provided an ABI from a file but without having `as const`, if he provided the ABI without saving it to a ts file first (with the `as const`), or if he provided an empty ABI, the user can call `contract.methods.ANY_METHOD_NAM()` but cannot pass any parameter to his called methods. Meaning he can call only methods that does not accept parameter or he need to do something like: `(contract.methods.transfer as any)('0xe4beef667408b99053dc147ed19592ada0d77f59', 12)`
This MR will fix that. 

## Type of change

## Checklist:


packages/web3-eth-contract/src/contract.ts
```diff
@@ -114,7 +114,7 @@ type ContractBoundMethod<
 	Abi extends AbiFunctionFragment,
 	Method extends ContractMethod<Abi> = ContractMethod<Abi>,
 > = (
-	...args: Method['Inputs']
+	...args: Abi extends AbiFunctionFragment ? Method['Inputs'] : any
```
jdevcs, line 116: we are sure about correct types and encoding based on provided ABI instead of using only param types in ts directly. so i'll suggest to use ABI. and we dnt add feature for "accept parameters even when no ABI was provided"

Muhammad-Altabba: Sorry, for the confusion. I now edited the title of the PR to describe the most actual practical use case. And also added elaboration in the PR description.

And many thanks for reviewing this PR even it is still in draft.

PR #6643, opened 2023-12-06
Increase Web3-contract Unit tests to above 90%

### LINKED ISSUE #6565: Increase Web3-contract Unit tests to above 90%
Current tests at 87%

Muhammad-Altabba:
## Description

Fixes #6565 

## Type of change

## Checklist:


avkos (approved):

packages/web3-eth-contract/test/unit/contract.test.ts
```diff
@@ -340,6 +345,131 @@ describe('Contract', () => {
...
+					newContract.options.address = deployedAddr;
+					expect(_tx.data).toBeDefined();
+					if (
+						_tx.data ===
+							'0xa41368620000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000548656c6c6f000000000000000000000000000000000000000000000000000000' ||
+						_tx.data ===
+							'0x4495ef8a00000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000548656c6c6f000000000000000000000000000000000000000000000000000000'
+					) {
+						// eslint-disable-next-line
+						expect(_tx.to).toStrictEqual(deployedAddr);
```
jdevcs, line 368: I'll suggest to check after each tx, that data sent is correct instead of checking `to`, this will ensure correct encoded to right function

Muhammad-Altabba: Yes, this block checks incoding (but not in the order of transactions) as we have `if(_tx.data === ... || ...)`. And this means the status we are retuning below at `return { status: '0x1'....` is only happening if the encoding was correct according to the data string in the `if` above. And if the encoding was wrong for `await deployedContract.methods.setGreeting(arg).send(sendOptions)` or `await deployedContract.methods.setGreeting(arg, true).send(sendOptions);`, the test will fail.

And also we have other tests for encoding in this file. So, I suggest to not add more logic to test the order of the transactions with their encoding here. And that testing the encoding not in-order with `if(_tx.data === ... || ...)` is sufficent in this test.

Thanks,

jdevcs (approved):

PR #6646, opened 2023-12-07
chore: add attribution for faster bytesToHex implementation

mpetrunic:
related to https://github.com/web3/web3.js/pull/6596#issuecomment-1844145778

### DIFF
--- a/packages/web3-validator/src/utils.ts
+++ b/packages/web3-validator/src/utils.ts
@@ -450,7 +450,7 @@ export function uint8ArrayToHexString(uint8Array: Uint8Array): string {
 }
 
 // for optimized technique for hex to bytes conversion
-// many thanks to @noble/hashes and viem from where this implementation is taken <3
+// many thanks to @noble/hashes and viem from where this implementation is taken <3 and then enhanced a bit
 const charCodeMap = {
 	zero: 48,
 	nine: 57,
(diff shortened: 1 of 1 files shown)


packages/web3-validator/src/utils.ts
```diff
@@ -450,6 +450,7 @@ export function uint8ArrayToHexString(uint8Array: Uint8Array): string {
 }
 
 // for optimized technique for hex to bytes conversion
+// many thanks to @noble/hashes and viem from where this implementation is taken <3
```
Muhammad-Altabba: A tiny suggestion to count for the tiny enhancements mentioned at: https://github.com/web3/web3.js/pull/6596#pullrequestreview-1736321024 :smile:. If you like to do so. 
```suggestion
// many thanks to @noble/hashes and viem from where this implementation is taken <3 and then enhanced a bit
```

Muhammad-Altabba (approved):
Sounds good!
But I suggest that only the code origin (@noble/hashes) need to be mentioned. If you think so. 

And also I had a small suggestion that is to mention that the code was further enhanced. So, other developers can take the enhanced version and then mention that they took the code from web3 and the original @noble/hashes.

luu-alex (approved):

avkos (approved):

PR #6688, opened 2024-01-04
fix: contract method types should allow any param if abi type not supplied

mpetrunic:
## Description

Previously type was [] (zero element tuple) instead of any[] which prevented passing parameters if contract aby type was not defined.

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-types/src/eth_abi_types.ts
+++ b/packages/web3-types/src/eth_abi_types.ts
@@ -316,7 +316,7 @@ export type ContractMethodInputParameters<Params extends ReadonlyArray<unknown> 
 			  [MatchPrimitiveType<H['type'], H['components']>, ...ContractMethodInputParameters<R>]
 			: ContractMethodInputParameters<R>
 		: Params extends undefined | unknown
-		? any[]
+		? []
 		: Params;
 
 export type ContractConstructor<Abis extends ContractAbi> = {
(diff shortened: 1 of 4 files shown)


packages/web3-types/src/eth_abi_types.ts
```diff
@@ -316,7 +316,7 @@ export type ContractMethodInputParameters<Params extends ReadonlyArray<unknown>
 			  [MatchPrimitiveType<H['type'], H['components']>, ...ContractMethodInputParameters<R>]
 			: ContractMethodInputParameters<R>
 		: Params extends undefined | unknown
-		? []
+		? any[]
```
Muhammad-Altabba: Thanks @mpetrunic,
Do you think this is better than https://github.com/web3/web3.js/pull/6636/files#diff-7fa8f782e68ce845dbb6e8446b86041697fa427db5485b9c63bf689d0b32ed81R116?

avkos (approved):

packages/web3-types/src/eth_abi_types.ts
mpetrunic: I'm not 100% but yours doesn't cover constructor args?

mpetrunic: But this should be fixed regardless of your PR

Muhammad-Altabba: Thanks Marin for the fix and the clarification.
It would be also great if you could add a test. As you see my PR for the methods args types did not work, so I left it for when having time for it. As I worked on it in my free time :smile: 


By the way, the constructor args types were handled as part of: https://github.com/web3/web3.js/pull/6635. Where even if the ABI was not provided, the types are inferred as possible.

Muhammad-Altabba (approved):
It will be great  to add a small test where typescript does not complain about the passed args.
And the Unit and integration tests are failing and needed to be checked, please.
Many thanks,

mpetrunic:
@Muhammad-Altabba Improved solution and added test^^

jdevcs (approved):

PR #6732, opened 2024-01-18
v4 Tests - Testing time only Github action config

### LINKED ISSUE #6686: v4 Tests - Testing time
Decreasing v4 tests time.

avkos:
## Description

Please include a summary of the changes and be sure to follow our [Contribution Guidelines](https://github.com/web3/web3.js/blob/4.x/CONTRIBUTIONS.md).

Fixes #6686 

## Type of change

## Checklist:


avkos:
Improvements I've applied:
- I’ve moved from artifacts to cache to save time from tar pack/unpack and uploading/downloading time.
CJS Build script time: `1m 12s. (was 1m 51s)`

My investigations to decrease test time:
- we can run all packages and all tests in the package in parallel. for this, I’ve fixed all tests to make separate refilled and unlocked wallets for each test. But it doesn’t make sense because the amount of parallel process jest count equals the of amount CPUs. In free GitHub actions, we have 2 CPUs. To use this opportunity we need to have custom GitHubActions machines that have more CPUs
- we can run each package as an individual process in GitHub Actions. But as we use a free plan we have limited concurrent process amount. Only 20 processes. https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration
- set this to jest.config
```
transform: {
  '^.+\\.(ts|tsx)$': ['ts-jest', { isolatedModules: true, }],
}
```
this option switched off type checking in tests which decreases processing time. In our case, it is just 1-5 seconds. That is why I’ve decided not to use this option.

.github/workflows/build.yml
```diff
@@ -1,301 +1,293 @@
...
-        with:
-            install: false
-            command: yarn test:e2e:ganache:ws:${{ matrix.browser }}
-            cache-key: node-v${{ matrix.node }}-on-${{ matrix.browser }}-hash-${{ hashFiles('yarn.lock') }}
+    build-esm:
+        name: Build ESM
+        needs: build
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
```
jdevcs: there is only one node version so no need of matrix

.github/workflows/build.yml
```diff
@@ -1,301 +1,293 @@
...
+                with:
+                   path: ./
+                   key: web3-${{ matrix.node }}-${{github.event.pull_request.base.sha}}
+            -   run: yarn build:esm
+    build-types:
+        name: Build Types
+        needs: build
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
```
jdevcs: there is only one node version so no need of matrix , same here and other places in this PR

.github/workflows/build.yml
```diff
@@ -1,301 +1,293 @@
...
+        runs-on: ubuntu-latest
+        env:
+            INFURA_GOERLI_HTTP: ${{ secrets.INFURA_GOERLI_HTTP }}
+            INFURA_GOERLI_WS: ${{ secrets.INFURA_GOERLI_WS }}
+        strategy:
+            fail-fast: false
+            matrix:
+                node: [18]
+                backend: ['ganache']
+                mode: ['http']
```
jdevcs: there is only one combination of these three so no need of matrix, use these directly in test command

.github/workflows/build.yml
```diff
@@ -1,301 +1,293 @@
...
-                  comment-always: false
-          -   name: Save main branch benchmark data
-              uses: actions/cache/save@v3
-              if: github.event_name == 'push' && github.ref == 'refs/heads/4.x'
+                path: ./
+                key: web3-${{ matrix.node }}-${{github.event.pull_request.base.sha}}
+            - run: yarn test:e2e:${{ matrix.backend }}:${{ matrix.mode }}
+              shell: bash
+
+    e2e-ipc:
```
jdevcs: its for all transports 'ipc', 'ws', 'http' , so i'll suggest to rename job instead of e2e-ipc

.github/workflows/build.yml
```diff
@@ -1,301 +1,293 @@
...
+                shell: bash
+
+    e2e:
+        name: End-to-End ganache:ws
+        needs: build
+        runs-on: ubuntu-latest
+        strategy:
+            fail-fast: false
+            matrix:
+                node: [18]
```
jdevcs: why we need node for tests in browser?

jdevcs (approved):

luu-alex (approved):

Issue #6805, opened 2024-02-09
Expand web3config guide [docs]


SantiagoDevRel:
We should expand this page of the docs [web3config](https://docs.web3js.org/guides/web3_config/)
- add code samples and explanation of each parameter that can be changed.
- parameter are listed [here](https://docs.web3js.org/guides/web3_config/#configuration-parameters)
Example:

### [handleRevert](https://docs.web3js.org/api/web3-core/class/Web3Config#handleRevert)

- Explain what handleRevert is, how it works, why is useful
- Code sample to set handleRevert at global leve
```js
new Web3({
  provider: 'https://mainnet.infura.io/v3/YOURID',
  config: {
    handleRevert: true,
  },
});
```
- Code sample to set handleRevert at package level
```js
const context = new Web3Context('http://127.0.0.1:7545');
context.setConfig({ handleRevert: true });
``` 

### [defaultAccount](https://docs.web3js.org/api/web3-core/class/Web3Config#defaultAccount)

- Explain what defaultAccount is, how it works, why is useful
- Code sample to set `defaultAccount` at global level 
- Code sample to set `defaultAccount` at package level...

MeenuyD:
Hello @SantiagoDevRel I would like to work on this issue can you please assign it to me.

MeenuyD:
Hello @SantiagoDevRel I would like to work on this issue can you please assign it to me

mmyyrroonn:
@SantiagoDevRel @jdevcs Hello, I noticed the previous pull request hasn't been updated for a while and is about to be closed. May I take over this issue?

SantiagoDevRel:
Hey @mmyyrroonn , feel free to take it over, thanks 💪!

mmyyrroonn:
@SantiagoDevRel Hi, I opened a PR to fix this issue, could you help review it?

PR #6839, opened 2024-02-27
Expanded web3config Guide

MeenuyD:
Expanded web3config Guide in Documentation: Added explanations and code samples for handleRevert and defaultAccount parameters.
#6805

### DIFF
--- a/docs/docs/guides/web3_config/index.md
+++ b/docs/docs/guides/web3_config/index.md
@@ -30,6 +30,55 @@ There is list of configuration params that can be set for modifying behavior of 
 
 ## Global level Config
 
+`handleRevert` is a configuration option that specifies whether the web3 provider should handle revert errors automatically. When set to true, the web3 provider will catch revert errors and return a more informative error message instead of throwing an exception.
+
+``` ts
+import { Web3 } from 'web3';
+
+const web3 = new Web3({
+  provider: 'https://mainnet.infura.io/v3/YOURID',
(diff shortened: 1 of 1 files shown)


docs/docs/guides/web3_config/index.md
```diff
@@ -30,6 +30,55 @@ There is list of configuration params that can be set for modifying behavior of
 
 ## Global level Config
 
+`handleRevert` is a configuration option that specifies whether the web3 provider should handle revert errors automatically. When set to true, the web3 provider will catch revert errors and return a more informative error message instead of throwing an exception.
```
jdevcs, line 33: @MeenuyD Thanks for working on this, I'll suggest to create a new section in this guide, for explaning config options, and we need desc of all of following config params:
image

jdevcs: & if you open docs for these params in https://docs.web3js.org/guides/web3_config/  there is already desc that can be reused in this new section, 

as second step I'll suggest to add code snippets as well in doc as examples.

MeenuyD: Ok

jdevcs: @MeenuyD let us know when updates are complete, thx

avkos (changes requested):
wait for changes

jdevcs:
@MeenuyD

PR #6882, opened 2024-03-11
fix: fromWei() parsing of large numbers > 10^20

### LINKED ISSUE #6880: Web3.utils.fromWei silently performs unsafe conversion for big numbers because of scientific notation
#### Expected behavior

Web3.utils.fromWei is parsing big numbers (>=10^21) without adding an unnecessary dot 

#### Actual behavior
```
Web3.utils.fromWei(19999990000000099000000, 'ether')  === "1..99999900000001e+22"
Web3.utils.fromWei(19999990000000000000000, 'ether') ===  "0.0000001.999999e+22"
Web3.utils.fromWei(19000000000000000000000, 'ether') === "0.000000000001.9e+22"
```

Web3.utils.fromWei insert an additional  dot in the scientific notation when dealing with Big numbers
In the example below the bumber 19999999999999990000000 is parsed to ether. this correcponds to 19999.99999999999 ETH.
 
When parsed we can see that there is an additional dot in the output: 1.9.99999999999999e+22. 
If we parse this string to a float then a 1.9 will be produced. 

This behaviour silently without errors converted a number representing ca 19999 ETH to 1.9 ETH. 

```
import Web3 from "web3";

const wei = 19999999999999990000000; // A value bigger than 10^21
console.log(wei) // 1.999999999999999e+22
console.log(BigInt(wei)) // 19999999999999991611392n

let ethConverted = Web3.utils.fromWei(wei, 'ether');
console.log(ethConverted) // 1.9.99999999999999e+22. //  Here a string is returned with a second dot

let ethParsed = parseFloat(ethConverted);
console.log(ethParsed) // 1.9 // but should be > 19999
```

#### Logs

```
1.999999999999999e+22
19999999999999991611392n
1.9.99999999999999e+22
1.9
```

#### Environment
node version 20
web3.js version 4.6.0

#### Reason of the error

The internal logic of fromWei is behaving wrongly because the number is parsed into a scientific notation string like "1.999999999999999e+22". This inroduceses errors into the output

sgerodes:
Prevent automatic conversion to scientific notation for numbers >= 10^21 by using BigInt, ensuring accurate string representations and avoiding parsing errors in fromWei function.

Fixes https://github.com/web3/web3.js/issues/6880

## Description

Current errornous behaviour examples:

```
Web3.utils.fromWei(19999990000000099000000, 'ether') === "1..99999900000001e+22"
Web3.utils.fromWei(19999990000000000000000, 'ether') === "0.0000001.999999e+22"
Web3.utils.fromWei(19000000000000000000000, 'ether') === "0.000000000001.9e+22"
```
The following code performs silently a wrong parsing because of this Issue  

```
const wei = 19999999999999991611392;
const ethString = Web3.utils.fromWei(wei, 'ether');
const ethFloat = parseFloat(eth);
console.log(ethFloat) // equals 1.9 ETH, but should be > 19999 ETH
```
19999 ETH was parsed as 1.9 ETH with little code, that felt very intutive. Just using parseFloat. Which  probably many of the devs use after they receive a string from "fromWei". 

## Explanation of the behaviour 

When converting large numerical values from wei to ether using fromWei, we encounter an issue where numbers greater than or equal to 10^21 are automatically represented in scientific notation by JavaScript. This conversion introduces an additional decimal point into the string representation (1.9.99999999999999e+22 for 19999999999999990000000), leading to significant parsing errors and value distortion upon further processing (e.g., when using parseFloat, 19999 ETH incorrectly becomes 1.9 ETH).

JavaScript's default behavior of converting large numbers to scientific notation when stringified is the underlying cause. This automatic conversion impacts the string manipulation logic within the fromWei function, resulting in malformed outputs for large wei values.

## Solution

I have implemented a conditional check to utilize BigInt for numbers that are at risk of being automatically converted to scientific notation (>= 10^21).

## Type of change

## Checklist:

- [x] I ran `npm run test:coverage` and my test cases cover all the lines and branches of the added code.
- [x] I have linked Issue(s) with this PR in "Linked Issues" menu.


packages/web3-utils/src/converters.ts
```diff
@@ -411,6 +411,13 @@ export const toHex = (
  */
 export const toNumber = (value: Numbers): number | bigint => {
 	if (typeof value === 'number') {
+            if (value > 100000000000000000000) {
```
avkos: Tried to use this code but receive incorrect result
```
toNumber(100000000000000000010)
returns  100000000000000000000
```

avkos (changes requested):
Thanks for your contribution. I found that this solution does not cover all cases because of JavaScript problem

sgerodes:
@avkos Suddenly there is nothing the library, me or you can do about it.

The precision is lost alredy **before** the number enters any function of the web3 library. If the user choses to use javascript numbers (instead of bigints or strings) then he will deal with precision issues because the javascript numbers system itself is not precise. 

Here is an example. You can see that the precision is already lost while initializing the number, not in the web3 library function.

```
const number = 100000000000000000010
console.log(number);
console.log(Web3.utils.toNumber(number));
```
logs:
```
100000000000000000000
100000000000000000000
```
Here is also a warning from my IDE that denotes exactly that issue. 

Screenshot 2024-03-14 at 16 20 41

These issues you have mentioned were already there before my changes. My commits have not changed any behaviour related to that. If the user chooses to use the javascript 'number' type the value arrives in the function imprecise and there is no way the function can fix that.

packages/web3-utils/src/converters.ts
sgerodes: This is a behaviour built in the javascript number system. This can not be fixed. More information in my comment below.

luu-alex:
Hey there, thanks for submitting this PR :) we talked about this with the team and I think that theres limitations using number which is why this is headache dealing with large/decimal numbers. 

My suggestion is that if a number is at the max that javascript will allow it (which i believe in this case its 100000000000000000000), an error should produce telling the user that their number is too large, and should use string/bigInts.

sgerodes:
@luu-alex This is indeed a viable approach to handle this situation. But this would probably break a lot of application that already use this library. There is really big amount of application already using this library. Backwards compatibility is a real concern here. 

I can of course implement that, but are you sure that the error solution is good for the library?

sgerodes:
@luu-alex Anyways, I propose you to merge this request, because this PR is not about precision, it is about a real bug in parsing.

We can open a new ticket with a PR to discuss and potentially implement something for the precision problem. These two problems are linked but still different. And the initial Ticket was not aimed to solve the precision problem.

luu-alex:
@sgerodes I think you bring up some good points, we can move forward with this PR. I'll create another issue on precision for discussion

luu-alex:
https://github.com/web3/web3.js/blob/4.x/packages/web3-utils/CHANGELOG.md include this fix in the changelog with issue #

sgerodes:
@luu-alex  done. I have assumed this fix would be released in version 4.6.1

luu-alex (approved):

luu-alex:
@sgerodes we'll change that number when we plan to do next release, i changed it for you. :) thanks for this!

avkos (approved):

Issue #6894, opened 2024-03-13
bug: `toWei` fail to convert number beyond 8th decimal point on a number


BeroBurny:
#### Expected behavior

To manage to convert value to wei

#### Actual behavior

Throw error while converting value beyond 8th decimal point

#### Steps to reproduce the behavior

```js
utils.toWei(0.0000001, 'ether');
```

#### Logs

Created Script to test scenarios
https://github.com/BeroBurny/web3js-v4-inting/blob/58c2389176c12c297e029f8e716d70ce8db3c637/src/index.ts#L5-L46

and this produced logs

```sh
 [toWei] Number works with: 1
 [toWei] Number works with: 0.1
 [toWei] Number works with: 0.01
 [toWei] Number works with: 0.001
 [toWei] Number works with: 0.0001
 [toWei] Number works with: 0.00001
 [toWei] Number works with: 0.000001
 [toWei] Number failed with: 0.0000001
 -  Cannot convert 1e-7 to a BigInt
 [toWei] Number failed with: 0.00000001
 -  Cannot convert 1e-8 to a BigInt
 [toWei] Number failed with: 0.000000001
 -  Cannot convert 1e-9 to a BigInt
 [toWei] String works with: 1
 [toWei] String works with: 0.1
 [toWei] String works with: 0.01
 [toWei] String works with: 0.001
 [toWei] String works with: 0.0001
 [toWei] String works with: 0.00001
 [toWei] String works with: 0.000001
 [toWei] String works with: 0.0000001
 [toWei] String works with: 0.00000001
 [toWei] String works with: 0.000000001
```

#### Environment

Web3 4.6.0
Node 20

luu-alex:
Looks like at some point when decimals become larger js turns it into scientific notation, creating a PR to handle this case. thanks

PR #6908, opened 2024-03-18
Fix towei scientific notation

luu-alex:
## Description
#6894

## Type of change

## Checklist:


packages/web3-utils/src/converters.ts
```diff
@@ -411,6 +412,7 @@ export const toHex = (
  */
 export const toNumber = (value: Numbers): number | bigint => {
 	if (typeof value === 'number') {
+			console.warn('Warning: Using type `number` with values that are large or contain many decimals may cause loss of precision, it is recommended to use type `string` or `BigInt` when using conversion methods')
```
luu-alex: I added a warning to `toNumber`, I think it would be good overall that users should be using bigint or strings to convert but it may be alot of warnings that a user will face if they use this method multiple times. thoughts?

avkos: I think it can be an annoying warning. What do you think if we add this warning they use a number larger than max?

avkos (approved):

packages/web3-utils/src/converters.ts
```diff
@@ -411,6 +412,7 @@ export const toHex = (
...
             if (value > 1e+20) {
```
Muhammad-Altabba: What about handling things here for small numbers as well?

Muhammad-Altabba (changes requested):
I think it is better to not show a warning nor throw an exception. I think we just need to check in this line if the value is less than the minimum acceptable floating point number, and use BigInt in that case. So, I suggest to simply do it similar to how we currently handle for large numbers. Or what do you think?

packages/web3-utils/src/converters.ts
luu-alex: good suggestion thanks!

packages/web3-utils/src/converters.ts
luu-alex: that makes sense to me 👍

packages/web3-utils/src/converters.ts
```diff
@@ -551,17 +553,31 @@ export const toWei = (number: Numbers, unit: EtherUnits): string => {
 	if (!denomination) {
 		throw new InvalidUnitError(unit);
 	}
+	let parsedNumber = number;
+	if (typeof parsedNumber === 'number'){
+		if (parsedNumber  < 1e-15){
+			console.warn('Warning: The type `numbers` that are large or contain many decimals may cause loss of precision, it is recommended to use type `string` or `BigInt` when using conversion methods')
+			throw new InvalidNumberDecimalPrecisionLossError(number as number);
```
Muhammad-Altabba: I think it should be either a warning or an exception, but not both.
And could you please double check that the CHANGELOG.md is updated to mention the exception thrown for this case.

luu-alex: as well the error is a breaking change, so i removed the error and kept the warning. thanks 👍

packages/web3-utils/test/fixtures/converters.ts
```diff
@@ -300,9 +300,21 @@ export const fromWeiValidData: [[Numbers, EtherUnits], string][] = [
     [[1.9999999999999991611392e+22, 'ether'], '19999.999999999991611392'],
 ];
 
-export const toWeiValidData: [[Numbers, EtherUnits], string][] = [
+export const toWeiValidData: [[Numbers, EtherUnits], Numbers][] = [
 	...conversionBaseData,
 	[['255', 'wei'], '0xFF'],
+	[['255', 'wei'], '0xFF'],
```
Muhammad-Altabba: This line is doublicate. Was it meant to have differnt values? Or, should this be deleted?

packages/web3-utils/src/converters.ts
```diff
@@ -555,17 +558,30 @@ export const toWei = (number: Numbers, unit: EtherUnits): string => {
 	if (!denomination) {
 		throw new InvalidUnitError(unit);
 	}
+	let parsedNumber = number;
+	if (typeof parsedNumber === 'number'){
+		if (parsedNumber  < 1e-15){
+			console.warn(PrecisionLossWarning)
+		}
```
Muhammad-Altabba, line 565: Just wondering, why this is no need to call `parsedNumber =  BigInt(parsedNumber);` here similar to the logic for `parsedNumber  < 1e-15`?

packages/web3-utils/src/converters.ts
```diff
@@ -555,17 +558,30 @@ export const toWei = (number: Numbers, unit: EtherUnits): string => {
...
+		if (parsedNumber > 1e+20) {
+			console.warn(PrecisionLossWarning)
 
+			parsedNumber =  BigInt(parsedNumber);
+		}
+	}
+	
+	parsedNumber = typeof number === 'number' ? number.toLocaleString('fullwide', {useGrouping: false, maximumFractionDigits: 20}) : number;
```
Muhammad-Altabba: I think it is more clear to put the variable assigmnet here and then no need to use `? :`
```suggestion
		parsedNumber = number.toLocaleString('fullwide', {useGrouping: false, maximumFractionDigits: 20});
	}

```

Muhammad-Altabba (approved):

packages/web3-utils/src/converters.ts
luu-alex: Large decimal numbers values become scientific notation (for example `1e-15`) and scientific notation is not accepted by BigInt. On line [573](https://github.com/web3/web3.js/pull/6908/files#diff-217301804b75cf6f71d5a808105e97fd258371e0846a38f44841668b5c852148R573) we will convert the number to a string, so there is no need to convert to a bigint

packages/web3-utils/src/converters.ts
luu-alex: Yes I believe i can simpllify this

PR #6927, opened 2024-03-25
update custom_rpc_methods.ts: more meaningful rpc method name

agodbug:
## Description
rename a example rpc method name: custom_rpc_method -> custom_rpc_method_without_parameters


jdevcs:
this change is not necessary.

Issue #6933, opened 2024-03-25
eth.getBlock() JSON Fields Unavailable in v4.6, yet Available in v1.9


keennay:
Hi all,

I'm in the process of migrating our tools from Web3JS v1.9.0 to v4.6.0. I'd noticed `eth.getBlock()` on v4 returns less fields than on v1, specifically `blobGasUsed`, `excessBlobGas`, and others. This is while pointed to the same Ethereum Mainnet Go-Ethereum archive node, and confirmed a CURL command direct to the GETH node provides those extra values.

Is there an extra parameter to send to `eth.getBlock`, in order to view all data?

Below is the JSON response pointed to the same node, for block 19513177, using Web3JS v1.9 & v4.6.

Web3JS v1.9.0:
https://pastebin.com/q4XXcfh1

Web3JS v4.6.0:
https://pastebin.com/tEd2bLYK

==========

EDIT: Was initially on Web3JS v4.4, upgraded to v4.6 and the problem persists.

SantiagoDevRel:
Hey @ymonye thanks for pointing this out, we'll give provide an answer by tomorrow!

luu-alex:
Hey there! @ymonye we format our responses for our rpc methods and it seems like we need to update the response for the blobs from eip-4844, we'll create PR for this. Thanks for bringing this to our attention :)

keennay:
> Hey there! @ymonye we format our responses for our rpc methods and it seems like we need to update the response for the blobs from eip-4844, we'll create PR for this. Thanks for bringing this to our attention :)

Thanks @luu-alex! I'd started some of that work on my end to see those changes reflected within my NodeJS ESM code. This was from initially modifying my local `node_modules` files: `eth_types.js`, `rpc_methods_wrappers.js`, & `schemas.js`.

Afterwards I'd forked the web3.js repo to make the same edits within those Typescript files, yet I however have no idea how to import the the ESM module from the forked web3.js repo into my package. Other than that, please review whenever you get a chance.

I'll continue comparing & contrasting our code which is still reliant on web3.js v1.9 to v4, and will update you guys (or open more pull requests) if there's anything amiss.

luu-alex:
> Afterwards I'd forked the web3.js repo to make the same edits within those Typescript files, yet I however have no idea how to import the the ESM module from the forked web3.js repo into my package. Other than that, please review whenever you get a chance.


To test this in a local project and using a local copy of web3js:
[...]

PR #6937, opened 2024-03-26
Added EIP-4844 Headers for BlockHeaderOutput. Issue: 6933

keennay:
Added EIP-4844 & EIP-4788 Headers for the BlockHeaderOutput, following Go-Ethereum headers:
https://github.com/ethereum/go-ethereum/blob/738b5a586e329965539877434b695bb61015d4c7/core/types/block.go#L65

getBlock() now uses blockHeaderSchema instead of blockSchema.

Original Issue: https://github.com/web3/web3.js/issues/6933

## Description

## Type of change

- [*] New feature (non-breaking change which adds functionality)

## Checklist:

- [*] I have selected the correct base branch.
- [*] I have performed a self-review of my own code.
- [*] My changes generate no new warnings.
- [*] I ran `npm run lint` with success and extended the tests and types if necessary.
- [*] I ran `npm run test:unit` with success.
- [*] I have tested my code on the live network.

### DIFF
--- a/packages/web3-eth/src/rpc_method_wrappers.ts
+++ b/packages/web3-eth/src/rpc_method_wrappers.ts
@@ -59,6 +59,7 @@ import { decodeSignedTransaction } from './utils/decode_signed_transaction.js';
 import {
 	accountSchema,
 	blockSchema,
+	blockHeaderSchema,
 	feeHistorySchema,
 	logSchema,
 	transactionReceiptSchema,
@@ -257,7 +258,7 @@ export async function getBlock<ReturnFormat extends DataFormat>(
 			hydrated,
 		);
 	}
-	return format(blockSchema, response as unknown as Block, returnFormat);
+	return format(blockHeaderSchema, response as unknown as Block, returnFormat);
 }
 
 /**
--- a/packages/web3-eth/src/rpc_method_wrappers.ts
+++ b/packages/web3-eth/src/rpc_method_wrappers.ts
@@ -59,7 +59,6 @@ import { decodeSignedTransaction } from './utils/decode_signed_transaction.js';
 import {
 	accountSchema,
 	blockSchema,
-	blockHeaderSchema,
 	feeHistorySchema,
 	logSchema,
 	transactionReceiptSchema,
@@ -258,7 +257,7 @@ export async function getBlock<ReturnFormat extends DataFormat>(
 			hydrated,
 		);
 	}
-	return format(blockHeaderSchema, response as unknown as Block, returnFormat);
+	return format(blockSchema, response as unknown as Block, returnFormat);
 }
 
 /**
--- a/packages/web3-eth/src/schemas.ts
+++ b/packages/web3-eth/src/schemas.ts
@@ -236,91 +236,6 @@ export const transactionInfoSchema = {
 	},
 };
 
-export const blockSchema = {
-	type: 'object',
-	properties: {
-		parentHash: {
-			format: 'bytes32',
-		},
-		sha3Uncles: {
-			format: 'bytes32',
-		},
-		miner: {
-			format: 'bytes',
-		},
-		stateRoot: {
-			format: 'bytes32',
-		},
-		transactionsRoot: {
-			format: 'bytes32',
-		},
-		receiptsRoot: {
-			format: 'bytes32',
-		},
-		logsBloom: {
-			format: 'bytes256',
-		},
-		difficulty: {
-			format: 'uint',
-		},
-		number: {
-			format: 'uint',
-		},
-		gasLimit: {
-			format: 'uint',
-		},
-		gasUsed: {
-			format: 'uint',
-		},
-		timestamp: {
-			format: 'uint',
-		},
-		extraData: {
-			format: 'bytes',
-		},
-		mixHash: {
-			format: 'bytes32',
-		},
-		nonce: {
-			format: 'uint',
-		},
-		totalDifficulty: {
-			format: 'uint',
-		},
-		baseFeePerGas: {
-			format: 'uint',
-		},
-		size: {
-			format: 'uint',
-		},
-		transactions: {
-			oneOf: [
-				{
-					type: 'array',
-					items: {
-						...transactionInfoSchema,
-					},
-				},
-				{
-					type: 'array',
-					items: {
-						format: 'bytes32',
-					},
-				},
-			],
-		},
-		uncles: {
-			type: 'array',
-			items: {
-				format: 'bytes32',
-			},
-		},
-		hash: {
-			format: 'bytes32',
-		},
-	},
-};
-
 export const withdrawalsSchema = {
 	type: 'object',
 	properties: {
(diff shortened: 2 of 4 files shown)


luu-alex:
Thanks for this, could u remove the yarn.lock and update the changelog for [web3-eth](https://github.com/web3/web3.js/blob/4.x/packages/web3-eth/CHANGELOG.md) and [web3-types](https://github.com/web3/web3.js/blob/4.x/packages/web3-types/CHANGELOG.md)
Could u give PR edit access to contributors of this project as well, just in case i need to update this pr

packages/web3-eth/src/rpc_method_wrappers.ts
```diff
@@ -257,7 +258,7 @@ export async function getBlock<ReturnFormat extends DataFormat>(
 			hydrated,
 		);
 	}
-	return format(blockSchema, response as unknown as Block, returnFormat);
```
luu-alex: I believe you would want to edit `blockSchema` and add the new properties there, not editing blockHeaderSchema and just using blockSchema

keennay: > I believe you would want to edit `blockSchema` and add the new properties there, not editing blockHeaderSchema and just using blockSchema

Ok, yeah I'd noticed both the `withdrawals` & `withdrawalsRoot` properties only appear under `blockHeaderSchema`, in the `schemas.ts` file. When I call getBlock() on web3.js v1.9, it outputs both of those properties. This is why I'd assumed `getBlock()` probably needed `blockHeaderSchema` defined within v4's `rpc_method_wrappers.ts` file instead of `blockSchema`, and thought it was a typo.

Should `blockSchema` & `blockHeaderSchema` have identical properties then?

keennay:
> Thanks for this, could u remove the yarn.lock and update the changelog for [web3-eth](https://github.com/web3/web3.js/blob/4.x/packages/web3-eth/CHANGELOG.md) and [web3-types](https://github.com/web3/web3.js/blob/4.x/packages/web3-types/CHANGELOG.md) Could u give PR edit access to contributors of this project as well, just in case i need to update this pr

I think I did this, gave you access to my forked repo

packages/web3-eth/src/rpc_method_wrappers.ts
luu-alex: Yea its a bit confusing, It seems like `blockSchema` needs to also have updated properties of `withdrawals` and `withdrawalsRoot`. `blockHeaderSchema` is only being used for [subscriptions](https://github.com/web3/web3.js/blob/526c6f518f6a23438907b119aadbc868b06fbcd6/packages/web3-eth/src/web3_subscriptions.ts#L121). We may need to update the blockHeaderSchema as well, if you want to check and confirm result of a subscription the result you can also edit blockheaderSchema, otherwise you can focus on just updating blockSchema

jdevcs (changes requested):
remove changes in yarn.lock,

keennay:
> remove changes in yarn.lock,

Done

packages/web3-eth/src/rpc_method_wrappers.ts
keennay: Ok, I've updated both blockSchema & blockHeaderSchema in `schemas.ts`, with blockHeaderSchema retaining the below author & excessDataGas properties while missing from blockSchema. Please let me know if blockSchema should also have these.

blockSchema & blockHeaderSchema also have different nested arrays for the `transactions` property, and were left as originally found.

```
author: {
   format: 'bytes32',
},
excessDataGas: {
   format: 'uint',
},
```
The BlockOutput & BlockHeaderOutput interfaces in `eth_types.ts` were also updated accordingly.

Under eth_types.ts, I'd noticed a number of properties are formatted as either `bigint | number` or `Numbers`, whether they are defined under BlockOutput or BlockHeaderOutput.

For example: totalDifficulty in BlockOutput:
`readonly totalDifficulty?: bigint | number;`

Whereas in BlockHeaderOutput, totalDifficulty is formatted as:
`readonly totalDifficulty?: Numbers; `

I'm unsure if it should be one or the other, but I've left as-is. Also the extra properties: author & excessDataGas were not added to the BlockOutput interface, as they are not present in blockSchema. However I did add everything else from BlockHeaderOutput interface into BlockOutput, while formatting the ones with `Numbers` as `bigint | number`. Please advise if that was the correct move.

Also, let me know if those author & excessDataGase properites should also be present across blockSchema/blockHeaderSchema & BlockOutput/BlockHeaderOutput.

`yarn.lock` had its changes reverted, as well as `rpc_method_wrappers.ts`.

keennay: Also, it appears the 'newHeads' & 'newBlockHeaders' subscriptions only return formatted data from the `BlockBase` interface in `eth_types.ts`, and not from the `BlockHeaderOutput` interface.

https://docs.web3js.org/api/web3-eth/class/NewHeadsSubscription

keennay: Adding a note here so I don't forget after sleep. Looks like more parameters are missing from v4.6 yet present in v1.9, outside the scope of EIP-4844 changes, specifically across other `eth_*` functions. I'll do a much proper rundown thru the next couple of days, likely into the weekend.

For ex: the `id` parameter is missing in the `logs` array, from eth_getTransactionReceipt. Example below from an ETH mainnet request:

web3 v4.6:
`{"address":"0xb584d4be1a5470ca1a8778e9b86c81e165204599","blockHash":"0x47c58ebbaaff9a9a9ee60d74029224df7cac63b5c09972d4306e7500d8078b55","blockNumber":"19529894","data":"0x0000000000000000000000000000000000000000000000000000022ecb25c000","logIndex":"469","removed":false,"topics":["0x90890809c654f11d6e72a28fa60149770a0d11ec6c92319d6ceb2bb0a4ea1a15","0x0000000000000000000000000a25e45792fef9fb600a81f0d7ac4ab9e893fc71","0x0000000000000000000000000000000000000000000000000000000000000064"],"transactionHash":"0xe2bad713807549fe3bb0bf0a69dce0ddc634c8dbec655660ea038279ef5bdb91","transactionIndex":"170"}`

web3 v1.9:
`{"address":"0xb584D4bE1A5470CA1a8778E9B86c81e165204599","blockHash":"0x47c58ebbaaff9a9a9ee60d74029224df7cac63b5c09972d4306e7500d8078b55","blockNumber":19529894,"data":"0x0000000000000000000000000000000000000000000000000000022ecb25c000","id":"log_ff3f9ddd","logIndex":469,"removed":false,"topics":["0x90890809c654f11d6e72a28fa60149770a0d11ec6c92319d6ceb2bb0a4ea1a15","0x0000000000000000000000000a25e45792fef9fb600a81f0d7ac4ab9e893fc71","0x0000000000000000000000000000000000000000000000000000000000000064"],"transactionHash":"0xe2bad713807549fe3bb0bf0a69dce0ddc634c8dbec655660ea038279ef5bdb91","transactionIndex":170}`

avkos (approved):

luu-alex:
sorry for the long awaited response, looks good :) this will be included in the next release. Thanks for your efforts!

luu-alex (approved):

packages/web3-eth/src/schemas.ts
```diff
@@ -236,46 +236,55 @@ export const transactionInfoSchema = {
...
+
+export const blockSchema = {
+	type: 'object',
+	properties: {
+        baseFeePerGas: {
+			format: 'uint',
 		},
-		logsBloom: {
-			format: 'bytes256',
+        blobGasUsed: {
```
jdevcs, line 263: reference: https://github.com/ethereum/execution-apis/commit/057892b256c181232ca22499d1801b9637192698#diff-7c184b1c6b52b3ebe701b069a613639ad30699c60b9b352be78b64ac0206f424R82

packages/web3-eth/src/schemas.ts
```diff
@@ -284,15 +293,33 @@ export const blockSchema = {
...
 		},
+		stateRoot: {
+			format: 'bytes32',
+		},
+		timestamp: {
+			format: 'uint',
+		},
+		totalDifficulty: {
+			format: 'uint',
+		},
```
jdevcs, line 322: these fields are already in current schema but this PR is only rearranging these to diff lines.

packages/web3-types/src/eth_types.ts
```diff
@@ -142,44 +149,57 @@ export interface BlockOutput {
...
+    // These fields are returned when the RPC client is Nethermind,
+    // but aren't available in other clients such as Geth
+    readonly author?: Address;
+    readonly totalDifficulty?: Numbers;
+    readonly size?: Numbers;
+    readonly excessDataGas?: Numbers;
+    readonly mixHash?: HexString32Bytes;
+    readonly transactions?: TransactionOutput[];
+    readonly uncles?: Uncles;
+    readonly withdrawals?: Withdrawals[];
```
jdevcs, line 202: this is also rearranging most of field except adding ( eip 4844, 4895, 4788 )s

jdevcs (approved):
lgtm, @luu-alex plus add changelog in release PR for this:
added support of eip4844, eip4895, eip4788 block fields

PR #6941, opened 2024-03-26
Plugin guide 6840 draft

SantiagoDevRel:


docs/docs/guides/web3_plugin_guide/index.md
```diff
@@ -1,50 +1,334 @@
...
+//4. Use the plugin methods
+web3.pluginExample.doSomething();
+//--> Hello web3!
+```
+
+## Initialize a provider for the plugin
+
+There are 2 ways to initialize a provider when you are using a plugin, the [first one](#set-a-provider-within-the-plugin) is within the plugin (internally using the `this` object and `setProvider()` method). And the [second one](#register-plugin), is done by the final user by using the `registerPlugin()` method.
+
+### Register plugin (External - Final user)
```
jdevcs: This section is overlapping with '## Use a plugin' so it should be removed.

docs/docs/guides/web3_plugin_guide/index.md
```diff
@@ -1,50 +1,334 @@
...
+web3.registerPlugin(new MyPlugin());
+
+//4. Use the plugin methods
+web3.pluginExample.doSomething();
+//--> Hello web3!
+```
+
+## Initialize a provider for the plugin
+
+There are 2 ways to initialize a provider when you are using a plugin, the [first one](#set-a-provider-within-the-plugin) is within the plugin (internally using the `this` object and `setProvider()` method). And the [second one](#register-plugin), is done by the final user by using the `registerPlugin()` method.
```
jdevcs: I think this section should also be removed as its overlapping with **Set a provider within the plugin (web3Context)**, and ideally plugin dev shouldnt worry about modifying provider unless there is any special use case and for that you added section : **Set a provider within the plugin (web3Context)**

docs/docs/guides/web3_plugin_guide/index.md
```diff
@@ -1,50 +1,334 @@
...
+
+//Step #2: Create class
+class MyPlugin extends Web3PluginBase {
+  pluginNamespace = "myPlugin";
+
+//Step #3: Set providers
+  async getChainIdMainnet() {
+    this.setProvider("https://eth.llamarpc.com") //or any other RPC
+    return await eth.getChainId(this)
+  }
```
jdevcs: these setprovider will only work for rest of lib if its context is applied via link override like: 
https://github.com/ChainSafe/web3-plugin-chainstack/blob/fbe7b2376ca8623db3801b41f96e4f0fb3b8ef2d/src/chainstack-plugin.ts#L43

and provider is set for parent context like:
https://github.com/ChainSafe/web3-plugin-chainstack/blob/fbe7b2376ca8623db3801b41f96e4f0fb3b8ef2d/src/chainstack-plugin.ts#L87

jdevcs:
This PR lgtm , only few changes suggested.

docs/docs/guides/web3_plugin_guide/index.md
avkos: I think, setting a provider inside a plugin is a bad practice. Users of the plugin should decide what provider they want to use

avkos (changes requested):

avkos (approved):

jdevcs (approved):

PR #6947, opened 2024-03-28
Add default return format to the context

### LINKED ISSUE #6281: Make possibility to setup return format in global web3 instance class.
The main problem for users to move from 1.x to 4.x is returning format. All numbers become a BigInt. So, it will be great to have the possibility to set a default returning format in the web3 instance and then all methods in all packages (eth, contract, etc) will return correct formatting

avkos:
## Description

Fixes #6281 

## Type of change

## Checklist:


packages/web3-eth-contract/src/contract.ts
```diff
@@ -799,7 +799,8 @@ export class Contract<Abi extends ContractAbi>
 			},
 			estimateGas: async <ReturnFormat extends DataFormat = typeof DEFAULT_RETURN_FORMAT>(
 				options?: PayableCallOptions,
-				returnFormat: ReturnFormat = DEFAULT_RETURN_FORMAT as ReturnFormat,
+				returnFormat: ReturnFormat = (this.defaultReturnFormat ??
+					DEFAULT_RETURN_FORMAT) as ReturnFormat,
```
jdevcs: need a test for format here

packages/web3-eth/src/utils/get_revert_reason.ts
```diff
@@ -80,7 +85,8 @@ export async function getRevertReason<
 	web3Context: Web3Context<EthExecutionAPI>,
 	transaction: TransactionCall,
 	contractAbi?: ContractAbi,
-	returnFormat: ReturnFormat = DEFAULT_RETURN_FORMAT as ReturnFormat,
+	returnFormat: ReturnFormat = (web3Context.defaultReturnFormat ??
```
jdevcs: as now we have default format coming from context so there isn't need of DEFAULT_RETURN_FORMAT form types package, and same of other places in this PR.

packages/web3-eth/src/web3_eth.ts
```diff
@@ -219,7 +220,8 @@ export class Web3Eth extends Web3Context<Web3EthExecutionAPI, RegisteredSubscrip
 	 * ```
 	 */
 	public async getHashRate<ReturnFormat extends DataFormat = typeof DEFAULT_RETURN_FORMAT>(
-		returnFormat: ReturnFormat = DEFAULT_RETURN_FORMAT as ReturnFormat,
+		returnFormat: ReturnFormat = (this.defaultReturnFormat ??
+			DEFAULT_RETURN_FORMAT) as ReturnFormat,
```
```suggestion
			web3Context.defaultReturnFormat) as ReturnFormat,
```
for all places as now we have update-able default set in context so that should be used instead of `DEFAULT_RETURN_FORMAT` from types package, so:
if some one is explicitly setting format via this.defaultReturnFormat it will work, but if not it will send DEFAULT_RETURN_FORMAT to rpc-wrapper, so rpc-wrapper will get value and will not use context default? in that case if its `local ?? context` then it will reflect format settings that can be updated.

jdevcs (approved):
(edited)
@avkos precedence will be following:
1. local function format ( if defined will be considered first )
2. else context format
for all packages

packages/web3-eth-ens/test/integration/ens.events.test.ts
```diff
@@ -120,15 +119,6 @@ describeIf(isSocket)('ens events', () => {
 		else provider = new ENS.providers.HttpProvider(clientUrl);
 
 		ens = new ENS(registry.options.address, provider);
-
-		const block = await getBlock(ens, 'latest', false, DEFAULT_RETURN_FORMAT);
```
luu-alex, line 124: why was this removed? edit: i see you don't set gas, instead just using type

docs/docs/guides/web3_config/index.md
```diff
@@ -198,4 +199,46 @@ console.log(web3.getContextObject().config)
   transactionPollingTimeout: 750000,
   ...
 */
-```
\ No newline at end of file
+```
+
+
+### defaultReturnFormat
```
luu-alex, line 205: Think there should be a guide about defaultReturnFormat under upgrading from 1.x too because this PR is targetted towards users upgrading

luu-alex (approved):

Issue #6864, opened 2024-03-06
Web3 Middleware


jdevcs:
- [x] Feature:

- Build feature of Pre and Post hooks for request processing to be used in plugin development.

- This should allow to optionally pre-process request just before sending to provider, and post-process request's result after receiving from provider and before passing back to lib. 

- The user should be able to set middleware using web3 config and requestmanager should be aware if its not undefined and use that. 

- Define specifications for this middleware so users can implement it and set in web3js lib's plugins implementation 

- [x] Tests:

Add unit tests

PR #6951, opened 2024-04-03
web3 middleware

jdevcs:
## Description

https://github.com/web3/web3.js/issues/6864
https://github.com/web3/web3.js/issues/6963

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-core/src/web3_request_manager.ts
+++ b/packages/web3-core/src/web3_request_manager.ts
@@ -52,6 +52,7 @@ import {
 	isWeb3Provider,
 } from './utils.js';
 import { Web3EventEmitter } from './web3_event_emitter.js';
+import { RequestManagerMiddleware } from './types.js';
 
 export enum Web3RequestManagerEvent {
 	PROVIDER_CHANGED = 'PROVIDER_CHANGED',
@@ -73,9 +74,12 @@ export class Web3RequestManager<
 }> {
 	private _provider?: SupportedProviders<API>;
 	private readonly useRpcCallSpecification?: boolean;
+	public middleware?: RequestManagerMiddleware<API>;
+
 	public constructor(
 		provider?: SupportedProviders<API> | string,
 		useRpcCallSpecification?: boolean,
+		requestManagerMiddleware?: RequestManagerMiddleware<API>
 	) {
 		super();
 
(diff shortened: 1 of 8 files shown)


luu-alex (approved):

packages/web3-core/src/web3_request_manager.ts
```diff
@@ -155,7 +167,17 @@ export class Web3RequestManager<
 		Method extends Web3APIMethod<API>,
 		ResponseType = Web3APIReturnType<API, Method>,
 	>(request: Web3APIRequest<API, Method>): Promise<ResponseType> {
-		const response = await this._sendRequest<Method, ResponseType>(request);
+
+		let requestObj = {...request};
+
+		if (!isNullish(this.middleware))
```
avkos, line 173: it is better to have brackets `{` and `}` even if it is very simple `if`. because in the future when someone extends this code they can make a mistake

tools/web3-plugin-example/src/custom_rpc_methods.ts
```diff
@@ -25,7 +26,24 @@ type CustomRpcApi = {
...
+	public web3Middleware: Web3Middleware<CustomRpcApi> | undefined;
 
+	public constructor(testMiddleware = false){
+		super();
+
+		if(testMiddleware)
+			this.web3Middleware = new Web3Middleware<CustomRpcApi>();
+	}
+
+public link(parentContext: Web3Context): void {
```
avkos: Could you please check the prettier config on your IDE? Looks like it is not correct code automatically

avkos (changes requested):

Muhammad-Altabba (approved):
Seems great! beside Oleksii's points.

However, I think it would be the best to use this feature in a plugin before we make it available in the next release. In order to see it in action and see if we need some changes that we can do without casing a breaking changes in the subsequent releases. So, can we for example set it as experimental till we try it inside some real case? And later we mark it as a read-to-use feature?

jdevcs:
> Seems great! beside Oleksii's points.
> 
> However, I think it would be the best to use this feature in a plugin before we make it available in the next release. In order to see it in action and see if we need some changes that we can do without casing a breaking changes in the subsequent releases. So, can we for example set it as experimental till we try it inside some real case? And later we mark it as a read-to-use feature?

I also thought same to not include this in next release and test it with [Swisstronik usecase](https://www.npmjs.com/package/@swisstronik/web3-plugin-swisstronik), so in case any breaking changes required we can do before merging it ,I think we should merge this PR after next release ( and for next sprint's release ). 

I'll update and close implementation issue and created [next issue](https://github.com/web3/web3.js/issues/6963) for tracking above.

avkos (approved):

Issue #6393, opened 2023-08-30
 Web3ValidatorError error when use sendSignedTransaction on HB blockchain


louloub:
### Description
As an blockchain developper, i'm actualy working on smart contract ERC-721 POC. I have a Hyperledger Besu blockchain working on my VM with 4 nodes, built with Qorum Quickstart (https://besu.hyperledger.org/stable/private-networks/tutorials/quickstart). This blockchain is configured for 0 fees.
I can develop, build, deploy, use my contract on Remix connected on my Hyperldeger Besu blockchain, but when i want to deploy the same smart contract from Visual Studio Code i have some mistakes with the signature of my transaction for deploy the contract. 

### Acceptance Criteria
Can build and sign my transaction for deploy the contract on my Hyperledger Besu blockchain.

### Steps to Reproduce (Bug)
1. Launch Hyperledger Besu blockchain with ```./resume.sh``` from the blockchain folder
2. Compile my smart contract and retrieve BIN, ABI and JSON files from bash script. Truffle dependencies on my VIsual Code had never worked, it's why i'm using this script. 
Contract code on Test.sol :
```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.17;

contract Test {
    uint public count;

    function increment() external {
        count += 1;
    }
}
```
command to launch the script : 
```./compile.sh contract.sol```
bash script : 

```# retrieve file name without path
fileNameWithoutpath=$(basename $1)
#retrive file name without extension
fileNameWithoutPathAndExtension=${fileNameWithoutpath%.*}

echo "${bold}*************************************"
echo "Compile smart contract to BIN and ABI files from SOL files"
echo "*************************************${normal}"
solcjs --bin --abi $1 -o ../../build/contracts/
echo "fileNameWithoutPathAndExtension => $fileNameWithoutPathAndExtension"
solc $1 --combined-json abi,bin > ../../build/contracts/$fileNameWithoutPathAndExtension.json

for file in ../../build/contracts/*; do
    # Rename each file from this folder
    mv "$file" "../../build/contracts/${file##*_}"
done 
```

3. Once my contract was built, i want to create transaction to deploy it on my Heperldger Besu Blockchain. 

This is the genesis file of it :
```
{
    "config" : {
     "chainId": 1337,
	    "homesteadBlock": 0,
	    "eip150Block": 0,
	    "eip150Hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
	    "eip155Block": 0,
[...]

PR #6958, opened 2024-04-05
fix: toHex not supporting Uint8Array

### LINKED ISSUE #6957: toHex is not supporting passing Uint8Array
#### Expected behavior

toHex(new Uint8Array([
  221, 128, 128, 128, 148, 186, 248,
  242, 159, 130, 231,  84, 254, 199,
  252,  69,  21,  58, 104, 102, 201,
  137, 255,   3, 196,  10, 128, 128,
  128, 128
])) = 0xdd80808094baf8f29f82e754fec7fc45153a6866c989ff03c40a80808080

#### Actual behavior

It gets Json.stringified first and results in:
0x7b2230223a3232312c2231223a3132382c2232223a3132382c2233223a3132382c2234223a3134382c2235223a3138362c2236223a3234382c2237223a3234322c2238223a3135392c2239223a3133302c223130223a3233312c223131223a38342c223132223a3235342c223133223a3139392c223134223a3235322c223135223a36392c223136223a32312c223137223a35382c223138223a3130342c223139223a3130322c223230223a3230312c223231223a3133372c223232223a3235352c223233223a332c223234223a3139362c223235223a31302c223236223a3132382c223237223a3132382c223238223a3132382c223239223a3132387d

1. [First step]
2. [Second step]
3. [and so on...]

mpetrunic:
## Description

Fixes #6957 

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-utils/src/converters.ts
+++ b/packages/web3-utils/src/converters.ts
@@ -362,6 +362,10 @@ export const toHex = (
 		return returnType ? 'bigint' : numberToHex(value);
 	}
 
+	if(value instanceof Uint8Array) {
+		return returnType ? 'bytes' : bytesToHex(value);
+	}
+
 	if (typeof value === 'object' && !!value) {
 		return returnType ? 'string' : utf8ToHex(JSON.stringify(value));
 	}
--- a/packages/web3-utils/src/converters.ts
+++ b/packages/web3-utils/src/converters.ts
@@ -362,7 +362,7 @@ export const toHex = (
 		return returnType ? 'bigint' : numberToHex(value);
 	}
 
-	if(value instanceof Uint8Array) {
+	if(isUint8Array(value)) {
 		return returnType ? 'bytes' : bytesToHex(value);
 	}
 
--- a/packages/web3-utils/CHANGELOG.md
+++ b/packages/web3-utils/CHANGELOG.md
@@ -206,3 +206,8 @@ Documentation:
 - fixed erroneous parsing of big numbers in the `toNumber(...)` function (#6880)
 
 ## [Unreleased]
+
+### Fixed
+
+- fixed toHex incorrectly hexing Uint8Arrays and Buffer (#6957)
+- fixed isUint8Array not returning true for Buffer
--- a/packages/web3-utils/src/uint8array.ts
+++ b/packages/web3-utils/src/uint8array.ts
@@ -18,7 +18,8 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 export function isUint8Array(data: unknown | Uint8Array): data is Uint8Array {
 	return (
 		data instanceof Uint8Array ||
-		(data as { constructor: { name: string } })?.constructor?.name === 'Uint8Array'
+		(data as { constructor: { name: string } })?.constructor?.name === 'Uint8Array' ||
+		(data as { constructor: { name: string } })?.constructor?.name === 'Buffer'
 	);
 }
 
--- a/packages/web3-utils/test/fixtures/converters.ts
+++ b/packages/web3-utils/test/fixtures/converters.ts
@@ -235,6 +235,22 @@ export const toHexValidData: [Numbers | Bytes | Address | boolean, [HexString, V
 	],
 	['-0x01', ['-0x1', 'int256']],
 	['123c', ['0x123c', 'bytes']],
+	[new Uint8Array([
+		221, 128, 128, 128, 148, 186, 248,
+		242, 159, 130, 231, 84, 254, 199,
+		252, 69, 21, 58, 104, 102, 201,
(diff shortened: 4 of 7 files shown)


jdevcs (approved):
need changelog update, Thanks

luu-alex (approved):

avkos (approved):

packages/web3-utils/src/converters.ts
```diff
@@ -362,6 +362,10 @@ export const toHex = (
 		return returnType ? 'bigint' : numberToHex(value);
 	}
 
+	if(value instanceof Uint8Array) {
```
Muhammad-Altabba: Could you please use `isUint8Array` from web3-validator because `instanceof` is not reliable at some rare edge cases. 

```suggestion
	if( isUint8Array(value) ) {
```

Thanks,

mpetrunic: isUint8Array implementation is incorrect as it returns false on buffer even though buffer is uint8array. 

In what cases does instanceof not work?

Muhammad-Altabba: `isUint8Array` checks for `data.constructor.name === 'Uint8Array'` in addition to checking the `instanceof`. So it is more inclusive than `instanceof`. 

The edge case were `instanceof` fails was reported at our library at: https://github.com/web3/web3.js/issues/6393. And it was also reported to be unreliable at one of our dependencies, where they applied our suggestion (https://github.com/paulmillr/noble-hashes/issues/25#issuecomment-1750106284) to fix it.

mpetrunic: 🤦🏻‍♂️ 
image
```
--rant
Usually when people complain that `x instanceOf Uint8Array`, you are supposed to close the issue and tell them not to use a shitty engine that modifies globals (that also means not using jest+jsdom when you are developing a cross-platform library). 😅 

I'm really worried how everyone is ok with libraries and engines like jest/jsdome modifying globals and everybody else is applying hacks to make it work instead of urging people to not use projects like that
-- end rant
```
We cannot pretend Node.js doesn't exist and ignore Buffers so added commit which add support for Buffer in isUint8Array function and web3-validator.

Muhammad-Altabba: Thanks for the modifications. It looks good now that is not solely depending on `instanceof`.
Thanks also for using the same `data?.constructor?.name` for `Buffers`. 


A note and a question on the side: It would never be the intention to fix for jest/jsdom because nobody cares about it. And in the issue that was open at our library (#6393) the guy stated ["I have a Hyperledger Besu blockchain working on my VM with 4 nodes, built with Qorum Quickstart"](https://github.com/web3/web3.js/issues/6393). And the guy stated that he did not face the issue when he used ethers.js. Do you think that the issue that the guy reported was with jest/jsdom but falsely reported and we should had closed the issue stating "not to use a shitty engine that modifies globals"?

Muhammad-Altabba (approved):

packages/web3-utils/src/converters.ts
mpetrunic: Hard to say. My assumption is that Besu is using some shitty js engine to execute his javascript script so he should've open issue on besu repo saying that their js execution illegally modifies globals.

I would call bullshit that it works with ethers because: https://github.com/ethers-io/ethers.js/blob/ad5f1c5fc7b73cfb20d9012b669e9dcb9122d244/src.ts/utils/data.ts#L29 😂 😂

Muhammad-Altabba: Thanks @mpetrunic , And thanks for looking ethers code.

However, just for clarification, using `instanceof` is not reliable does not seem to be because of someone modifies globals. But because `Uint8Array` is "serializable, but not transferable" (https://github.com/web3/web3.js/issues/6393#issuecomment-1750095884).  

And following the provided links in the issues leads to [this article](https://medium.com/@simonwarta/limitations-of-the-instanceof-operator-f4bcdbe7a400) that mentions one of the cases for the failure of using `instanceof` when the guy is working on a browser extension. This is another example of the edge case that comes because of the nature of Uint8Array being "not transferable" as [mozilla documentation stated](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects).

mpetrunic: But doesn't that just mean that somebody illegally transferred "non-transferable" Uint8Array and then it was surprised that intanceof didn't work. Solution would then be to transfer underlying ArrayBuffer and on destination create new Uint8Array(arrayBuffer...) after which instanceof would work?

mpetrunic: But for jsdom: https://github.com/jsdom/jsdom/issues/1769

(for what it's worth happy-dom doesn't have those issues so if you are using vitest it's better to switch)

Issue #6741, opened 2024-01-22
Events Guide


mconnelly8:
2024 H1 Action Items: Expand web3.js documentation with 5 guides explaining common usage and/or best practices

SantiagoDevRel:
References:
- Change this [`introduction`](https://docs.web3js.org/guides/events_subscriptions/) section for `Mastering events subcriptions` and include sample codes for every supported subscriptions
- add a sample code to subscribe to a smart contract events using the contract instance  (with comments and more organized)
```ts
import {Web3} from "web3";
  
const web3 = new Web3("provider");

const contract = new web3.eth.Contract(abi, address);

const subscription = contract.events.RandomNo()
subscription.on("data", console.log);
```

Yazi27:
Also, chaining .on's does not work anymore as stated on the 1.2.0 docs

PR #6983, opened 2024-04-17
Events guide 6741

SantiagoDevRel:
Issue #6741 

pls confirm is this is correct like this:

- `on("data")` - Fires on each incoming log with the log object as argument.
```js
 subcription.on("data", (data) => console.log(data));
```

- `on("changed")` - Fires on each log which was removed from the blockchain. The log will have the additional property "removed: true".
```js
 subcription.on("changed", (changed) => console.log(changed));
```

- `on("error")` - Fires when an error in the subscription occurs.
```js
 subcription.on("error", (error) => console.log(error));
```

- `on("connected")` - Fires once after the subscription successfully connected. Returns the subscription id.
```js
 subcription.on("connected", (connected) => console.log(connected));
```


avkos (approved):

docs/docs/guides/events_subscriptions/index.md
```diff
@@ -1,17 +1,155 @@
...
 sidebar_position: 1
-sidebar_label: 'Introduction'
+sidebar_label: 'Mastering Events Subcriptions'
 ---
 
 # Events Subscription
 
+## Subscribing to smart contracts events
+
+```js
```
luu-alex:
```suggestion
```ts
```
should change all js to ts

docs/docs/guides/events_subscriptions/index.md
```diff
@@ -1,17 +1,155 @@
...
+
+async function subscribe() {
+  //create subcription
+  const subcription = await web3.eth.subscribe("logs");
+
+  //print logs of the latest mined block
+  subcription.on("data", (data) => console.log(data));
+
+  //unsubscribe
+  //await subcription.unsubscribe();
```
luu-alex: we should showcase unsubscribe rather than have it as a comment, maybe create a function called `unsubscribe` that takes in a subscription to unsubscribe

luu-alex (approved):

PR #7009, opened 2024-04-28
Added troubleshooting example

### LINKED ISSUE #6533: Add a troubleshooting page in our documentation
It is good to add a troubleshooting page in our documentation.
We can start the first section in the troubleshooting page by documenting the workaround for the issue with BigInt at expo. Here is the steps for expo shimming and poly-filling `BigInt`: https://github.com/web3/web3.js/issues/6495#issuecomment-1771011375

SantiagoDevRel:
closes #6533 

@Muhammad-Altabba pls check if this is what you have in mind for the troubleshooting page and feel free to request any changes💪🙌


Muhammad-Altabba (approved):
Looks great!
However, it could be also to split the Troubleshooting into a separate page since it may grow overtime.

SantiagoDevRel:
> Looks great! However, it could be also to split the Troubleshooting into a separate page since it may grow overtime.

Sure! since we agreed in the call a few days ago, let's leave it like this for now, if it grows too much we will split it for sure

Issue #7021, opened 2024-04-30
RPC Error: err: insufficient funds for gas * price + value


luu-alex:
trying to send a transaction with no data may error using rpc providers when there is sufficient funds for gas and values.

For example on metamask RPC Error: err: insufficient funds for gas * price + value, with unexpectedly high gas prices .

This issue is happening due to the sendtransaction method using eth_call to check if the smart contract will revert, but with metamask, eth_call will return an error expecting more gas to be provided.

Transaction sent:
```
object { from: "0xEd965cEAca9DE3DDE4e98109f52c3E195e3F2C91", to: "0xd8da6bf26964af9d7eed9e03e53415d37aa96045", value: "0x5af3107a4000", gasPrice: undefined, maxPriorityFeePerGas: "0x9502f900", maxFeePerGas: "0x951425ae" }
```
Metamask error 
```
MetaMask - RPC Error: err: insufficient funds for gas * price + value: address 

0xEd965cEAca9DE3DDE4e98109f52c3E195e3F2C91 have 508654303021506364 want 1250662775000000000 (supplied gas 500000000) 
```
Metamask is returning a revert issue wanting 1.250662775 eth when the transaction only needs atleast 0.005 eth in balance
Bug reported in https://github.com/web3/web3.js/issues/6997
#### Steps to reproduce the behavior

1. [First step]
2. [Second step]
3. [and so on...]

PR #7043, opened 2024-05-15
add gas conditional if simple transaction

luu-alex:
## Description
#7021 
There seems to be a bug when using sendTransaction with no gas property in the rpc method eth_call in `checkRevertBeforeSending`. Testing with both metamask and an infura provider, with a simple transaction (with no data or input so no contract call)
eth_call will revert and say there is insufficient funds when there should be easily enough funds to cover the gas and value. 
The transaction looks like 
```
const tx = {
        to: 0,
        value:1,
        from: accounts[0].address,
    }
```
```
InvalidResponseError: Returned error: err: insufficient funds for gas * price + value: address 0xBe1320eA9Fd1b34C0cf664eFbb29CF2dBD3cC81b have 100000000000000000 want 2859580732500000001 (supplied gas 50000000)
    at Web3RequestManager._processJsonRpcResponse (/Users/alexluu/practice/v4web3js/node_modules/web3/node_modules/web3-core/lib/commonjs/web3_request_manager.js:258:23)
    at Web3RequestManager.<anonymous> (/Users/alexluu/practice/v4web3js/node_modules/web3/node_modules/web3-core/lib/commonjs/web3_request_manager.js:156:29)
    at fulfilled (/Users/alexluu/practice/v4web3js/node_modules/web3/node_modules/web3-core/lib/commonjs/web3_request_manager.js:21:58)
  innerError: {
    code: -32000,
    message: 'err: insufficient funds for gas * price + value: address 0xBe1320eA9Fd1b34C0cf664eFbb29CF2dBD3cC81b have 100000000000000000 want 2859580732500000001 (supplied gas 50000000)'
```  

## Solution

Within `checkRevertBeforeSending`, A simple fix to this is to check if the transaction does not have data set we explicitly set the gas to 21000  

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-eth/src/utils/send_tx_helper.ts
+++ b/packages/web3-eth/src/utils/send_tx_helper.ts
@@ -122,7 +122,14 @@ export class SendTxHelper<
 
 	public async checkRevertBeforeSending(tx: TransactionCall) {
 		if (this.options.checkRevertBeforeSending !== false) {
-			const reason = await getRevertReason(this.web3Context, tx, this.options.contractAbi);
+			let formatTx = tx;
+			if (tx.data === undefined && tx.gas === undefined) { // eth.call will run into an error if data isnt filled and gas is not defined, simple transaction so we fill it with 21000
+				formatTx = {
+					...tx,
+					gas: 21000
+				}
+			}
+			const reason = await getRevertReason(this.web3Context, formatTx, this.options.contractAbi);
 			if (reason !== undefined) {
 				throw await getTransactionError<ReturnFormat>(
 					this.web3Context,
--- a/packages/web3-eth/src/utils/send_tx_helper.ts
+++ b/packages/web3-eth/src/utils/send_tx_helper.ts
@@ -123,7 +123,7 @@ export class SendTxHelper<
 	public async checkRevertBeforeSending(tx: TransactionCall) {
 		if (this.options.checkRevertBeforeSending !== false) {
 			let formatTx = tx;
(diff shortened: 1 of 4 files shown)


packages/web3-eth/src/utils/send_tx_helper.ts
```diff
@@ -122,7 +122,14 @@ export class SendTxHelper<
 
 	public async checkRevertBeforeSending(tx: TransactionCall) {
 		if (this.options.checkRevertBeforeSending !== false) {
-			const reason = await getRevertReason(this.web3Context, tx, this.options.contractAbi);
+			let formatTx = tx;
+			if (isNullish(tx.data) && isNullish(tx.input) && isNullish(tx.gas)) { // eth.call runs into error if data isnt filled and gas is not defined, its a simple transaction so we fill it with 21000
```
jdevcs, line 126: when data is populated, is gas limit always populated in that case?

jdevcs: and its only missing in case of value transfers `eth_calls` only?

jdevcs (approved):

packages/web3-eth/src/utils/send_tx_helper.ts
jdevcs: additionally, if tx has gas limit only for eth_call rpc( for revert check) and not in actual Tx that will be sent after it, 
eth_call may pass while actual Tx may fail in some cases, as both will differ, so I think gas limit should be applied in all cases

luu-alex: eth_call with no data and no gas will fail from provider, but send_transaction gas will be filled automatically. i belive this is just a weird edge case, and with our tests passing i dont believe this will be an issue.

PR #7070, opened 2024-05-28
fix(docs): mastering providers

danforbes:
Improvements to Mastering Provider section


jdevcs (approved):

docs/docs/guides/web3_providers_guide/index.md
```diff
@@ -29,64 +18,136 @@ const web3 = new Web3(/* PROVIDER*/);
...
+```
+
 ### WebSocket provider
 
+WebSockets support a persistent connection between a client and a server, which means they are suitable for use cases that require real-time [event subscriptions](/guides/events_subscriptions/).
+
 ``` ts title='Initialize WS Provider'
 import { Web3, WebSocketProvider } from 'web3';
 
+//supply a WebSocket provider as a URL string
```
luu-alex, line 97:
```suggestion
// supply a WebSocket provider as a URL string
```
nitpick: comments should always have a space before the text

danforbes: Yes, I completely agree 💯 That doesn't seem to be the case in most of the existing docs, though, so I chose to stay consistent. My next PR will be various formatting nitpicks like that 👍🏻

PR #7083, opened 2024-06-04
fix(docs): introduction

danforbes:
## Description

- In header navbar, libdocs link has been renamed from "Documentation" to "Packages"
- Added a "Using These Docs" section to Introduction
- Added a section about plugins to Introduction

### DIFF
--- a/docs/docs/guides/getting_started/introduction.md
+++ b/docs/docs/guides/getting_started/introduction.md
@@ -31,7 +31,7 @@ There is a lot to learn about Web3.js! Here are some tips for developers of diff
 
 ### For Beginner Web3.js Developers
 
-New Web3.js developers should proceed to the [Quickstart](/guides/getting_started/quickstart) section to learn how to get started with Web3.js. Once you understand the basics, you may want to consider learning more about [providers](/guides/web3_providers_guide/), [wallets and accounts](/guides/wallet/), [smart contracts](/guides/smart_contracts/), and how to [use Web3.js with the Hardhat development environment](/guides/hardhat_tutorial/).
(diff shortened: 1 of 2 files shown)


docs/docs/guides/getting_started/introduction.md
```diff
@@ -61,6 +58,31 @@ Web3.js is a modular collection of packages, each of which serves a specific nee
...
+
+Web3.js supports [plugins](/guides/web3_plugin_guide/), which are another way to encapsulate capabilities that support a specific need. There are plugins that exist to support native features, like those described by [EIPs](https://eips.ethereum.org/) as well as plugins that are designed to support specific smart contracts, middleware, or even other Ethereum-compatible networks. Visit the [Web3.js plugins homepage](https://web3js.org/plugins) to view a list of the most important Web3.js plugins, which includes:
+
+- [EIP-4337 (Account Abstraction) Plugin](https://www.npmjs.com/package/@chainsafe/web3-plugin-eip4337)
+
+- [EIP-4844 (Blob Transactions) Plugin](https://www.npmjs.com/package/web3-plugin-blob-tx)
+
+- [zkSync Plugin](https://www.npmjs.com/package/web3-plugin-zksync)
+
+## Features of Web3.js v4
```
jdevcs: I think we should keep this overview under introduction section

jdevcs: but at same time if we want to highlight "Using These Docs" section , what do you think about using https://docusaurus.io/docs/markdown-features/admonitions ?

danforbes: Oh, I just moved the section back to its original location 😅 Are you suggesting that "Features of Web3.js v4" get moved back to its original location and then we use an admonition for the "Using These Docs" section? I do not necessarily like the idea of using an admonition for "Using These Docs" as I think it's too much text to go into an admonition...

jdevcs (approved):
looks good, just added few updates ideas, Thanks @danforbes

Issue #7033, opened 2024-05-06
Increase testing coverage - web3-core


mconnelly8:
[Codecov](https://app.codecov.io/gh/web3/web3.js/tree/4.x/packages%2Fweb3-core%2Fsrc)

Goal should be 100%

PR #7091, opened 2024-06-07
add test coverage to web3-core

luu-alex:
## Description

## Type of change
https://github.com/web3/web3.js/issues/7033

## Checklist:

### DIFF
--- a/packages/web3-core/src/web3_request_manager.ts
+++ b/packages/web3-core/src/web3_request_manager.ts
@@ -429,22 +429,12 @@ export class Web3RequestManager<
 		) {
 			return this._buildResponse(payload, response, error);
 		}
-
 		if (jsonRpc.isBatchRequest(payload) && !Array.isArray(response)) {
 			throw new ResponseError(response, 'Got normal response for a batch request.');
 		}
 
 		if (!jsonRpc.isBatchRequest(payload) && Array.isArray(response)) {
 			throw new ResponseError(response, 'Got batch response for a normal request.');
-		}
-
-		if (
-			(jsonRpc.isResponseWithError(response) || jsonRpc.isResponseWithResult(response)) &&
-			!jsonRpc.isBatchRequest(payload)
-		) {
-			if (response.id && payload.id !== response.id) {
-				throw new InvalidResponseError<ErrorType>(response);
-			}
 		}
 
 		throw new ResponseError(response, 'Invalid response');
--- a/packages/web3-core/src/web3_subscriptions.ts
+++ b/packages/web3-core/src/web3_subscriptions.ts
@@ -30,7 +30,6 @@ import {
 	Web3APISpec,
 } from 'web3-types';
 import { jsonRpc } from 'web3-utils';
-import { SubscriptionError } from 'web3-errors';
 
 // eslint-disable-next-line import/no-cycle
 import { Web3SubscriptionManager } from './web3_subscription_manager.js';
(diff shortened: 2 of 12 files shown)


packages/web3-core/src/web3_request_manager.ts
```diff
@@ -438,15 +437,6 @@ export class Web3RequestManager<
 			throw new ResponseError(response, 'Got batch response for a normal request.');
 		}
 
-		if (
-			(jsonRpc.isResponseWithError(response) || jsonRpc.isResponseWithResult(response)) &&
```
luu-alex, line 442: removed, This conditional can never happen as we check `jsonRpc.isResponseWithError(response)` and  ` jsonRpc.isResponseWithResult(response)` seperately

packages/web3-core/src/web3_subscriptions.ts
```diff
@@ -82,16 +81,6 @@ export abstract class Web3Subscription<
 		super();
 		const { requestManager } = options as { requestManager: Web3RequestManager<API> };
 		const { subscriptionManager } = options as { subscriptionManager: Web3SubscriptionManager };
-		if (requestManager && subscriptionManager) {
```
luu-alex, line 85: removed, in the [constructor](https://github.com/web3/web3.js/blob/c5d47f8c68f6b69114dc9708759d30449315c015/packages/web3-core/src/web3_subscriptions.ts#L63)
only one of `requestManager` or `subscriptionManager` must be defined

packages/web3-core/src/web3_subscriptions.ts
```diff
@@ -82,16 +81,6 @@ export abstract class Web3Subscription<
...
-			throw new SubscriptionError(
-				'Only requestManager or subscriptionManager should be provided at Subscription constructor',
-			);
-		}
-		if (!requestManager && !subscriptionManager) {
```
luu-alex, line 90: removed, in the [constructor](https://github.com/web3/web3.js/blob/c5d47f8c68f6b69114dc9708759d30449315c015/packages/web3-core/src/web3_subscriptions.ts#L63)
either `requestManager` or `subscriptionManager` must be defined

packages/web3-types/src/json_rpc_types.ts
```diff
@@ -55,8 +55,9 @@ export interface JsonRpcNotification<T = JsonRpcResult> {
 	readonly jsonrpc: JsonRpcIdentifier;
 	readonly method: string; // for subscription
 	readonly params: SubscriptionParams<T>; // for subscription results
-	readonly result: never;
+	readonly result?: never;
```
luu-alex, line 58: Needed to add this otherwise we can't instantiate JsonRpcNotification

jdevcs: this will be breaking change & , the result property is marked as a required property of type never. This means that any object conforming to this interface must have a result property, but it can never be assigned a value because the never type represents a state that should never occur, so its a contradictory definition.

jdevcs (commented):
lgtm, except a change suggested

avkos (approved):
LGTM except Junaid comment

packages/web3-types/src/json_rpc_types.ts
luu-alex: This is a bug,
Within [processSubscriptionData](https://github.com/web3/web3.js/blob/eab5a68eb5e97864a6b3fba7c3763ec7a2d09bb9/packages/web3-core/src/web3_subscriptions.ts#L107) data must be 
```
data:
			| JsonRpcSubscriptionResult
			| JsonRpcSubscriptionResultOld<Log>
			| JsonRpcNotification<Log>,
```
and within type `JsonRpcSubscriptionResult` result is a string,
we run into an issue in data is `JsonRpcNotification<Log>` when result is never, saying result is missing
```
 test/unit/web3_subscription.test.ts:251:10 - error TS2741: Property 'result' is missing in type '{ jsonrpc: "2.0"; params: { subscription: string; result: { id: string; }; }; method: string; }' but required in type 'JsonRpcNotification<Log>'.

    251    const responseWithNotification: JsonRpcNotification<Log> = {
                 ~~~~~~~~~~~~~~~~~~~~~~~~

      ../web3-types/lib/commonjs/json_rpc_types.d.ts:36:14
        36     readonly result: never;
                        ~~~~~~
        'result' is declared here.
```
Because result mandatory in `JsonRpcSubscriptionResult` 
and never in `responseWithNotification` using `result ?: never` will still enforce the type and allow processSubscriptionData to pass in type responseWithNotification

We do this already in other types already
https://github.com/web3/web3.js/blob/eab5a68eb5e97864a6b3fba7c3763ec7a2d09bb9/packages/web3-types/src/json_rpc_types.ts#L31

PR #7106, opened 2024-06-17
fix(docs): metamask-tutorial

danforbes:
Updates to MetaMask tutorial


avkos (approved):

jdevcs (approved):

luu-alex:
I took a look at the new docs, and I think the content is great! its in-depth and goes through alot of the actions that a user would do when trying to when using metamask + web3.js, maybe the only thing missing is sending transactions and contact interactions but i think thats out of scope for this.
The only thing that worries me is that alot of the users will be developing using a framework like react and when you have an example created in html, it’ll be confusing or not helpful to these users.
I think we can keep this, but i’d also like to still have the original example and live code editor as well that oleksii created. if you think oleksii’s code example is not up to standard i would not mind helping out and fleshing it out to make it within your vision

These tutorials can seperated by tabs:

Tutorial: Connecting to metamask with HTML
Tutorial: Connecting to metamask within a react project

SantiagoDevRel:
> I took a look at the new docs, and I think the content is great! its in-depth and goes through alot of the actions that a user would do when trying to when using metamask + web3.js, maybe the only thing missing is sending transactions and contact interactions but i think thats out of scope for this. The only thing that worries me is that alot of the users will be developing using a framework like react and when you have an example created in html, it’ll be confusing or not helpful to these users. I think we can keep this, but i’d also like to still have the original example and live code editor as well that oleksii created. if you think oleksii’s code example is not up to standard i would not mind helping out and fleshing it out to make it within your vision
> 
> These tutorials can seperated by tabs:
> 
> Tutorial: Connecting to metamask with HTML Tutorial: Connecting to metamask within a react project

Exactly! in the slack thread I commented the same, we should keep this HTML guide and also add a React guide for the same (if Im not mistaken this React guide was already there in `Connecting with metamask` section previously but I think this got changed) anyway we should have both HTML & React

luu-alex (approved):

PR #7119, opened 2024-06-21
feat(docs): external-provider

### LINKED ISSUE #7115: Create documentation for Quicknode provider feature
- [ ] Update docs : QuickNodes is now default provider.
- [ ] Update docs : there is rate limit of QuickNode provider & create new user account in case of limit reached by using web3.js referral link for supporting web3.js lib
- [ ] Adding External providers guide: 
how to create instance of Quicknode provider and use with web3 for custom Network, Transport and token , example code is https://github.com/web3/web3.js/blob/bfe276956517245d580af53c2c28203825572bc8/packages/web3/test/integration/web3RPCProviders.test.ts#L32

danforbes:
Add documentation for external providers.

Closes #7115


docs/docs/guides/web3_providers_guide/index.md
```diff
@@ -272,3 +274,38 @@ The following example should be run in a browser with the MetaMask extension ins
...
+
+## Web3.js External Providers
+
+The `web3-rpc-providers` package provides helpful utilities for working with certain well-known remote providers. The following example demonstrates using this package to create a WebSocket connection for the Ethereum [Sepolia](https://ethereum.org/en/developers/docs/networks/#sepolia) test network with [QuickNode](https://www.quicknode.com/):
+
+```js
+import { Web3 } from "web3";
+import { Network, QuickNodeProvider, Transport } from "web3-rpc-providers";
+
+const web3 = new Web3(new QuickNodeProvider(Network.ETH_SEPOLIA, Transport.WebSocket));
```
SantiagoDevRel, line 286: Here we should have `Transport.HTTP` since the example is using `getChainId()` and not listening/subscribing to events

docs/docs/guides/web3_providers_guide/index.md
```diff
@@ -272,3 +274,38 @@ The following example should be run in a browser with the MetaMask extension ins
...
+  new QuickNodeProvider(
+    // omit network parameter
+    undefined,
+    Transport.WebSocket,
+    defaultSepoliaToken,
+    defaultSepoliaHost,
+  ),
+);
+
+console.log(await web3.eth.getChainId());
```
SantiagoDevRel, line 309: Here instead of `getChainId()` again, I think it could be better to use a method which actually demostrates `how to use account tokens and custom hosts`? to explain the above statement

docs/docs/guides/web3_providers_guide/index.md
danforbes: I just wanted to demonstrate that this is how the user would select a different transport. There are [other places in the docs](https://docs.web3js.org/guides/web3_providers_guide/#websocket-provider) where we demonstrate using a WebSocket provider without making use of subscriptions. I think it's important that we show users how to select the transport mechanism, but I think adding in a subscription to this part would make the example unnecessarily complicated. Does that make sense to you?

docs/docs/guides/web3_providers_guide/index.md
danforbes: I don't think that using an account token and custom host provides the user with any new capabilities. I think it's just for authentication purposes to get around rate limits. Maybe I am missing something. What method did you have in mind?

docs/docs/guides/web3_providers_guide/index.md
SantiagoDevRel: Perfect! lets just leave it like that, i just said because one-time request like `getChainid()` are made with HTTP, but its fine💪

docs/docs/guides/web3_providers_guide/index.md
SantiagoDevRel: Ohhh when I read `account tokens` I though about ERC20 tokens and that there was any new capability of doing something with tokens by adding the RPC providers, that's why I was confused, thanks Dan!!

jdevcs (commented):
In some next PR when RPC providers package is added in libdocs, I think its link and one line intro also needs to be in https://docs.web3js.org/#packages--plugins section.

jdevcs:
this PR looks good, Thanks @danforbes

Redoudou:
@danforbes @jdevcs  you can use the code WEB3JS and user will get 20% off for 2 months when they switch to a paid RPC provider with Quick Nodes.

docs/docs/guides/getting_started/quickstart.md
```diff
@@ -43,25 +43,27 @@ import { Web3 } from 'web3';
...
-// ↳ 18849658n
+console.log(await web3.eth.getChainId());
+// ↳ 1n
 ```
 ## Querying the Blockchain
 
-After instantiating the `Web3` instance with a provider, the [`web3-eth`](/libdocs/Web3Eth) package can be used to fetch data from the Ethereum network:
+After instantiating the `Web3` instance, the [`web3-eth`](/libdocs/Web3Eth) package can be used to fetch data from the Ethereum network:
 
 ```ts
```
jdevcs, line 68: RPC Providers package should also be under https://42f7c920.web3-js-docs.pages.dev/#additional-supporting-packages

danforbes: All set, @jdevcs https://380b2f6f.web3-js-docs.pages.dev/#additional-supporting-packages

jdevcs (approved):

jdevcs (approved):

avkos (approved):

danforbes:
Still active

PR #7197, opened 2024-08-16
Refactor some parts of contract and accounts packages

### LINKED ISSUE #7196: Refactor some parts of Contract, Accounts and Eth packages
Refactor some parts of Contract, Accounts and Eth packages to be more usable inside web3.js ZKSync plugin .

Muhammad-Altabba:
## Description

Fixes #7196

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-eth-contract/src/contract.ts
+++ b/packages/web3-eth-contract/src/contract.ts
@@ -15,6 +15,8 @@ You should have received a copy of the GNU Lesser General Public License
 along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+/* eslint-disable max-classes-per-file */
+
 import {
 	Web3Context,
 	Web3EventEmitter,
@@ -87,6 +89,7 @@ import {
 	TransactionReceipt,
 	FormatType,
 	DecodedParams,
+	TransactionCall,
 } from 'web3-types';
 import {
 	format,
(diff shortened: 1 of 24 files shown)


packages/web3-eth-contract/src/contract.ts
```diff
@@ -215,6 +218,182 @@ const contractSubscriptions = {
 	newBlockHeaders: NewHeadsSubscription,
 };
 
+
+/*
+ * This class is only supposed to be used for the return of `new Contract(...).deploy(...)` method.
+ */
+export class DeployerMethodClass<FullContractAbi extends ContractAbi> {
```
jdevcs: why there is need of moving `deploy` to different class?

jdevcs: (from usability perspective for external users)

Muhammad-Altabba: Good question,
I needed to add `populateTransaction`. But the `Contract` class was already about 1500 lines of code. So abstracting the logic into multiple classes will ease code maintenance for us. And if a user wanted to extend the `Contract` class he will have to deal with simpler code structures... 
And we would in the future also create a class for the return of the method `_createContractMethod` as it also returns a class like structure.

avkos (approved):

packages/web3-eth-contract/src/contract.ts
```diff
@@ -15,6 +15,8 @@ You should have received a copy of the GNU Lesser General Public License
 along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+/* eslint-disable max-classes-per-file */
```
jdevcs: I'll suggest to move new class in different file

jdevcs: LGTM just need above change

avkos:
I've refactored plugin a little bit. now it is not necessary to register a new transaction type. so you can make changes in this PR too

Muhammad-Altabba:
> I've refactored plugin a little bit. now it is not necessary to register a new transaction type. so you can make changes in this PR too

Thanks @avkos ,
The changes in this MR for the transaction type registration still fixes an issue that might happen if someone used the this functionality.

PR #7239, opened 2024-09-05
feat(requestEIP6963Providers): return-type

### LINKED ISSUE #7238: Add Return Type to `requestEIP6963Providers` Function Signature
The static `Web3.requestEIP6963Providers` function has an untyped return value. The return type of this function should be `Map<string, EIP6963ProviderDetail<API>>`.

danforbes:
Add return type to `requestEIP6963Providers` function signature.

Closes #7238


jdevcs (approved):
Thanks Dan, also could you update changelog for `EIP6963ProviderResponse`

PR #7270, opened 2024-09-18
Export EIP-6963 Types

### LINKED ISSUE #6952: export eip6963 types
eip6963 types are not being exported to the package, this makes using eip6963 hard for users,
https://github.com/web3/web3.js/blob/4.x/packages/web3/src/web3_eip6963.ts types should be exported and available for ease

1. [First step]
2. [Second step]
3. [and so on...]

danforbes:
Export EIP-6963 types


packages/web3-types/src/eip6963_types.ts
```diff
@@ -0,0 +1,38 @@
...
+	rdns: string;
+}
+
+export interface EIP6963ProviderDetail<API = Web3APISpec> {
+	info: EIP6963ProviderInfo;
+	provider: EIP1193Provider<API>;
+}
+
+export type EIP6963ProviderResponse = Map<string, EIP6963ProviderDetail>;
+
```
jdevcs, line 34: why these types are moved from web3 to types package?

packages/web3-types/src/eip6963_types.ts
```diff
@@ -0,0 +1,38 @@
...
+export interface EIP6963ProvidersMapUpdateEvent extends CustomEvent {
```
jdevcs: why these types are moved from web3 to types package?

danforbes: Because I think they should be exported and it seemed like the best way to go about it. I can modify the PR export them directly from the `web3` package, but this seemed like a better approach to me.

jdevcs: usually we move types in web3-types package if these are used by 1+ package / are common types. 
As web3-types package is a dependency of almost all other packages so we try to keep it light for usecase of other individual packages usage.

( same for utils package , for future changes)

jdevcs (commented):
also EIP-6963 types are exported under provider namespace:
https://github.com/web3/web3.js/blob/6f9a485e0715a70170bf170ca1e1d1b26b43f6c3/packages/web3/src/providers.exports.ts#L22
https://github.com/web3/web3.js/blob/6f9a485e0715a70170bf170ca1e1d1b26b43f6c3/packages/web3/src/index.ts#L353

but I think we may also export directly

danforbes:
Thank you for the clarification, @jdevcs - I had gotten confused. I need to revert #7271 as well.

packages/web3-types/src/eip6963_types.ts
luu-alex: I believe since we created web3-types, it should encompass all the neccesary types to use the package. I think as well remain consistent, as we have EIP1193Provider types EIP6693 provider types should be included. This will allow these types to be exported through the main package which would make alot of since for most users, as I believe not very many users would know that EIP6693 types would be exported through providers namespace. We had a discussion on this a while back on this issue about doing this change https://github.com/web3/web3.js/issues/6952

danforbes:
Closes #6952

packages/web3-types/src/eip6963_types.ts
jdevcs: > as I believe not very many users would know that EIP6693 types would be exported through providers namespace.

I agree on this, that these types should be directly exported from web3 package, instead of via namspace. 

But regarding first point : only shared types shouldbe in web3-types package, and eip6963 types are used in web3 main package only? Same applies to other packages non-shared types like https://github.com/web3/web3.js/blob/4.x/packages/web3-eth-contract/src/types.ts

@luu-alex

luu-alex: I see what you are getting at, I think id still prefer it being within web3-types package as it feels intuitive for an eip type to be there, but it should be fine as long as this is documented and users are able to bring these types in properly

jdevcs (approved):

PR #7352, opened 2024-10-23
Update dRPC Provider with API Key

danforbes:
Modeled after https://github.com/web3/web3.js/blob/4.x/packages/web3-rpc-providers/src/web3_provider_quicknode.ts


avkos (approved):

jdevcs (approved):
looks good, plus before merging update changelog

Issue #7188, opened 2024-08-08
add eth_newPendingTransactionFilter to rpc method wrappers


luu-alex:
eth_newPendingTransactionFilter exists in rpc methods, its just not being exposed by web3.eth. Double check and make sure other methods are implemented as well 
https://github.com/web3/web3.js/issues/4483
https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_newpendingtransactionfilter

- [x] List all RPC methods
- [x] Move to library
- [x] Doc Update

mconnelly8:
9/3 Meeting - should have this to support execution level.  Just need to expose it to the public in the lib.

jdevcs:
following function are available in RPC Methods package but are not exported via Eth class:


| Function Name          | [JSON RPC API](https://ethereum.org/en/developers/docs/apis/json-rpc/) | [EL Specs](https://github.com/ethereum/execution-apis) |
|------------------------|----------|----------|
| getCompilers        |    x     |     x     |
| compileSolidity     |    x      |    x      |
| compileLLL          |     x     |     x     |
| compileSerpent      |     x     |     x     |
| `newFilter`           |    ✓     |     ✓     |
| `newBlockFilter`      |     ✓     |     ✓    |
| `uninstallFilter`     |     ✓     |    ✓      |
| `getFilterChanges`    |    ✓      |   ✓       |
| `getFilterLogs`       |     ✓     |    ✓      |
| submitHashrate      |    x      |     x     |
| `newPendingTransactionFilter` |  ✓     |    ✓      |

PR #7353, opened 2024-10-24
newPendingTransactionFilter

jdevcs:
## Description

https://github.com/web3/web3.js/issues/7188

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-eth/src/filtering_rpc_method_wrappers.ts
+++ b/packages/web3-eth/src/filtering_rpc_method_wrappers.ts
@@ -1,0 +1,171 @@
+/*
+This file is part of web3.js.
+
+web3.js is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+web3.js is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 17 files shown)


packages/web3-eth/src/filtering_rpc_method_wrappers.ts
```diff
@@ -0,0 +1,171 @@
...
+import { logSchema } from './schemas.js';
+
+/**
+ * View additional documentations here: {@link Web3Eth.createNewPendingTransactionFilter}
+ * @param web3Context ({@link Web3Context}) Web3 configuration object that contains things such as the provider, request manager, wallet, etc.
+ * @param returnFormat ({@link DataFormat}) Return format
+ */
+export async function createNewPendingTransactionFilter<ReturnFormat extends DataFormat>(
+	web3Context: Web3Context<EthExecutionAPI>,
+	returnFormat: ReturnFormat,
```
danforbes, line 32: Should this use `defaultReturnFormat` as a default value so that this can be an optional parameter? This same question would apply to all other cases where there is a `returnFormat` parameter without a default value.

danforbes (approved):
Thank you for the great documentation 🙏🏻 Just one question.

packages/web3-eth/src/filtering_rpc_method_wrappers.ts
jdevcs: Default return format is specified for these methods in Web3Eth class, and we are following this for all other rpc method wrappers.  This can be a feature request to specify default return format for all rpc wrappers.

avkos (approved):

PR #7374, opened 2024-11-07
fix: export Web3Account, Wallet and signature related types

### LINKED ISSUE #7341: Export all return types such as `Web3Account` or the signature type
Hi, I’ve noticed that some types returned by methods aren’t exported as types, such as `Web3Account` or `Signature` (`ReturnType<typeof account.sign>`). This makes it more difficult to structure larger TypeScript codebases effectively. Please see the example below:

```ts
// this is not possible
import { Signature, Web3Account } from "web3"

function processSignature(signature: Signature) {
  // Do something with the signature
}

function prepareAndSignTransaction(account: Web3Account, to: Address) {
  // Prepare the transaction
}
```

krzysu:
## Description

Export types that represent the return values of method calls but aren’t currently available as standalone types. Specifically, this includes `Web3Account`, `Wallet`, and signature-related types like `SignatureObject`, `SignResult`, and `SignTransactionResult`.

Fixes #7341

## Type of change

## Checklist:

### DIFF
--- a/packages/web3-eth-accounts/src/types.ts
+++ b/packages/web3-eth-accounts/src/types.ts
@@ -16,37 +16,7 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { Web3BaseWalletAccount, HexString } from 'web3-types';
-import { FeeMarketEIP1559TxData, AccessListEIP2930TxData, TxData } from './tx/types.js';
 import { AccessListEIP2930Transaction, FeeMarketEIP1559Transaction, Transaction } from './tx';
-
-export type SignatureObject = {
-	messageHash: string;
-	r: string;
-	s: string;
-	v: string;
-};
-
-export type SignTransactionResult = SignatureObject & {
-	rawTransaction: string;
-	transactionHash: string;
-};
-
-export type SignTransactionFunction = (
-	transaction:
-		| TxData
-		| AccessListEIP2930TxData
-		| FeeMarketEIP1559TxData
-		| Record<string, unknown>,
-) => SignTransactionResult;
-
-export type SignResult = SignatureObject & {
-	message?: string;
-	signature: string;
-};
-
-export type SignFunction = (data: string, privateKey: string) => SignResult;
-
-// https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition
 
 export interface Web3Account extends Web3BaseWalletAccount {
 	address: HexString;
--- a/packages/web3-eth-accounts/src/types.ts
+++ b/packages/web3-eth-accounts/src/types.ts
@@ -15,8 +15,16 @@ You should have received a copy of the GNU Lesser General Public License
 along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { Web3BaseWalletAccount, HexString } from 'web3-types';
+import {
+	HexString,
+	SignatureObject,
+	SignResult,
+	SignTransactionResult,
+	Web3BaseWalletAccount,
+} from 'web3-types';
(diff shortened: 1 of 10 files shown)


packages/web3-eth-accounts/src/types.ts
```diff
@@ -16,38 +16,8 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
-	s: string;
-	v: string;
-};
-
-export type SignTransactionResult = SignatureObject & {
-	rawTransaction: string;
-	transactionHash: string;
-};
-
-export type SignTransactionFunction = (
```
krzysu, line 34: this is not used anywhere

packages/web3-eth-accounts/src/types.ts
```diff
@@ -16,38 +16,8 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
...
-		| FeeMarketEIP1559TxData
-		| Record<string, unknown>,
-) => SignTransactionResult;
-
-export type SignResult = SignatureObject & {
-	message?: string;
-	signature: string;
-};
-
-export type SignFunction = (data: string, privateKey: string) => SignResult;
```
krzysu, line 47: this is not used anywhere

packages/web3/src/index.ts
```diff
@@ -364,3 +364,4 @@ export * as validator from 'web3-validator';
  */
 export * from 'web3-errors';
 export * from 'web3-types';
+export type { Web3Account, Wallet } from 'web3-eth-accounts';
```
krzysu: if there is any better place to export these, let me know!

packages/web3-eth-accounts/src/types.ts
```diff
@@ -16,38 +16,8 @@ along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { Web3BaseWalletAccount, HexString } from 'web3-types';
-import { FeeMarketEIP1559TxData, AccessListEIP2930TxData, TxData } from './tx/types.js';
 import { AccessListEIP2930Transaction, FeeMarketEIP1559Transaction, Transaction } from './tx';
 
-export type SignatureObject = {
```
jdevcs, line 22: I'll suggest to move only shared types that are used in 1+ packages in web3-types package, and for moved types re-export those from web3-eth-accounts package for backward compatibility.

packages/web3/src/index.ts
jdevcs: from web3-accts pkg

krzysu: all account types are exported from `web3-eth-accounts` but not from `web3`. See https://github.com/web3/web3.js/blob/7341-export-types/packages/web3-eth-accounts/src/index.ts#L42
How to solve that?

packages/web3-eth-accounts/src/types.ts
krzysu: the problem is that `SignatureObject` is used in web3-eth-accounts but is also part of the `Web3BaseWalletAccount` type in `web3-types`. 
See the type details below:

```ts
export interface Web3BaseWalletAccount {
	// ..
	readonly signTransaction: (tx: Transaction) => Promise<SignTransactionResult>;
	readonly sign: (data: Record<string, unknown> | string) => SignResult;
	//
}
```

Also, none of the `web3-eth-accounts` types are exported from the umbrella `web3` package, but I will reexport these for `web3-eth-accounts` backwards compatibility.

packages/web3/src/index.ts
krzysu: I moved the export of these to packages/web3/src/types.ts

luu-alex (approved):

PR #7418, opened 2024-12-07
Added documentation about create2 method

pamellix:
## Description
https://github.com/web3/web3.js/issues/7369
Added the documentation for `create2ContractAddress`:
- Included JSDoc comments directly in the function definition.
## Type of change

-   [x] Documentation (changes that only address relatively minor typographical errors are not accepted)

## Checklist:

-   [ ] I have selected the correct base branch.
-   [ ] I have performed a self-review of my own code.
-   [ ] I have commented my code, particularly in hard-to-understand areas.
-   [x] I have made corresponding changes to the documentation.
-   [ ] My changes generate no new warnings.
-   [ ] Any dependent changes have been merged and published in downstream modules.
-   [x] I ran `npm run lint` with success and extended the tests and types if necessary.
-   [ ] I ran `npm run test:unit` with success.
-   [ ] I ran `npm run test:coverage` and my test cases cover all the lines and branches of the added code.
-   [ ] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
-   [ ] I have tested my code on the live network.
-   [ ] I have checked the Deploy Preview and it looks correct.
-   [ ] I have updated the `CHANGELOG.md` file in the root folder.
-   [ ] I have linked Issue(s) with this PR in "Linked Issues" menu.

PR #7421, opened 2024-12-10
doc: add docs of createContractAddress

jasonandjay:
## Description

Added the documentation for createContractAddress just like https://github.com/web3/web3.js/pull/7418:

Included JSDoc comments directly in the function definition.

## Type of change

-   [x] Documentation (changes that only address relatively minor typographical errors are not accepted)

## Checklist:

-   [ ] I have selected the correct base branch.
-   [ ] I have performed a self-review of my own code.
-   [ ] I have commented my code, particularly in hard-to-understand areas.
-   [x] I have made corresponding changes to the documentation.
-   [ ] My changes generate no new warnings.
-   [ ] Any dependent changes have been merged and published in downstream modules.
-   [x] I ran `npm run lint` with success and extended the tests and types if necessary.
-   [ ] I ran `npm run test:unit` with success.
-   [ ] I ran `npm run test:coverage` and my test cases cover all the lines and branches of the added code.
-   [ ] I ran `npm run build` and tested `dist/web3.min.js` in a browser.
-   [ ] I have tested my code on the live network.
-   [ ] I have checked the Deploy Preview and it looks correct.
-   [ ] I have updated the `CHANGELOG.md` file in the root folder.
-   [ ] I have linked Issue(s) with this PR in "Linked Issues" menu.

### DIFF
--- a/packages/web3-eth-contract/src/utils.ts
+++ b/packages/web3-eth-contract/src/utils.ts
@@ -235,14 +235,14 @@ export const getCreateAccessListParams = ({
  *
  * @param from The sender’s Ethereum {@link Address}, from which the contract will be deployed.
  * @param nonce The transaction count (or {@link Numbers}) of the sender account at the time of contract creation.
- *              This is used to ensure uniqueness of the generated address.
+ *              You can get it here: https://docs.web3js.org/api/web3/class/Web3Eth#getTransactionCount.
  * @returns An Ethereum {@link Address} of the contract in checksum address format.
  * @throws An {@link InvalidAddressError} if the provided address ('from') is invalid.
  * @throws An {@link InvalidNumberError} if the provided nonce value is not in a valid format.
  * @example
  * ```ts
  * const from = "0x1234567890abcdef1234567890abcdef12345678";
- * const nonce = 1; // The nonce value for the transaction
+ * const nonce = (await web3.eth.getTransactionCount(from)) + 1; // The nonce value for the transaction
  *
  * const res = createContractAddress(from, nonce);
  *
--- a/packages/web3-eth-contract/src/utils.ts
+++ b/packages/web3-eth-contract/src/utils.ts
@@ -227,6 +227,29 @@ export const getCreateAccessListParams = ({
 	return txParams;
 };
 
+/**
+ * Generates the Ethereum address of a contract created via a regular transaction.
+ *
+ * This function calculates the contract address based on the sender's address and nonce,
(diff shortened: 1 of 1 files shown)


packages/web3-eth-contract/src/utils.ts
```diff
@@ -227,6 +227,29 @@ export const getCreateAccessListParams = ({
...
 
+/**
+ * Generates the Ethereum address of a contract created via a regular transaction.
+ *
+ * This function calculates the contract address based on the sender's address and nonce,
+ * following Ethereum's address generation rules.
+ *
+ * @param from The sender’s Ethereum {@link Address}, from which the contract will be deployed.
+ * @param nonce The transaction count (or {@link Numbers}) of the sender account at the time of contract creation.
+ *              This is used to ensure uniqueness of the generated address.
```
danforbes: I think this can be simplified as suggested below. Can you link to `Web3Eth.getTransactionCount` here since that is most likely how users will be acquiring the nonce? https://docs.web3js.org/api/web3/class/Web3Eth#getTransactionCount

```suggestion
 * @param nonce The transaction count (or {@link Numbers}) of the sender account at the time of contract creation.
```

packages/web3-eth-contract/src/utils.ts
```diff
@@ -227,6 +227,29 @@ export const getCreateAccessListParams = ({
...
+ * @returns An Ethereum {@link Address} of the contract in checksum address format.
+ * @throws An {@link InvalidAddressError} if the provided address ('from') is invalid.
+ * @throws An {@link InvalidNumberError} if the provided nonce value is not in a valid format.
+ * @example
+ * ```ts
+ * const from = "0x1234567890abcdef1234567890abcdef12345678";
+ * const nonce = 1; // The nonce value for the transaction
```
danforbes: This is probably more helpful, but please ensure it works and is correct before accepting the suggestion.

```suggestion
 * const nonce = (await web3.eth.getTransactionCount(from)) + 1; // The nonce value for the transaction
```

packages/web3-eth-contract/src/utils.ts
jasonandjay: Of course, that would be clearer.

packages/web3-eth-contract/src/utils.ts
jasonandjay: Although nonce is a basic knowledge that all developers must have, it is better to understand

danforbes (approved):

PR #7462, opened 2025-02-21
Fix Incorrect NPM Links for web3-errors and web3-rpc-methods in README

savvar9991:


jdevcs (commented):
Thanks for your contributions. we are in the process of sunsetting Web3.js, Our focus during this transition will be on addressing only critical issues to ensure a smooth migration for our users.
For further details regarding the sunsetting of Web3.js, please refer to the official announcement here: [Web3.js Sunset Announcement](https://blog.chainsafe.io/web3-js-sunset/)

